PR #125, opened 2021-10-15
60fps redo

oziphantom:
This has the RC1 60fps code as it left my linux vm, I then grabbed all your changes from master and patched up the ASF->RSF ( I think) i


oziphantom:
this has 30/60 fps as a toggle on the title screen.
save files are compatible with each other (lightly tested)
theoretically 30fps mode is "unchanged" again lightly tested

oziphantom:
it just modifies all the files and doesn't say anything? 

ok I will copy the files back to windows and check in

src/3dsystem/3d_gen.c
```diff
@@ -364,7 +364,7 @@ void phd_InitWindow(
     PhdCenterX = width / 2;
     PhdCenterY = height / 2;
     PhdNearZ = nearz << W2V_SHIFT;
-    PhdFarZ = farz << W2V_SHIFT;
+    PhdFarZ = 2147483647; // farz << W2V_SHIFT;
```
rr-, line 367: Why?

.gitignore
```diff
@@ -1,2 +1,16 @@
...
+*.log
+*.ilk
+*.idb
+*.sln
+v15/
+v16/
+Release/
+Debug/
+*.vcxproj*
+
```
rr-, line 16: These changes do not belong in this pull request

src/global/vars.h
```diff
@@ -296,4 +296,7 @@ extern int16_t InvChosen;
 
 extern int16_t BarOffsetY[6];
 
+extern int32_t LaraSpeedError;
+extern int16_t ANIM_SCALE;
```
rr-, line 300: Global variables should be named in PascalCase. This looks like a macro.

src/game/camera.c
```diff
@@ -502,6 +508,8 @@ void FixedCamera()
     }
 }
 
+TEXTSTRING *cameraText = NULL;
```
rr-, line 511: Global variables should be named in PascalCase.

src/game/camera.c
```diff
@@ -655,6 +663,31 @@ void CalculateCamera()
...
+        T_ChangeText(cameraText, ammostring);
+    } else {
+        
+        cameraText = T_Print(
+            -screen_margin_h - text_offset_x, (text_height*3) + screen_margin_v,
+            ammostring);
+        T_SetScale(cameraText, PHD_ONE * scale, PHD_ONE * scale);
+        //T_RightAlign(LaraText2, 1);
+    }
+#endif
```
rr-, line 690: We do not commit in dead code.

src/game/cinema.c
```diff
@@ -17,6 +17,8 @@
 static int32_t OldSoundIsActive;
 static const int32_t CinematicAnimationRate = 0x8000;
 
+extern PHD_3DPOS_F lara_float_pos;
```
rr-, line 20: Global variables should be named in PascalCase.

src/game/cinema.c
```diff
@@ -110,7 +116,7 @@ int32_t DoCinematic(int32_t nframes)
         CalculateCinematicCamera();
         CineFrame++;
 
-        if (CineFrame >= NumCineFrames) {
+        if ((CineFrame / ANIM_SCALE) >= NumCineFrames - 1) {
```
rr-, line 119: Why `-1`?

src/specific/file.c
```diff
@@ -273,12 +286,50 @@ static int32_t LoadObjects(MYFILE *fp)
     AnimRanges = game_malloc(
         sizeof(ANIM_RANGE_STRUCT) * AnimRangeCount, GBUF_ANIM_RANGES);
     FileRead(AnimRanges, sizeof(ANIM_RANGE_STRUCT), AnimRangeCount, fp);
+    for (int32_t i = 0; i < AnimRangeCount; ++i) // DAN
```
rr-, line 289: What does DAN mean?…

src/specific/file.c
```diff
@@ -273,12 +286,50 @@ static int32_t LoadObjects(MYFILE *fp)
...
+        AnimRanges[i].start_frame *= ANIM_SCALE;
+        AnimRanges[i].link_frame_num *= ANIM_SCALE;
+    }
 
     FileRead(&AnimCommandCount, sizeof(int32_t), 1, fp);
     LOG_INFO("%d anim commands", AnimCommandCount);
     AnimCommands =
         game_malloc(sizeof(int16_t) * AnimCommandCount, GBUF_ANIM_COMMANDS);
     FileRead(AnimCommands, sizeof(int16_t), AnimCommandCount, fp);
+    for (uint32_t j = 0; j < AnimCount; ++j) // DAN
```
rr-, line 301: What does DAN mean?

src/specific/file.c
```diff
@@ -273,12 +286,50 @@ static int32_t LoadObjects(MYFILE *fp)
...
+    {
+        if (Anims[j].number_commands > 0) {
```
rr-, line 303: Use continue to reduce indentation

src/specific/file.c
```diff
@@ -273,12 +286,50 @@ static int32_t LoadObjects(MYFILE *fp)
...
+            int16_t *command = &AnimCommands[Anims[j].command_index];
+            for (uint32_t i = Anims[j].number_commands; i > 0; i--) {
+                switch (*(command++)) {
+                case AC_MOVE_ORIGIN:
+                    /* Translate (allowing for y-axis rotation) */
```
rr-, line 308: Avoid `/**/`

src/specific/file.c
```diff
@@ -273,12 +286,50 @@ static int32_t LoadObjects(MYFILE *fp)
...
+                    command += 3;
+                    break;
+
+                case AC_JUMP_VELOCITY:
+                    // item->fall_speed = *(command++);        // Get Upward
+                    // Velocity item->speed = *(command++);         // Get
```
rr-, line 314: Remove dead code

src/global/const.h
```diff
@@ -207,10 +213,12 @@
 #define LOW_LIGHT 0x1400 // = 5120
 #define HIGH_LIGHT 0x1000 // = 4096
 
-#define VIEW_NEAR 127
+#define VIEW_NEAR 127 / ANIM_SCALE
```
rr-, line 216: Missing braces around a macro

src/global/const.h
```diff
@@ -207,10 +213,12 @@
...
 #define VIEW_FAR (0x14 * 0x400)
-#define DEPTH_Q_END (20 * 1024) // = 20480
-#define DEPTH_Q_START ((DEPTH_Q_END) - (8 * 1024)) // = 12288
+//#define DEPTH_Q_END (20 * 1024)
+#define DEPTH_Q_END (40 * 1024)
```
rr-, line 219: Why?

src/global/const.h
```diff
@@ -207,10 +213,12 @@
...
+//#define DEPTH_Q_START ((DEPTH_Q_END) - (8 * 1024))
+#define DEPTH_Q_START ((DEPTH_Q_END - 1024)) // - (16 * 1024))
```
rr-, line 221: Why?

src/global/const.h
```diff
@@ -14,7 +14,9 @@
 #define MAX_UI_SCALE 2.0
 #define DEFAULT_UI_SCALE 1.0
 
-#define FRAMES_PER_SECOND 30
+//#define ANIM_SCALE 2 //this is now in vars
```
rr-, line 17: Then remove it from here

src/game/traps/thors_hammer.c
```diff
@@ -63,7 +63,7 @@ void ThorsHandleControl(int16_t item_num)
 
     case THS_ACTIVE: {
         int32_t frm = item->frame_number - Anims[item->anim_number].frame_base;
-        if (frm > 30) {
+        if (frm > (30 * ANIM_SCALE)) {
```
rr-, line 66: No need for braces

src/game/traps/rolling_ball.c
```diff
@@ -38,7 +38,7 @@ void RollingBallControl(int16_t item_num)
         if (item->pos.y < item->floor) {
             if (!item->gravity_status) {
                 item->gravity_status = 1;
-                item->fall_speed = -10;
+                item->fall_speed = -10; //(-10/ANIM_SCALE);
```
rr-, line 41: Dead code

src/game/traps/movable_block.c
```diff
@@ -137,7 +137,7 @@ void MovableBlockCollision(
             Lara.gun_status = LGS_HANDSBUSY;
         }
     } else if (lara_item->current_anim_state == AS_PPREADY) {
-        if (lara_item->frame_number != AF_PPREADY) {
+        if (lara_item->frame_number != (AF_PPREADY * ANIM_SCALE)) {
```
rr-, line 140: Unnecessary braces

src/game/traps/falling_ceiling.c
```diff
@@ -33,7 +33,7 @@ void FallingCeilingControl(int16_t item_num)
         && item->pos.y >= item->floor) {
         item->goal_anim_state = TRAP_WORKING;
         item->pos.y = item->floor;
-        item->fall_speed = 0;
+        item->fall_speed = 0 / ANIM_SCALE;
```
rr-, line 36: Pointless division

src/game/savegame.c
```diff
@@ -434,6 +438,8 @@ void ExtractSaveGameInfo()
 
     ReadSG(&FlipEffect, sizeof(int32_t));
     ReadSG(&FlipTimer, sizeof(int32_t));
+
+    LaraSetFloatPosFromFixed(); // fix up her float positions
```
rr-, line 442: The comment repeats the function name, no need for it

src/game/objects/scion.c
```diff
@@ -142,7 +142,8 @@ void PickUpScionCollision(
 
     if (lara_item->current_anim_state == AS_PICKUP) {
         if (lara_item->frame_number
-            == Anims[lara_item->anim_number].frame_base + AF_PICKUPSCION) {
+            == Anims[lara_item->anim_number].frame_base
+                + (AF_PICKUPSCION * ANIM_SCALE)) {
```
rr-, line 146: Unnecessary braces

src/game/objects/puzzle_hole.c
```diff
@@ -45,7 +45,7 @@ void PuzzleHoleCollision(
             return;
         }
 
-        if (lara_item->frame_number == AF_USEPUZZLE) {
+        if (lara_item->frame_number == (AF_USEPUZZLE * ANIM_SCALE)) {
```
rr-, line 48: Unnecessary braces (I will stop with this comment here, please go through the diff and apply to all cases)

src/game/traps/lava.c
```diff
@@ -72,7 +73,7 @@ void LavaEmitterControl(int16_t item_num)
         fx->pos.y = item->pos.y;
         fx->pos.z = item->pos.z;
         fx->pos.y_rot = (GetRandomControl() - 0x4000) * 2;
-        fx->speed = GetRandomControl() >> 10;
+        fx->speed = GetRandomControl() / ANIM_SCALE >> 10;
```
rr-, line 76: Missing braces

src/game/laramisc.c
```diff
@@ -14,10 +14,19 @@
...
 #include <stddef.h>
 #include <stdint.h>
 
+int32_t lara_jump_error = 0;
+TEXTSTRING *LaraText = NULL;
+TEXTSTRING *LaraText2 = NULL;
+
+PHD_3DPOS_F lara_float_pos;
+double lara_fall_speed_f = 0.0;
+double lara_speed_F = 0.0;
```
rr-, line 28: PascalCase for all of the above variables

Consider moving these to `ITEM_INFO`, it's not like we need to worry about the game using these structures as well – everything was decompiled anyway

src/game/laramisc.c
```diff
@@ -215,6 +238,8 @@ void LaraSwapMeshExtra()
     }
 }
 
+int32_t frameCounter = 0;
```
rr-, line 241: PascalCase

src/global/vars.h
oziphantom: it was a macro, but I made it a toggle. I can do a find and replace but it will modify a lot.

src/specific/file.c
oziphantom: that I changed it so I can find it again, removed.

src/global/const.h
oziphantom: fixes object clipping near the camera I believe

rr-:
Could you elaborate what kind of bug it is?

src/game/laramisc.c
oziphantom: From memory the ITEM_INFO are loaded or partially loaded, but modifying it did case crashes or issues,

oziphantom:
ping

oziphantom:
okay, fetched up stream, merged in master, merged master into this branch and resolved the conflicts

rr- (commented):
Very good work. I added some cosmetic updates. But I have one major pain point with this PR - overall it introduces way too much micromanagement of Lara's floating point coordinates and has lack of compatibility for other objects such as enemies or traps.

I have two proposals how to remedy this:

- either switching entirely to floating point arithmetic for ITEM_INFO/FX_INFO's x/y/z / speed / fallspeed. This sounds insane, but may actually not be so bad given the scope of the changes that were already done.
- another solution would be to amend the ITEM_INFO structure to store floating point coordinates. Then we could make helper functions for these - ItemSetPos that would update both integer and fp coordinates, and GetItemPosF and GetItemPosI to retrieve them; same for fall speed. Then use those functions across the codebase once rather than repeating the same assignments twice.

While needing some adjustments, this in general should be compatible with savegame files, because we write each item's properties manually rather than just dumping the entire `ITEM_INFO` content.

This way we get floating point logic for every ingame object for free. Additionally, we can remove `ShiftItemLara` and `TranslateItem_f` specializations that feel, well I'll be honest, quite awkward. I'd rather not introduce these to the codebase.

Other points:

- Please do something about the hanging from ledge quadrant hacks. I really don't want to commit it in like this.
- Same about that `item->fall_speed + 3 - DAMAGE_START` magic number.
- There are many cases where we divide by AnimScale and cast to (double) and many cases where we don't do that. IMO we should settle for a single and consistent way.

Finally I also added a few individual comments.

src/game/laramisc.c
```diff
@@ -234,19 +250,25 @@ void AnimateLara(ITEM_INFO *item)
...
+                        &LaraFloatPos, command[0], command[1], command[2],
+                        item->pos.y_rot);
                     command += 3;
                     break;
 
-                case AC_JUMP_VELOCITY:
+                case AC_JUMP_VELOCITY: {
                     item->fall_speed = command[0];
-                    item->speed = command[1];
+                    LaraFallSpeedF = (double)command[0];
```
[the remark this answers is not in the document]
oziphantom: speed is not scaled and is always the same, we modify it when you add it to her position.

src/game/cinema.c
```diff
@@ -125,7 +131,7 @@ void CalculateCinematicCamera()
     PHD_VECTOR campos;
     PHD_VECTOR camtar;
 
-    int16_t *ptr = &Cine[8 * CineFrame];
+    int16_t *ptr = &Cine[8 * CineFrame / AnimScale];
```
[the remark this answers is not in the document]
oziphantom: we probably do, just not tested far enough for it to happen. Or it just so happens the compiler in that case did it the right order and hence it didn't crash. Good catch.

src/game/cinema.c
```diff
@@ -110,7 +116,7 @@ int32_t DoCinematic(int32_t nframes)
         CalculateCinematicCamera();
         CineFrame++;
 
-        if (CineFrame >= NumCineFrames) {
+        if (CineFrame / AnimScale >= NumCineFrames - 1) {
```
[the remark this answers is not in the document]
oziphantom: doesn't fix and still gives a T pose frame. We can remove 1 frame from only 60fps if you want?

src/game/lara.c
```diff
@@ -2264,9 +2389,11 @@ int32_t LaraLandedBad(ITEM_INFO *item, COLL_INFO *coll)
     item->floor = height;
     item->pos.y = height;
     TestTriggers(TriggerIndex, 0);
-    item->pos.y = oy;
+    item->pos.y = oy; // restores Y so don't touch double y
```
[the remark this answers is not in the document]
oziphantom: I don't make a separate cache for the double pos.Y because this restores Y and the code in-between doesn't modify the floating point position. I.e this only does fp Y not f Y because fp Y is ultimately unchanged.

rr-:
>When you make a game you go by what works when you play it on the machine it needs to run and ship it. Nobody sits around and goes,"but yes is this mathematically pure, what if say we tried to run this on a I don't know, lets imagine a Pentium III comes out with I don't know, some insane speed like 1Ghz would this still hold up?" I can tell you now, this game won't hold up :) 

Sorry this is unintelligible to me.

>Without that addition you can't grab unless you are dead on a 90degree multiple angle, and even then sometimes being dead on it not enough. The jump pushing you forward a little bit is not a mistake, I feel the engine needs that penetration to make it work, just with the errors in the maths and I also feel that the level collision data is not quite aligned properly, possibly even its 1000 big not 1024 big, and "its just so happens to work thanks to maths error". Basically I know of no way to get rid of the magic + for this quadrant to make it work on all the cases. The number is possibly too high, if you hit with a hammer make it Mjolnir, especially when people are tired of testing the whole game to make sure it was fixed in every instance. Even then Ash found another one it failed on.

I really want to get rid of this particular code as soon as possible. We're not TR5.

>I did ponder making a function to handle fixed -> float and then float -> fixed and putting them everywhere, but I felt it was against the nature of the code. Happy to do it though, however I feel this is a separate tidy up task and not something to halt the merge as it stands.

You've been working on this for a few weeks, I'm seeing this for a few days and it really feels like you're trying to rush to get this merged, while it's really not production-quality code. I do understand the sense of urgency related to the TR25 event so I'm ok with merging it as-is… as long as you promise to address my remarks later. This is quite big technical debt we're taking, which combined with Core's design becomes quite significant. I don't want to build more features on top of a crumbling foundation.

>I found adding things to the end of ITEM_INFO was instant death, I don't know why,

You did recompile the entire thing from scratch, yes?

rr-:
I'm really not appreciating this attitude. Your issues with using modern version control software are really not of my concern. You can see my changes (that are mostly variable renames) in a single commit – I decided that just making the changes would be faster than hear you complain about having to switch back and forth between your VS build branches every time I raise a change request. If these break things, sorry about it. What is for you tons of work done in testing, for me is still very much experimental feature with multiple layers of hacks that I don't want.

You really can't blame me for your decision to go forth with your feature without consulting with upstream and with complete disregard to the existing naming conventions, coding style and build pipeline that you ostentatiously agreed to uphold upon my request, while also pressuring me in every communication channel (discord, github, unrelated github issues) to merge this due to your personal issues with the build system.

This sort of interaction with the community makes me regret going back to this project.

rr-:
Closing due to lack of activity

PR #132, opened 2021-10-18
Miscfixes

oziphantom:
A few minor fixes. 
- fixes demo mode not loading the right level
- fixes uninitialized code warning
- fixes compile guard define


rr-:
Thanks

PR #133, opened 2021-10-18
MSVC build fixes

oziphantom:
This has fixes unique to MSVC and Clang doesn't have a sln, but permits one to live ;)


src/specific/init.c
```diff
@@ -139,7 +139,7 @@ void *game_malloc(int32_t alloc_size, GAMEALLOC_BUFFER buf_index)
     void *result = GameAllocMemPointer;
     GameAllocMemFree -= aligned_size;
     GameAllocMemUsed += aligned_size;
-    GameAllocMemPointer += aligned_size;
+    GameAllocMemPointer = (char *)GameAllocMemPointer + aligned_size;
```
rr-, line 142: Hmm, what do you think about changing this pointer to `char*` in its definition?

oziphantom: no issues from Clang either.

rr- (approved):

PR #141, opened 2021-10-24
Port a few HWR_ routines

rr-:


src/specific/ati.h
```diff
@@ -359,7 +359,6 @@ typedef struct {
     C3D_BOOL bClampS;
     C3D_BOOL bClampT;
     C3D_BOOL bAlphaBlend;
-    C3D_ETEXTILE eTexTiling;
```
thatguyfrombb, line 362: Both the documentation and the SDK header used to build GLRage include this line. Why was it removed?

src/specific/hwr.c
```diff
@@ -1296,8 +1693,11 @@ void T1MInjectSpecificHWR()
     INJECT(0x00409F44, HWR_InsertObjectG4);
     INJECT(0x0040A01D, HWR_InsertObjectG3);
     INJECT(0x0040A6B1, HWR_ClipVertices2);
```
thatguyfrombb, line 1695: "HWR_ZedClipper" not injected (I know... doesn't really matter ;) )

thatguyfrombb (changes requested):

PR #146, opened 2021-10-28
3d pickups

oziphantom:
here they are. 

I'm doing this against main, because against your branch was 321 check-ins and basically a few hundred files just 100% changed. 

Unless there is a better way?


rr-:
The general consensus on Discord (:roll_eyes:) seems to be keep it as-is and I'm fine with this decision too. @thatguyfrombb @oziphantom what do you think?

thatguyfrombb:
Everybody that posted in the Discord preferred the unscaled version, so I'd say unscaled it is.

thatguyfrombb (approved):

oziphantom:
its moot, but yeah, 1:1

PR #156, opened 2021-10-31
Remove injections + inline some of the vars.h

rr-:
Now it's possible to remove most of the INJECT macros because 99% of the game runs from our code, with two exceptions. I had to leave injection macros for WinMain so that it actually boots our entrypoint, not the original game's, and I still had to leave ATI3DCIF injections due to integration with glrage, which works by patching the OG routines. I didn't inline all of the variables, I want to do it in incremental fashion and span it across a few PRs – especially that for some of the easier stuff I want to refactor it as we go (see HiRes and GameHiRes encapsulation routines in this PR).


rr-:
(The reason why I put forth so many commits is to make potential bughunting with `git bisect` easier in the future – it proved to be very valuable in the past.)

src/global/types.h
```diff
@@ -2051,6 +2051,12 @@ typedef struct SAMPLE_DATA {
 
 #pragma pack(pop)
 
+typedef struct PICTURE {
```
thatguyfrombb, line 2054: Do we care about the packing on this?

thatguyfrombb (changes requested):

src/game/traps/teeth_trap.c
```diff
@@ -12,6 +13,19 @@ BITE_INFO Teeth2B = { 71, 10, -1718, 1 };
 BITE_INFO Teeth3A = { -23, -10, -1718, 0 };
 BITE_INFO Teeth3B = { 71, -10, -1718, 1 };
 
+static void BaddieBiteEffect(ITEM_INFO *item, BITE_INFO *bite);
+
+static void BaddieBiteEffect(ITEM_INFO *item, BITE_INFO *bite)
```
oziphantom, line 18: why teeth_trap and not control which holds all the global "creature" code? Or in blood.c of which this creates and sets up?

src/game/settings.c
```diff
@@ -33,7 +34,14 @@ static int32_t S_ReadUserSettingsATI()
...
     FileRead(Layout[1], sizeof(int16_t), 13, fp);
     FileRead(&RenderSettings, sizeof(int32_t), 1, fp);
-    FileRead(&GameHiRes, sizeof(int32_t), 1, fp);
+
+    {
+        int32_t resolution_idx;
+        FileRead(&resolution_idx, sizeof(int32_t), 1, fp);
+        CLAMP(resolution_idx, 0, RESOLUTIONS_SIZE - 1);
+        SetGameScreenSizeIdx(resolution_idx);
+    }
```
oziphantom, line 43: why is this scoped?

src/specific/display.c
```diff
@@ -10,10 +10,66 @@
 
 #include <stdlib.h>
 
+// The screen resolution is controlled by two variables are indices
+// within an array of predefined screen resolutions.
+// Actual screen resolution, sometimes different from the game
+// resolution (during FMVs, main menu sequence, static pictures etc.)
```
oziphantom: this comment block could be written a lot better.

src/specific/display.c
```diff
@@ -10,10 +10,66 @@
...
+    if (idx >= 0 && idx < RESOLUTIONS_SIZE) {
+        GameHiRes = idx;
+        return true;
+    }
+    return false;
+}
+
+bool SetPrevGameScreenSize()
+{
+    if (GameHiRes - 1 >= 0) {
```
oziphantom, line 33: GameHiRes > 0

oziphantom: which is just if GameHiRes

src/specific/display.c
```diff
@@ -10,10 +10,66 @@
...
+        GameHiRes--;
+        return true;
+    }
+    return false;
+}
+
+bool SetNextGameScreenSize()
+{
+    if (GameHiRes + 1 < RESOLUTIONS_SIZE) {
```
oziphantom, line 42: RESOLUTIONS_SIZE is a constant, so do GameHiRes < RESOLUTIONS_SIZE - 1 as the -1 can be done at compile time while the + 1 has to be done at run time.

src/game/settings.c
```diff
@@ -42,7 +42,10 @@ static int32_t S_ReadUserSettingsATI()
         SetGameScreenSizeIdx(resolution_idx);
     }
 
-    FileRead(&GameSizer, sizeof(double), 1, fp);
+    {
+        float game_sizer;
+        FileRead(&game_sizer, sizeof(double), 1, fp);
+    }
```
oziphantom, line 48: if this is read to be discarded, possibly Seek over Read?

src/specific/display.c
oziphantom: so this was fixed up later, the fix-up could still do with some work on the second part.

src/game/traps/teeth_trap.c
```diff
@@ -12,6 +13,19 @@ BITE_INFO Teeth2B = { 71, 10, -1718, 1 };
 BITE_INFO Teeth3A = { -23, -10, -1718, 0 };
 BITE_INFO Teeth3B = { 71, -10, -1718, 1 };
 
+static void BaddieBiteEffect(ITEM_INFO *item, BITE_INFO *bite);
```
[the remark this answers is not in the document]
rr-: It's not but isn't it in good taste to have declarations for all functions?

src/global/types.h
rr-: Not really, this structure belongs to us only

src/game/invvars.c
```diff
@@ -0,0 +1,827 @@
...
+int16_t InvKeysObjects;
+int16_t InvKeysQtys[24] = {
+    1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+};
+
+INVENTORY_ITEM *InvKeysList[23] = {
+    &InvItemLeadBar,
+    &InvItemPuzzle1,
+    &InvItemPuzzle2,
+    &InvItemPuzzle4,
```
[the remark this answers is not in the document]
rr-: Good catch, fixed

src/game/invvars.c
```diff
@@ -0,0 +1,827 @@
...
+    NULL,
+    NULL,
+    NULL,
+    NULL,
+    NULL,
+};
+
+int16_t InvOptionCurrent;
+int16_t InvOptionObjects = 5;
+INVENTORY_ITEM *InvOptionList[] = {
```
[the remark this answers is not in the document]
rr-: Yes, fixed, I was playing with `InvItemGamma` (it crashes the game immediately so it doesn't make sense to keep it).

src/game/traps/teeth_trap.c
rr-: I'm not sure I understand your question but I moved it to this file, because it is the only usage of this function across the entire codebase.

src/game/settings.c
rr-: To reduce the logical scope of `resolution_idx` (it's not an optimization, only for readability not to inflate the function's state should more variables come into play). I plan on using this  idiom more in the future.

src/specific/display.c
rr-: I'm open to suggestions.

src/specific/display.c
rr-: Microoptimization that saves maybe a few nanoseconds at the cost of readability

src/specific/display.c
rr-: True, but I wanted this condition to mirror that of `SetNextGameScreenSize`'s.

src/game/settings.c
rr-: Excellent remark, I'll change it

rr-: Fixed

src/game/settings.c
```diff
@@ -33,8 +34,19 @@ static int32_t S_ReadUserSettingsATI()
...
+        int32_t resolution_idx;
+        FileRead(&resolution_idx, sizeof(int32_t), 1, fp);
+        CLAMP(resolution_idx, 0, RESOLUTIONS_SIZE - 1);
+        SetGameScreenSizeIdx(resolution_idx);
+    }
+
+    // Skip GameSizer from TombATI, which is no longer used in T1M.
+    // In the original game, it's expected to be 1.0 everywhere and changing it
+    // to any other value results in uninteresting window clipping anomalies.
+    FileSeek(fp, sizeof(double), SEEK_CUR);
```
thatguyfrombb, line 48: `SEEK_CUR` is undeclared

src/game/option.c
```diff
@@ -591,9 +550,7 @@ void DoDetailOptionHW(INVENTORY_ITEM *inv_item)
             max_row = DETAIL_HW_UI_BAR_SCALE;
         } else {
             static char tmp[10];
-            sprintf(
-                tmp, "%dx%d", AvailableResolutions[GameHiRes].width,
-                AvailableResolutions[GameHiRes].height);
+            sprintf(tmp, "%dx%d", GetScreenWidth(), GetScreenHeight());
```
thatguyfrombb, line 553: The in-game resolution changes and saves correctly, but it always shows "640x480" for me.

thatguyfrombb (changes requested):

src/game/settings.c
rr-: Fixed

src/game/option.c
rr-: Fixed

thatguyfrombb (approved):
It looks good to me for now.

PR #159, opened 2021-11-01
specific: fix unloading textures (closes #158)

### LINKED ISSUE #158: T1M bug: (master) Black screen after inventory with FPS counter on
Passport>New game>New game. F2, FPS counter displays. Press ESC twice and screen goes pretty much black.

It doesn't happen on latest release.

BlackScreenSmall

[Tomb1Main_BlackScreen.log](https://github.com/rr-/Tomb1Main/files/7449918/Tomb1Main_BlackScreen.log)

rr-:
The call to `HWR_DownloadTextures` seems to be some artifact from another platform – ATI3DCIF handles changing render settings on the fly just fine, as evidenced by the F3 button working in the OG TombATI.exe. Removing it makes the problem go away with seemingly no side effects. Since we now know the reason why the chunks that were removed in recent decompilation efforts existed, and those routines no longer seem to be necessary, I also removed the NOTE comments.

Also the reported issue demonstrates that it doesn't make sense to keep those prim strip renders out of the `if` blocks, since they're all black anyway, so I pulled them back into the `if` blocks like they were arranged in the original .exe.

### DIFF
--- a/src/global/vars.h
+++ b/src/global/vars.h
@@ -106,7 +106,6 @@ extern int32_t KeyChange;
 #define IsHardwareRenderer      VAR_U_(0x00463610, int32_t)
 #define IConfig                 VAR_U_(0x0045A23C, int32_t)
 #define RenderSettings          VAR_U_(0x0045A240, uint32_t)
-#define OldRenderSettings       VAR_U_(0x0045E960, uint32_t)
 #define ScanCodeNames           ARRAY_(0x00454A40, char*, [])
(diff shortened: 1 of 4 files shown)


thatguyfrombb (approved):
Code looks good and it fixes the issue. Haven't had the chance to try to break it yet, but it should be good to go.

rr-:
Thanks, I'll wait till it's tested :)

thatguyfrombb:
> Thanks, I'll wait till it's tested :)

I've been unable to break it :)

Issue #160, opened 2021-11-01
T1M bug: wrong key flashes


Trxyebeep:
The X here flashes as if it conflicts with another key:
![image](https://user-images.githubusercontent.com/38836940/139745808-d02217b0-faef-4573-907e-a336280301e6.png)

Which, on first look at the layout, it doesn't. However, X is the default key for one of the cheats, which is the reason for the flash.
When enabling cheats, I am able to remap the cheat buttons so the problem disappears.

The same problem happens when mapping `P` (because of pause), which actually can't be remapped (currently)

Issue #162, opened 2021-11-01
T1M bug: level skip cheat works in inventory.


Trxyebeep:
When pressing the level skip button in any inventory menu, seemingly nothing happens, but when the inventory is closed, the level is still skipped.

Issue #161, opened 2021-11-01
T1M bug: Pause key can't be remapped


Trxyebeep:
We are unable to remap the key that pops up the pause screen, which by default is `P`.
Side effect: P key flashing when mapped to other controls, as in #160

PR #168, opened 2021-11-02
Change Input and InputDB to an union; fix remapping issues

rr-:
Fixes issue #160, #161 and #162.

Simple adding of new `IN_` constants would require us to change `Input` size from `int32_t` to `int64_t`. This would break `GetDemoInput` that relies on `Input` being able to handle `(int32_t)-1`.  A better solution was to get rid of unused `IN_` constants, but this meant I had to hardcode TombATI key mapping in `GetDemoInput`. Since I dislike `Input` storing `-1` to signal end of demo playback, I refactored it to return `bool` and renamed it to `ProcessDemoInput` that I think is more fitting.

Since by now I already changed `IN_` handling significantly with regard to the original game, and we already have had inconsistent codebase with regard to usage of `CHK_ANY` / `CHK_ALL` macros vs. manual bitwise operations, I figured I could streamline the input handling as well while at it. So now the `INPUT_STATE` is an union that holds 1-bit states for each button. In places that need to check if any button was pressed, the code can query the `.any` union member. I tried to limit usages of `.any` to absolute minimum for code clarity.

Since we're using C, this usage of unions is safe:

>6.5.2.3 Structure and union members
>95) If the member used to read the contents of a union object is not the same as the member last used to store a value in the object, the appropriate part of the object representation of the value is reinterpreted as an object representation in the new type as described in 6.2.6 (a process sometimes called ‘‘type punning’’). This might be a trap representation.


thatguyfrombb (approved):
Looks good. The only thing that weirded me out while testing (not caused by this PR) is that the default layout still applies even when the user layout is selected.

rr-:
Oh this is OG behavior. We can address it in a different PR because it'd require more changes; for example, you can't rebind to arrows or escape in the user layout, we'd have to make sure that is possible for example before killing this fallback mechanism.

PR #169, opened 2021-11-02
Fix OG glitches (disabled by default)

rr-:


thatguyfrombb (commented):
QWOP I'm unable to test. Code looks good, testing does disable the sidestep bug.

Bad stuff: For wall bug I've been testing with this (https://youtu.be/H-JHRzFSJ6c?t=1318), which I can realiably do on master. I can't do it here, even with all fixes disabled. Also, jumping through the corners of a moving pushblock still works, but I'm not sure if any of this was supposed to fix it.

Just as a thought, it wouldn't surprise me if the "AF_" stuff that was replaced with frame_base had "incorrect" numbers.

EDIT: And I'll sink more time into this in a few hours.

rr-:
@thatguyfrombb you're right about `AF_*`. An easy way to verify this would be to dump all `LOG_INFO("%d", Anims[*].frame_base)`  after loading the level file, and compare it against `AF_*`.

thatguyfrombb:
> @thatguyfrombb you're right about `AF_*`. An easy way to verify this would be to dump all `LOG_INFO("%d", Anims[*].frame_base)` after loading the level file, and compare it against `AF_*`.

I was planning on doing so "Soon™" :)

thatguyfrombb:
So I took every anim in `LARA_ANIMATION` and logged its name and frame base for every retail level. All levels (including gold) contain the same data, except title and cutscenes. I haven't messed at all with guns/arms animations. As suspected, some `AF_` values don't match.

"wrong_anim_frames" contains only the data for animations where the values don't match, unless I've messed it up (could be, manual process). "lara_anims_1" is the whole dump for "Caves", same as any other retail level.

[wrong_anim_frames.txt](https://github.com/rr-/Tomb1Main/files/7467000/wrong_anim_frames.txt)
[lara_anims_1.txt](https://github.com/rr-/Tomb1Main/files/7467003/lara_anims_1.txt)

rr-:
@thatguyfrombb I can confirm, here's my results for Caves:

Animation            | Anims[*].frame_base | AF_* value
---                  | ---                 | ---
AA_VAULT12           | 759                 | 759
AA_VAULT34           | 614                 | 614
**AA_FASTFALL**      | **480**                 | **481**
AA_STOP              | 185                 | 185
AA_FALLDOWN          | 492                 | 492
AA_STOP_LEFT         | 58                  | 58
AA_STOP_RIGHT        | 74                  | 74
AA_HITWALLLEFT       | 800                 | 800
AA_HITWALLRIGHT      | 815                 | 815
AA_RUNSTEPUP_LEFT    | 837                 | 837
AA_RUNSTEPUP_RIGHT   | 830                 | 830
AA_WALKSTEPUP_LEFT   | 844                 | 844
AA_WALKSTEPUP_RIGHT  | 858                 | 858
AA_WALKSTEPD_LEFT    | 887                 | 887
AA_WALKSTEPD_RIGHT   | 874                 | 874
AA_BACKSTEPD_LEFT    | 899                 | 899
AA_BACKSTEPD_RIGHT   | 930                 | 930
AA_LANDFAR           | 358                 | 358
AA_GRABLEDGE         | 1493                | 1493
**AA_GRABLEDGE_OLD** | **480**                 | **621**
AA_SWIMGLIDE         | 1431                | 1431
AA_FALLBACK          | 1473                | 1473
**AA_HANG**          | **1493**                | **1514**
**AA_HANG_OLD**      | **483**                 | **642**
**AA_STARTHANG**     | **1493**                | **1505**
**AA_STARTHANG_OLD** | **483**                 | **634**
**AA_STOPHANG**      | **439**                 | **448**
AA_SLIDE             | 1133                | 1133
AA_SLIDEBACK         | 1677                | 1677
AA_TREAD             | 1736                | 1736
AA_SURFTREAD         | 1937                | 1937
AA_SURFDIVE          | 2041                | 2041
AA_SURFCLIMB         | 1849                | 1849
AA_JUMPIN            | 1895                | 1895
**AA_ROLL**          | **3855**                | **3857**
AA_RBALL_DEATH       | 3561                | 3561
AA_SPIKE_DEATH       | 3887                | 3887
AA_GRABLEDGEIN       | 3974                | 3974
AA_SPAZ_FORWARD      | 2666                | ?
AA_SPAZ_BACK         | 2701                | ?
AA_SPAZ_RIGHT        | 2736                | ?
AA_SPAZ_LEFT         | 2771                | ?

So I will revert this change, at least for now. Are we good to go with the rest of the PR?

rr-:
It's possible that the wall jump bug was caused by… the game hardcoding wrong `AF_` values :D

thatguyfrombb (approved):
Can confirm, other than QWOP which I can't do :)

rr-:
I'll trust Troye on his analysis which is consistent with the state of my knowledge about the nature of that bug. Thank you for testing

walkawayy:
Here's a list of all base and end frame numbers for Lara's animations from Caves. I can try to put these in order in `LARA_ANIMATION_FRAME`. Also add all 160 animations to `LARA_ANIMATION` in order. And probable name them to match WadTool.
[anim-frames-tr1.txt](https://github.com/rr-/Tomb1Main/files/7494740/anim-frames-tr1.txt)

walkawayy:
(random aside) Couldn't figure out why in `pickup.c` there was a check for frame_number equal to AF_PICKUP 3443 instead of 3466 when the animation actually ends. But it's because the pickup gets taken away at that frame because Lara is bent over. I'm thinking LARA_ANIMATION_FRAME should contain all the end frames like LARA_SHOTGUN_ANIMATION_FRAME does. But then there would need to be a magic number for 3443 like there is for some frame checks in `lara1gun.cpp`. For example there's `AF_SG_RECOIL + 12` and `AF_SG_UNDRAW + 20` checks in the shotgun file.
lara_pickup

Issue #129, opened 2021-10-15
OG bug: bat height


Richard-L:
Maybe just something to capture for later. Apparently hanging bats that aren't flying yet are positioned too high in the PC version, so that they're sometimes hidden by the ceiling.

TombATI/TR1Main
![](https://cdn.discordapp.com/attachments/811974815672959046/898684956824453140/unknown.png)

PS1
![](https://cdn.discordapp.com/attachments/811974815672959046/898685065922478120/unknown.png)

In fact this can sometimes create some artefacts in TR1Main. Bat coming through next to her left foot here:

https://user-images.githubusercontent.com/4934209/137557520-fbd8f07d-7b5c-4646-aa2b-90552aa2899d.mp4

oziphantom:
are they in the wrong position or is this a manifestation of the sorting bug?

rr-:
According to Troye they're in a wrong position.

PR #182, opened 2021-11-03
Fix bats position

thatguyfrombb:
See issue #129. In the retail levels, bats that are supposed to be at ceiling level are most of the times inside or above it, causing rendering problems.

This calculates the correct bat height for a bat at ceiling level and only moves bats that are above the calculated position. Moving all bats to the ceiling is a bad idea, Palace Midas already has bats that aren't intended to be at the ceiling. Gold levels don't have bats, so they're unaffected.

### DIFF
--- a/src/game/ai/bat.c
+++ b/src/game/ai/bat.c
@@ -99,45 +99,48 @@ void InitialiseBat(int16_t item_num)
     // Almost all of the bats in the OG levels are embedded in the ceiling.
     // This will move all bats up to the ceiling of their rooms and down
     // by the height of their hanging animation.
-    if (T1MConfig.fix_bats_position) {
-        ITEM_INFO *item;
-        FLOOR_INFO *floor;
-        int32_t x, y, z;
-        int16_t room_number, ceiling, old_anim, old_frame, bat_height;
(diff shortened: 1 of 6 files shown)


src/game/ai/bat.c
```diff
@@ -88,3 +91,53 @@ void BatControl(int16_t item_num)
...
 }
+
+void InitialiseBat(int16_t item_num)
+{
+    InitialiseCreature(item_num);
+
+    // Almost all of the bats in the OG levels are embedded in the ceiling.
+    // This will move all bats up to the ceiling of their rooms and down
+    // by the height of their hanging animation.
+    if (T1MConfig.fix_bats_position) {
```
rr-, line 102: I don't think we need to have this check in the config, but can we move the entire calculation below to a separate static function? `FixEmbeddedBatPosition`

thatguyfrombb: Fixed. Declared that function in the .c file since it has no use outside it.

rr-:
@thatguyfrombb one last thing before we merge, let's add mention this fix somewhere in the README :)

rr- (approved):

PR #183, opened 2021-11-03
Work towards 1.3.0

rr-:


rr-:
Actually I think I hate that `DynArray` thing. WDYT of replacing it back with a static array but of size `MAX_ROOMS`?

thatguyfrombb:
I just did some quick testing before looking at the code. Both Caves and City of Vilcabamba were crashing very, very early. It happened for about 3 or 4 runs, but it's not happening anymore.

Caves crashed after Lara starts looking at the snowy climb at the start (after the darts). Vilcabamba, after crossing the starting door, just a few tiles forward. I'll check the code now.

rr-:
@thatguyfrombb I'm pretty sure it's some uninitialized variable. Sadly I can't reproduce the crashes. Maybe bisecting the code would find the culprit faster?

src/dynarray.c
```diff
@@ -0,0 +1,73 @@
...
+#include <string.h>
+
+struct DYNARRAY {
+    char *data;
+    size_t elem_size;
+    size_t used;
+    size_t size;
+};
+
+DYNARRAY *DynArray_Create(size_t elem_size)
```
thatguyfrombb, line 15: Not having the option to start at a decent size will cause lots of unnecessary reallocs when used for some structures.

rr-: Like I mentioned in a previous comment I'm not sure I even want this, the code that uses this looks pretty dense. WDYT about making `RoomsToDraw`, the only struct that currently uses it, a static array of `MAX_ROOMS`?

thatguyfrombb: The interface is a bit heavy on the eyes, yes :) I like the idea of unlimited rooms, items and everything, but I don't like it making debugging harder for now. So I'd just make some big arrays for now and keep it for the future if there's a want/need.

src/game/option.c
```diff
@@ -70,6 +78,121 @@ static TEXTSTRING *CtrlText[2] = { 0 };
...
+        case DIK_W:            return "W";
+        case DIK_E:            return "E";
+        case DIK_R:            return "R";
+        case DIK_T:            return "T";
+        case DIK_Y:            return "Y";
+        case DIK_U:            return "U";
+        case DIK_I:            return "I";
+        case DIK_O:            return "O";
+        case DIK_P:            return "P";
+        case DIK_LBRACKET:     return "<";
```
thatguyfrombb, line 110: Microsoft says `DIK_LBRACKET` and `DIK_RBRACKET` are square brackets `[]`

src/game/option.c
```diff
@@ -70,6 +78,121 @@ static TEXTSTRING *CtrlText[2] = { 0 };
...
+        case DIK_G:            return "G";
+        case DIK_H:            return "H";
+        case DIK_J:            return "J";
+        case DIK_K:            return "K";
+        case DIK_L:            return "L";
+        case DIK_SEMICOLON:    return ";";
+        case DIK_APOSTROPHE:   return "\'";
+        case DIK_GRAVE:        return "`";
+        case DIK_LSHIFT:       return "SHIFT";
+        case DIK_BACKSLASH:    return "#";
```
thatguyfrombb, line 127: DIK_BACKSLASH would be `\`

src/game/option.c
```diff
@@ -47,16 +47,24 @@ typedef enum COMPASS_TEXT {
...
-    DETAIL_HW_UI_TEXT_SCALE = 4,
-    DETAIL_HW_UI_BAR_SCALE = 5,
+    DETAIL_HW_PERSPECTIVE = 0,
+    DETAIL_HW_BILINEAR = 1,
+    DETAIL_HW_UI_TEXT_SCALE = 2,
+    DETAIL_HW_UI_BAR_SCALE = 3,
+    DETAIL_HW_TITLE = 4,
+    DETAIL_HW_TITLE_BORDER = 5,
     DETAIL_HW_RESOLUTION = 6,
     DETAIL_HW_NUMBER_OF = 7,
```
thatguyfrombb, line 57: Pressing down on `UI bar scale` takes you to the title (Select Detail), pressing down again makes the border disappear and down again gets you to the resolution.

thatguyfrombb (changes requested):
Sadly, I haven't found the bug in the code review. Will keep at it, though.

src/dynarray.c
rr-: Changed back, and made sure we're not running into buffer overflows.

thatguyfrombb:
Oops, that's a mixup. I reviewed the code with DynArray in :)

src/game/option.c
rr-: That's the value in the OG, and binding to `[` correctly displays `[`.

src/game/option.c
rr-: Fixed

src/specific/init.c
```diff
@@ -91,24 +92,45 @@ void S_ExitSystem(const char *message)
...
-    GameAllocMemFree = GameMemorySize;
+    GameAllocMemFree = MALLOC_SIZE;
     GameAllocMemUsed = 0;
 }
 
+void game_malloc_shutdown()
+{
+    if (GameMemoryPointer) {
+        free(GameMemoryPointer);
+    }
```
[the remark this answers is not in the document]
rr-: Fixed

src/global/vars_platform.h
```diff
@@ -1,47 +1,40 @@
...
+extern float DDrawSurfaceMinY;
+extern float DDrawSurfaceMaxX;
+extern float DDrawSurfaceMaxY;
+extern int32_t DDrawSurfaceWidth;
+extern int32_t DDrawSurfaceHeight;
+extern LPDIRECTDRAWSURFACE Surface1;
+extern LPDIRECTDRAWSURFACE Surface2;
+extern LPDIRECTDRAWSURFACE Surface3;
+extern LPDIRECTDRAWSURFACE Surface4;
+extern LPDIRECTDRAWSURFACE TextureSurfaces[];
```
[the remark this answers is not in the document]
rr-: Actually I'd rather not duplicate it here, it's sufficient to specify it at the definition level.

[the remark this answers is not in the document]
rr-: After some thought it makes sense to specify it in the header as well so that the users do not have to search for the specification of these variables in its implementation file. Added

src/global/vars_platform.h
```diff
@@ -1,47 +1,40 @@
...
+extern void *Surface1DrawPtr;
+extern void *Surface2DrawPtr;
+
+extern HMODULE HATI3DCIFModule;
+extern C3D_HRC ATIRenderContext;
+extern C3D_3DCIFINFO ATIInfo;
+extern C3D_HTX ATITextureMap[];
+extern C3D_HTXPAL ATITexturePalette;
+extern C3D_PALETTENTRY ATIPalette[];
```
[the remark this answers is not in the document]
rr-: Added

src/specific/hwr.c
```diff
@@ -23,9 +23,15 @@ typedef struct HWR_LIGHTNING {
...
-#define HWR_LightningCount VAR_U_(0x00463618, int32_t)
-
+HWR_LIGHTNING HWR_LightningTable[100];
+int32_t HWR_LightningCount = 0;
+
+static bool HWR_IsPaletteActive;
+static bool HWR_IsRendering;
+static bool HWR_IsRenderingOld;
+static bool HWR_IsTextureMode;
+static int32_t HWR_SelectedTexture;
```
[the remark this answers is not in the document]
rr-: Not sure, but it won't hurt. Added

src/game/option.c
rr-: Fixed + allowed changing the resolution from within the game, but it requires closing the inventory to apply.

rr-:
Another approach to debug the crashes would be to attach IDA to the game whenever it starts, try loading Caves, and if it crashes, we'll have our stacktrace.

thatguyfrombb:
The crash apparently came from the `realloc()` inside `DynArray_Append()`, so that's not an issue anymore. Current issue is half the runs I get very bad performance (until it fixes itself). I thought it was the DynArray calling realloc during gameplay, but that can't be the case anymore.

I'll keep going at this tomorrow.

thatguyfrombb:
I just got the performance issue on the shotgun PR #184 so whatever causes it has to be on master (which never gave me issues before, but oh well). So, this one should be ready to go.

thatguyfrombb (changes requested):

src/game/collide.c
```diff
@@ -455,8 +456,10 @@ int32_t CollideStaticObjects(
 void GetNearByRooms(
     int32_t x, int32_t y, int32_t z, int32_t r, int32_t h, int16_t room_num)
 {
-    RoomsToDraw[0] = room_num;
-    RoomsToDrawNum = 1;
+    RoomsToDrawCount = 0;
+    if (RoomsToDrawCount + 1 < MAX_ROOMS_TO_DRAW) {
```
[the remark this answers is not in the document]
rr-: I wanted to be consistent across appends ^^

src/global/vars.h
```diff
@@ -1,130 +1,60 @@
 #ifndef T1M_GLOBAL_VARS_H
 #define T1M_GLOBAL_VARS_H
 
+#include "dynarray.h"
```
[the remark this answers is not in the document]
rr-: Yes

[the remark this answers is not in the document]
rr-: Fixed

scripts/ida_import.py
```diff
@@ -57,35 +56,6 @@ def import_variables() -> None:
...
-            elif flags == "x":
-                idc.set_color(offset, idc.CIC_FUNC, 0xD8D8D8)
-            else:
-                idc.set_color(offset, idc.CIC_FUNC, idc.DEFCOLOR)
-    print("    done")
-
-
 if __name__ == "__main__":
     import_types()
     import_variables()
```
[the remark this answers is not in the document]
rr-: Good catch; fixed

thatguyfrombb (approved):

PR #184, opened 2021-11-04
lara: fix shotgun shooting when target out of sight.

### LINKED ISSUE #51: OG bugfix: Shotgun ammo waste when aiming
When Lara is aiming an enemy with Shotgun and it goes out of sight, Lara keeps shooting forward wasting ammo. With other weapons she makes her guns idle and waiting for an enemy to appear again. It should be fixed I think.

walkawayy:
Closes #51


rr-:
@walkawayy can we make a setting for it in T1MConfig and mention the fix in the readme?

walkawayy:
Ya sure. I guess it changes the original game behavior so it should be a setting. Should I default it to false?

I also wanted to look into the shotgun frames, so I might make a separate issue and fix for that when I get to it. When Lara doesn't have a target, she shoots the shotgun faster than when she's locked on to a target. It's like she's skipping some frames. I can probably tie both fixes to a generic shotgun fix setting in the JSON.

I'll look at how you implemented some of the other optional fixes and add a check.

rr-:
Sounds fair about the other check and having a single trigger for both of shotgun issues. I think we should however set the default setting to true.

Please ping me once you're done with the changes :)

walkawayy:
Here's a video comparison. First clip is with the fix disabled (original TR behavior), and the second clip is with the fix enabled. Both clips I was holding down action the whole time to keep the target locked. Some animation frames are skipped in the original when firing without a target, and she will shoot at nothing while not locked on to a target.

https://streamable.com/619hgu

cfg/Tomb1Main.json5
```diff
@@ -24,6 +24,9 @@
     // Draws flame when firing a shotgun, like for other guns.
     "enable_shotgun_flash": true,
 
+    // Shotgun no longer fires when a locked target moves out of Lara's sight, like for other guns.
+    "enable_shotgun_target_fix": false,
```
thatguyfrombb, line 28: Should default to true here.

src/game/lara1gun.c
```diff
@@ -164,27 +164,22 @@ void AnimateShotgun()
...
-        if (ani == AF_SG_AIM && Input.action) {
-            ani++;
-        } else if (ani > AF_SG_AIM && ani < AF_SG_DRAW) {
-            ani++;
-            if (ani == AF_SG_DRAW) {
-                ani = Input.action ? AF_SG_RECOIL : AF_SG_UNAIM;
-            }
-        } else {
-            if (ani == AF_SG_RECOIL) {
+        if (T1MConfig.enable_shotgun_target_fix && Input.action
```
thatguyfrombb, line 167: Do we want the `Input.Action` check here? The original checks `Input.Action` after this, not always for true.

src/game/lara1gun.c
```diff
@@ -164,27 +164,22 @@ void AnimateShotgun()
...
+            && !Lara.target) {
+            if (ani >= AF_SG_AIM && ani < AF_SG_DRAW) {
+                ani++;
+                if (ani == AF_SG_DRAW) {
+                    ani = AF_SG_RECOIL;
```
thatguyfrombb, line 172: Case where `Input.Action` is checked for true and false. Here, the false case seems unhandled.

src/game/lara1gun.c
```diff
@@ -164,27 +164,22 @@ void AnimateShotgun()
...
+                }
+            } else if (ani == AF_SG_RECOIL) {
                 if (Input.action) {
```
thatguyfrombb, line 175: `Input.Action` is checked here for both true and false.

src/game/lara1gun.c
```diff
@@ -164,27 +164,22 @@ void AnimateShotgun()
...
-                    ani = AF_SG_UNAIM;
                 }
             } else if (ani > AF_SG_RECOIL && ani < AF_SG_UNDRAW) {
                 ani++;
-                if (ani == AF_SG_RECOIL + 12 + 1) {
-                    ani = AF_SG_AIM;
-                } else if (ani == AF_SG_UNDRAW) {
-                    ani = AF_SG_UNAIM;
+                if (ani == AF_SG_UNDRAW) {
+                    ani = AF_SG_RECOIL;
```
thatguyfrombb, line 182: In this case, OG goes to `AF_SG_UNAIM`. Why the change?

thatguyfrombb (dismissed):
Current PR code causes the shotgun animation to stop and resume when I let go of Action when there's no target and the fix is enabled.

A small example of this is visible in your video, around 0:26. Last shot against the wall isn't followed by a shotgun pump. Then, before shooting at the wolf, Lara pumps the shotgun.

cfg/Tomb1Main.json5
walkawayy: Ah sorry. Fixed.

src/game/lara1gun.c
walkawayy: I tried to make it kind of match the pistols in lara2gun.c AnimatePistols.
`if (Lara.right_arm.lock || (Input.action && !Lara.target))`

src/game/lara1gun.c
walkawayy: I don't see where it's checked for false? Isn't it actually double checking for true again which it already has to be based on `if (T1MConfig.enable_shotgun_target_fix && Input.action && !Lara.target)`? I might be misunderstanding something though since I'm new to the code.

src/game/lara1gun.c
walkawayy: So I changed the fixed to be more like the `if (Lara.left_arm.lock)` part of the code, and less like what is now `else if (!T1MConfig.enable_shotgun_target_fix)`. I definitely do notice what you are saying now about the unhandled case where you stop holding action and Lara's animation gets stuck at whatever point you were last holding action at. I'll do some more tweaking.

Originally I thought they devs had meant to implement a third case where Lara is meant to fire quicker with subsequent shots after the first shot at a locked target. But, instead it was firing quicker when there is no locked target. However, I decided to go with the original behavior which is a full animation for every shot, and fix the fast shooting at no target and the shooting when a locked target is out of her sight.

src/game/lara1gun.c
thatguyfrombb: I was comparing to the code for `if (!T1MConfig.enable_shotgun_target_fix)`, which doesn't have a big `if` for `Input.Action` and looks like:
```
                if (ani == AF_SG_RECOIL) {
                    if (Input.action) {
                        FireShotgun();
                        ani++;
                    } else {
                        ani = AF_SG_UNAIM;
                    }
                }
```
When we get there in the old code, we're not inside any `if` that checks `Input.Action`, so the `else` case can run too. Guns/arms anim code is not exactly my forte, so I'm a bit confused and not sure right now if we need that :)

src/game/lara1gun.c
thatguyfrombb: Now I understand it better :) I'll give all of this a re-read tomorrow and really try to get how it works, my head is done for today. Of course, feel free to push any changes.

cfg/Tomb1Main.json5
thatguyfrombb: Fix won't be here until you push it, so I'm not resolving this one yet.

rr-:
LMK @walkawayy when the branch is ready for a re-review

walkawayy:
> 
> 
> LMK @walkawayy when the branch is ready for a re-review

No rush but it's ready now. I just merged the latest master and pushed.

walkawayy:
I made the mistake of amending my first few fixes so now the code conversations are outdated since I had to force push the new commits. Now I realize I can just push up new commits and we can squash it when it's approved.I'm going to close the existing conversations because they are commented under outdated code because of the force push.

bin/cfg/Tomb1Main.json5
```diff
@@ -24,6 +24,9 @@
     // Draws flame when firing a shotgun, like for other guns.
     "enable_shotgun_flash": true,
 
+    // Shotgun no longer fires when a locked target moves out of Lara's sight, like for other guns.
+    "enable_shotgun_target_fix": true,
```
rr-, line 28: I think we should call this `fix_shortgun_targeting` to be consistent with the other OG fix configuration switches.

rr- (commented):
I tested it and it's working fine, but the function is pretty hard to follow. What do you think about making merging branches from both conditions?
The differences when shotgun targeting fix is enabled are:

- The entire `g_Lara.target` branch
- The `ani > AF_SG_RECOIL && ani < AF_SG_UNDRAW` condition

Otherwise they seem to be the same.

Suggested patch that merges these conditions:

```patch
diff --git a/src/game/lara1gun.c b/src/game/lara1gun.c
index 113bbb62..ddd5f56e 100644
--- a/src/game/lara1gun.c
+++ b/src/game/lara1gun.c
@@ -168,93 +168,64 @@ void AnimateShotgun()
                 ani = AF_SG_AIM;
             }
         }
+    } else if (g_Config.fix_shotgun_targeting && g_Lara.target) {
+        if (ani == AF_SG_RECOIL) {
+            ani = AF_SG_UNAIM;
+            ani++;
+        } else if (ani > AF_SG_RECOIL && ani < AF_SG_UNDRAW) {
+            ani++;
+            if (ani == AF_SG_RECOIL + 10) {
+                Sound_Effect(SFX_LARA_RELOAD, &g_LaraItem->pos, SPM_NORMAL);
+            } else if (ani == AF_SG_RECOIL + 16) {
+                ani = AF_SG_AIM;
+            } else if (ani == AF_SG_UNDRAW) {
+                ani = AF_SG_UNAIM;
+            }
+        } else if (ani >= AF_SG_UNAIM && ani < AF_SG_END) {
+            ani++;
+            if (ani == AF_SG_END) {
+                ani = AF_SG_AIM;
+            }
+        } else if (ani > AF_SG_AIM && ani <= AF_SG_DRAW) {
+            ani--;
+        }
     } else {
-        if (g_Config.fix_shotgun_targeting) {
-            if (g_Lara.target) {
-                if (ani == AF_SG_RECOIL) {
-                    ani = AF_SG_UNAIM;
-                    ani++;
-                } else if (ani > AF_SG_RECOIL && ani < AF_SG_UNDRAW) {
-                    ani++;
-                    if (ani == AF_SG_RECOIL + 10) {
-                        Sound_Effect(
-                            SFX_LARA_RELOAD, &g_LaraItem->pos, SPM_NORMAL);
-                    } else if (ani == AF_SG_RECOIL + 16) {
-                        ani = AF_SG_AIM;
-                    } else if (ani == AF_SG_UNDRAW) {
-                        ani = AF_SG_UNAIM;
-                    }
-                } else if (ani >= AF_SG_UNAIM && ani < AF_SG_END) {
-                    ani++;
-                    if (ani == AF_SG_END) {
-                        ani = AF_SG_AIM;
-                    }
-                } else if (ani > AF_SG_AIM && ani <= AF_SG_DRAW) {
-                    ani--;
-                }
-            } else {
-                if (g_Input.action && ani == AF_SG_AIM) {
-                    ani++;
-                } else if (ani > AF_SG_AIM && ani < AF_SG_DRAW) {
-                    ani++;
-                    if (ani == AF_SG_DRAW) {
-                        ani = g_Input.action ? AF_SG_RECOIL : AF_SG_UNAIM;
-                    }
-                } else if (ani == AF_SG_RECOIL) {
-                    if (g_Input.action) {
-                        FireShotgun();
-                        ani++;
-                    } else {
-                        ani = AF_SG_UNAIM;
-                    }
-                } else if (ani > AF_SG_RECOIL && ani < AF_SG_UNDRAW) {
-                    ani++;
-                    if (ani == AF_SG_RECOIL + 10) {
-                        Sound_Effect(
-                            SFX_LARA_RELOAD, &g_LaraItem->pos, SPM_NORMAL);
-                    } else if (ani == AF_SG_UNDRAW) {
-                        ani = g_Input.action ? AF_SG_RECOIL : AF_SG_UNAIM;
-                    }
-                } else if (ani >= AF_SG_UNAIM && ani < AF_SG_END) {
-                    ani++;
-                    if (ani == AF_SG_END) {
-                        ani = AF_SG_AIM;
-                    }
-                }
+        if (g_Input.action && ani == AF_SG_AIM) {
+            ani++;
+        } else if (ani > AF_SG_AIM && ani < AF_SG_DRAW) {
+            ani++;
+            if (ani == AF_SG_DRAW) {
+                ani = g_Input.action ? AF_SG_RECOIL : AF_SG_UNAIM;
             }
-        } else if (!g_Config.fix_shotgun_targeting) {
-            if (ani == AF_SG_AIM && g_Input.action) {
-                ani++;
-            } else if (ani > AF_SG_AIM && ani < AF_SG_DRAW) {
+        } else if (ani == AF_SG_RECOIL) {
+            if (g_Input.action) {
+                FireShotgun();
                 ani++;
-                if (ani == AF_SG_DRAW) {
+            } else {
+                ani = AF_SG_UNAIM;
+            }
+        } else if (ani > AF_SG_RECOIL && ani < AF_SG_UNDRAW) {
+            ani++;
+            if (g_Config.fix_shotgun_targeting) {
+                if (ani == AF_SG_RECOIL + 10) {
+                    Sound_Effect(SFX_LARA_RELOAD, &g_LaraItem->pos, SPM_NORMAL);
+                } else if (ani == AF_SG_UNDRAW) {
                     ani = g_Input.action ? AF_SG_RECOIL : AF_SG_UNAIM;
                 }
             } else {
-                if (ani == AF_SG_RECOIL) {
-                    if (g_Input.action) {
-                        FireShotgun();
-                        ani++;
-                    } else {
-                        ani = AF_SG_UNAIM;
-                    }
-                } else if (ani > AF_SG_RECOIL && ani < AF_SG_UNDRAW) {
-                    ani++;
-                    if (ani == AF_SG_RECOIL + 12 + 1) {
-                        ani = AF_SG_AIM;
-                    } else if (ani == AF_SG_UNDRAW) {
-                        ani = AF_SG_UNAIM;
-                    } else if (ani == AF_SG_RECOIL + 10) {
-                        Sound_Effect(
-                            SFX_LARA_RELOAD, &g_LaraItem->pos, SPM_NORMAL);
-                    }
-                } else if (ani >= AF_SG_UNAIM && ani < AF_SG_END) {
-                    ani++;
-                    if (ani == AF_SG_END) {
-                        ani = AF_SG_AIM;
-                    }
+                if (ani == AF_SG_RECOIL + 12 + 1) {
+                    ani = AF_SG_AIM;
+                } else if (ani == AF_SG_UNDRAW) {
+                    ani = AF_SG_UNAIM;
+                } else if (ani == AF_SG_RECOIL + 10) {
+                    Sound_Effect(SFX_LARA_RELOAD, &g_LaraItem->pos, SPM_NORMAL);
                 }
             }
+        } else if (ani >= AF_SG_UNAIM && ani < AF_SG_END) {
+            ani++;
+            if (ani == AF_SG_END) {
+                ani = AF_SG_AIM;
+            }
         }
     }
```

As a side bonus we get reduced indentation.

rr-:
Also I'd move `fix_shotgun_targeting` near other fixes.

rr-:
Also I'd move `fix_shotgun_targeting` near other fixes.

oziphantom:
if (ani == AF_SG_RECOIL) {
+            ani = AF_SG_UNAIM;
+            ani++; <- why?

oziphantom:
after staring out this code for a while, I think the only real way to clean it up is to pull all the ani > < == checks out to the base level which then form a "state" and then put all the logic for each state inside them. this way there is no global "fix, target fix + target, no target" splits. just tiny code changes in the 2 states that need it. And the copy paste factor is greatly removed. The code will be easier to follow as the animations can be checked in order which will allow one to follow the flow better, I think... I could just be hallucinating at this point, I did watch Last night in Soho earlier and it was more coherent than this Core special.

walkawayy:
> I tested it and it's working fine, but the function is pretty hard to follow. What do you think about making merging branches from both conditions? The differences when shotgun targeting fix is enabled are:
> 
>     * The entire `g_Lara.target` branch
> 
>     * The `ani > AF_SG_RECOIL && ani < AF_SG_UNDRAW` condition
> 
> 
> Otherwise they seem to be the same.
> 
> Suggested patch that merges these conditions:
> ...

This was my first PR, so I wasn't sure if we wanted to keep all the fixes separate at that point and preserve original code. I'm ok with condensing. The code is really specific and confusing because it relies on exact frame numbers checks. I had to do a lot of back and forth with WAD Tool to view animations to see what was going on lol. I'll condense them like you suggested and test the behavior again to make sure it still works as expected.

walkawayy:
> if (ani == AF_SG_RECOIL) { ani = AF_SG_UNAIM; ani++; <- why?

Current code is confusing is why lol. So basically in its current form, the `if (g_Lara.left_arm.lock)` case covers when Lara is actively aiming at locked target that's in her sight. The big else case covers everything else.

If the target moves out of sight but the player still has the action button held down to shoot, and she reaches the animation frame where Lara normally fires, this makes her not shoot and instead go to the `AF_SG_UNAIM`.

```
if (g_Config.fix_shotgun_targeting) {
    if (g_Lara.target) {
        if (ani == AF_SG_RECOIL) {
```

oziphantom:
> > if (ani == AF_SG_RECOIL) { ani = AF_SG_UNAIM; ani++; <- why?
> 
> Current code is confusing is why lol. So basically in its current form, the `if (g_Lara.left_arm.lock)` case covers when Lara is actively aiming at a locked target that's in her sight. The big else case covers everything else.
> 
> If the target moves out of sight but the player still has the action button held down to shoot, and she reaches the animation frame where Lara normally fires, this makes her not shoot and instead go to the `AF_SG_UNAIM`.
> 
> ```
> if (g_Config.fix_shotgun_targeting) {
>     if (g_Lara.target) {
>         if (ani == AF_SG_RECOIL) {
> ```

I feel you have missed my point. My point was why does one do Ani++ right after explicitly setting it to a value. 
i.e
~~~
ani = AF_SG_UNAIM;
ani++;
~~~
is the same as 
~~~
ani = AF_SG_UNAIM + 1;
~~~
however I also feel that the `ani++` is an error looking at the original code.

walkawayy:
> > > if (ani == AF_SG_RECOIL) { ani = AF_SG_UNAIM; ani++; <- why?
> > 
> > 
> > Current code is confusing is why lol. So basically in its current form, the `if (g_Lara.left_arm.lock)` case covers when Lara is actively aiming at a locked target that's in her sight. The big else case covers everything else.
> > If the target moves out of sight but the player still has the action button held down to shoot, and she reaches the animation frame where Lara normally fires, this makes her not shoot and instead go to the `AF_SG_UNAIM`.
> > ```
> > if (g_Config.fix_shotgun_targeting) {
> >     if (g_Lara.target) {
> >         if (ani == AF_SG_RECOIL) {
> > ```
> 
> I feel you have missed my point. My point was why does one do Ani++ right after explicitly setting it to a value. i.e
> 
> ```
> ani = AF_SG_UNAIM;
> ani++;
> ```
> 
> is the same as
> 
> ```
> ani = AF_SG_UNAIM + 1;
> ```
> 
> however I also feel that the `ani++` is an error looking at the original code.

Ah I think you are right about that. I think `ani++` in that case should be removed. It's a very edge case, so I probably didn't notice the one frame difference.

walkawayy:
Let me know what you guys think about the pics above. It's only a one frame difference.

Also just debugging some of the cleaned up version because it's buggy rn. Lara stops shooting after facing away from a target and then can't shoot without a target. I'll push when it's fixed.

walkawayy:
Everything seems to be working after I rearranged some of the cleanup. But, it seems like something else changed. Even on the previous commit before the `AnimateShotgun`, with the targeting fix disabled, Lara maintains her target and the camera locks on the target even if it dies or is out of sight. She will still shoot when not aimed at it, but the camera locks and she can't acquire a new target until action is let go. Before, she would lose the target when her arm was no longer locked.

rr- (approved):
Tested and the code LGTM. Can we merge it?

walkawayy:
Ya I think it's good to go!

I just double checked TombATI regarding my [comment](https://github.com/rr-/Tomb1Main/pull/184#issuecomment-995223448) and we are actually working the same. I think I just got confused because the target's healthbar is now visible.

PR #185, opened 2021-11-05
inv: add T1M version in the corner

rr-:

PR #186, opened 2021-11-05
Audio refactor

rr-:
@thatguyfrombb please note that #185 comes before this PR as this is not a part of the 1.3.0 scope.

This pull request refactors specific/sndpc.c, game/mmsound.c and game/sound.c into game/music.c (game logic), specific/music.c (winmm layer), game/sound.c (game logic), specific/sound.h (directsound). It also brings neat naming to the refactored functions. The game code no longer calls platform-specific code, all the specific/ calls happen from either game/sound.c or game/music.c.

It also introduces a concept of private state, as a sort of namespace, to reduce amount of global variables and the likelihood of generating a conflict with another module (especially `globals/vars.h`):

```
static struct {
    blablabla
} S = { 0 };
```

Some of the global variables are removed or inlined. Setting the volume got simplified. I'm not very happy about the calculations of the target volume happening in game/sound.c rather than specific/sound.c, but the idea of master volume is a part of the game logic so I suppose it needs to stay there.


src/specific/music.c
```diff
@@ -43,14 +43,14 @@ int32_t MusicInit()
     return 1;
 }
 
-void S_MusicVolume(int16_t volume)
+void S_Music_AdjustVolume(int16_t volume)
```
oziphantom, line 46: Adjust implies a relative operation, this is an absolute function. S_Music_SetVolume

src/game/music.c
```diff
@@ -5,7 +5,11 @@
 #include "global/vars.h"
 #include "specific/music.h"
 
-static int16_t MusicTrackLooped = 0;
+static struct {
+    bool loop;
+    int16_t track;
+    int16_t track_looped;
+} S = { 0 };
```
oziphantom, line 12: a single letter struct instance is asking for trouble.

src/specific/music.c
rr-: Point, fixed

src/game/music.c
rr-: I saw in TEN people do `g_BlaBla` to make globals stand out even more, I like it. Especially that it would remove the clash with `FunctionNames`. Maybe instead of going with this kind of structure, we could adopt this `g_` idiom with T1M, but designate module-local globals somehow? I was thinking of `pg` as in `private global` but that probably can be confused with `pointer`. Or `lg` `local global`. WDYT?

(Related reading: https://www.joelonsoftware.com/2005/05/11/making-wrong-code-look-wrong/ )

oziphantom: warning : I'm one of the last few card carrying member of the "Hungarian notation was good" club ;) However I accept that advancements in colour coding, type on hover and some people also have editors that put the type inline in the code view have made it mostly redundant.

I fell that 'l' is still mostly bad because of iIlL ( all 4 are different ). Another trap with Hungarian is to talk about what it is, rather than what it does. I.E. yes MS `lpstrMyFile` is great but they kind of all are, `pathMyFile` is better as it tells me the string should be a path to file rather than just a filename, is more useful info.

To this end while everything is global, and we are doing some nice smoke and mirrors linker tricks to pretend it's not global, conceptually it isn't. So gs_Blah or sg_Blah to denote it as global static or static global is "what it is" but we are using it more as a "this file is now a class and this is the data only for this "class/type", at which borrowing the C++ nomenclature of m_Blah might be nicer. 
Giving us
g_Blah ; true global
m_Blah ; data only for this file
blah ; only for this function
or perhaps
f_Blah for file scope

rr-: I like g_Blah and m_Blah for file scope. Let's go with that.
And I agree that hiding data behind a static struct is silly, but I'm glad I tried it.

rr-: Done, went with `m_` after all

oziphantom (approved):

rr-:
(Fixed small `#include` conflict with the newest master)

thatguyfrombb (approved):

PR #187, opened 2021-11-06
Text refactor

rr-:
Similar refactors to what we have in #186 . One curiosity is the introduction of `on_remove` hook in `TEXSTRING*` to achieve inversion of control with regards to global pointer invalidation for static textstrings (details in the commit message).

### DIFF
--- a/src/game/overlay.c
+++ b/src/game/overlay.c
@@ -378,15 +378,16 @@ void Overlay_DrawPickups()
 
 void Overlay_DrawFPSInfo()
 {
-    static char fps_buf[20];
     static int32_t elapsed = 0;
 
     if (T1MConfig.render_flags.fps_counter) {
         if (ClockGetMS() - elapsed >= 1000) {
             if (FPSText) {
+                char fps_buf[20];
                 sprintf(fps_buf, "%d FPS", FPSCounter);
                 Text_ChangeText(FPSText, fps_buf);
             } else {
+                char fps_buf[20];
                 sprintf(fps_buf, "? FPS");
                 FPSText = Text_Create(10, 30, fps_buf);
             }
--- a/src/game/overlay.c
+++ b/src/game/overlay.c
@@ -80,6 +80,8 @@ static RGB888 ColorBarMap[][COLOR_STEPS] = {
 static void Overlay_GetBarLocation(
     int8_t bar_location, int32_t width, int32_t height, int32_t *x, int32_t *y);
 static void Overlay_DrawBar(int32_t value, int32_t value_max, int32_t bar_type);
+static void Overlay_OnAmmoTextRemoval(const TEXTSTRING *textstring);
+static void Overlay_OnFPSTextRemoval(const TEXTSTRING *textstring);
 
 static void Overlay_GetBarLocation(
     int8_t bar_location, int32_t width, int32_t height, int32_t *x, int32_t *y)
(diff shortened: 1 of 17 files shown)


thatguyfrombb:
Error detected in testing phase: Title screen: Enable FPS counter. Start new game. Exit game. FPS counter doesn't display, but is still active as it requires pressing F2 twice to get it showing again.

thatguyfrombb (changes requested):
I'm not sure what causes the error with the FPS counter :(

rr-:
Yes, otherwise the project wouldn't compile. And we do not rely on the order within bitfields as they're not portable in this manner.

rr-:
> Error detected in testing phase: Title screen: Enable FPS counter. Start new game. Exit game. FPS counter doesn't display, but is still active as it requires pressing F2 twice to get it showing again.
> 
> Everything else seems to work OK.

@thatguyfrombb fixed. Now to give some background…

The main problem with the previous implementation of text was this:
1. Code X calls `Text_Create()` / `T_Print()`, caches its pointer internally
2. Code Y calls `Text_RemoveAll` / `T_RemoveAllPrints`.
3. Code Z calls `Text_Create()` again.
Because `TEXTSTRING` is a static array, code Z has a chance of obtaining the same pointer as code X. Then both of them write to the same `TEXTSTRING`. I mitigated this problem by adding the `on_remove` observer and giving code X the ability to detach its cached pointer. However, those `on_remove` observers are called in `Text_RemoveAll` / `T_RemoveAllPrints` rather than `Text_Init`. 

The issue was that one place still used `Text_Init` to reset the `TEXTSTRING` array, thus never executing those observers. It resulted in exactly the scenario described above:

1. We start the game.
2. FPS text is created via `Text_Create` and cached internally.
3. We start Caves, we remove all texts via `Text_RemoveAll`.
4. `FPSText` observes `on_remove` and is reset, which is why it spells `? FPS` right after quitting the inventory.
5. We quit the game to main menu.
6. The code removes all texts via `Text_Init` (instead of `Text_RemoveAll`).
7. `FPSText` still contains old pointer and does not realize it should call `Text_Create` again.
8. Version text is created via `Text_Create` (among others). It obtains a pointer to the same index within the `TEXSTRING` array as FPSText.
9. Version tries to write `T1M 1.3.0` once.
10. FPS text now updates the same `TEXTSTRING` every second.

If you look closely, you'll see that indeed FPS counter still is displayed, but in the version slot.

This is now fixed by calling `Text_Init` in `S_InitialiseSystem()` like all other components, and `Text_RemoveAll` in other places. I hope you enjoyed this little story :)

src/game/overlay.c
```diff
@@ -17,8 +17,12 @@
...
 #define MAX_PICKUPS 16
 
-static int16_t BarOffsetY[6] = { 0 };
-static DISPLAYPU Pickups[MAX_PICKUPS] = { 0 };
+static struct {
+    TEXTSTRING *ammo_text;
+    TEXTSTRING *fps_text;
+    int16_t bar_offset_y[6];
+    DISPLAYPU pickups[MAX_PICKUPS];
+} S = { 0 };
```
oziphantom, line 25: are you not going to convert the S -> m_ as well?

src/global/types.h
```diff
@@ -1554,387 +1515,378 @@ typedef struct SAVEGAME_INFO {
...
-    /* 0032 */ int32_t scale_v;
-    /* 0034 */ char *string;
-    /* 0038 end */
+    union {
+        uint32_t all;
+        struct {
+            uint32_t active : 1;
+            uint32_t flash : 1;
+            uint32_t rotate_h : 1;
+            uint32_t centre_h : 1;
```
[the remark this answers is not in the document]
rr-: Yes, it's unused and unimplemented.

src/game/overlay.c
rr-: Good point, fixed

oziphantom (approved):

thatguyfrombb (approved):
Testing phase passed. Code review passed.

PR #189, opened 2021-11-06
build: support local branches versioning

rr-:
This improves the version output when compiling
on a branch that is not master.

Let's say we're on a branch called `audio_refactor`,
12 commits ahead of `master`.

Before the version would say:

    T1M 1.3.0

After this commit, the version would say:

    T1M 1.3.0-12-gf65378c

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ WINDRES=i686-w64-mingw32-windres
 LDFLAGS=-ldbghelp -lwinmm -ldsound -lddraw -ldinput8 -ldxguid
 CFLAGS=-Wall -Isrc -DVERSION='"T1M ${VERSION}"'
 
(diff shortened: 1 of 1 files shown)


oziphantom (commented):
not really able to test this one, as I don't build with the makefile. However I have mine coded to show 'T1M windows' which just barely fits on the screen, does 'T1M 1.3.0-12-gf65378c' all fit in the current position?

rr-:
20211110_122312_nsd
Here's how it looks for me. Please note that menu is always rendered in 640x480 with the `TempVideoAdjust(2)` call so we don't have to worry about different screen resolutions.

oziphantom (approved):

thatguyfrombb:
I was already getting the long output for my docker builds, but it's nice knowing the actual version on the branch.

thatguyfrombb (approved):

PR #191, opened 2021-11-06
input, lara: improve tr3 sidesteps

### LINKED ISSUE #190: T1M bug: walk turning not working with the TR3 sidestep feature
You can't seem to turn while walking forwards without holding the up key, but only if the TR3 sidestep feature is enabled.

TR2Main implementation of the feature: 
https://github.com/Arsunt/TR2Main/commit/32acf1ccccb68df2cf001416a5fde0edceb18e65

And as said on Discord, to fix: it would better to remove all T1MConfig.enable_tr3_sidesteps occurrences from lara.c, there must be one single check. In input code.

rr-:
Closes #190

### DIFF
--- a/src/game/lara.c
+++ b/src/game/lara.c
@@ -145,17 +145,15 @@ void LaraAsWalk(ITEM_INFO *item, COLL_INFO *coll)
         return;
     }
 
-    if (!T1MConfig.enable_tr3_sidesteps || (Input.slow && Input.forward)) {
-        if (Input.left) {
-            Lara.turn_rate -= LARA_TURN_RATE;
-            if (Lara.turn_rate < -LARA_SLOW_TURN) {
-                Lara.turn_rate = -LARA_SLOW_TURN;
-            }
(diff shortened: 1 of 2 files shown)


oziphantom (approved):
much nicer

thatguyfrombb (approved):
We match TR2Main's implementation, which is nice, but I don't think it closes the issue. In fact, TR2Main wouldn't either. You can't turn while on the walking forwards without holding up with the feature enabled, not here, not in TR2Main AFAIK.

EDIT: I don't have much experience with TR2Main, there might be an exact "set" of settings which might work. No idea.

walkawayy:
I just loaded TR4, and this walking behavior seems to match that. If you hold walk and left or right, Lara sidesteps. Once you start walking forward, you can use left and right to walk in those directions. If you let go if forward, she will stop walking and start sidestepping. You have to press left or right without walk for her to turn in place.

walkawayy (approved):

PR #202, opened 2021-11-09
ENUM type mismatch

oziphantom:
probably doesn't happen for you because Linux lives in the A world, while my Windows is W so they mismatch for me.


rr- (approved):

rr-:
Your compiler probably sets `UNICODE` and/or `_UNICODE` by default, this is why it's different.

PR #209, opened 2021-11-10
game/draw: simplify 3d pickup lighting

rr-:
`CalculateObjectLighting` in itself calls `S_CalculateLight` so the call underneath felt extraneous. I couldn't get `CalculateObjectLighting` to calculate the light correctly, though, so I just decided to throw mesh and matrix information away and go with just the intensity of the light at the pickup location. Pickups are supposed to be small anyway so it shouldn't matter.

20211110_152020_abt
20211110_152216_dyn
20211110_152605_pfr
20211110_151725_ikl
20211110_151840_aul

### DIFF
--- a/src/game/draw.c
+++ b/src/game/draw.c
@@ -462,14 +462,9 @@ void DrawPickupItem(ITEM_INFO *item)
     phd_PushMatrix();
     phd_TranslateAbs(item->pos.x, offset, item->pos.z);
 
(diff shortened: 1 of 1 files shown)


oziphantom (approved):
I remember doing this. Looking at the other function deciding it did the same thing anyway and throwing it away ( TFFan94 code), I must have done it as part of the of Scale code which got nuked.

PR #210, opened 2021-11-10
Anim frame setting and testing tidy

oziphantom:
Pulls them into a function to do the checks and sets, where sane.

### DIFF
--- a/src/game/ai/bat.c
+++ b/src/game/ai/bat.c
@@ -133,8 +133,7 @@ static void FixEmbeddedBatPosition(int16_t item_num)
         Item_SetAnimBase(item, Objects[item->object_number].anim_index);
         bounds = GetBoundsAccurate(item);
 
-        item->anim_number = old_anim;
-        item->frame_number = old_frame;
+        Item_SetAnim(item, old_anim, old_frame);
 
         bat_height = ABS(bounds[FRAME_BOUND_MIN_Y]);
 
--- a/src/game/ai/bat.c
+++ b/src/game/ai/bat.c
@@ -6,6 +6,7 @@
 #include "game/control.h"
 #include "game/draw.h"
 #include "game/effects/blood.h"
+#include "game/items.h"
 #include "game/lot.h"
 #include "global/types.h"
 #include "global/vars.h"
@@ -129,8 +130,7 @@ static void FixEmbeddedBatPosition(int16_t item_num)
         old_anim = item->anim_number;
         old_frame = item->frame_number;
 
-        item->anim_number = Objects[item->object_number].anim_index;
-        item->frame_number = Anims[item->anim_number].frame_base;
+        Item_SetAnimBase(item, Objects[item->object_number].anim_index);
         bounds = GetBoundsAccurate(item);
 
         item->anim_number = old_anim;
--- a/src/game/ai/abortion.c
+++ b/src/game/ai/abortion.c
@@ -122,9 +122,8 @@ void AbortionControl(int16_t item_num)
         case ABORTION_TURN_L:
             if (!abortion->flags) {
                 abortion->flags = item->frame_number;
-            } else if (
-                item->frame_number - abortion->flags > 13
-                && item->frame_number - abortion->flags < 23) {
+            } else if (Item_AnimInRange(
+                           item, abortion->flags + 14, abortion->flags + 22)) {
                 item->pos.y_rot -= PHD_DEGREE * 9;
             }
 
@@ -137,8 +136,7 @@ void AbortionControl(int16_t item_num)
             if (!abortion->flags) {
                 abortion->flags = item->frame_number;
             } else if (
-                item->frame_number - abortion->flags > 16
-                && item->frame_number - abortion->flags < 23) {
+                Item_AnimInRange(item, abortion->flags + 17, abortion->flags + 22)) {
                 item->pos.y_rot += PHD_DEGREE * 14;
             }
 
--- a/src/game/ai/abortion.c
+++ b/src/game/ai/abortion.c
@@ -135,8 +135,8 @@ void AbortionControl(int16_t item_num)
         case ABORTION_TURN_R:
             if (!abortion->flags) {
                 abortion->flags = item->frame_number;
-            } else if (
-                Item_AnimInRange(item, abortion->flags + 17, abortion->flags + 22)) {
+            } else if (Item_AnimInRange(
+                           item, abortion->flags + 17, abortion->flags + 22)) {
                 item->pos.y_rot += PHD_DEGREE * 14;
             }
 
--- a/src/game/ai/abortion.c
+++ b/src/game/ai/abortion.c
@@ -49,9 +49,8 @@ void AbortionControl(int16_t item_num)
     if (item->hit_points <= 0) {
         if (item->current_anim_state != ABORTION_DEATH) {
             item->current_anim_state = ABORTION_DEATH;
-            item->anim_number =
-                Objects[O_ABORTION].anim_index + ABORTION_DIE_ANIM;
-            item->frame_number = Anims[item->anim_number].frame_base;
+            Item_SetAnimBase(
+                item, Objects[O_ABORTION].anim_index + ABORTION_DIE_ANIM);
         }
     } else {
         AI_INFO info;
@@ -169,9 +168,7 @@ void AbortionControl(int16_t item_num)
                 || LaraItem->hit_points <= 0) {
                 item->goal_anim_state = ABORTION_KILL;
 
-                LaraItem->anim_number = Objects[O_LARA_EXTRA].anim_index;
-                LaraItem->frame_number =
-                    Anims[LaraItem->anim_number].frame_base;
+                Item_SetAnimBase(LaraItem, Objects[O_LARA_EXTRA].anim_index);
                 LaraItem->current_anim_state = AS_SPECIAL;
                 LaraItem->goal_anim_state = AS_SPECIAL;
                 LaraItem->room_number = item->room_number;
--- a/src/game/ai/alligator.c
+++ b/src/game/ai/alligator.c
@@ -51,9 +51,8 @@ void AlligatorControl(int16_t item_num)
     if (item->hit_points <= 0) {
         if (item->current_anim_state != ALLIGATOR_DEATH) {
             item->current_anim_state = ALLIGATOR_DEATH;
-            item->anim_number =
-                Objects[O_ALLIGATOR].anim_index + ALLIGATOR_DIE_ANIM;
-            item->frame_number = Anims[item->anim_number].frame_base;
+            Item_SetAnimBase(
+                item, Objects[O_ALLIGATOR].anim_index + ALLIGATOR_DIE_ANIM);
             item->hit_points = DONT_TARGET;
         }
 
@@ -63,9 +62,8 @@ void AlligatorControl(int16_t item_num)
             item->object_number = O_CROCODILE;
             item->current_anim_state = CROCODILE_DEATH;
             item->goal_anim_state = CROCODILE_DEATH;
-            item->anim_number =
-                Objects[O_CROCODILE].anim_index + CROCODILE_DIE_ANIM;
-            item->frame_number = Anims[item->anim_number].frame_base;
+            Item_SetAnimBase(
+                item, Objects[O_ALLIGATOR].anim_index + CROCODILE_DIE_ANIM);
             room_num = item->room_number;
             floor = GetFloor(item->pos.x, item->pos.y, item->pos.z, &room_num);
             item->pos.y =
@@ -134,8 +132,7 @@ void AlligatorControl(int16_t item_num)
         item->object_number = O_CROCODILE;
         item->current_anim_state = Anims[item->anim_number].current_anim_state;
         item->goal_anim_state = item->current_anim_state;
-        item->anim_number = Objects[O_CROCODILE].anim_index;
-        item->frame_number = Anims[item->anim_number].frame_base;
+        Item_SetAnimBase(item, Objects[O_CROCODILE].anim_index);
         item->pos.y = item->floor;
         item->pos.x_rot = 0;
         gator->LOT.step = STEP_L;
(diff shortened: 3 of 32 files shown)


rr- (approved):
@oziphantom I added two extra cosmetic commits, can you check if they're fine?

src/game/ai/bat.c
```diff
@@ -133,8 +133,7 @@ static void FixEmbeddedBatPosition(int16_t item_num)
         Item_SetAnimBase(item, Objects[item->object_number].anim_index);
         bounds = GetBoundsAccurate(item);
 
-        item->anim_number = old_anim;
-        item->frame_number = old_frame;
+        Item_SetAnim(item, old_anim, old_frame);
```
oziphantom, line 136: No, I will have to undo this later.

rr-: How is this wrong?

oziphantom (commented):
does this make it work?

rr-:
@oziphantom the changes related to AnimScale do not exist yet so they do not apply at the moment. The point of this PR is to use uniform interface and it does it neatly in almost all cases; if the remaining cases need special treatment to make 60 FPS work, it should happen in the 60 FPS branch or perhaps we should introduce a different interface here. But here IMO there is nothing preventing us from applying the new idiom everywhere.

@thatguyfrombb let me know your thoughts on this.

walkawayy (approved):

walkawayy:
I went through all the files and everything looks good. I also agree with rr that this is a good animation framework. Once this gets approved and merged, you can add the `AnimScale` to the `SetAnim` functions in the 60 FPS branch.

thatguyfrombb:
Hmm... I have some mixed feelings. I get what rr- and walkaway are saying and they're right, *but*, oziphantom is right too. We need to think of the project now and what's coming, and 60FPS is coming.

So i have a question. @oziphantom Would you absolutely need to cache the scaled anims? Is there so much loss against just doing it live at runtime, which should allow (please correct me) just to add `* AnimScale` to one of this functions when 60FPS is enabled?

oziphantom:
@thatguyfrombb 

> So I have a question.Would you absolutely need to cache the scaled anims? Is there so much loss against just doing it live at runtime, which should allow (please correct me) just to add * AnimScale to one of this functions when 60FPS is enabled?

Not sure I understand the question. The issue is we save the current animation number and frame, because the code modifies them for some test, then restore then to what they where. The live numbers will be scaled, they have to be as frames numbers are int, I can't do a half frame, so the frame counters have to be scaled. Then we restore them which we can't scale them again. We can't do /AnimScale because we would loose the odd frame ( to which I mean bottom bit is set; odd ). So we need to restore the values verbatim. So another function could be made, Item_RestoreAnim ? but at this point is this "neater" than doing the code? There are 3 instances, but now when you read the code you have to ask yourself what does this function do different, while the 2 lines of code are explicit. If the ITEM does restore then I would also expect it to do the `Cache` but then you have to add two fields to each struct for every object for 3 cases in the code. Not that the memory matters in this day an age. We know that the code isn't threaded, so it can't do more than 1 cache, so the cache could 'm_' in the Item that holds the "one cache" but that sounds swapping a fry pan for fire.

rr-:
Do I understand the problem correctly?

1. We run the game at 60 FPS
2. `frame_number`s progress twice as fast due to faster FPS
3. To make the routines work in a sane way you want to make `Item_SetAnim` divide the current frame by two (which it does not do in this PR)
4. Then the game calls whatever routine it needs, sees 30-FPS-compatible frame numbers
5. Then restore the original, "faster" frame number by `Item_SetAnim` with multiplied value?

And that's why you want to cache the extra bit for odd frames?

Indeed that is terrible design. A better approach would be to keep everything from this PR and add following two functions:

`Item_NormalizeAnim`
`Item_DenormalizeAnim`

that would manipulate the extra bit.

(The name is a WIP – other pairs of words that popped to my mind: downscale/upscale, push/pop, `AnimTo30FPS`/`AnimFrom30FPS`…)

Then again why do we let the frames progress twice as fast if we can't deal with that and have to do ugly hacks? Why not increment them only on even frames?

oziphantom:
`Do I understand the problem correctly?`
No

1.) We run the game at 60 FPS
2.) Frame_numbers progress twice as fast due to faster FPS
3.) When we SetAnim I multiply the frame number by AnimScale 
4.) Then the game calls whatever routine it needs, sees 60-FPS-compatible frame numbers

And for all the code that **Sets** the anim this holds true. However your extra uses of SetAnim don't _set_ it, but rather _restore a previously cached value_ of that animation. 
From your code
~~~
        // The bats animation and frame have to be changed to the hanging
        // one to properly measure them. Save it so it can be restored
        // after.
        old_anim = item->anim_number;
        old_frame = item->frame_number;

        Item_SetAnimBase(item, Objects[item->object_number].anim_index);

        Item_SetAnim(item, old_anim, old_frame);
~~~
So here the old values are cached, now in 60fps this will be 60fps values, this is fine.
We SetAnimBase to modify the animation on the object for a Bounds grab. SetAnimBase will convert the base frame to a 60fps frame and GetBoundsAccurate will work just fine with the 60fps frame. No issues. 
But you then want to set the anim back to the old values, while it just so happens that SetAnim in 30fps does the same 2 lines of code, it is a different concept. We are not _setting_ the anim, we are _restoring cached anim values_. To which the restored values must be the same that were cached. We need to RestoreAnimFrame not SetAnimFrame these are different concepts that currently can be done with identical code. In 60fps they can't be done with identical code and need each concept to be treated specifically. 

What I'm saying with the divide is we can't just do this 
~~~
        // The bats animation and frame have to be changed to the hanging
        // one to properly measure them. Save it so it can be restored
        // after.
        old_anim = item->anim_number;
        old_frame = item->frame_number / AnimScale;

        Item_SetAnimBase(item, Objects[item->object_number].anim_index);

        Item_SetAnim(item, old_anim, old_frame);
~~~
I.e we divide by AnimScale and then use SetItem to restore the cached values because the frame_number can be odd to which we hold the animation in locked position as it will go to frame 1 get set to frame 0, to frame 1 get set to frame 0 and that will break the animation for those objects. 

I think thatguyfrombb is saying "Why don't we only keep the frame number at 30fps?", you also ask "Why don't we just update the animation frame every 2nd frame?". 
Because the point of the exercise is to get the game running and animating at 60fps, if we only update at 30fps then the game will only animate at 30fps. 
We can't keep 30fps frame numbers and then do 60fps updates because If you have 
FrameNumber = 37 and then 
FrameNumber = 37
One of those needs an extra frame of interpolation calculated, which one? We do know you would need an extra bit somewhere that keeps count of this is frame 37 normal and 37 extra. Or `LastFrameNumber` and if `LastFrameNumber==FrameNumber` then its an extra frame. Which is a horrible mess and if you look at how interpolation is calculated then you will see that is going to make a lot of messy hacky code to get this "Oh but sometimes we need to pretend there is a half frame here only its all ints so lets pretend its the next full frame and add them and div by two and pray we don't overflow". Because of 3 places that each have 2 lines of code, of which said 3 places are basically hacks :D

To which we can either 
a.) Do Nothing
~~~
        // The bats animation and frame have to be changed to the hanging
        // one to properly measure them. Save it so it can be restored
        // after.
        old_anim = item->anim_number;
        old_frame = item->frame_number;

        Item_SetAnimBase(item, Objects[item->object_number].anim_index);

        item->anim_number = old_anim;
        item->frame_number = old_frame;
~~~
I.e this is the code that is already there, and is clear as day what it does, and will require no modifications to work in 60fps. 
or
b.) make another function that handles the 3 instances of Restore
~~~
        // The bats animation and frame have to be changed to the hanging
        // one to properly measure them. Save it so it can be restored
        // after.
        old_anim = item->anim_number;
        old_frame = item->frame_number;

        Item_SetAnimBase(item, Objects[item->object_number].anim_index);

        Item_RestoreAnimFrame(old_anim, old_frame);
~~~
and ponder why does Item Restore it but doesn't Cache it in the first place. 

And as a reminder – this PR is from 60fps, as I proposed it as a method to help clean up the 60fps branch, only you asked me to pull it out so it could go into 30fps now.

rr-:
> Which is a horrible mess 

I would say that making frame number store two things is equal mess. How do I know when it is okay to use 30 FPS and when it is okay to use 60 FPS without spending 5 minutes analyzing the code flow every time?

Please store only one value and every place it needs one or another, use dedicated getter methods, one that uses raw values and the other that uses downscaled value. Then you don't need to restore anything. In case this fails and the "30 FPS" methods need to assign 30 FPS values, make two setters as well.

rr- (changes requested):
^

rr-:
These can be named `Item_GetDisplayAnim` (60 FPS) and `Item_GetStorageAnim` (30 FPS). There probably will be a couple of such methods. Either way the idea is to never use `->frame_number` directly.

oziphantom:
You can make a getter for '->FrameNumber' but that is all it will be. 
~~~
Item_GetFrameNumber(ITEM* item) {
    return item->FrameNumber;
}
~~~
There is no Display and Storage, the FrameNumber is always what the frame number is.
It doesn't store two things, it stores the frame_number. All initialization and comparison numbers are `*AnimScale`, this is "the setter" that handles this. This setter will handle converting original values to *AnimScale values. The `Anim_InRange` function will scale the values you are checking by AnimScale for you. I.e 
~~~
if (item->FrameNumber > a * AnimScale && item->FrameNumber < b * AnimScale) 
~~~
any other comparison will also need to have `*AnimScale` added to it. 
You never have 30 and 60 values at the same time, you never compare a 30 to a 60 or a 60 to a 30 because all values initialization and comparison values are modified with a *AnimScale. 
So these functions clean up
~~~
    item->anim_number = AA_STOP;
    item->frame_number = AF_STOP * AnimScale;
~~~
now this just becomes AF_STOP and the original numbers are preserved in the code. 
But 
~~~
            Lara.hit_frame++;
            if (Lara.hit_frame > 34 * AnimScale) {
                Lara.hit_frame = 34 * AnimScale;
            }
~~~
stays as it is, we could convert the > to `InRange` and `34, SINT_MAX` or make a `GreaterThan`, `LessThan`, `LessThanEqual`, `GreaterThanEqual`, `Equal` and `NotEqual` but at this point I don't feel there are enough to warrant such functions. this is `hitframe` not `animframe`. Hard to see as these fixes don't exist in the 60fps and the *AnimScale don't exist in the 30fps branch. 

The other alternative is we ditch 30 and everything goes 60, so all the constants are doubled, the loaded values are doubled and then we don't have to do an AnimScale on frame numbers just do inverted AnimScale on frame advancement, but this risks and potentially changes 30's behavior and diverges the code from what is in the data and the original.

rr-:
My understanding was flawed, but I still think we can use two setters – it's just a matter of semantics, you call it `RestoreAnim` and I call it something else. Let me expand on my thinking. Animation files have hardcoded number of frames that are meant to play at 30 frames per second. That's what I meant by `storage` animations. Rendered animations use interpolation to achieve more frames.  That's what I meant by `display` animations. So

```c
Item_SetAnimBaseStorage(item, Objects[item->object_number].anim_index);
```

can be

```c
void Item_SetAnimBaseStorage(ITEM_INFO *item, const int16_t anim_num)
{
    Item_SetAnimStorage(item, anim_num, Anims[anim_num].frame_base);
}

void Item_SetAnimStorage(
    ITEM_INFO *item, const int16_t anim_num, const int16_t anim_frame)
{
    assert(item);
    item->anim_number = anim_num;
    // convert storage frame number, which is expected to render at 30 FPS,
    // to ingame FPS
    item->frame_number = anim_frame * AnimScale;
}
```

and the restoring part can be

```c
old_anim = item->anim_number;
old_frame = item->frame_number;
...
Item_SetAnimDisplay(item, old_anim, old_frame);
```

which will do

```c
void Item_SetAnim(
    ITEM_INFO *item, const int16_t anim_num, const int16_t anim_frame)
{
    assert(item);
    item->anim_number = anim_num;
    item->frame_number = anim_frame;
}
```

...yet another, completely different, approach would be to just apply scaling at the time of setting the variable and do not multiply the item functions.

```c
Item_SetAnim(LaraItem, AA_RUN, AF_RUN * AnimScale);

old_anim = item->anim_number;
old_frame = item->frame_number;
...
Item_SetAnim(item, old_anim, old_frame);
```

...yet another approach would be to scale everything like you said at the time of initializing things.


Before I can make any informed decisions, can you remind me which place is the one that figures that frame 147 in 60 FPS means we need to load frame 73 from the animation and interpolate it? Are there many such places or more like 1, 2? Because I'm leaning towards multiplying everything and let only those 1,2 places use this downscaling with tons of comments.

rr-:
Or is my concept of animations flawed and they supply 3-4 keyframes even if they last 50 frames and we use interpolation ALL THE TIME regardless?

oziphantom:
draw.c : GetFrame gets the interpolation frac values, it is then pushed through a few functions but if you search for GetFrame you will find all the cases that it is used. 

They way I jig it is in file.c when we load the animation I then do this 
~~~
if (Anims[i].interpolation == 0) {
            Anims[i].interpolation = 1;
        } else {
            Anims[i].interpolation *= AnimScale;
        }
~~~
this is because an animation only has a fixed interpolation between **all** frames in it, you can't set a per frame value. 
However in the scale everything by 2 to make it all 60, we don't change the GetFrame function, we do
`item->FrameNumber += 2` for 30 and `item->FrameNumber += 1` for 60, i.e inverted AnimFrame.(There are a few other cases that need the change as well) However not sure if that would involve some of the interpolation update functions when they normally wouldn't be used even though we skip the middle frame as it now always has an interpolation value.

rr-:
Thanks. Could you do a small test with a single animation to see if it
would break things?

oziphantom:
I've realized this question has two parts. 
1.) can one do a test with 1 animation
2.) what animation does one test with
To which thinking about it, there is no way to just test 1 anim, the whole code and engine needs to be switched over to the new system. I could make it just "lara" and let the other enemies do weird things. But it is still a large rework.

rr-:
Closing due to lack of activity

Issue #145, opened 2021-10-27
Feature request: Saturn style water


Richard-L:
https://user-images.githubusercontent.com/4934209/139142220-60e3ee7a-aa53-433c-ac38-6083696d15bb.mp4

Since we talked about it, why not collect it.

In the Saturn version, farther away water is gradually drawn with more saturation of the given water colour, and also seems to look a bit darker. This creates an effect of murky water which is quite nice, and could be an optional visual feature for later.

Richard-L:
For reference, here's a nice first attempt by Troye in his TR3 project to do something like this:

~~https://cdn.discordapp.com/attachments/1028717372841934879/1068869220198264892/Tomb_Raider_III_2023-01-28_06-08-34.mp4~~

Reupload:

https://github.com/user-attachments/assets/26bbfae4-a9de-4518-9b9d-b01238bbc2f2



![comp](https://github.com/user-attachments/assets/1c4ff635-076a-46d0-ac94-f54740507655)

Richard-L:
Curiously the PowerVR version seems to have had a similar effect on the PC - compare the underwater moment in the Lost Valley demo here: https://youtu.be/J20pDT8Y6Yg?t=51

filipe-maia:
I have always heard that the water in the Saturn version was better than in the other versions. This would be a cool addition.

rr-:
![Image](https://github.com/user-attachments/assets/ee69e314-7335-4634-8abf-f58a8d5a4b3f)

Video: https://nebula.wind.garden/serve/pub/2025-05-20_00-13-06.mp4

Am I doing this right?

```diff
diff --git a/data/tr1/ship/shaders/common.glsl b/data/tr1/ship/shaders/common.glsl
index b493b4c31..0498cba3c 100644
--- a/data/tr1/ship/shaders/common.glsl
+++ b/data/tr1/ship/shaders/common.glsl
@@ -41,6 +41,19 @@ float shadeFog(float shade, float depth, vec2 fog)
     }
 }
 
+vec3 colorizeFog(vec3 color1, vec3 color2, float depth, vec2 fog)
+{
+    float fogBegin = fog.x;
+    float fogEnd = fog.y;
+    if (depth < fogBegin) {
+        return color1;
+    } else if (depth >= fogEnd) {
+        return color2;
+    } else {
+        return color1 + (depth - fogBegin) * (color2 - color1) / (fogEnd - fogBegin);
+    }
+}
+
[...]

PR #211, opened 2021-11-10
hwr: add per-level fog settings

### LINKED ISSUE #118: Feature Request: Per-Level Distance Fog settings
This is TRLE related.
It would be nice to have a per-level setting of the Distance Fog and Cull settings.
TombATI had a setting for that but only for the whole game.

rr-:
Closes #118.


oziphantom:
Should the water stuff be part of this commit?

rr-:
No, I don't think so, I only ported GLRage's distance settings override patch. Plus #145 probably will require different approach.

thatguyfrombb (commented):
Default settings in this PR makes the wall/sprite popping very apparent. This is also an issue on TombATI when changing the draw distance. Otherwise, the values in the JSON can be negative, which makes for some really cool effects. If there's no foreseeable issue with that, I'd love to leave it in.
Causes popping issues

rr-:
What would be the default values you recommend @thatguyfrombb ?

thatguyfrombb:
Fade at `16384` and max at `24576` is the ceiling for absolutely safe settings. Anything over that and the sprites popping in are visible.
Fade at `20480` and max at `28672` is the limit for buildings/walls. Sprites already pop at visible distance, but walls fade in nicely. Anything over this and walls start popping in.

With the current code, I'd go for `16384` so it should just work everywhere and add a note in the README so users can change it at will. Whenever we get it to render better we can change our defaults.

rr-:
Sounds good. I'll amend the settings. Could you create two new issues regarding the hard z limit for sprites and for the geometry?

rr-:
Done, but haven't tested the new settings

thatguyfrombb:
So I talked about TombATI from what I remembered it doing. Turns out, TombATI does handle the original values it ships with correctly. I remembered it having issues because I tried very large values. So, this is *our* bug. Will investigate tomorrow.

rr-:
@thatguyfrombb should be good now, improved on multiple ends
- distances in the config are now in tiles, not in pixels
- ported tr2main's CalculateFogShade to support fog depths other than 8 tiles
- fixed the clipping issues

cfg/Tomb1Main_gameflow.json5
```diff
@@ -19,6 +19,17 @@
     // [0.45, 1.0, 1.0] is default TombATI filter.
     "water_color": [0.45, 1.0, 1.0],
 
+    // The distance (in tiles) at which objects and the world
+    // start to fade into blackness.
+    // The default hardcoded value in TR1 is 12.
+    // The default hardcoded value in TombATI is 80.
```
thatguyfrombb, line 25: In TombATI fade is 72, not 80, and the whole feature is off by default to "not ruin the ambience" or something like that.

src/game/gameflow.c
```diff
@@ -201,6 +201,30 @@ static int8_t S_LoadScriptMeta(struct json_object_s *obj)
             json_array_get_number_double(tmp_arr, 2, GF.water_color.b);
     }
 
+    if (json_object_get_value(obj, "draw_distance_fade")) {
+        double value = json_object_get_number_double(
+            obj, "draw_distance_fade", JSON_INVALID_NUMBER);
+        if (value == JSON_INVALID_NUMBER) {
+            LOG_ERROR("'draw_distance_fade' must be an integer");
```
thatguyfrombb, line 208: Incorrect error message, we take doubles now

src/game/gameflow.c
```diff
@@ -201,6 +201,30 @@ static int8_t S_LoadScriptMeta(struct json_object_s *obj)
...
+        GF.draw_distance_fade = value;
+    } else {
+        GF.draw_distance_fade = 12;
+    }
+
+    if (json_object_get_value(obj, "draw_distance_max")) {
+        double value = json_object_get_number_double(
+            obj, "draw_distance_max", JSON_INVALID_NUMBER);
+        if (value == JSON_INVALID_NUMBER) {
+            LOG_ERROR("'draw_distance_max' must be an integer");
```
thatguyfrombb, line 220: Same as above

src/game/gameflow.c
```diff
@@ -201,6 +201,30 @@ static int8_t S_LoadScriptMeta(struct json_object_s *obj)
...
+            return 0;
+        }
+        GF.draw_distance_max = value;
+    } else {
+        GF.draw_distance_max = 20;
```
thatguyfrombb, line 225: It's stored in a float, let's make that clear by assigning `20.0f`

cfg/Tomb1Main_gameflow.json5
```diff
@@ -19,6 +19,17 @@
...
+    "draw_distance_fade": 22,
```
thatguyfrombb, line 26: Should we change it to `22.0` and max to `30.0` to show that it accepts floats?

thatguyfrombb (changes requested):
Testing works great!

src/game/gameflow.c
rr-: Fixed

cfg/Tomb1Main_gameflow.json5
rr-: Fixed

src/game/gameflow.c
rr-: Fixed

src/game/gameflow.c
rr-: Fixed

cfg/Tomb1Main_gameflow.json5
rr-: Good call, changed

thatguyfrombb:
Nearly forgot, do we want this in README?

rr-:
Yes, because the per-level functionality wasn't there originally. I'll add it

PR #213, opened 2021-11-10
option: add adjustable ingame brightness

rr-:


oziphantom (approved):
more things I need to rework for 60/30 toggle :D

thatguyfrombb (approved):

thatguyfrombb:
Forgot one thing, should we add a line about it in README? Mainly so users know it's available by the time the next release comes out.

rr-:
Good catch, added :)

PR #223, opened 2021-11-12
build: use meson

rr-:
This is an effort to sanitize our toolchain – hand-rolled makefiles are not really scalable (I am currently experimenting with adding glrage's source code to the repository.). Inflexibility of meson configuration is a good thing IMO, and overall my experience with it is "it just works", whereas with Makefiles I had pretty lousy experience.

Other changes:
- removed dead modules
- fixed some warnings
- improved github actions so that they're testable locally (they weren't)

For Docker users it doesn't matter as they still can run `make docker_build`.
As a bonus there apparently is some integration with Visual Studio, so VS users might benefit from this change.

Another bonus we get parallel builds.


rr-:
Build time with meson (outside docker): 5.1 s
Build time with make: 18.5 s

oziphantom:
you can get parallel with make, you just do make -j<number of cores * 1.5> for best results

rr-:
Yup – but you have to do that, and adding building of additional modules / recursive Makefiles is just painful

Dockerfile
```diff
@@ -3,12 +3,10 @@ FROM ubuntu:latest
...
+    python3 -m pip install pyjson5 meson ninja
 
 RUN mkdir /app
 ADD . /app
 WORKDIR /app
 
-RUN chmod 0666 /app/src/init.c
-
-CMD ["make", "build"]
+CMD ["make", "debug_linux", "build"]
```
thatguyfrombb, line 12: This includes "make clean" which tries to remove the "build" directory from inside Docker, which it has no permissions for (if it even exists).

thatguyfrombb: "generate_init" lacks permissions to write to `src/init.c` under Docker (only used when the in-repo `init.c` is outdated)

src/game/sound.c
```diff
@@ -51,13 +51,13 @@ static int16_t m_AmbientLookup[MAX_AMBIENT_FX] = { 0 };
...
 static SOUND_SLOT *Sound_GetSlot(
-    int32_t sfx_num, uint32_t loudness, PHD_3DPOS *pos, int16_t mode);
+    int32_t sfx_num, int32_t loudness, PHD_3DPOS *pos, int16_t mode);
 static void Sound_UpdateSlotParams(SOUND_SLOT *slot);
 static void Sound_ClearSlot(SOUND_SLOT *slot);
 static void Sound_ClearSlotHandles(SOUND_SLOT *slot);
 
 static SOUND_SLOT *Sound_GetSlot(
-    int32_t sfx_num, uint32_t loudness, PHD_3DPOS *pos, int16_t mode)
+    int32_t sfx_num, int32_t loudness, PHD_3DPOS *pos, int16_t mode)
```
thatguyfrombb, line 60: This change breaks some water sound on Lost Valley, start that level and go forwards until the water. You'll hear the problem right there.

walkawayy (approved):

thatguyfrombb (changes requested):

src/game/sound.c
rr-: Waaat. Great find. Went back to uint32_t, but added a cast for SOUND_NOT_AUDIBLE.

Dockerfile
rr-: Fixed

rr-:
Also added back iterative builds with Docker so it doesn't run `clean` every run.

thatguyfrombb (approved):
Sound is fixed, docker_build works, rebuilding works at least as good as it usually does :)

PR #224, opened 2021-11-12
objects/pickup: make pickup radius broader

### LINKED ISSUE #195: OG bug: strange pickup radius
from pickup.c

```
int16_t PickUpBounds[12] = {
    -256, +256, -100, +100, -256, +100, -10 * PHD_DEGREE, +10 * PHD_DEGREE,
    0,    0,    0,    0,
};

int16_t PickUpBoundsUW[12] = {
    -512,
    +512,
    -512,
    +512,
    -512,
    +512,
    -45 * PHD_DEGREE,
    +45 * PHD_DEGREE,
    -45 * PHD_DEGREE,
    +45 * PHD_DEGREE,
    -45 * PHD_DEGREE,
    +45 * PHD_DEGREE,
};
```
in the first one, the +100 after the -256, is that correct? Does kinda of explain why it hard to pick one up sometimes until you move elsewhere.

rr-:
In TR1 and TR2, possibly in the later games too, Lara has a larger pickup radius if she's facing away from the item. This is counter-intuitive. This commit adjusts it so that the radius is the same regardless of her orientation in reference to the collectibles.

Closes #195

### DIFF
--- a/src/game/objects/pickup.c
+++ b/src/game/objects/pickup.c
@@ -13,7 +13,7 @@ PHD_VECTOR PickUpPosition = { 0, 0, -100 };
 PHD_VECTOR PickUpPositionUW = { 0, -200, -350 };
 
 int16_t PickUpBounds[12] = {
-    -256, +256, -100, +100, -256, +100, -10 * PHD_DEGREE, +10 * PHD_DEGREE,
+    -256, +256, -100, +100, -256, +256, -10 * PHD_DEGREE, +10 * PHD_DEGREE,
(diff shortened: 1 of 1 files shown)


walkawayy:
Looks good. When the walk to pickups/switches gets added, I'll thinking I'll add a different set of bounds to use if that feature is enabled to make the radius slightly bigger.

walkawayy (approved):

oziphantom (approved):

PR #230, opened 2021-11-16
s/*: prefix filenames with s_

rr-:
When a file `specific/foo.c` does `#include "bar.h"`, it refers to `specific/bar.h` instead of `bar.h`. Rather than doing one-off exceptions like `smain.c`, we apply uniform approach and prefix all files in `specific/` with `s_`, mainly due to their universal naming.

The thing that prompted me to create this PR was `init.h` conflict with `specific/init.h`.

I think this PR is controversial because of the following factors:

- `game/` stuff is not prefixed with `g_`, so we're being inconsistent.
- the file names look ugly.
- we _could_ use `#include <...>` to always refer to absolute path, because we get `-I` from meson anyway, but that is weird too IMO.

Let me know your thoughts about this.


oziphantom:
`game/ stuff is not prefixed with g_, so we're being inconsistent.`
however Game is "the actual folder" its just in a folder for the dream of the engine being used nicely for something else. But also is not prefixed with g_ yet ;) but also g_ means global which is unfortunate. 
`the file names look ugly.`
much better than the 8.3 originals I bet ;)
`we could use #include <...> to always refer to absolute path, because we get -I from meson anyway, but that is weird too IMO.`
No, <> must be "library" level and only be used for such files, if we start using it for our local files as well, then you may have clashes with some random library stored on some machine. Then include path order becomes very important and you hope that the linkers put them in the same order and not reverse order. 
~~~
#include "myfile.h"
#include <myOtherFile.h>
#include "startwithstobeafter.h"

#include <somelib.h>
~~~
It also looks rubbish. 

`The thing that prompted me to create this PR was init.h conflict with specific/init.h.`
great I will be able to get rid of sinit.c hack which causes me to have a bunch of missing strings ;)

rr-:
thatguyfrombb and walkawayy also voiced their approval on Discord. The majority seems to be in favor of this new naming scheme, so let's go with it. I'll fix the conflicts and then I'll need an approval before go with the merge.

walkawayy:
I think it's fine that `game` and `global` aren't prefixed with a `g_` equivalent. Unless I overlooked something, only the `specific` folder has files that have the same name as other files, so there's not the same confusion or issues that come up.

src/specific/s_clock.h
```diff
@@ -1,5 +1,5 @@
-#ifndef T1M_SPECIFIC_CLOCK_H
-#define T1M_SPECIFIC_CLOCK_H
+#ifndef T1M_SPECIFIC_S_CLOCK_H
+#define T1M_SPECIFIC_S_CLOCK_H
```
walkawayy, line 2: Not a huge deal to me, but the `SPECIFIC` and `S` are redundant. Just double checking if this was supposed to change to `_S_` only or not.

rr-: True – it is redundant – but I wanted this scheme to be very simple and have no exceptions.

walkawayy: Ya that works.

walkawayy (approved):

PR #231, opened 2021-11-16
ATI3DCIF refactors

rr-:
- Fix compilation warnings on master
- Simplify loading ATI3DCIF procedures
- Use conforming names for the public function names, apart from library routines which were not touched
- Remove useless calls that are not implemented in glrage such as viewport scissoring, fog, dithering etc.
- Remove ATI3DCIF specific variables from `var_platforms.h` and put them in proper modules
- Remove no longer needed injections of ATI pointers


walkawayy:
I don't really understand the `ATI3DCIF` code in all honesty. But it compiles for me in Docker now without warnings, and I didn't notice any issues in a quick play test through the beginning of the levels. Looks good to me, but I'll wait for someone else to double check and formally approve.

rr-:
ATI3DCIF calls stuff from ati3dcif.dll that in TombATI is supplied by [glrage](https://github.com/ata4/glrage). Certain things in the code that were made for the original ATI3DCIF are no longer relevant, because glrage never implements them, or enables them always, etc.

walkawayy (approved):

PR #233, opened 2021-11-16
s/hwr: let people play even with texture errors

rr-:
This is a safety mechanism so that when the texture loader fails for any reason, the player can see cel-shaded polys instead of an unfriendly error message.

The reason why drawing untextured triangles used to be not working in the past was because we should have been calling `HWR_DisableTextureMode()`.

### DIFF
--- a/src/specific/s_hwr.c
+++ b/src/specific/s_hwr.c
@@ -314,14 +314,20 @@ void HWR_SelectTexture(int tex_num)
         return;
     }
 
+    if (!HWR_TextureLoaded[tex_num]) {
+        return;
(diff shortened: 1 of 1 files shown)


walkawayy:
Is there a way to break the texture loader intentionally to test this?

rr-:
Yeah – replace call `if (ATI3DCIF_TextureReg(…))` with `if (1)` to enter the other branch.

walkawayy:
Looks good once the merge conflicts are resolved.

walkawayy (approved):

PR #235, opened 2021-11-16
Support arbitrary PCX backdrops

rr-:
PNG support will be handled in another pull request.


rr-:
For now not touching screen resolution parts.

PR #236, opened 2021-11-17
Tidy build process

rr-:


rr-:
Merging as it is really blocking me with further glrage/libav work. Tested for releases, incremental builds and release/debug configurations and things seem to be working fine. Any problems can be addressed in follow-up pull requests.

PR #241, opened 2021-11-18
Input refactors

rr-:
Apart from the usual renaming refactors, one potentially harmful action was the removal of the DIK_ checks in option.c, which in theory should allow rebinding to arrows and keys, but I haven't tested that. This change was motivated by decoupling option.c and DInput.


rr-:
Tested and it's working as best as it can with the TR1 input conflict resolution mechanics.

PR #254, opened 2021-11-26
s/sound: ditch DSound, use new SDL backend

rr-:
This PR is a part of the efforts to make TR1 cross-platform and implements SDL audio backend for the samples that replaces the DirectSound-based approach. It implements the following:
- pitch control. This is done without any interpolation but sounds good enough to me.
- pan control. This sounds good to me, but the calculation are not true to the OG.
- volume control. This sounds off to me, and the calculations are not true to the OG - DirectSound approach to setting the volume works by adding -100.0 dB (which produces silence) to 0.0 dB (which plays the sample at its normal loudness) to each sample. Us, on the other hand, mix the audio by hand, so instead of adding decibel values, we multiply individual sample values by a multiplier. I was looking for a good formula that we could use here, but certain things sound either a bit too quiet (especially the footsteps) or too loud.
- looping.
- resampling (the original samples are 22050 Hz U8, our mixer is 44100 F32LE).

Despite my earlier comments I believe the current volume control is good enough, I found the playing experience to be enjoyable. However, what prevents me from pushing this pull request is the resampling – it produces clicks at the end of the playback for certain samples. This is especially audible near the Scion, where there's a looped humming sound. It produces 2-3 clicks per second on each loop.

What I tried to fix this:
- resample on the fly during mixing. This actually produces even more clicks (?). It's possible I botched this, but even if it worked, it'd essentially make our pitch shift algorithm unusable and very difficult to implement.
- cut off the silence from the end of the resampled tracks.

I am looking for opinions how to tackle this issue.


oziphantom:
so `int samples_requested =
            len / sizeof(AUDIO_WORKING_FORMAT) / AUDIO_WORKING_CHANNELS;` I can't see how this is correct, there isn't a define for AUDIO_WORKING_FORMAT in the code checked in, so I assume it is something defined by SDL 'SDL_AudioFormat' but you are getting the size of it, and not the size that the format represents. A look at AudioFormat suggests is uses 16bits, but I can't see the actual strict definition of it, to know it if its 16bits or will be stored in 32 or 64 bits. Which might put you in a case of "it just so happens to work" when you compile on 32bit as your using floats, but it might then go horribly wrong on 64bit or anything else. 
Speaking of which you also cast 'void*' to 'int' and back, if everything uses int make the functions int, int and void* are not always the same size and this can cause you to loose a lot of bits. Again on x86 32bit, it "works" but it can go horribly wrong, please remove it. 

The other ye olde flaw that happens, is the extra fence posts error.  lets say we have some samples that are 1 2 3 4 and we are resampling at 3x the rate and I'm just going to duplicate the samples.  so we get the samples and do 
`1 1 1 2 2 2 3 3 3 4`
because once we reach 4 we are the the last sample and the thus the sample number overflows and you get `1 1 1 2 2 2 3 3 3 4 X X` where X is whatever just happened to be in the buffer last time and you get clicks and noise. So you either duplicate the last sample N times, or reduce the play length such that its not `originalSamples*3` but `originalSamples*3 - (3-1)`.

rr-:
Thank you for the in depth review.

On Sat, Nov 27, 2021 at 03:07:14AM -0800, oziphantom wrote:
> so int samples_requested = len / sizeof(AUDIO_WORKING_FORMAT) /
> AUDIO_WORKING_CHANNELS; I can't see how this is correct,

len is in bytes. We are supposed to fill the buffer in full. Our mixer
works with F32LE stereo samples. To get the number of samples, we need
to divide the size of the buffer by channel count and the size of our
working format which is 4 bytes. So we divide here by 8.

> there isn't a define for AUDIO_WORKING_FORMAT in the code checked in

It's defined in specific/s_audio.h :)

> Speaking of which you also cast 'void*' to 'int' and back, if
> everything uses int make the functions int, int and void* are not
> always the same size and this can cause you to loose a lot of bits.
> Again on x86 32bit, it "works" but it can go horribly wrong, please
> remove it.

I agree casting to (void*) does not make any sense, the only reason I
left it is because I gave up on this pull request due to audible clicks
I mentioned. In general apart from this issue, I'd like to merge
specific/s_sound.c and game/sound.c because keeping both does not make
sense IMO with the new common backend.

> The other ye olde flaw that happens, is the extra fence posts error.
> lets say we have some samples that are 1 2 3 4 and we are resampling
> at 3x the rate and I'm just going to duplicate the samples. so we get
> the samples and do 1 1 1 2 2 2 3 3 3 4 because once we reach 4 we are
> the the last sample and the thus the sample number overflows and you
> get 1 1 1 2 2 2 3 3 3 4 X X where X is whatever just happened to be in
> the buffer last time and you get clicks and noise. So you either
> duplicate the last sample N times, or reduce the play length such that
> its not originalSamples*3 but originalSamples*3 - (3-1).

The resampling from the source format (U8 22050 Hz) to the target format
(44100 Hz F32LE) is handled by SDL
(http://wiki.libsdl.org/SDL_NewAudioStream and similar functions). We're
getting clicks as early as right after this conversion: I dumped the raw
PCM data that we got after the converting and played it in ffmpeg - got
a click at the end; note that this was at the sample decoding +
resampling phase, before any of the mixer code. I suspect
http://wiki.libsdl.org/SDL_AudioStreamFlush appends some data that we do
not want. I have no idea how to tackle this. I'm pretty sure, however,
that SDL resamples the source using proper interpolation.

About dumb duplication of the samples - I agree that we do it with my
pitch shift implementation on the mixing level, however, in my tests
that was good enough + I am not sure I understand your fix.

oziphantom:
> Thank you for the in depth review.
> On Sat, Nov 27, 2021 at 03:07:14AM -0800, oziphantom wrote: so int samples_requested = len / sizeof(AUDIO_WORKING_FORMAT) / AUDIO_WORKING_CHANNELS; I can't see how this is correct,
> len is in bytes. We are supposed to fill the buffer in full. Our mixer works with F32LE stereo samples. To get the number of samples, we need to divide the size of the buffer by channel count and the size of our working format which is 4 bytes. So we divide here by 8.
> there isn't a define for AUDIO_WORKING_FORMAT in the code checked in
> It's defined in specific/s_audio.h :)

ah, which is not in the diff above, but looking at the file it defines it as AUDIO_F32 which is SDL define. 
however we also have
~~~
    stream->sdl.stream = SDL_NewAudioStream(
        sdl_format, sdl_channels, sdl_sample_rate, AUDIO_WORKING_FORMAT,
        sdl_channels, AUDIO_WORKING_RATE);
~~~
so if that is a defined as `float` you can't call this function with 
~~~
    stream->sdl.stream = SDL_NewAudioStream(
        sdl_format, sdl_channels, sdl_sample_rate, float,
        sdl_channels, AUDIO_WORKING_RATE);
~~~
where that function is defined as
~~~
SDL_AudioStream * SDL_NewAudioStream(const SDL_AudioFormat src_format,
                   const Uint8 src_channels,
                   const int src_rate,
                   const SDL_AudioFormat dst_format,
                   const Uint8 dst_channels,
                   const int dst_rate);
~~~
so `AUDIO_WORKING_FORMAT` is of type `SDL_AudioFormat`, so the sizeof gets the size of a `SDL_AudioFormat` which has nothing to do with the actual size the _sample_ in said format would be right? 

So rather than `sizeof(AUDIO_WORKING_RATE)` you want to do `(SDL_AUDIO_BITSIZE(AUDIO_WORKING_RATE)/8)` to get the right values. 

> Speaking of which you also cast 'void*' to 'int' and back, if everything uses int make the functions int, int and void* are not always the same size and this can cause you to loose a lot of bits. Again on x86 32bit, it "works" but it can go horribly wrong, please remove it.
> I agree casting to (void*) does not make any sense, the only reason I left it is because I gave up on this pull request due to audible clicks I mentioned. In general apart from this issue, I'd like to merge specific/s_sound.c and game/sound.c because keeping both does not make sense IMO with the new common backend.
> The other ye olde flaw that happens, is the extra fence posts error. lets say we have some samples that are 1 2 3 4 and we are resampling at 3x the rate and I'm just going to duplicate the samples. so we get the samples and do 1 1 1 2 2 2 3 3 3 4 because once we reach 4 we are the the last sample and the thus the sample number overflows and you get 1 1 1 2 2 2 3 3 3 4 X X where X is whatever just happened to be in the buffer last time and you get clicks and noise. So you either duplicate the last sample N times, or reduce the play length such that its not originalSamples*3 but originalSamples*3 - (3-1).
> The resampling from the source format (U8 22050 Hz) to the target format (44100 Hz F32LE) is handled by SDL (http://wiki.libsdl.org/SDL_NewAudioStream and similar functions). We're getting clicks as early as right after this conversion: I dumped the raw PCM data that we got after the converting and played it in ffmpeg - got a click at the end; note that this was at the sample decoding + resampling phase, before any of the mixer code. I suspect http://wiki.libsdl.org/SDL_AudioStreamFlush appends some data that we do not want. I have no idea how to tackle this. I'm pretty sure, however, that SDL resamples the source using proper interpolation. About dumb duplication of the samples - I agree that we do it with my pitch shift implementation on the mixing level, however, in my tests that was good enough + I am not sure I understand your fix.

My dumb duplication is just to show there are now 3 samples for the 1, purely for demonstration purposes. The point being that when you get to the last sample, its important to still write 2 more samples because the "num of samples" has been multiplied by 3( in my simple example), but most people write code that stops on the last sample and so you get 2 missing samples at the end. Another way to fix this is to sub the extra samples from the total number of samples to play. You would expect SDL to handle this however. 
My next step would be don't convert the sample, get it to play it in U8, and see if you still get the clicking.

rr-:
Good news: I fixed the volume calculations, they should be 100% consistent with DirectDraw. (Took the formulas from wine)
Clicking is still present.

rr-:
I just noticed that the crackling near Scion happens with OG as well, so it's a matter of a botched sample. Merging.

PR #258, opened 2021-12-08
alligator: fix alligator dealing no damage if Lara is still

### LINKED ISSUE #194: OG bug: alligators can get stuck in an attack animation and deal no damage
Alligators can get stuck in their animation of attacking Lara. They look like they are attacking but deal no damage. If Lara moves, they will move and fix themselves. I think this is an OG bug, but I haven't confirmed it yet to add the label.

https://streamable.com/ld1l6d

May be related to #44.

walkawayy:
Closes #194


rr- (approved):
LGTM. I understand that you went with approach #1 from your solutions mentioned in the original issue?

walkawayy:
> 

This is actually solution 3. Like this:
https://streamable.com/3rgkrw

I actually think I need to fix one outstanding problem. This [line](https://github.com/rr-/Tomb1Main/blob/3bddf4f45d18e9f2e55610868d005c8bec544163/src/game/ai/alligator.c#L123) means the croc does a double bite of damage only the first time it ever does damage because `item->required_anim_state` is initialized to `ALLIGATOR_EMPTY`.

rr- (approved):
LGTM. Can you ask on Discord for someone to test it?

rr- (changes requested):
Revoking the approval in light of the recent discussions in the relevant ticket.

src/game/ai/alligator.c
```diff
@@ -105,11 +106,19 @@ void AlligatorControl(int16_t item_num)
...
         break;
 
     case ALLIGATOR_ATTACK:
-        if (item->frame_number == g_Anims[item->anim_number].frame_base) {
+        if (g_Config.fix_alligator_ai
+            && item->frame_number == ALLIGATOR_BITE_AF) {
+            item->required_anim_state = ALLIGATOR_EMPTY;
+        } else if (
+            !g_Config.fix_alligator_ai
+            && item->frame_number == g_Anims[item->anim_number].frame_base) {
```
rr-, line 121: I believe these can be simplified to
`if (frame == frame_base (fix croc && frame == alligator_bite_af))`

rr-:
Tested and LGTM. There is a trivial conflict in the README. Left one cosmetic remark.

src/game/ai/alligator.c
walkawayy: I don't think so because the fix version will have the extra damage bug because it can also pass the `item->frame_number == g_Anims[item->anim_number].frame_base` check which is true for multiple animations.

rr-: Right, then how about `if (frame == fix alligator ? ALLIGATOR_BITE_AF : frame_base) {`?

rr- (approved):

PR #263, opened 2021-12-10
Bridge grab fix

walkawayy:


src/game/objects/bridge.c
```diff
@@ -4,6 +4,17 @@
 #include "game/draw.h"
 #include "game/objects/cog.h"
 #include "global/vars.h"
+#include "src/config.h"
+
+bool isSameSector(int32_t x, int32_t y, int32_t z, ITEM_INFO *item)
```
rr-, line 9: This function doesn't adhere to the naming conventions outlined in contributing guidelines.
ITEM_INFO * can be const.

src/game/objects/bridge.c
```diff
@@ -4,6 +4,17 @@
...
+{
+    int32_t a_Sector_x = x / WALL_L;
+    int32_t a_Sector_z = z / WALL_L;
+    int32_t b_Sector_x = item->pos.x / WALL_L;
+    int32_t b_Sector_z = item->pos.z / WALL_L;
```
rr-, line 14: Same for variables

src/game/lara.c
```diff
@@ -2145,8 +2157,10 @@ bool LaraTestHangJumpUp(ITEM_INFO *item, COLL_INFO *coll)
     }
 
     if (angle & (PHD_90 - 1)) {
+        LOG_DEBUG("Fail 5 angle: %d", angle);
         return false;
     }
+    LOG_DEBUG("PASSED Fail 5 angle: %d", angle);
```
rr-, line 2163: We shouldn't commit debug messages

src/game/objects/bridge.c
```diff
@@ -103,14 +114,31 @@ void DrawBridgeCollision(
...
+        if (y > item->pos.y) {
+            return;
+        }
+
         *height = item->pos.y;
+    } else {
+        if (y <= item->pos.y) {
+            *height = item->pos.y;
+        }
     }
```
rr-, line 130: Current logic is hard to follow and should be simplified to something like

```c
if (g_Config.fix_bridge_grabs && !Bridge_IsSameSector(x, y, z, item)) {
    return;
}
if (y > item->pos.y) {
    return;
}
*height = item->pos.y;
```

src/game/objects/bridge.c
```diff
@@ -103,14 +114,31 @@ void DrawBridgeCollision(
...
 }
 
 void BridgeFlatCeiling(
     ITEM_INFO *item, int32_t x, int32_t y, int32_t z, int16_t *height)
 {
+    if (g_Config.fix_bridge_grabs) {
+        if (!isSameSector(x, y, z, item)) {
```
rr-, line 137: Let's merge both conditions.

src/game/objects/bridge.c
```diff
@@ -133,6 +161,21 @@ void BridgeTilt1Floor(
...
+        if (y > level) {
+            return;
+        }
+
+        *height = level;
+    } else {
+        if (y <= level) {
+            *height = level;
+        }
+    }
```
rr-, line 178: Again this is pretty unreadable.

src/game/objects/bridge.c
```diff
@@ -151,6 +200,21 @@ void BridgeTilt2Floor(
...
```
rr-, line 217: Ditto

src/game/objects/bridge.c
```diff
@@ -141,6 +184,12 @@ void BridgeTilt1Floor(
 void BridgeTilt1Ceiling(
     ITEM_INFO *item, int32_t x, int32_t y, int32_t z, int16_t *height)
 {
+    if (g_Config.fix_bridge_grabs) {
+        if (!isSameSector(x, y, z, item)) {
```
rr-, line 188: Let's merge conditions

src/game/objects/bridge.c
```diff
@@ -159,6 +223,12 @@ void BridgeTilt2Floor(
 void BridgeTilt2Ceiling(
     ITEM_INFO *item, int32_t x, int32_t y, int32_t z, int16_t *height)
 {
+    if (g_Config.fix_bridge_grabs) {
+        if (!isSameSector(x, y, z, item)) {
```
rr-, line 227: Let's merge conditions

rr- (changes requested):

src/game/lara.c
walkawayy: Whoops fixed.

src/game/objects/bridge.c
walkawayy: How are these new names?

src/game/objects/bridge.c
walkawayy: Great point. Looks much cleaner and easier to follow.

src/game/objects/bridge.h
```diff
@@ -2,6 +2,7 @@
 
 #include "global/types.h"
 
+bool IsSameBridgeSector(int32_t x, int32_t y, int32_t z, const ITEM_INFO *item);
```
rr-, line 5: I believe we can make this function static within `bridge.c`, as it seems to be unused in other modules?

src/game/lara.c
```diff
@@ -2124,14 +2124,12 @@ bool LaraTestHangJumpUp(ITEM_INFO *item, COLL_INFO *coll)
 
     bounds = GetBoundsAccurate(item);
     hdif = coll->front_floor - bounds[FRAME_BOUND_MIN_Y];
-    if (hdif < 0 && hdif + item->fall_speed < 0) {
-        return false;
-    }
+    if (hdif < 0 && hdif + item->fall_speed < 0) { }
```
rr-, line 2127: This change looks sketchy, can you double check? If it's needed to be a no-op for the fix to work, let's hide it behind `g_Config` as well

rr- (commented):
Thanks for addressing my earlier remarks, found 2 more things

src/game/lara.c
walkawayy: I think I screwed up on this change. I pushed all my code because I was switching hard drives and clearly forgot I changed some time. I’ll review it later tonight since I’m busy right now.

src/game/objects/bridge.h
walkawayy: Yep. Done.

rr- (approved):
LGTM but let's have someone external test it :)

walkawayy:
Fixed the invisible wall edge case. Also ported all height variables to use int32_t since there was some int16_t back and forth. Tomb5 used all int32_t.

This will probably need some testing. I couldn't find any more edge cases or issues, but I don't know the game as well as others.

src/game/objects/bridge.c
```diff
@@ -137,27 +137,38 @@ void BridgeFlatCeiling(
...
+        offset = -z & 0x3FF;
+        // Fixes edge case of an invisible wall on the tilt2 bridge floor
+        // Offset would get set to 0 on a specific z pos on bottom of slope
+        // This sets the offset to the max value (1023) when Lara's at the
+        // bottom of the slope
+        if (g_Config.fix_bridge_grabs && offset == 0) {
+            if (y < item->pos.y) {
+                offset = (-z - 1) & 0x3FF;
+            }
+        }
```
oziphantom, line 159: use the constants not hard coded values ;)

oziphantom: -x and WALL_L-x are not the same, does this not cause more issues? as it seems from your comments the only issue was in the last case?

oziphantom (changes requested):

rr- (commented):
Why did you change the types to int32_t everywhere?
AFAIR ROOM_INFO is not capable of storing x/y/z bigger than int16_t.

src/game/objects/bridge.c
```diff
@@ -101,65 +113,105 @@ void DrawBridgeCollision(
...
-        return (WALL_L - x) & (WALL_L - 1);
-    } else if (item->pos.y_rot == -PHD_180) {
-        return x & (WALL_L - 1);
-    } else if (item->pos.y_rot == PHD_90) {
-        return z & (WALL_L - 1);
-    } else {
-        return (WALL_L - z) & (WALL_L - 1);
-    }
+    int32_t offset = 0;
+    if (item->pos.y_rot == 0)
```
rr-, line 143: We **do not** omit curly braces. Never.

rr- (changes requested):

src/game/objects/bridge.c
walkawayy: Not intentional. I'll fix. I was testing Tomb5's version of this function.

src/game/objects/bridge.c
```diff
@@ -150,15 +181,25 @@ void BridgeTilt1Ceiling(
...
+    }
+
     int32_t level = item->pos.y + (GetOffset(item, x, z) >> 1);
-    if (y <= level) {
-        *height = level;
+    if (y > level) {
+        return;
     }
+
+    *height = level;
```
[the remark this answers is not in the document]
rr-: I'm very afraid of this change, I feel like it has potential of having enormous impact and produce weird bugs.
Unless we specifically need it for this fix, I'd made a separate PR for it. WDYT?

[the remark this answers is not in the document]
walkawayy: Ok. I made the change thinking it was related to the bug and wanted to match [Tomb5's version](https://github.com/Trxyebeep/TOMB5/blob/ac24665b4f7a4e1c321614d8a906c8fab5f49f17/TOMB5/game/objects.cpp#L215). I'm happy to remove the change because the bug was unrelated.

Though it is odd that the code mixes the types around. But you're right, if the change is needed, it should be a separate PR.

src/game/objects/bridge.c
walkawayy: Fixed.

src/game/objects/bridge.c
walkawayy: Changed back to TR1 version of this.

src/game/objects/bridge.c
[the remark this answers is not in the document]
walkawayy: Changed height back to int16_t and restored the use of constants and braces here.

rr- (approved):
I think the conditions
```
        if (g_Config.fix_bridge_collision && offset == 0) {
            if (y < item->pos.y) {
```
can be condensed.

Also I'm not sure if we need to have this switch configurable, but let's leave it configurable for now IMO to have a trace in the git history.

walkawayy:
> I think the conditions
> 
> ```
>         if (g_Config.fix_bridge_collision && offset == 0) {
>             if (y < item->pos.y) {
> ```
> 
> can be condensed.
> 
> Also I'm not sure if we need to have this switch configurable, but let's leave it configurable for now IMO to have a trace in the git history.

Done. Ya makes sense to reduce it especially if the fix will no longer be optional in the future.

rr-:
I tested it and LGTM. Can we merge it? @walkawayy

walkawayy:
> I tested it and LGTM. Can we merge it? @walkawayy

Yep let's do it. Next release we can see if anyone reports issues with it, and remove the optional fix if there are no problems. Want me to or do you normally do the merging?

rr-:
For now I'd prefer to merge the PRs myself :)

PR #265, opened 2021-12-13
demo: fix crash if user inputs during demo ring close

### LINKED ISSUE #264: OG Bug: Crash when pressing any button while demo mode is loading
In the main menu, wait for the demo to load. When the menu ring is closing to start demo mode press ctrl.

tombati crashes to desktop. T1M crashes with this error message
Screenshot 2021-12-10 232916

walkawayy:
Close #264.

### DIFF
--- a/src/game/inventry.c
+++ b/src/game/inventry.c
@@ -108,7 +108,6 @@ int32_t Display_Inventory(int inv_mode)
 
         if (g_InvMode != INV_TITLE_MODE || g_Input.any || g_InputDB.any) {
             g_NoInputCount = 0;
-            g_ResetFlag = false;
         } else {
             if (!g_Config.disable_demo) {
                 g_NoInputCount++;


oziphantom (approved):

rr- (approved):
Can you describe the nuances of this fix?

walkawayy:
> Can you describe the nuances of this fix?
g_ResetFlag gets set to true after there's no user input for a while. This starts the ring closing animation. While that was happening, if the user pressed any button, the `if (g_InvMode != INV_TITLE_MODE || g_Input.any || g_InputDB.any) {` check was getting set to true, causing the `g_ResetFlag` to set to false. This is because the inventory loop ending condition of `} while (imo.status != RNG_DONE);` does not get set to done until the ring finishes closing.

So, once the loop finished and `` was no longer true, this check failed:
```
    if (g_ResetFlag) {
        return GF_START_DEMO;
    }
```
Then the game checks what item was chosen (nothing):
```
    switch (g_InvChosen) {
    case O_PASSPORT_OPTION:
```
so the case statement falls all the way to `return GF_NOP;` and the shell exits.

I was going to just make a `O_DEMO_OPTION` a case to check for in the `g_InvChosen` switch, but the `GAME_OBJECT_ID` enums are used in a bunch of places and there is a `O_NUMBER_OF = 191` cap, so adding to that didn't make sense.

walkawayy:
I think I'm the only one who tested. But it's not like we have a lot of testers to help.

rr-:
Worst case we'll get more bug reports, these seem easier to get a hold of…

PR #267, opened 2021-12-13
ui: fix default healthbar and airbar behavior

### LINKED ISSUE #266: Bug: Default healthbar and airbar not implemented
Setting `"xbar_showing_mode": "default",` in the JSON doesn't do anything. Instead, `flashing-or-default` behavior is used.

`T1M_BSM_DEFAULT` is missing from `m_BarShowingModes`.

walkawayy:
Closes #266.


src/game/overlay.c
```diff
@@ -243,18 +244,21 @@ void Overlay_DrawHealthBar()
     int32_t show = g_HealthBarTimer > 0 || hit_points <= 0
         || g_Lara.gun_status == LGS_READY;
     switch (g_Config.healthbar_showing_mode) {
-    case T1M_BSM_ALWAYS:
-        show = 1;
-        break;
-    case T1M_BSM_NEVER:
-        show = 0;
+    case T1M_BSM_DEFAULT:
+        show;
```
oziphantom, line 248: what is this suppose to be?

src/game/overlay.c
```diff
@@ -268,18 +272,21 @@ void Overlay_DrawAirBar()
     int32_t show = g_Lara.water_status == LWS_UNDERWATER
         || g_Lara.water_status == LWS_SURFACE;
     switch (g_Config.airbar_showing_mode) {
-    case T1M_BSM_ALWAYS:
-        show = 1;
-        break;
-    case T1M_BSM_NEVER:
-        show = 0;
+    case T1M_BSM_DEFAULT:
+        show;
```
oziphantom, line 276: and this one

oziphantom (changes requested):
at this point should one not also make show a bool since we have it now, as you are changing it anyway.

src/game/overlay.c
walkawayy: The DEFAULT cases don't need any extra options like the other cases, but I wanted to represent that it is a case.

oziphantom: so default doesn't need to modify show. 
which this "does" but also looks like 'somebody forgot to set the default value' at which point you could put a comment to explain, at which you could then just put the comment and not worry about odd "compiler doesn't mind" cases.

walkawayy: I’ll just change it to only a break. Idk why I thought to leave show; lol. Late night brain fart.

walkawayy: I should just remove that DEFAULT case and make a (C keyword) default case that just breaks.

bin/cfg/Tomb1Main.json5
```diff
@@ -93,17 +93,17 @@
...
     // Changes how the airbar is displayed. Possible values:
-    // - always: always show the airbar
-    // - default: show the airbar only in the water
-    // - flashing-or-default: show the airbar when Lara's oxygen is 20% or
-    //   below or under normal circumstances
-    // - flashing-only: show the airbar *only* when Lara's oxygen is 20% or /
+    // - default: show the airbar only in the water (as in OG)
+    // - flashing-or-default: flash the airbar when Lara's oxygen is 20% or
+    //   below, or show under normal circumstances
+    // - flashing-only: show and the airbar *only* when Lara's oxygen is 20% or /
```
rr-, line 99: No need for this trailing slash

bin/cfg/Tomb1Main.json5
```diff
@@ -93,17 +93,17 @@
...
     //   below (useful for challenge runs)
+    // - always: always show the airbar
     // - never: never display the airbar (for challenge runs)
     // In case of airbar, the flashing-or-default mode is there mostly for
     // consistency and the only practical difference is that it's shown when
     // Lara dies. When using the fly cheat, the airbar is never shown.
-    "airbar_showing_mode": "default",
+    "airbar_showing_mode": "flashing-or-default",
```
rr-, line 106: This mode does not make much sense TBH for the airbar. You'll never get a flashing airbar outside of a water body.

src/game/overlay.c
```diff
@@ -172,7 +172,8 @@ static void Overlay_DrawBar(int32_t value, int32_t value_max, int32_t bar_type)
     int32_t blink =
         percent <= blink_threshold && blink_time > blink_interval / 2;
 
-    if (percent && !blink) {
+    if ((percent && !blink)
+        || (g_Config.healthbar_showing_mode == T1M_BSM_DEFAULT)) {
```
rr-, line 176: Can you describe the rationale of this change?

walkawayy: In the PS1 OG game, the bars don’t flash when they are low. This makes it so they don't flash if they are low.

bin/cfg/Tomb1Main.json5
walkawayy: I didn't add this mode. Just rearranged the order to match the order of the json, enums, and case statement. Want me to remove it?

rr-: I'm not sure, do you think it makes sense to have it around?
When I added it I only made it to be consistent, but I think that might have been a mistake.

src/game/overlay.c
rr-: ! That's the thing I needed :)

src/game/overlay.c
```diff
@@ -243,17 +244,19 @@ void Overlay_DrawHealthBar()
...
         break;
     case T1M_BSM_NEVER:
         show = 0;
         break;
-    case T1M_BSM_FLASHING_OR_DEFAULT:
-        show |= hit_points <= (LARA_HITPOINTS * 20) / 100;
-        break;
-    case T1M_BSM_FLASHING_ONLY:
-        show = hit_points <= (LARA_HITPOINTS * 20) / 100;
+    default:
```
rr-, line 259: This `default: break` idiom is to avoid the compiler complaining about the switch missing certain cases right?

rr- (approved):

src/game/overlay.c
walkawayy: I don't think there's a compiler error for no default, but IMO it makes it clear that the default case is just `show` and we didn't forget about it.

src/game/overlay.c
walkawayy: I just checked. HP bar does flash if health is low, but it does not stay up on the screen.

bin/cfg/Tomb1Main.json5
walkawayy: I don't see a use for it, but I get your point about consistency. It's not harmful other than lengthening the list of options.

rr- (changes requested):
I tested it and I have two comments:

The healthbar in the default mode should be flashing for consistency with OG, whereas this PR removes it.
Please change the airbar back to `default`. The consequence of `flashing-or-default` is that the game will render empty airbar when Lara dies on the ground (her air is 0), which is a deviation from the OG.

src/game/overlay.c
rr-: Let's restore this behavior then.

walkawayy:
> 

Ok. I was thinking PS1 behavior was the default. If you see the 5 min mark at this [video](https://www.youtube.com/watch?v=UUNOovpJCho), there's no flashing for air or health bar. TombATI behaves slightly different. If we want the default behavior to flash when low, we should prob change the description. Currently it says:
```
   // - default: show the healthbar at the beginning of a level, after
    //   getting hit or while having weapons equipped (as in OG)
```
While `flashing-or-default` says:
```
   // - flashing-or-default: show the airbar when Lara's oxygen is 20% or
    //   below or under normal circumstances
```

That's why I changed `show` to `flash` on this. I think we just need to clean up the descriptions to more clearly define the behavior we want, then we can implement the code from there once we know exactly what we want.

walkawayy:
@rr- let me know what you prefer for default behavior on this. I'll update the description and behavior to match.

rr-:
I would like to keep compatibility with TombATI with `default` (possibly rename it to something else) and if we want to have compatibility with PS1, introduce a separate mode for it, e.g. not have the flashing part removed.

By default I would like to have the healthbar shown when Lara's health changes, or if it's under 20% in which case I'd like it to flash. This is `flashing-or-default`. I would like to have the airbar shown when she's underwater or on the water surface, which is `default`. (`default-or-flashing` doesn't make sense here as the only difference is that it will only be shown when Lara's dead on the ground, which looks like a bug. I'm OK with removing this mode altogether for the airbar.)

walkawayy:
How ugly is... Healthbar:
```
// Changes how the healthbar is displayed. Possible values:
// - default: show the healthbar at the start of a level, after getting
//   hit, or while having weapons equipped; flash the healthbar if it's already
//   on screen when Lara's health is 20% or below (TombATI)
// - flashing-or-default: constantly flash the healthbar when Lara's
//   health is 20% or below, or show under default circumstances
// - flashing-only: show the healthbar *only* when Lara's health is 20% or
//   below (useful for challenge runs)
// - always: always show the healthbar
// - never: never display the healthbar (for challenge runs)
// - psx: show the healthbar under default circumstances with no flashing
"healthbar_showing_mode": "flashing-or-default",
```

It's just slightly different for airbar because `flashing-or-default` would be `default` for air. Maybe:
```
// Changes how the airbar is displayed. Possible values:
// - default: show the airbar only in the water and flash the airbar when
//   Lara's oxygen is 20% or below (TombATI)
// - flashing-only: show the airbar *only* when Lara's oxygen is 20% or
//   below (for challenge runs)
// - never: never display the airbar (for challenge runs)
// - psx: show the healthbar under default circumstances with no flashing
```

rr-:
Looks good to me. I like that we don't break backwards compatibility with this.

walkawayy:
Lol idk I use PS1 and PSX. The reasoning is the console was code named PlayStation X. The gaming media called it PSX, and even though the X was dropped the name stuck. There was also no PS2 yet, so it wasn't called PS1 until PS2 started coming out. Let's just call it PS1 since that's clearer.


I also want to consider moving the blink code:
```
    const int32_t blink_interval = 20;
    const int32_t blink_threshold = bar_type == BT_ENEMY_HEALTH ? 0 : 20;
    int32_t blink_time = blink_counter++ % blink_interval;
    int32_t blink =
        percent <= blink_threshold && blink_time > blink_interval / 2;
```
to `Overlay_DrawHealthBar` and `Overlay_DrawAirBar` and calculate it in the case statements. Then pass blink as a param to `Overlay_DrawBar`. Any thoughts?

Also what is the difference between:
```
    case T1M_BSM_FLASHING_ONLY:
        show = g_Lara.air <= (LARA_AIR * 20) / 100;
        break;

    if (!show) {
        return;
    }
```
and `if ((percent && !blink)`? Does the `show` in the case statements cause the entire healthbar (including the border) to flash while the `blink` causes only the inner red color to flash?

rr-:
>Moving the code

OK so long as the new code guarantees us to have this:



and not this:

rr-:
Also to reduce amount of module-level variables consider wrapping the bits for each bar in its own struct.
Also consider moving `g_HealthBarTimer` to a module-level constant and add setter functions for use in other modules (`Overlay_SetHealtBarTimer`, `Overlay_TimerTick` or something like that).

src/game/overlay.c
```diff
@@ -78,12 +80,90 @@ static RGB888 m_ColorBarMap[][COLOR_STEPS] = {
...
+}
+
 static void Overlay_GetBarLocation(
     int8_t bar_location, int32_t width, int32_t height, int32_t *x, int32_t *y);
-static void Overlay_DrawBar(int32_t value, int32_t value_max, int32_t bar_type);
+static void Overlay_DrawBar(BAR_INFO *draw_bar);
 static void Overlay_OnAmmoTextRemoval(const TEXTSTRING *textstring);
 static void Overlay_OnFPSTextRemoval(const TEXTSTRING *textstring);
 
+static void Overlay_GetBlink(BAR_INFO *blink_bar)
```
walkawayy, line 143: Should I make the function bars (like `blink_bar`) all the same name since they are passed as pointers? I never know what people prefer to name as far as different function but same variables go.

src/game/overlay.c
```diff
@@ -12,6 +12,8 @@
 
 #include <stdio.h>
 
+#include "log.h"
```
rr-, line 15: I don't think we need this include, and if we do, it should be put with the rest of the `""` includes.

src/game/overlay.c
```diff
@@ -78,12 +80,90 @@ static RGB888 m_ColorBarMap[][COLOR_STEPS] = {
...
+    m_HealthBar.max_value = LARA_HITPOINTS;
+    m_HealthBar.show_mode = g_Config.healthbar_showing_mode;
+    m_HealthBar.show = 0;
+    m_HealthBar.blink = 0;
+    m_HealthBar.timer = 40;
+    m_HealthBar.color = g_Config.healthbar_color;
+    m_HealthBar.bar_location = g_Config.healthbar_location;
+}
+
+static BAR_INFO m_AirBar = { 0 };
```
rr-, line 109: Let's lump all static variables together, then all static functions, then all public functions.
Also all functions should have a forward declaration (I know they may not be needed, but this is for consistency so that we don't have to think about it.)

src/game/overlay.c
rr-: Let's just name it `bar` or `bar_info` across all functions :)

src/game/overlay.c
```diff
@@ -78,12 +80,90 @@ static RGB888 m_ColorBarMap[][COLOR_STEPS] = {
...
 static void Overlay_OnAmmoTextRemoval(const TEXTSTRING *textstring);
 static void Overlay_OnFPSTextRemoval(const TEXTSTRING *textstring);
 
+static void Overlay_GetBlink(BAR_INFO *blink_bar)
+{
+    if (blink_bar->show_mode == T1M_BSM_PS1) {
+        blink_bar->blink = 0;
+    }
+
+    static int32_t blink_counter = 0;
```
rr-, line 149: I think we can move this to `BAR_INFO`.

src/game/overlay.c
```diff
@@ -78,12 +80,90 @@ static RGB888 m_ColorBarMap[][COLOR_STEPS] = {
...
+    const int32_t percent_max = 100;
+
+    if (blink_bar->value < 0) {
+        blink_bar->value = 0;
+    } else if (blink_bar->value > blink_bar->max_value) {
+        blink_bar->value = blink_bar->max_value;
+    }
```
rr-, line 156: I think we can use `CLAMP` macro here.

src/game/overlay.c
```diff
@@ -78,12 +80,90 @@ static RGB888 m_ColorBarMap[][COLOR_STEPS] = {
...
 static void Overlay_OnFPSTextRemoval(const TEXTSTRING *textstring);
 
+static void Overlay_GetBlink(BAR_INFO *blink_bar)
+{
+    if (blink_bar->show_mode == T1M_BSM_PS1) {
+        blink_bar->blink = 0;
+    }
+
+    static int32_t blink_counter = 0;
+    const int32_t percent_max = 100;
```
rr-, line 150: I'm not sure if it makes sense to keep it as a variable, below we're using `* 100` as a literal numeral.

src/game/overlay.c
```diff
@@ -78,12 +80,90 @@ static RGB888 m_ColorBarMap[][COLOR_STEPS] = {
...
-static void Overlay_DrawBar(int32_t value, int32_t value_max, int32_t bar_type);
+static void Overlay_DrawBar(BAR_INFO *draw_bar);
 static void Overlay_OnAmmoTextRemoval(const TEXTSTRING *textstring);
 static void Overlay_OnFPSTextRemoval(const TEXTSTRING *textstring);
 
+static void Overlay_GetBlink(BAR_INFO *blink_bar)
+{
+    if (blink_bar->show_mode == T1M_BSM_PS1) {
+        blink_bar->blink = 0;
+    }
```
rr-, line 147: I think we're missing `return;` here to skip triggering the rest of the blink calculations.

src/game/overlay.c
```diff
@@ -78,12 +80,90 @@ static RGB888 m_ColorBarMap[][COLOR_STEPS] = {
...
+
+    if (blink_bar->value < 0) {
+        blink_bar->value = 0;
+    } else if (blink_bar->value > blink_bar->max_value) {
+        blink_bar->value = blink_bar->max_value;
+    }
+    int32_t percent = blink_bar->value * 100 / blink_bar->max_value;
+
+    const int32_t blink_interval = 20;
+    const int32_t blink_threshold = blink_bar->type == BT_ENEMY_HEALTH ? 0 : 20;
```
rr-, line 160: TBH I'd move this check (for the enemy healthbar) to the beginning of the function and make a similar early exit.

src/game/overlay.c
```diff
@@ -113,40 +193,27 @@ static void Overlay_GetBarLocation(
...
-        value = 0;
-    } else if (value > value_max) {
-        value = value_max;
+    if (draw_bar->value < 0) {
+        draw_bar->value = 0;
+    } else if (draw_bar->value > draw_bar->max_value) {
+        draw_bar->value = draw_bar->max_value;
     }
-    int32_t percent = value * 100 / value_max;
+    int32_t percent = draw_bar->value * 100 / draw_bar->max_value;
```
rr-, line 205: We can use `CLAMP` value. Also since we duplicate the code from above, maybe it's good to introduce a function for this `Overlay_BarGetPercent(BAR_INFO *bar)`

src/game/overlay.c
```diff
@@ -113,40 +193,27 @@ static void Overlay_GetBarLocation(
...
-        bar_color = g_Config.healthbar_color;
-    } else if (bar_type == BT_LARA_AIR) {
-        Overlay_GetBarLocation(g_Config.airbar_location, width, height, &x, &y);
-        bar_color = g_Config.airbar_color;
-    } else if (bar_type == BT_ENEMY_HEALTH) {
-        Overlay_GetBarLocation(
-            g_Config.enemy_healthbar_location, width, height, &x, &y);
-        bar_color = g_Config.enemy_healthbar_color;
-    }
+    Overlay_GetBarLocation(draw_bar->bar_location, width, height, &x, &y);
```
rr-, line 216: Consider making `Bar` a common prefix across all bar functions i.e. `Overlay_BarGetLocation` etc. to lump those functions together.

src/game/overlay.c
```diff
@@ -214,98 +278,114 @@ void Overlay_Init()
...
+void Overlay_SetHealtBarTimer(int16_t health_bar_timer)
+{
+    m_HealthBar.timer = health_bar_timer;
+}
+
+void Overlay_TimerTick()
+{
+    m_HealthBar.timer--;
+    if (m_HealthBar.timer < 0) {
+        Overlay_SetHealtBarTimer(0);
```
rr-, line 296: I think we can use one of the variants of the `CLAMP` function here.

src/game/overlay.c
```diff
@@ -214,98 +278,114 @@ void Overlay_Init()
...
+    }
 }
 
 void Overlay_DrawHealthBar()
 {
     static int32_t old_hit_points = 0;
 
-    for (int i = 0; i < 6; i++) {
+    for (int32_t i = 0; i < 6; i++) {
```
rr-, line 304: Let's leave it as `int` IMO, although no strong feeling about this

src/game/overlay.c
```diff
@@ -214,98 +278,114 @@ void Overlay_Init()
...
-    int hit_points = g_LaraItem->hit_points;
-    if (hit_points < 0) {
-        hit_points = 0;
-    } else if (hit_points > LARA_HITPOINTS) {
-        hit_points = LARA_HITPOINTS;
+    m_HealthBar.value = g_LaraItem->hit_points;
+    if (m_HealthBar.value < 0) {
+        m_HealthBar.value = 0;
+    } else if (m_HealthBar.value > m_HealthBar.max_value) {
+        m_HealthBar.value = m_HealthBar.max_value;
```
rr-, line 312: We can use `CLAMP` for this

src/game/overlay.c
```diff
@@ -214,98 +278,114 @@ void Overlay_Init()
...
     }
 
-    if (old_hit_points != hit_points) {
-        old_hit_points = hit_points;
-        g_HealthBarTimer = 40;
+    if (old_hit_points != m_HealthBar.value) {
+        old_hit_points = m_HealthBar.value;
+        Overlay_SetHealtBarTimer(40);
```
rr-, line 317: I think it's OK to just use `m_HealthBar.timer` directly here, since it's just a simple assignment, but no strong feelings about it.

src/game/overlay.c
```diff
@@ -214,98 +278,114 @@ void Overlay_Init()
...
         || g_Lara.gun_status == LGS_READY;
-    switch (g_Config.healthbar_showing_mode) {
+    switch (m_HealthBar.show_mode) {
+    case T1M_BSM_FLASHING_OR_DEFAULT:
+        m_HealthBar.show |=
+            m_HealthBar.value <= (m_HealthBar.max_value * 20) / 100;
+        break;
+    case T1M_BSM_FLASHING_ONLY:
+        m_HealthBar.show =
+            m_HealthBar.value <= (m_HealthBar.max_value * 20) / 100;
```
rr-, line 329: I think `20` should be moved as a module-level constant for all usages. A name such as `BLINK_THRESHOLD`.

src/game/overlay.c
```diff
@@ -78,12 +80,90 @@ static RGB888 m_ColorBarMap[][COLOR_STEPS] = {
       { 120, 60, 70 } },
 };
 
+typedef struct BAR_INFO {
+    int32_t type;
```
rr-, line 84: Consider making it a `BAR_TYPE`

src/game/overlay.c
```diff
@@ -78,12 +80,90 @@ static RGB888 m_ColorBarMap[][COLOR_STEPS] = {
...
+    int32_t value;
+    int32_t max_value;
+    int32_t show;
+    int32_t show_mode;
```
rr-, line 88: Consider making this a `T1M_BAR_SHOW_MODE`

src/game/overlay.c
```diff
@@ -78,12 +80,90 @@ static RGB888 m_ColorBarMap[][COLOR_STEPS] = {
...
+    int32_t blink;
```
rr-, line 89: Consider making it a `bool`

src/game/overlay.c
```diff
@@ -78,12 +80,90 @@ static RGB888 m_ColorBarMap[][COLOR_STEPS] = {
...
+    int32_t timer;
+    int32_t color;
+    int32_t bar_location;
```
rr-, line 92: Consider making it a `T1M_BAR_LOCATION`

src/game/overlay.c
```diff
@@ -214,98 +278,114 @@ void Overlay_Init()
...
 
 void Overlay_DrawAirBar()
 {
-    int32_t show = g_Lara.water_status == LWS_UNDERWATER
+    m_AirBar.value = g_Lara.air;
+    if (m_AirBar.value < 0) {
+        m_AirBar.value = 0;
+    } else if (g_Lara.air > m_AirBar.max_value) {
+        m_AirBar.value = m_AirBar.max_value;
+    }
```
rr-, line 354: We can use `CLAMP`

src/game/overlay.h
```diff
@@ -4,6 +4,8 @@
 
 void Overlay_Init();
 
+void Overlay_SetHealtBarTimer(int16_t health_bar_timer);
```
rr-, line 7: Just `timer` will do

src/game/setup.c
```diff
@@ -119,7 +119,7 @@ int32_t InitialiseLevel(int32_t level_num, GAMEFLOW_LEVEL_TYPE level_type)
 
     Overlay_Init();
 
-    g_HealthBarTimer = 100;
+    Overlay_SetHealtBarTimer(100); // TODO Move this to Overlay_Init()?
```
rr-, line 122: Not sure. On one hand this is a good decision, on another here we're very explicit about flashing the health for 100 frames(?) at the start of the level. IMO let's leave it as-is, maybe move the blank line above under this line to lump `Overlay_` calls together.

rr- (changes requested):
Great changes, I have only cosmetic comments.

src/game/overlay.c
walkawayy: Honestly I don't understand the point of `blink_counter`. It's only used in one line:
`    int32_t blink_time = blink_counter++ % blink_interval;`

Isn't it always just going to be `1 % blink_interval`?

walkawayy: Also went with `bar_info` for easier differentiation when searching and reading the code IMO.

src/game/overlay.c
walkawayy: Enemy health bars should never blink? I was considering making it behave the same as the user health bar setting. But it probably doesn't make sens e to do that. I also need to make the blink threshold a constant.

src/game/overlay.c
walkawayy: Ok. I changed it since most of the ints are given defined types and it seemed inconsistent. But it shouldn't matter for a loop counter.

src/game/overlay.c
walkawayy: I wasn't sure for this either. I decided to use the function because every other time it's set uses that. But local in the function it could be the simple assignment since that's how the rest of it is set I guess.

src/game/overlay.c
walkawayy: I originally had it as that, but then it became `m_HealthBar.bar_type` which seemed like unneeded characters.

src/game/overlay.c
walkawayy: Didn't even realize / think to do that lol. Mind blown.

src/game/overlay.c
rr-: It's not a bad practice.

Because `blink_counter` is declared as static, it behaves like a module-level variable. So it'll be 1, 2, 3, … MAX_INT, 0, 1,   etc. Declaring static variables in function bodies is a bad practice, because it introduces global state in a stealthy way. Hence my proposal to move it to `BAR_INFO`.

walkawayy:
I think I addressed all of the review comments. Tried to list them all in the commit message.

bin/cfg/Tomb1Main.json5
```diff
@@ -61,15 +61,17 @@
...
+    // - flashing-or-default: constantly flash the healthbar when Lara's
+    //   health is 20% or below, or show under default circumstances
     // - flashing-only: show the healthbar *only* when Lara's health is 20% or
-    //   below (useful for challenge runs)
+    //   below (for challenge runs)
+    // - always: always show the healthbar
     // - never: never display the healthbar (for challenge runs)
-    "healthbar_showing_mode": "flashing-or-default",
+    // - ps1: show the healthbar under default circumstances with no flashing
+    "healthbar_showing_mode": "default",
```
rr-, line 74: I think we agreed to keep it as `flashing-or-default`; from yesterday's comment:
>By default I would like to have the healthbar shown when Lara's health changes, or if it's under 20% in which case I'd like it to flash. This is `flashing-or-default`

src/game/overlay.c
```diff
@@ -78,14 +79,105 @@ static RGB888 m_ColorBarMap[][COLOR_STEPS] = {
...
+    int32_t value;
+    int32_t max_value;
+    int32_t percent;
+    bool show;
+    T1M_BAR_SHOW_MODE show_mode;
+    bool blink;
+    int32_t blink_counter;
+    int32_t timer;
+    int32_t color;
+    T1M_BAR_LOCATION bar_location;
```
rr-, line 93: IMO we can drop `T1M_` prefix from these enums, and simplify `bar_location` to `location`

src/game/overlay.c
```diff
@@ -78,14 +79,105 @@ static RGB888 m_ColorBarMap[][COLOR_STEPS] = {
...
 };
 
-static void Overlay_GetBarLocation(
-    int8_t bar_location, int32_t width, int32_t height, int32_t *x, int32_t *y);
-static void Overlay_DrawBar(int32_t value, int32_t value_max, int32_t bar_type);
+typedef struct BAR_INFO {
+    BAR_TYPE type;
+    int32_t value;
+    int32_t max_value;
+    int32_t percent;
```
rr-, line 86: `percent` duplicates what can be calculated from `value` and `max_value`, I'd rather we'd store only one source of truth – in the current form it is unclear which is more valid during calculations.

src/game/overlay.c
```diff
@@ -78,14 +79,105 @@ static RGB888 m_ColorBarMap[][COLOR_STEPS] = {
...
+    m_EnemyBar.blink = false;
+    m_EnemyBar.blink_counter = 0;
+    m_EnemyBar.timer = 0;
+    m_EnemyBar.color = g_Config.enemy_healthbar_color;
+    m_EnemyBar.bar_location = g_Config.enemy_healthbar_location;
+}
+
+static void Overlay_BarGetPercent(BAR_INFO *bar_info)
+{
+    bar_info->percent = bar_info->value * 100 / bar_info->max_value;
```
rr-, line 159: To continue my earlier comment – here I'd prefer for us to just return the calculated value instead of store it in the `BAR_INFO`.

src/game/overlay.c
```diff
@@ -113,40 +205,20 @@ static void Overlay_GetBarLocation(
...
-    static int32_t blink_counter = 0;
-    const int32_t percent_max = 100;
-
-    if (value < 0) {
-        value = 0;
-    } else if (value > value_max) {
-        value = value_max;
-    }
-    int32_t percent = value * 100 / value_max;
+    int32_t percent = bar_info->value * 100 / bar_info->max_value;
```
rr-, line 210: Here we could use `int32_t percent = Overlay_BarGetPercent(bar_info);`

rr- (commented):
Almost there :)

bin/cfg/Tomb1Main.json5
walkawayy: Ya sorry. Changed it to TombATI and forgot to change it back.

rr- (commented):
I didn't notice one more change in the config

walkawayy:
> I didn't notice one more change in the config – by changing the default bar location we lose compatibility with OG TombATI, TR2 and TR3; possibly more games.

Ah sorry again. I keep mixing up my windows with the config for my game folder and my dev folder. Happens because I use a script to copy everything out of WSL.

rr- (approved):
LGTM.

PR #269, opened 2021-12-14
bin: change .rpl to .avi

### LINKED ISSUE #259: Feature request: Add option near top of gameflow notepad to change format for all FMV videos.
Hi, I just wanted to suggest that it might be more user-friendly if there's one overall option in gameflow to choose a FMV format.  For example choosing "avi" would mean all avi type FMVs would play instead of rpl ones.

rr-:
Closes #259

We have a mechanism in place that guesses the FMV file extension. This pull request ensures that if the user has both .rpl and high res .avi movies, the game will choose .avi first. If .avi does not exist, the game will try FMVs sequentially according to the predefined list of hardcoded file extensions, until it reaches .rpl which it will play as usual.

### DIFF
--- a/bin/cfg/Tomb1Main_gameflow.json5
+++ b/bin/cfg/Tomb1Main_gameflow.json5
@@ -56,7 +56,7 @@
 
             // list of actions to execute when this level is played
             "sequence": [
-                {"type": "play_fmv", "fmv_path": "fmv\\mansion.rpl"},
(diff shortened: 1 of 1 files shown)


walkawayy (approved):

walkawayy:
Idk if you want to add the 2.1 milestone or not. Changes look good.

rr-:
Let's go with it, since it's a small change.

PR #271, opened 2021-12-14
ai/evil-lara: fix malformed savegame flags

rr-:
The main gist of this pull request is the removal of `obj->save_flags = 1;`, the rest are debugging leftovers which are either aesthetic improvements, or useful functions as is the case with `File_GetPos` which I long meant to add.

@walkawayy could you please test, and confirm that the savegame behavior around Bacon Lara is consistent with OG?

### DIFF
--- a/src/filesystem.c
+++ b/src/filesystem.c
@@ -130,6 +130,12 @@ void File_Seek(MYFILE *file, size_t pos, FILE_SEEK_MODE mode)
     }
 }
 
+size_t File_GetPos(MYFILE *file)
+{
+    assert(file);
+    return ftell(file->fp);
+}
+
 size_t File_Size(MYFILE *file)
 {
     size_t old = ftell(file->fp);
--- a/src/filesystem.h
+++ b/src/filesystem.h
@@ -29,6 +29,7 @@ size_t File_Write(
     const void *data, size_t item_size, size_t count, MYFILE *file);
 size_t File_Size(MYFILE *file);
 void File_Seek(MYFILE *file, size_t pos, FILE_SEEK_MODE mode);
+size_t File_GetPos(MYFILE *file);
 void File_Close(MYFILE *file);
 int File_Delete(const char *path);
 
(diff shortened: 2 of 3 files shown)


walkawayy:
After a save and reload, Bacon Lara no longer moved or faced me. The only thing she did was pull out her weapon and shoot if I did but she didn't aim at me.

She also broke when I was trying the fly cheat but idk if that's really a bug.

rr-:
How is this behaving in the OG?

walkawayy:
I was not able to load my `ai/evil-lara: fix` save in Atlantis on the master branch. I also was not able to load my master branch save on `ai/evil-lara: fix`.

Also, I did some testing on master and evil lara actually does properly move during the fly cheat. She also properly behaves after a save is reloaded. So, something you changed is breaking her and doesn't seem to fix the save incompatibility issue.

rr-:
You'd need to test 100% with TombATI. SetupEvilLara had wrong flags since forever, and saves made in T1M until this fix are incompatible.

rr-:
I reverted some casts, could you check again?

walkawayy:
> 

Bacon Lara still doesn't do anything after a save and reload.

rr-:
So what do we do? It's do or die – either we lose backwards compatibility with OG saves for Atlantis and decide to save Bacon Lara flags, fixing the behavior above, or we keep the compatibility but purposefully break Bacon Lara upon saving and reloading.

walkawayy:
So I just double checked TombATI, and Bacon Lara definitely breaks on save/reload.

Is there any way to recover Bacon Lara if an OG/ATI save is imported? Like, if a save is loaded in Atlantis, can we check if the evil Lara flags are there? And if not, somehow set her to be mirrored to Lara's loaded position? Sounds kind of crazy, but I don't fully know how you implemented the fix that broke compatibility.

Honestly, I would say making her behave properly is more important than broken backward compatibility? Because if she doesn't work right, the level is not able to be finished anyway. I would think the subset of people moving from an OG Atlantis save to Tomb1-Main would be really small. The only use case I could see would be downloadable saves from something like Stella's Tomb Raider site. But I don't think OG saves work in TombATI anyway.

Issue #273, opened 2021-12-15
Bug: load game selector is off screen If your last save is in slot 13+


nfraihat:
Steps to reproduce:

Create a new save in save slot 13 (saveati.12)
F6 > you can't see the selector
press up twice, or down 5 times to see it

-----
Current behaviour
![image](https://user-images.githubusercontent.com/49598539/146262546-6155b9b6-2958-4de9-a732-0c18a7133b27.png)

Expected
![image](https://user-images.githubusercontent.com/49598539/146262452-b1912c4c-7bb1-499d-9d33-aeb693358d61.png)

walkawayy:
This also seems to be broken for the save game selection. I started to take a look at it, and the ring inventory code is pretty ugly lol. Will probably refactor some of it while doing this.

oziphantom:
this has been on my fix list for a while, I was going to tackle it when I doubled the number of save slots. 
Oddly it does have the right row selected, so you can just press "action" to select it, its just off screen.

rr-:
Not sure if OG bug or not, assuming ours. Fix in #274

walkawayy:
Noticed the inverse of this happening. If you load a game in the last few slots, but your most recent save is at the top, the most recent save is selected at the top of the list off screen.

![image](https://user-images.githubusercontent.com/81546780/148446275-41da430c-ec51-4b5e-a2f9-18f04fb01ba6.png)

rr-:
@walkawayy can you please close this bug and open a new one instead? Reopening tickets that were already "released" makes it problematic to track milestones properly.

PR #274, opened 2021-12-16
Fix offscreen save selection

rr-:
Related: #273


walkawayy:
Works great for me! Tested it with load and save in different slots including the last slot. I also tested at the lowest resolution of 320x300.

walkawayy (approved):

walkawayy:
Noticed the inverse of this happening. If you load a game in the last few slots, but your most recent save is at the top, the most recent save is selected at the top of the list off screen.

image

PR #282, opened 2021-12-19
lara: fix hang bug for stacked rooms

### LINKED ISSUE #115: OG Bugfix: Stacked rooms allow climb ledge onto wall
When you have 2 rooms stacked, where the upper room has a ledge and the lower room has a wall, Lara can climb into the wall when she climbs leftwards.
This is an original TR1 Bug and appears in St. Francis Folly. For demonstration I made a Level with TE.
This is NOT an "invisible ledge" where ceiling/floor are on the same height. It is an actual wall sector!

[Video](https://streamable.com/h3ygdn)

[St. Francis Folly](https://streamable.com/d46uia)

walkawayy:
Resolves #115.

### DIFF
--- a/src/game/lara.c
+++ b/src/game/lara.c
@@ -1809,9 +1809,9 @@ void LaraHangTest(ITEM_INFO *item, COLL_INFO *coll)
     bounds = GetBoundsAccurate(item);
     int32_t hdif = coll->front_floor - bounds[FRAME_BOUND_MIN_Y];
 
-    if ((ABS(coll->left_floor - coll->right_floor) >= SLOPE_DIF)
-        || (coll->mid_ceiling >= 0) || (coll->coll_type != COLL_FRONT)
-        || (hdif < -SLOPE_DIF) || (hdif > SLOPE_DIF) || (flag)) {
+    if (ABS(coll->left_floor - coll->right_floor) >= SLOPE_DIF
+        || coll->mid_ceiling >= 0 || coll->coll_type != COLL_FRONT
+        || hdif < -SLOPE_DIF || hdif > SLOPE_DIF || flag) {
         item->pos.x = coll->old.x;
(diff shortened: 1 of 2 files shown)


walkawayy:
I'm unsure if the ceiling check is needed. I would also like more places to test if anyone has any suggestions.

src/game/lara.c
```diff
@@ -1806,8 +1806,12 @@ void LaraHangTest(ITEM_INFO *item, COLL_INFO *coll)
...
     }
 
-    if (ABS(coll->left_floor - coll->right_floor) >= SLOPE_DIF
-        || coll->mid_ceiling >= 0 || coll->coll_type != COLL_FRONT || flag) {
+    bounds = GetBoundsAccurate(item);
+    int32_t hdif = coll->front_floor - bounds[FRAME_BOUND_MIN_Y];
+
+    if ((ABS(coll->left_floor - coll->right_floor) >= SLOPE_DIF)
+        || (coll->mid_ceiling >= 0) || (coll->coll_type != COLL_FRONT)
+        || (hdif < -SLOPE_DIF) || (hdif > SLOPE_DIF) || (flag)) {
```
rr-, line 1814: We don't need all those braces :)

rr- (approved):
Apart from one cosmetic remark, @walkawayy are we ready to merge it? I saw your comment in the original issue that you tested the fix on a lot of ledges, just wanted to confirm that it's up to date.

src/game/lara.c
walkawayy: Lol it was for my sanity to read it clearly when testing.

I couldn't find any more issues in the St. Francis Folly room. I also tested the invisible crack ledges and they still work. I think it's good to merge.

rr- (approved):

PR #285, opened 2021-12-22
output: improve VBuf fidelity

rr-:
Closes #283.

Copied TR2Main's approach to VBUFs.

PR #286, opened 2021-12-22
Make anisotropy filter value configurable

rr-:
The motivation for this PR has two sources:

- TombATI makes it possible to change the anisotropy filter value
- I am facing a problem on wine that makes the game unplayable if there are many textures and the anisotropy filter is enabled

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,8 @@ lint:
 
 test_base:
 	cp build/*.exe test/
+
+test_bin:
 	rsync -r bin/ test/
 
 test: build test_base
@@ -38,4 +40,4 @@ test: build test_base
 test_gold: build test_base
 	WINEARCH=win32 MESA_GL_VERSION_OVERRIDE=3.3 wine test/Tomb1Main.exe -gold
 
-.PHONY: debug debugopt release clean lint test_base test test_gold
+.PHONY: debug debugopt release clean lint test_base test_bin test test_gold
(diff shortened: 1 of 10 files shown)


walkawayy:
I tested this in Caves, and the wolf tracks look a lot better with anisotropic filtering. However, there are now black line artifacts between the floor textures that seem to show up. I merged in the Vbuf fidelity to see if that was needed to fix it, but it's still there. Here's some screenshots. Look at the white floor in the distance with the filter on. There's like black dashing lines. It's really obvious in motion.

Here's a filter value of 0:
screenshot0003


Here's a filter value of 16.0:
screenshot0000

walkawayy:
I zoomed in to make it more obvious. This is with the 16 filter on:
image

rr-:
@walkawayy are you saying that #285 breaks the anisotropic filter?
I agree that disabling the filter looks worse; my only motivation for this PR  is to let me play the game on Wine, because without this fix the game runs at 1 frame per minute…

walkawayy:
Hmm you're right. The issue is present on master. I must not have noticed before and only noticed because I was looking for the filtering? Would the fix you applied to the walls having this issue work for the floors too?

Seems like this PR is good then.

walkawayy (approved):

rr-:
>BTW: Does this make the perspective setting from the sunglasses in the menu work?

No :D that is hardcoded and it makes sense to do so, because who would want to watch the textures morph? But Ialready have plans on what to put there instead :)

PR #291, opened 2022-01-04
gfx: remove tiny resolutions

### LINKED ISSUE #275: Bug: at 320x200 resolution the details screen is scaled improperly
This is super low priority, and I don't know if anyone even plays at a resolution this low, but the details screen is improperly scaled at the lowest resolution of 320x300. I noticed while checking the load/save game slot fix. I checked TombATI to see if it happens there, but they only have perspective and bilinear options in game.

image

rr-:
Closes #275

This will force the players to set their resolution again after the update, because we store indexes within this array rather than values.


walkawayy:
Good riddance to those low resolutions!

walkawayy (approved):

PR #292, opened 2022-01-04
s/output: clear picture surface before blit

### LINKED ISSUE #288: TR1M bug: No black background on main menu after end of manor level
Hi, hope everyone have had a good Christmas!

Just to report a bug I've encountered, when Lara climbs out of the swimming pool and the manor level ends, this is what the main menu looks like:

screenshot0001

This happens on TR1M version 2.1

rr-:
Closes #288.

I couldn't reproduce the original issue so this is a blind fix.

### DIFF
--- a/src/specific/s_output.c
+++ b/src/specific/s_output.c
@@ -747,7 +747,10 @@ void S_Output_DownloadPicture(const PICTURE *pic)
         .bottom = target_rect.top + new_height,
     };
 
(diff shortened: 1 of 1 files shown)


walkawayy:
Ya I can't get it to happen either. Code looks fine to me and the user said the build you gave them fixed the issue. So, I say let's go for it.

walkawayy (approved):

Issue #287, opened 2021-12-24
Bug: Game crashes after fmv sequence on a specific folder in version 2.1


vini1264:
I have Tomb1Main installed on a folder called Tomb Raider in this specific path "C:\Games\Tomb Raider", and when the intro sequence ends the main menu doesn't load and it crashes to the desktop. If I move the folder or rename it to something else the game loads fine after the FMV sequence and doesn't crash. This bug wasn't present in the previous version, it's only present in version 2.1.

walkawayy:
I tried this, and it still works for me when I run from the folder.

![image](https://user-images.githubusercontent.com/81546780/148010105-b884efed-f627-4ed5-9e0c-9ca878649d67.png)

rr-:
We'll need more details what kind of crash. Missing files? Hard crash?

vini1264:
Sorry for taking so long to respond. For me the application simply exits out of nowhere after the intro sequence, and again, it's only on the directory "C:\Games\Tomb Raider" that I told before, if I move or rename it it runs as expected and doesn't crash after the intro movie. I have checked all the files on the folder and nothing seems to be missing, i've tested the latest version and the same thing happens.

Here's the log file if you want to take a look:
[Tomb1Main.log](https://github.com/rr-/Tomb1Main/files/7876465/Tomb1Main.log)

viniciusferrao:
Same issue here, but it's D:\ instead of C:\. The space on the directory name may be the one to blame.

walkawayy:
The default Steam directory `C:\Program Files (x86)\Steam\steamapps\common\Tomb Raider (I)` has a space which works fine. So something else is probably causing the issue I'd think?

rr-:
Idea: does the game have permissions to write things in that directory? What happens if we try to run a game from a directory marked as read-only?

vini1264:
Well, I've found another crash on the game related to the directory. I have renamed the folder to "TombRaider", without spaces and located on the same "C:\Games" directory. This time the game doesn't crash after the intro, but when I load the Lara's House tutorial level, pause the game and try to exit to the main menu the game ends up crashing instead of going to the menu, on the normal game this crash doesn't happen, only on Lara's House. This is happening on the latest version 2.4 btw.

Here's the log file related to the crash to take a look:
[Tomb1Main.log](https://github.com/rr-/Tomb1Main/files/7899423/Tomb1Main.log)

walkawayy:
[...]

PR #293, opened 2022-01-05
lava: fix Lara getting set on fire on trapdoors over Lava

### LINKED ISSUE #2: OG bugfix: fix trapdoor setting you on fire
Current behavior:

Sometimes under rare circumstances jumping onto a drawbridge above a lava pit can cause Lara to catch on fire.

Desired behavior:

Lara doesn't catch fire.

walkawayy:
Resolves #2.


walkawayy:
Lol I fixed the tilt lava issue, but didn't realize Lara is now setting on fire on the bridge again. o.o Made this a draft PR.

rr- (approved):

rr-:
@walkawayy do we wait for #287 with 2.2? Doesn't sound like we'll be getting anywhere with that ticket.

walkawayy:
I think we need more info on that one to solve it.

Also I have no clue how to solve the audio popping.

rr-:
Yes, my point is that I think these can wait for 2.3 (or some other future version) and we can release 2.2 as-is.

walkawayy:
Yep agreed.

PR #298, opened 2022-01-05
shell: save screenshots in a folder

walkawayy:
Saves screenshots in a `screenshots` folder in the game directory. They are named based on the level name from the gameflow file plus a number.


src/game/text.c
```diff
@@ -56,6 +56,14 @@ static int8_t m_TextASCIIMap[95] = {
 
 static void Text_DrawText(TEXTSTRING *textstring);
 
+void Text_ReplaceChar(char *str, char orig_char, char rep_char)
+{
+    char *ix = str;
+    while ((ix = strchr(ix, orig_char)) != NULL) {
```
rr-, line 62: No need to be explicit about `!= NULL` but no strong opinion on this.

src/game/text.c
```diff
@@ -56,6 +56,14 @@ static int8_t m_TextASCIIMap[95] = {
 
 static void Text_DrawText(TEXTSTRING *textstring);
 
+void Text_ReplaceChar(char *str, char orig_char, char rep_char)
```
rr-, line 59: I don't think this should be put here. This module is about displaying TEXTSTRING not manipulation of `char*`. I'd move this to `game/shell.c` and inline this logic under a new function, `GetScreenshotName`.

src/game/shell.c
```diff
@@ -199,9 +199,21 @@ bool Shell_MakeScreenshot()
...
     for (int i = 0; i < 10000; i++) {
-        sprintf(path, "screenshot%04d.%s", i, ext);
+#if defined(_WIN32)
+    #include <direct.h>
+        _mkdir("screenshots");
+#else
+    #include <sys/types.h>
+    #include <sys/stat.h>
+        mkdir("screenshots", 0664);
+#endif
```
rr-, line 211: We don't do `#ifdefs` in `game/`. Also this logic doesn't belong here. Please make a `S_File_CreateDirectory` function and create a proxy for it in `filesystem.c` that gets called here.

src/game/shell.c
```diff
@@ -199,9 +199,21 @@ bool Shell_MakeScreenshot()
...
+        char level_title[100];
+        sprintf(
+            level_title, "%s", g_GameFlow.levels[g_CurrentLevel].level_title);
+        Text_ReplaceChar(level_title, ' ', '-');
```
rr-, line 215: I'd prefer underscores over `-`.

src/game/shell.c
```diff
@@ -199,9 +199,21 @@ bool Shell_MakeScreenshot()
...
+    #include <direct.h>
+        _mkdir("screenshots");
+#else
+    #include <sys/types.h>
+    #include <sys/stat.h>
+        mkdir("screenshots", 0664);
+#endif
+        char level_title[100];
+        sprintf(
+            level_title, "%s", g_GameFlow.levels[g_CurrentLevel].level_title);
```
rr-, line 214: This has potential of creating illegal filenames.

https://stackoverflow.com/a/31976060
I suggest stripping all characters that are not A-Z or 0-9.

rr- (changes requested):

src/game/shell.c
walkawayy: Ok. Ngl I’m not sure how a proxy works.

rr-: Proxy in context of T1M is when we avoid inclusion of platform-specific code by game modules. Example: `File_GetGameDirectory` is a proxy for `S_File_GetGameDirectory`.

This concept is less than ideal and I want to do without it at some point, but for now it's better to continue this single concept rather than coming up with new ones to track.

src/game/shell.c
walkawayy: Level title named `zs<> .,..,?!911 -lol-Caves` saved as `zs_911_lolCaves-0000.jpg` (haven't done timestamp yet). So I think I got the replace + strip done. Not as trivial to do in C as it is in Python haha.

rr-: I knowwww I hate C for this. Thanks for persevering :heart:

walkawayy: I'm going hard and naming it `Shell_ValidateLevelName`. Seems odd to only do it on screenshots. Right now, empty string levels and only bad char levels are accepted.

walkawayy:
Alright everything works.  I'm not sure how exactly you want the proxy clock code handled, so idk if you will be happy with that part lol. `yyyymmdd-hhmmss` timestamps are added to the screenshots with `0000`, `0001`, etc added if you rapidly press print screen during the same second. The game should also now handle custom level names better and format them correctly for screenshots.

The only edge case I noticed was if multiple `.` or `'` are in the level name in a row. The game needs to support those for stuff like `St. Francis' Folly`. The screenshots will work but like `Test...Level` would screenshot as `Test___Level` with multiple underscores. Tbh I don't think it's a huge deal, but it's the one thing I noticed and wanted to mention.

src/game/clock.c
```diff
@@ -21,3 +23,42 @@ int32_t Clock_SyncTicks(int32_t target)
...
+}
+
+char *Clock_GetDateTime(char *date_time)
+{
+    sprintf(
+        date_time, "%04d%02d%02d_%02d%02d%02d", Clock_GetYear(),
+        Clock_GetMonth(), Clock_GetDay(), Clock_GetHours(), Clock_GetMinutes(),
+        Clock_GetSeconds());
+    return date_time;
+}
```
rr-, line 64: I don't think we use `Clock_GetMinutes` etc. publicly. Since I think it's safe to rely on `libc`, I'd inline all of these in this function, `Clock_GetDateTime` and not bother putting these in `s_clock.c`.

`Clock_GetDateTime` `sprintf`s the datetime and as such its signature is confusing. I'd either make it return a pointer to a static buffer (which is not thread-safe) or make it `void` to avoid confusion.

This file is missing a newline character at its end (GH highlights this with a stop sign). [See this to understand why it's needed](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline)

src/game/clock.h
```diff
@@ -7,3 +7,10 @@ bool Clock_Init();
 int32_t Clock_GetMS();
 int32_t Clock_Sync();
 int32_t Clock_SyncTicks(int32_t target);
+int32_t Clock_GetYear();
+int32_t Clock_GetMonth();
+int32_t Clock_GetDay();
+int32_t Clock_GetHours();
+int32_t Clock_GetMinutes();
+int32_t Clock_GetSeconds();
+char *Clock_GetDateTime(char *date_time);
```
rr-, line 16: Missing newline character. 
Like mentioned earlier, most of `Clock_Get*` don't need to be public.

src/game/gameflow.c
```diff
@@ -606,7 +606,9 @@ static bool GameFlow_LoadScriptLevels(struct json_object_s *obj)
             LOG_ERROR("level %d: 'title' must be a string", level_num);
             return false;
         }
-        cur->level_title = strdup(tmp_s);
+        char *title_check = strdup(tmp_s);
+        Shell_ValidateLevelTitle(title_check);
+        cur->level_title = strdup(title_check);
```
rr-, line 611: We don't free the memory anywhere and we don't check if the allocation succeeded.

src/game/shell.c
```diff
@@ -187,6 +189,60 @@ void Shell_Wait(int nticks)
...
+            *ix++;
+        } else {
+            memmove(&str[idx], &str[idx + 1], strlen(str) - idx);
+        }
+    }
+
+    // Make sure title between 0 and LEVEL_TITLE_SIZE chars
+    if (strlen(str) == 0) {
+        memmove(str + 1, str, 2);
+        str[0] = 'x';
```
rr-, line 211: `x`? Do we ever run into this case?

src/game/shell.c
```diff
@@ -187,6 +189,60 @@ void Shell_Wait(int nticks)
...
+            *check++;
+        }
+    }
+
+    // Make sure title has at least 1 char
+    if (strlen(str) == 0) {
+        memmove(str + 1, str, 2);
+        str[0] = 'x';
+        return;
+    }
```
rr-, line 243: Let's merge consecutive `__` into a single `_`

src/game/shell.c
```diff
@@ -187,6 +189,60 @@ void Shell_Wait(int nticks)
...
+}
```
rr-, line 244: I'd expect this function to return a complete screenshot name and not have to amend its return value in the client code later. In other words I'd make sure this functions does the `sprintf("%s_%s.%s", date_time, level_title, ext)` part.

I'd also allocate memory here and free it in the client code (where we use this function) via `Memory_Free`.

So something like `char *Shell_GetScreenshotName(const char *level_title, const char *ext)` where `level_title` is the unprocessed level title.

src/game/clock.c
rr-: Also consider allocating the memory for the return value here, making it `char *Clock_GetDateTime()` and make sure whoever uses it `Memory_Free`s its return value later.

src/game/shell.c
```diff
@@ -199,19 +255,39 @@ bool Shell_MakeScreenshot()
...
+                path, "screenshots/%s_%s_%04d.%s", date_time, level_title, i,
+                ext);
+            char *full_path = NULL;
+            File_GetFullPath(path, &full_path);
+
+            if (!File_Exists(full_path)) {
+                bool result = Output_MakeScreenshot(full_path);
+                Memory_Free(full_path);
+                full_path = NULL;
+                return result;
```
rr-, line 288: We can `break;` to avoid the double call to `Memory_Free`

src/specific/s_clock.c
```diff
@@ -50,3 +51,45 @@ int32_t S_Clock_SyncTicks(int32_t target)
...
+{
+    struct tm *tptr = S_Clock_GetLocalTime();
+    return tptr->tm_min;
+}
+
+int32_t S_Clock_GetSeconds()
+{
+    struct tm *tptr = S_Clock_GetLocalTime();
+    return tptr->tm_sec;
+}
```
rr-, line 95: Missing newline character

src/specific/s_clock.h
```diff
@@ -7,3 +7,10 @@ bool S_Clock_Init();
 int32_t S_Clock_GetMS();
 int32_t S_Clock_Sync();
 int32_t S_Clock_SyncTicks(int32_t target);
+struct tm *S_Clock_GetLocalTime();
+int32_t S_Clock_GetYear();
+int32_t S_Clock_GetMonth();
+int32_t S_Clock_GetDay();
+int32_t S_Clock_GetHours();
+int32_t S_Clock_GetMinutes();
+int32_t S_Clock_GetSeconds();
```
rr-, line 16: Missing newline character

src/specific/s_filesystem.c
```diff
@@ -17,3 +17,15 @@ const char *S_File_GetGameDirectory()
...
 }
+
+void S_File_CreateDirectory(const char *path)
+{
+#if defined(_WIN32)
+    #include <direct.h>
+    _mkdir(path);
+#else
+    #include <sys/types.h>
+    #include <sys/stat.h>
```
rr-, line 28: We shouldn't `#include` in function bodies AFAIK.

src/specific/s_filesystem.h
```diff
@@ -1,3 +1,4 @@
 #pragma once
 
 const char *S_File_GetGameDirectory();
+void S_File_CreateDirectory(const char *path);
```
rr-, line 4: Missing newline character

src/specific/s_filesystem.c
```diff
@@ -17,3 +17,15 @@ const char *S_File_GetGameDirectory()
...
+    mkdir(path, 0664);
+#endif
+}
```
rr-, line 31: Missing newline character

rr- (changes requested):

src/game/shell.c
walkawayy: If the user names the level title `""` in the gameflow the level title would just be a NULL string. ("title": "Caves") I decided to at least give it a one char temp name to be safe since we are checking for other bad level titles already. And `x` is like the generic algabraic symbol idk lol. I could change it to `Level X` or something.

src/game/clock.c
walkawayy: Ok I wasn't sure where to allocate the memory and made the clock functions for each hours, mins, secs, etc seprate in case they were ever needed elsewhere.

src/game/shell.c
rr-: Does it make sense to name the screenshots `20220106_165959_x.jpg`? IMO `20220106_165959.jpg` would suffice.

walkawayy: So remove the level title? I guess we could, but I kind of like having it. It's up to you though. I'm not a heavy screenshot end user, so it doesn't really matter to me.

rr-: I mean, only remove the level title if the author failed to supply it instead of providing a dummy placeholder. I'd prefer to keep it under normal circumstances as well.

walkawayy: I have an idea. I'll just call it `Level X` where `X` is the level number.

src/game/shell.c
walkawayy: Done. Also did a check to remove any trailing underscore before the `.jpg`/extension.

src/game/clock.c
```diff
@@ -2,6 +2,9 @@
 
 #include "specific/s_clock.h"
 
+#include <stdio.h>
+#include <windows.h>
```
rr-, line 6: Yh no, we don't include `<windows.h>` outside `specific/`. We previously used libc for this. Please restore the previous version without the milliseconds.

src/game/gameflow.c
rr-: Let's rebase the branch and use the new `Memory_Dup` utility function. (We don't need to check its results.)

src/game/shell.c
```diff
@@ -187,6 +190,100 @@ void Shell_Wait(int nticks)
     }
 }
 
+void Shell_ValidateLevelTitle(char *str)
```
rr-, line 193: The function is actually about sanitizing (mutating the input data), not validating (checking the input data for conformance).

src/game/shell.c
```diff
@@ -187,6 +190,100 @@ void Shell_Wait(int nticks)
...
+        str[LEVEL_TITLE_SIZE] = '\0';
+        return;
+    }
+
+    // If title totally invalid, name it based on level number
+    if (strlen(str) == 0) {
+        char new_title[10]; // 3 digit level num max
+        sprintf(new_title, "Level_%d", g_CurrentLevel);
+        memmove(str, new_title, strlen(new_title));
+        Memory_Free(new_title);
```
rr-, line 220: We don't free memory that was allocated on the stack.

src/game/shell.c
```diff
@@ -187,6 +190,100 @@ void Shell_Wait(int nticks)
...
+    // Large level titles write outside compass and passport bounds
+    if (strlen(str) > LEVEL_TITLE_SIZE) {
+        str[LEVEL_TITLE_SIZE] = '\0';
+        return;
+    }
+
+    // If title totally invalid, name it based on level number
+    if (strlen(str) == 0) {
+        char new_title[10]; // 3 digit level num max
+        sprintf(new_title, "Level_%d", g_CurrentLevel);
```
rr-, line 218: Let's name it with a space since this is for presentation, not filenames.

src/game/shell.c
```diff
@@ -187,6 +190,100 @@ void Shell_Wait(int nticks)
...
+    int idx = 0;
+    while (*ix != '\0') {
+        if ((*ix >= 'A' && *ix <= 'Z') || (*ix >= 'a' && *ix <= 'z')
+            || (*ix >= '0' && *ix <= '9') || *ix == '\'' || *ix == '.'
+            || *ix == ' ') {
+            idx++;
+            *ix++;
+        } else {
+            memmove(&str[idx], &str[idx + 1], strlen(str) - idx);
+        }
```
rr-, line 206: I am not sure about this. I originally meant the sanitization to only apply to filenames, whereas this now covers also the presentation ingame. I know for a fact that the French community uses T1M to display text in French, and this uses non-alphanumeric characters outside the ASCII range. Let's do this elsewhere, in `Shell_GetScreenshotName`.

src/game/shell.c
```diff
@@ -187,6 +190,100 @@ void Shell_Wait(int nticks)
...
+    }
+
+    // If title totally invalid, name it based on level number
+    // Check again for valid level titles with only ' ', '\'', or '.'
+    if (strlen(level_title) == 0) {
+        char new_title[10]; // 3 digit level num max
+        sprintf(new_title, "Level_%d", g_CurrentLevel);
+        memmove(level_title, new_title, strlen(new_title));
+        Memory_Free(new_title);
```
rr-, line 266: We don't `Memory_Free` memory that was allocated on the stack.

src/game/gameflow.c
rr-: We don't need to `strdup` again when assigning to `cur->level_title`. Once is enough.

src/game/shell.c
```diff
@@ -199,21 +296,26 @@ bool Shell_MakeScreenshot()
         break;
     }
 
-    char path[20];
-    for (int i = 0; i < 10000; i++) {
-        sprintf(path, "screenshot%04d.%s", i, ext);
-        char *full_path = NULL;
-        File_GetFullPath(path, &full_path);
+    // Screenshot name
+    char ss_name[LEVEL_TITLE_SIZE + 25]; // 50
```
rr-, line 300: Please extract the magic number to a module-level constant, use more meaningful variable names, and remove verbose comments in favor of variable names

src/game/shell.c
```diff
@@ -199,21 +296,26 @@ bool Shell_MakeScreenshot()
...
+    char ss_folder[SS_FOLDER_NAME_SIZE];
+    sprintf(ss_folder, "screenshots");
+    File_CreateDirectory(ss_folder);
+
+    // Screenshot folder/name path
+    char path[LEVEL_TITLE_SIZE + SS_FOLDER_NAME_SIZE + 63]; // 100
+    sprintf(path, "%s/%s", ss_folder, ss_name);
 
+    char *full_path = NULL;
+    File_GetFullPath(path, &full_path);
```
rr-, line 313: Missing `Memory_Free` in the unsuccessful case.

src/game/shell.c
```diff
@@ -199,21 +296,26 @@ bool Shell_MakeScreenshot()
...
+    if (!File_Exists(full_path)) {
```
rr-, line 314: Please flip the condition so that we exit early in case of errors to reduce indentation.

src/specific/s_clock.c
```diff
@@ -50,3 +50,10 @@ int32_t S_Clock_SyncTicks(int32_t target)
     } while (elapsed < (double)target);
     return elapsed;
 }
+
+SYSTEMTIME S_Clock_GetLocalTime()
+{
+    SYSTEMTIME lt;
+    GetLocalTime(&lt);
+    return lt;
```
rr-, line 58: Please restore the libc version.

rr- (changes requested):

src/game/clock.c
rr-: The problem I had with this function was that it both returned a `char*` and used an out parameter essentially making it confusing about which is the real return value (and if the input argument being declared as `char*` instead of `const char*` is not a typo).

src/game/clock.c
walkawayy: Well I added ms to avoid adding in a duplicate name digit. This way the name is all made in `Shell_GetScreenshotName`. But ya it's Windows specific. This is so annoying lol. I'll see if there's a better way.

src/specific/s_clock.h
```diff
@@ -2,6 +2,7 @@
 
 #include <stdint.h>
 #include <stdbool.h>
+#include <windows.h>
```
rr-, line 5: This include is no longer needed.

src/specific/s_clock.c
```diff
@@ -2,7 +2,7 @@
 
 #include "global/vars.h"
 
-#include <windows.h>
+#include <time.h>
```
rr-, line 5: This needs to stay as <windows.h> because we use its high precision counters here (used to maintain desired FPS).

src/game/shell.c
```diff
@@ -194,20 +271,40 @@ bool Shell_MakeScreenshot()
...
+    // Screenshot folder/name path
+    char path
+        [SS_FOLDER_NAME_SIZE + TIMESTAMP_SIZE + LEVEL_TITLE_SIZE
+         + EXTENSION_SIZE];
+    sprintf(path, "%s/%s.%s", ss_folder, ss_name, ext);
+    char *full_path = NULL;
+    File_GetFullPath(path, &full_path);
+    if (!File_Exists(full_path)) {
+        bool result = Output_MakeScreenshot(full_path);
+        Memory_FreePointer(full_path);
```
rr-, line 292: This should crash, we're supposed to pass variables by reference to `Memory_FreePointer` (almost always)

src/game/shell.c
```diff
@@ -194,20 +271,40 @@ bool Shell_MakeScreenshot()
...
+    // Name already exists so add number to name
+    for (int i = 2; i < 1000; i++) {
+        sprintf(path, "%s/%s_%d.%s", ss_folder, ss_name, i, ext);
         char *full_path = NULL;
         File_GetFullPath(path, &full_path);
-
         if (!File_Exists(full_path)) {
             bool result = Output_MakeScreenshot(full_path);
-            Memory_FreePointer(&full_path);
+            Memory_FreePointer(full_path);
```
rr-, line 303: Ditto

src/game/shell.c
```diff
@@ -182,6 +187,78 @@ void Shell_Wait(int nticks)
...
+            idx++;
+            prev_us = false;
+        }
+    }
+
+    // If title totally invalid, name it based on level number
+    if (strlen(level_title) == 0) {
+        char new_title[LEVEL_TITLE_SIZE];
+        sprintf(new_title, "Level_%d", g_CurrentLevel);
+        memmove(level_title, new_title, strlen(new_title));
```
rr-, line 241: I think `strcpy`  here would be more readable

src/game/shell.c
```diff
@@ -182,6 +187,78 @@ void Shell_Wait(int nticks)
     }
 }
 
+void Shell_GetScreenshotName(char *str)
+{
+    // Get level title of unknown length
+    char level_title[100];
+    sprintf(level_title, "%s", g_GameFlow.levels[g_CurrentLevel].level_title);
```
rr-, line 194: Nitpick: this doesn't check for `LEVEL_TITLE_SIZE`

src/game/shell.c
```diff
@@ -182,6 +187,78 @@ void Shell_Wait(int nticks)
...
+            } else {
+                *check++ = '_';
+                idx++;
+                prev_us = true;
+            }
+        } else if (*check == '\'' || *check == '.') {
+            // Strip ' and .
+            memmove(
+                &level_title[idx], &level_title[idx + 1],
+                strlen(level_title) - idx);
```
rr-, line 229: I think this condition can be merged with the first one, simplifying the first condition to
```c
        if ((*check < 'A' || *check > 'Z') && (*check < 'a' || *check > 'z')
            && (*check < '0' || *check > '9') && *check != ' ') {
```

src/game/shell.c
```diff
@@ -182,6 +187,78 @@ void Shell_Wait(int nticks)
...
+
+    while (*check != '\0') {
+        if ((*check < 'A' || *check > 'Z') && (*check < 'a' || *check > 'z')
+            && (*check < '0' || *check > '9') && *check != '\'' && *check != '.'
+            && *check != ' ') {
+            // Strip non alphanumeric chars
+            memmove(
+                &level_title[idx], &level_title[idx + 1],
+                strlen(level_title) - idx);
+        } else if (*check == ' ') {
```
rr-, line 214: Consider checking for this first so that you don't need to duplicate the check for space in the first `if`.

src/game/shell.c
```diff
@@ -182,6 +187,78 @@ void Shell_Wait(int nticks)
     }
 }
 
+void Shell_GetScreenshotName(char *str)
```
rr-, line 190: Let's add an assertion `assert(str)`.

src/specific/s_filesystem.c
```diff
@@ -17,3 +24,12 @@ const char *S_File_GetGameDirectory()
     }
     return m_GameDir;
 }
+
+void S_File_CreateDirectory(const char *path)
+{
+#if defined(_WIN32)
```
rr-, line 30: Let's add an assertion `assert(path)`.

src/game/shell.c
oziphantom: A - Z a-z 0-9 can be replaced with `iaslnum()` because you also have to handle Extended ASCII such as á

rr-: AFAIK sprite numbers for characters such as á in TR1 do not conform to any known encoding.

src/game/shell.c
walkawayy: The game crashes on screenshot if the user provides a title longer than `char level_title[100]` size in the gameflow title because I removed the gameflow validate / sanitize function. So I made this a longer buffer just in case and then trim it to `LEVEL_TITLE_SIZE` for the screenshot. That's why I put in the comment that the level title is unknown length.

src/game/shell.c
walkawayy: Ok. My C is super rusty, and I try to be super cautious lol. Thanks for your patience on this.

src/specific/s_clock.c
walkawayy: Ya I moved it to the header when I added that Windows time function that's now gone. Good catch.

src/game/shell.c
walkawayy: I have no idea on language support other than English if I'm being honest. Just let me know if it needs changed or not.

oziphantom: from Arsunt 
image
which is most definitely not a standard encoding, but probably something that needs to be handled, to convert the encoding to EASCII/UTF16 for the filename.

src/game/shell.c
rr-: I meant using `snprintf` or similar. `sprintf` will copy the entire thing and may go over 100 characters.

src/game/shell.c
```diff
@@ -194,20 +275,40 @@ bool Shell_MakeScreenshot()
         break;
     }
 
-    char path[20];
-    for (int i = 0; i < 10000; i++) {
-        sprintf(path, "screenshot%04d.%s", i, ext);
+    // Screenshot folder
+    char ss_folder[SS_FOLDER_NAME_SIZE];
+    sprintf(ss_folder, "screenshots");
```
rr-, line 280: `const char ss_folder[SS_FOLDER_NAME_SIZE] = "screenshots";`

src/game/shell.c
```diff
@@ -194,20 +275,40 @@ bool Shell_MakeScreenshot()
...
         if (!File_Exists(full_path)) {
             bool result = Output_MakeScreenshot(full_path);
             Memory_FreePointer(&full_path);
             return result;
         }
-
-        Memory_FreePointer(&full_path);
     }
 
+    Memory_FreePointer(&full_path);
```
rr-, line 312: This is now in the wrong place, we allocate memory with `File_GetFullPath` in each iteration but free it only once after all iterations have finished.

src/specific/s_clock.c
rr-: And we don't need `<time.h>`.

src/game/shell.c
rr-: The comment about merging second and third branch still stands.

walkawayy: Ok so do we need to open an issue for that?

Also oof I see what you mean now. When I removed the validate gameflow function and moved this here, I didn't clean it up properly. I think it's good now?

walkawayy:
Hmm apparently I missed some warnings. I'll fix.

src/game/shell.c
rr-: Regarding the encoding, if you want… I don't think I'll ever want to touch this part myself.

rr- (approved):
LGTM but I had to create the folder for `screenshots` manually under Wine. Can you confirm that `mkdir` is working fine on Windows?

src/game/shell.c
walkawayy: Oh wait... I just realized I don't think I need `|| *check == '\'' || *check == '.'` at all anymore. Before they were allowed in the gameflow level titles but not in the screenshot name. Now I just need to strip. I'll push.

src/game/shell.c
```diff
@@ -182,6 +188,78 @@ void Shell_Wait(int nticks)
...
+                    strlen(level_title) - idx);
+            } else {
+                *check++ = '_';
+                idx++;
+                prev_us = true;
+            }
+        } else if (
+            ((*check < 'A' || *check > 'Z') && (*check < 'a' || *check > 'z')
+             && (*check < '0' || *check > '9'))
+            || *check == '\'' || *check == '.') {
```
rr-, line 226: I think this should be `&& *check != '\'' && check != '.'`?

walkawayy: I posted a comment above, but I don't think I need this at all anymore. We don't want `'` or `.` in the screenshot name right? I think I just need:

```
        } else if (
            ((*check < 'A' || *check > 'Z') && (*check < 'a' || *check > 'z')
             && (*check < '0' || *check > '9'))
```
?

rr-: I mean if we want to strip these characters then we don't need these checks altogether. If we want to keep them, they need to be reversed.

walkawayy: Well keeping `.` in the screenshot name could potentially screw up the file extension?

Also we still want to strip any spaces, and symbols like `< > ? !` etc right?

walkawayy:
> LGTM but I had to create the folder for `screenshots` manually under Wine. Can you confirm that `mkdir` is working fine on Windows?

Hmm is your Wine running Windows 7? It makes the directory for me fine on Windows 10.

walkawayy:
I just pushed a change so mkdir checks for errors. Let me know what the log says.

rr-:
On mobile for 3 days. Can you please check if the screenshots folder is
created in the proper place if the game is launched from a directory
other than its parent directory? I have a suspicion that it creates the
screenshots directory always in CWD. To fix it, we should call
File_GetFullPath on the directory path rather than the final result.

rr-:
Yes we want to strip them. I'd strip the dot as well.

walkawayy:
> 

Ok so I'll just do and keep only letters and numbers.
```
        } else if (
            ((*check < 'A' || *check > 'Z') && (*check < 'a' || *check > 'z')
             && (*check < '0' || *check > '9'))) {
```

Alright I'll test that. And we should probably do the full path anyway. I'll push that up too.

This is no rush. Enjoy your time afk!

walkawayy:
> 

Ya the screenshots folder is made in the CWD. I'll make it use the full game folder path.

src/game/shell.c
```diff
@@ -194,18 +271,40 @@ bool Shell_MakeScreenshot()
         break;
     }
 
-    char path[20];
-    for (int i = 0; i < 10000; i++) {
-        sprintf(path, "screenshot%04d.%s", i, ext);
+    // Screenshot folder
+    const char ss_folder[SS_FOLDER_NAME_SIZE] = "screenshots";
+    char *full_ss_path = NULL;
+    File_GetFullPath(ss_folder, &full_ss_path);
```
rr-, line 277: I think we can just say `File_GetFullPath("screenshots", &full_ss_path)` here and remove `SS_FOLDER_NAME_SIZE` const.

src/game/shell.c
```diff
@@ -194,18 +271,40 @@ bool Shell_MakeScreenshot()
...
+    char ss_name[TIMESTAMP_SIZE + LEVEL_TITLE_SIZE];
+    Shell_GetScreenshotName(ss_name);
+
+    // Screenshot folder/name path
+    char path
+        [SS_FOLDER_NAME_SIZE + TIMESTAMP_SIZE + LEVEL_TITLE_SIZE
+         + EXTENSION_SIZE];
+    sprintf(path, "%s/%s.%s", ss_folder, ss_name, ext);
+    char *full_path = NULL;
+    File_GetFullPath(path, &full_path);
```
rr-, line 291: Now it's not needed to `File_GetFullPath` for the second time, we can just `char path[strlen(full_ss_path) + ...]; sprintf(...);` and not have to `Memory_FreePointer` again.

src/specific/s_filesystem.c
```diff
@@ -17,3 +27,17 @@ const char *S_File_GetGameDirectory()
...
+{
+    assert(path);
+    int32_t err_check = 0;
+#if defined(_WIN32)
+    err_check = _mkdir(path);
+#else
+    err_check = mkdir(path, 0664);
+#endif
+    if (err_check != 0) {
+        LOG_ERROR("mkdir failed! %s\n", strerror(errno));
```
rr-, line 41: We don't put `\n` in `LOG_*` calls, it does it for us.

walkawayy: I'm going to remove this because every screenshot the user takes after the first one will log an "error message" saying the directory already exists.

src/game/shell.c
```diff
@@ -194,20 +270,41 @@ bool Shell_MakeScreenshot()
...
+    File_CreateDirectory(ss_folder_path);
+
+    // Screenshot name
+    char *ss_name =
+        Memory_Alloc(TIMESTAMP_SIZE + LEVEL_TITLE_SIZE + EXTENSION_SIZE);
+    Shell_GetScreenshotName(ss_name);
+
+    // Screenshot folder/name path
+    char *path = Memory_Alloc(strlen(ss_folder_path) + strlen(ss_name));
+    sprintf(path, "%s/%s.%s", ss_folder_path, ss_name, ext);
```
rr-, line 285: Now you're writing beyond the memory variable size. We need to allocate more bytes:
- `+strlen(ext)`
- `+1` for `/`
- `+1` for `.`
- `+1` for the `\0` string terminator
- `+1` for the underscore (the additional number)
- `+3` for maximum of 999 (the additional number)

This gives us `Memory_Alloc(strlen(ss_folder_path) + strlen(ss_name) + 7)`.

Since we're using timestamps the maximum number can be reduced from 1000 to 100 because we will never exceeds 60 screenshots per second. This assumption will reduce the number above by 1 down to `Memory_Alloc(strlen(ss_folder_path) + strlen(ss_name) + 6)`.

Consider doing these `Memory_FreePointer` once like so:

```c
bool result = false;
if (!File_Exists(path)) {
    result = Output_MakeScreenshot(path);
    goto cleanup;
}

for (int i = 2; i < 1000; i++) {
    sprintf(...);
    if (!File_Exists(path)) {
        result = Output_MakeScreenshot(path);
        goto cleanup;
    }
}

cleanup:
Memory_FreePointer(…);
Memory_FreePointer(…);
Memory_FreePointer(…);
return result;
```

walkawayy: Ya I had the extension size defined as 20 to account for all that. Is it better to do that, or magic number the `+ 6`?

And yes definitely need to clean up the freeing. I don't personally like goto, so I'm going to push up a slightly different approach.

walkawayy:
Alright let me know what you think. I cleaned up the memory freeing. I also put the folder name, screenshot name, and extension in order since that matches the order the full path is built.

src/specific/s_filesystem.c
```diff
@@ -17,3 +26,14 @@ const char *S_File_GetGameDirectory()
...
 }
+
+void S_File_CreateDirectory(const char *path)
+{
+    assert(path);
+    int32_t err_check = 0;
+#if defined(_WIN32)
+    err_check = _mkdir(path);
+#else
+    err_check = mkdir(path, 0664);
```
rr-, line 37: Since we don't do anything with this variable, assignments are no longer needed.

rr- (approved):
LGTM with one last remark

rr- (approved):

PR #301, opened 2022-01-06
s/audio: fix audio clicks

rr-:
Replacing SDL resampler with libavcodec helped removed the clicks.


walkawayy:
Wow sounds so much better! Awesome!

The only thing I noticed was this issue again. I think when the mutants explode, all the pieces have their own explosion noises? The Uzi sounds go mute for a little bit while the mutant explodes. I have no idea if this is an OG bug or not, or if it's easily fixable.

https://streamable.com/m1hqub

rr-:
@walkawayy I can confirm that 20 was too little, and that it was caused by all sample buffers being in use. After adding 5 more the issue no longer seems to occur.

walkawayy (approved):

walkawayy:
Hmm wait the Uzis might still be cutting out in that one spot.

walkawayy:
This is with 50 samples. Maybe it's a mixing thing since there's so many noises happening at once.

https://streamable.com/tfiqxa

rr-:
The problem is with game/sound.c and it's an OG bug. Our mixer is fine.

rr-:
Fixed for good. Also, mass explosions are now even more explosive which I like a lot.

PR #307, opened 2022-01-07
collide: fix Lara glitching through static objects into the void

### LINKED ISSUE #299: Bug: Jumping onto suit of armour causes Lara to fall into the void
Just came aross this bug when playing on version 2.2.1, not sure if this is an OG or T1M bug.

https://user-images.githubusercontent.com/94776985/148386011-a866e60a-2404-48de-9b3c-1a9d1f0c42b9.mp4

walkawayy:
Resolves #299.

Based on [TR2's collision function](https://github.com/asasas9500/TR2Main/blob/864b4a185e494b88a145a30a2ba45961eb9d5792/game/collide.cpp#L162).


rr- (approved):
Let's not merge yet, I'll have to test it

rr-:
LGTM but we need to resolve conflicts.

PR #311, opened 2022-01-08
music: fix secret chime when collecting consecutive secrets if no fix

### LINKED ISSUE #310: Bug: no secret chime when collecting consecutive secrets
If `fix_secrets_killing_music` is set to false, the secret chime doesn't play when picking up a second secret if no other audio track played in between.

walkawayy:
Resolves #310.

This is when `fix_secrets_killing_music` is false. This fixed the chimes for me, and it still stops a music track if one is currently playing. Let me know if this fix would cause other problems though.


rr-:
IMO this will cause a problem with one-shot music triggers that are not marked as such with trigger flags. Example of such music trigger is on the torso platform in The Great Pyramid. After the battle music finishes, stepping on the tile the level starts on should NOT retrigger the music. Ideally this is achieved by marking the music triggers as one-shot, but this particular behavior is expected by the OG and thus the level designers.

IMO this fix will break it and we need something different. I'm OK with hardcoding track ID since we already do this in other places.

walkawayy:
Ok. Ya hard coding it is ugly, but I think that's the best way to do it. This broken audio mode is annoying to have to support lol.

walkawayy:
To clarify @rr- are you saying to do something like this?

```
static void Music_StreamFinished(int stream_id, void *user_data)
{
    // When a stream finishes, play the remembered background BGM.
    if (stream_id == m_AudioStreamID) {
        m_AudioStreamID = -1;
        if (m_Track == 13) {
            m_track = 0;
        }
        if (m_TrackLooped >= 0) {
            Music_PlayLooped(m_TrackLooped);
        }
    }
}
```

rr-:
Yes. In case you're feeling extra patient, we could add an enum for all the music tracks.

walkawayy:
I can do that. Are there official names for the tracks? I can try and match them up with `SoundTool`'s names. Our indices are different, so I'll have to listen to each track and match it.

rr-:
That's correct, there are two versions of the same sound, one is the high definition music file and the other is a low definition PCM U8 8? kHz sample file in each level. Incidentally they sound exactly the same, but one breaks the music and the other plays like a regular sample.

walkawayy:
Going to name the prefix based on [these abbreviations](https://sound.stackexchange.com/questions/28594/what-do-the-acronyms-dx-fx-mx-stand-for). 

Some of the music tracks are dialog, but since they all share the same numbering, I'm going to define them all as `MX`.

src/global/types.h
```diff
@@ -473,6 +473,70 @@ typedef enum SOUND_EFFECT_ID {
...
+    MX_BALDY_SPEECH = 51,
+    MX_COWBOY_SPEECH = 52,
+    MX_LARSON_SPEECH = 53,
+    MX_NATLA_SPEECH = 54,
+    MX_PIERRE_SPEECH = 55,
+    MX_SKATEKID_SPEECH = 56,
+    MX_ST_FRANCIS_FOLLY_AMBIENCE = 57,
+    MX_CISTERN_AMBIENCE = 58,
+    MX_WINDY_AMBIENCE = 59,
+    MX_ATLANTIS_AMBIENT = 60,
```
rr-, line 537: `AMBIENCE`

rr- (approved):
Almost there :+1: thanks for looking up the track names.

src/global/types.h
walkawayy: Oops fixed. I kinda guessed on the ambience names since I couldn't find official names for them, but I think the names works well enough.

rr- (approved):

PR #313, opened 2022-01-09
input: allow holding down forward/back to move through saves

### LINKED ISSUE #171: Feature request: Allow to hold up/down key in savestate list
Currently have to step through them with key presses. Would be better UX if you could hold up or down. TR2Main added this sometime last year I think.

walkawayy:
Resolves #171.

### DIFF
--- a/src/game/input.c
+++ b/src/game/input.c
@@ -8,6 +8,7 @@
 INPUT_STATE g_Input = { 0 };
 INPUT_STATE g_InputDB = { 0 };
 INPUT_STATE g_OldInputDB = { 0 };
+
 static int32_t m_HoldBack = 0;
 static int32_t m_HoldForward = 0;
 
--- a/src/game/input.c
+++ b/src/game/input.c
@@ -2,11 +2,14 @@
 
 #include "specific/s_input.h"
 
+#define DELAY_FRAMES 12
+#define HOLD_FRAMES 3
+
 INPUT_STATE g_Input = { 0 };
 INPUT_STATE g_InputDB = { 0 };
 INPUT_STATE g_OldInputDB = { 0 };
-static int32_t m_HoldBack = -12;
-static int32_t m_HoldForward = -12;
+static int32_t m_HoldBack = 0;
+static int32_t m_HoldForward = 0;
 
 static INPUT_STATE Input_GetDebounced(INPUT_STATE input);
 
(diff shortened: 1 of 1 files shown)


rr- (changes requested):
I'm glad that you went with the tested TR2Main approach, but it needs a bit of polishing.

- I don't understand why these are negative numbers? Why not counting to `15` from `0`?
- We don't put `static` variables inside functions, let's make it a module level variable to highlight its global scope.
- Similarly we don't name variables with camelCase. This should be `snake_case` for normal variables and `m_PascalCase` for module-level globals.

rr- (commented):
I understand that we need initial buildup, what I fail to understand is why we're starting at -12 and waiting till it passes 3 instead of starting at 0 and waiting till it passes 15. Also this should be a constant or a module-level `#define`.

walkawayy:
Hmm ya that is odd. I tried to be consistent with how TR2Main did it. I was thinking maybe the negative number was to indicate that when the value is below 0, there is that initial delay. But that would use values of -15 and 0.

rr-:
Yes, it will need to be scaled.

src/game/input.c
```diff
@@ -2,16 +2,37 @@
 
 #include "specific/s_input.h"
 
+#define DELAY_FRAMES 12
+#define HOLD_FRAMES 3
```
rr-, line 6: OK now it's readable, we wait 12 frames till letting the user cycle fast, and then we only do it once every 3 frames.

src/game/input.c
```diff
@@ -2,16 +2,37 @@
...
+
 INPUT_STATE g_Input = { 0 };
 INPUT_STATE g_InputDB = { 0 };
 INPUT_STATE g_OldInputDB = { 0 };
+static int32_t m_HoldBack = 0;
```
rr-, line 11: Let's put a blank above since the vars before are truly global, and these are only modular

rr- (commented):
Thanks for the refactor

src/game/input.c
walkawayy: Ya. Then we can just scale this for 60 FPS in the future.

rr- (approved):

Richard-L:
@walkawayy any chance you could look into this again? 🙏 It seems to have broken in v2.15

PR #315, opened 2022-01-10
inventory: fix black screen frames between inventory mode

### LINKED ISSUE #75: OG bugfix: blinking when switching between inventory modes
When navigating from the main menu ring to the keys or the options or back, the game renders a black screen briefly.

walkawayy:
Resolves #75.

I don't know much about rendering, so let me know if this is an acceptable fix or not. If it's not, I hope this can help you guys figure out a different solution.

### DIFF
--- a/src/gfx/2d/2d_surface.c
+++ b/src/gfx/2d/2d_surface.c
@@ -165,13 +165,9 @@ bool GFX_2D_Surface_Flip(GFX_2D_Surface *surface)
         return false;
     }
 
-    bool rendered = GFX_Context_IsRendered();
-
     // don't re-upload surfaces if external rendering was active after
     // lock() has been called, since it wouldn't be visible anyway
-    if (rendered) {
-        surface->is_dirty = false;
-    }
+    surface->is_dirty = false;
 
     // swap front and back buffers
     uint8_t *buffer_tmp = surface->back_buffer->buffer;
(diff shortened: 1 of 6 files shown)


rr-:
I think this is acceptable but now I don't think we need the calls to `GFX_Context_SetRendered` in `GFX_3D_Renderer_RenderPrimList` and `GFX_3D_Renderer_RenderPrimStrip`.

rr- (approved):
See above, + we need to remove the comment

walkawayy:
Ok definitely let me know if this is what you suggested. The game still seems to perform fine?

walkawayy:
Actually the one thing I noticed is the Eidos screen at the game's boot up seems to show for only a split second instead of the normal length.

src/gfx/2d/2d_surface.c
```diff
@@ -165,13 +165,9 @@ bool GFX_2D_Surface_Flip(GFX_2D_Surface *surface)
...
     }
 
-    bool rendered = GFX_Context_IsRendered();
-
     // don't re-upload surfaces if external rendering was active after
     // lock() has been called, since it wouldn't be visible anyway
-    if (rendered) {
-        surface->is_dirty = false;
-    }
+    surface->is_dirty = false;
```
rr-, line 170: The comment above no longer makes sense and can be removed.

src/gfx/2d/2d_surface.c
```diff
@@ -191,23 +187,14 @@ bool GFX_2D_Surface_Flip(GFX_2D_Surface *surface)
 
     // swap buffer now if there was external rendering, otherwise the
     // surface would overwrite it
-    if (rendered) {
-        GFX_Context_SwapBuffers();
-    }
+    GFX_Context_SwapBuffers();
```
rr-, line 190: The comment above no longer makes sense and can be removed.

rr-:
Actually let's revert my earlier suggestions, I understand GLRage's approach. I can't think of anything better at the moment without complicating the code more than it is. 

One thing I'd like is to avoid making `game` depend directly on `gfx`. To that end I suggest adding a new routine, `void GFX_3D_Renderer_RenderEmpty()` that calls `GFX_Context_SetRendered`. Proxy it through `S_Output_DrawEmpty()` and finally `Output_DrawEmpty` to trigger an "external redraw" in GLRage's terminology.

walkawayy:
Ok cool. I'm going to push up a clean branch and PR with that different approach.

PR #329, opened 2022-01-14
bin: downgrade shaders to GLSL version 1.30

### LINKED ISSUE #327: OG bug: can't run on older hardware
Running tomb1main on an old computer yields:

```
../src/gfx/context.c 137 GFX_Context_SetWindowSize Window size: 1366x768
../src/game/random.c 10 Random_SeedControl 41937
../src/game/random.c 22 Random_SeedDraw 40483
../src/gfx/context.c 56 GFX_Context_Attach Attaching to HWND 002C01C0
../src/specific/s_shell.c 46 S_Shell_ShowFatalError Shader compilation failed:
0:1(10): error: GLSL 3.30 is not supported. Supported versions are: 1.10, 1.20, 1.30, 1.00 ES, and 3.00 ES
```

Instead of exiting, tomb1main should use previous GLSL versions.

rr-:
Resolves #327. I'm not sure about the compatibility range. It's working fine on Wine both with and without the bilinear filter.

### DIFF
--- a/bin/shaders/2d.fsh
+++ b/bin/shaders/2d.fsh
@@ -1,4 +1,5 @@
-#version 330 core
+#version 130
+#extension GL_ARB_explicit_attrib_location: enable
 
 in vec2 vertTexCoords;
 
--- a/bin/shaders/2d.vsh
+++ b/bin/shaders/2d.vsh
@@ -1,4 +1,5 @@
-#version 330 core
+#version 130
+#extension GL_ARB_explicit_attrib_location: enable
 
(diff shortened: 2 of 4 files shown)


walkawayy:
Works for me, but might need more data from lower end hardware users.

rr-:
Merging, as it's working for 3+ people and shouldn't cause problems to users with higher end machines. Bad case – this won't change anything to users with older machines. Worst case – some new GPUs dropped support for GLSL 1.30 and we'll have to revert this or go with supplying multiple shader versions.

PR #335, opened 2022-01-15
gameflow: fix fallback game strings not working

### LINKED ISSUE #331: T1M bug: default strings get removed and are useless
So in 
`Init.c T1MInit` will install all the default strings. 

`gameflow.c GameFlow_LoadFromFileImpl` does `GameFlow_Shutdown()` which deletes them all and set it to NULL, so if you have a default string that is not in the users cfg file the game will crash with a null pointer trying to use the string. 

Ideally the strings are not removed first or after the Shutdown the defaults are installed again first.

rr-:
Resolves #331.

### DIFF
--- a/src/game/gameflow.c
+++ b/src/game/gameflow.c
@@ -254,10 +254,16 @@ static bool GameFlow_LoadScriptGameStrings(struct json_object_s *obj)
         if (!value || !value->string || key < 0 || key >= GS_NUMBER_OF) {
             LOG_ERROR("invalid string key %s", strings_elem->name->string);
         } else {
-            Memory_FreePointer(&g_GameFlow.strings[key]);
             g_GameFlow.strings[key] = Memory_Dup(value->string);
         }
         strings_elem = strings_elem->next;
+    }
+
+    for (const GAMEFLOW_DEFAULT_STRING *def = g_GameFlowDefaultStrings;
+         def->string; def++) {
+        if (!g_GameFlow.strings[def->key]) {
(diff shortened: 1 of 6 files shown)


src/game/gameflow.c
```diff
@@ -254,12 +254,18 @@ static bool GameFlow_LoadScriptGameStrings(struct json_object_s *obj)
         if (!value || !value->string || key < 0 || key >= GS_NUMBER_OF) {
             LOG_ERROR("invalid string key %s", strings_elem->name->string);
         } else {
-            Memory_FreePointer(&g_GameFlow.strings[key]);
```
rr-, line 257: This is handled by calling `GameFlow_Shutdown()` before re-initialization of the gameflow.

src/game/gameflow.c
```diff
@@ -254,12 +254,18 @@ static bool GameFlow_LoadScriptGameStrings(struct json_object_s *obj)
...
         }
         strings_elem = strings_elem->next;
     }
 
+    for (const GAMEFLOW_DEFAULT_STRING *def = g_GameFlowDefaultStrings;
+         def->string; def++) {
+        if (!g_GameFlow.strings[def->key]) {
+            g_GameFlow.strings[def->key] = Memory_Dup(def->string);
+        }
+    }
```
rr-, line 267: In case loading the gameflow fails with an unrecoverable error, the game exits anyway, so it should be safe to go with this approach (i.e. placing the fallback initialization only in this place and nowhere else).

walkawayy:
LGTM but I'm gonna let @oziphantom do an official review since he reported the issue.

PR #340, opened 2022-01-16
lara: let Lara start with other guns than pistols

rr-:


walkawayy:
When skipping levels with the shotgun, this happened in Cistern. I'm guessing because Palace Midas starts with Lara underwater and she was supposed to start the level with a shotgun and I skipped there. So maybe not a normal use case but still seems like a bug.

20220116_114353_The_Cistern

rr-:
:D good catch. Also good observation about the water status. I'll see what I can do about it.

rr-:
@walkawayy ready for a retest

src/game/laramisc.c
```diff
@@ -576,18 +576,17 @@ void LaraInitialiseMeshes(int32_t level_num)
...
-    int16_t gun_type = start->gun_type;
-    if (gun_type == LGT_SHOTGUN) {
-        if (start->flags.got_uzis || start->flags.got_magnums
-            || start->flags.got_pistols) {
-            gun_type = LGT_PISTOLS;
-        } else {
-            gun_type = LGT_UNARMED;
-        }
+    if (g_Lara.gun_status == LGS_ARMLESS) {
+        return;
```
walkawayy, line 580: You need to move up the shotgun back torso check before this. If Lara has the shotgun in her inventory but not equipped, it doesn't appear on her mesh until she draws it at the start of a new level.

It does load correctly on a saved game, so that must be restoring it from a saved state and not calling this function?

Everything else looks good.

walkawayy (changes requested):

src/game/laramisc.c
rr-: Fixed

walkawayy (approved):

PR #345, opened 2022-01-16
movable_block: fix pushing blocks through doors

### LINKED ISSUE #46: OG bug: Lara can push a block through a closed door
In Atlantis level there is area with a block next to a door. If player cheats Lara to this area (normally you can't get there that early) he can push the block through a door, camera acts weirdly during that. After pusing animation ends, player getting pushed back behind the door

Снимок (3)

walkawayy:
Resolves #46.

Let me know how ugly this is :D.

### DIFF
--- a/src/game/traps/movable_block.c
+++ b/src/game/traps/movable_block.c
@@ -9,6 +9,7 @@
 #include "game/lara.h"
 #include "game/sound.h"
 #include "global/vars.h"
+#include "src/game/objects/door.h"
 
 void SetupMovableBlock(OBJECT_INFO *obj)
 {
@@ -105,6 +106,11 @@ void MovableBlockCollision(
         }
 
         if (!TestLaraPosition(g_MovingBlockBounds, item, lara_item)) {
+            return;
+        }
+
+        // OG fix: stop pushing blocks through doors
+        if (CheckLaraDoorCollision()) {
             return;
         }
 
--- a/src/game/traps/movable_block.c
+++ b/src/game/traps/movable_block.c
@@ -34,10 +34,8 @@ static bool TestDoorCoveringBlock(ITEM_INFO *lara_item, COLL_INFO *coll)
             continue;
         }
 
-        if (TestBoundsCollide(item, lara_item, coll->radius)) {
-            return true;
-        }
-        if (TestCollision(item, lara_item)) {
+        if (TestBoundsCollide(item, lara_item, coll->radius)
+            && TestCollision(item, lara_item)) {
             return true;
         }
     }
--- a/src/game/traps/movable_block.c
+++ b/src/game/traps/movable_block.c
@@ -9,7 +9,40 @@
 #include "game/lara.h"
 #include "game/sound.h"
 #include "global/vars.h"
-#include "src/game/objects/door.h"
+#include "src/game/collide.h"
+#include "src/game/sphere.h"
+
+static bool TestDoorCoveringBlock(ITEM_INFO *lara_item, COLL_INFO *coll);
+
+static bool TestDoorCoveringBlock(ITEM_INFO *lara_item, COLL_INFO *coll)
+{
+    // OG fix: stop pushing blocks through doors
+    int32_t max_dist = SQUARE((WALL_L * 2) >> 8);
+    for (int item_num = 0; item_num < g_LevelItemCount; item_num++) {
+        ITEM_INFO *item = &g_Items[item_num];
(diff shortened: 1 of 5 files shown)


rr-:
Yeah it's quite bad. I'd refactor this. I'd try this:

- Extract the logic related to getting neraby objects from `LaraBaddieCollision` to a separate function, `void Lara_GetNearbyItems(ITEM_INFO *target[MAX_BADDIE_COLLISION]);` from the line `int16_t numroom = 0;` up to  `&& z < TARGET_DIST) {`
- In `MovableBlockCollision`, call this function, iterate through each item and check if it's a door by `item->object_number >= O_DOOR_TYPE1 && item->object_number <= O_DOOR_TYPE2`, then check if `TestBoundsCollide(g_LaraItem, item, coll)` and `TestCollision(g_LaraItem, item`. If both do, disallow moving the block

walkawayy:
The problem with that is I need `lara_item` and `coll` to do that logic? I'd need to pass `ITEM_INFO *lara_item, COLL_INFO *coll` right?

walkawayy:
Based the new logic on `OpenNearestDoors` because none of the door structs store item number and `g_RoomInfo` doesn't include doors.

src/game/traps/movable_block.c
```diff
@@ -9,6 +9,40 @@
...
+
+        if ((item->object_number < O_DOOR_TYPE1
+             || item->object_number > O_DOOR_TYPE8)) {
+            continue;
+        }
+
+        if (TestBoundsCollide(item, lara_item, coll->radius)) {
+            return true;
+        }
+        if (TestCollision(item, lara_item)) {
```
rr-, line 40: Don't we need to have both of these conditions satisfied at once in order to detect a collision?
In that case I'd say

if (!Test…()) continue;
if (!Test…()) continue;
return true;

rr- (approved):
Thanks, it looks much better. The reason why it looks better is because it doesn't introduce more global state and the logic is kept only in a single place.

src/game/traps/movable_block.c
walkawayy: I don't think both need to be satisfied at once? Here's DoorCollision (which is for a single door):
```
void DoorCollision(int16_t item_num, ITEM_INFO *lara_item, COLL_INFO *coll)
{
    ITEM_INFO *item = &g_Items[item_num];

    if (!TestBoundsCollide(item, lara_item, coll->radius)) {
        return;
    }
    if (!TestCollision(item, lara_item)) {
        return;
    }

    if (coll->enable_baddie_push) {
        if (item->current_anim_state != item->goal_anim_state) {
            ItemPushLara(item, lara_item, coll, coll->enable_spaz, 1);
        } else {
            ItemPushLara(item, lara_item, coll, 0, 1);
        }
    }
}
```

rr-: DoorCollision does not push Lara if TestBoundsCollide fails or TestCollision fails. Applying De Morgan's law to flip this expression, we get: DoorCollision pushes Lara if TestBoundsCollide succeeds and TestCollision succeeds.

They need to be both satisfied.

walkawayy: Yes true. Still not used to the backwards if checking lol. So what about:
```
static bool TestDoorCoveringBlock(ITEM_INFO *lara_item, COLL_INFO *coll)
{
    // OG fix: stop pushing blocks through doors
    int32_t max_dist = SQUARE((WALL_L * 2) >> 8);
    for (int item_num = 0; item_num < g_LevelItemCount; item_num++) {
        ITEM_INFO *item = &g_Items[item_num];
        int32_t dx = (item->pos.x - lara_item->pos.x) >> 8;
        int32_t dy = (item->pos.y - lara_item->pos.y) >> 8;
        int32_t dz = (item->pos.z - lara_item->pos.z) >> 8;
        int32_t dist = SQUARE(dx) + SQUARE(dy) + SQUARE(dz);

        if (dist > max_dist) {
            continue;
        }

        if ((item->object_number < O_DOOR_TYPE1
             || item->object_number > O_DOOR_TYPE8)) {
            continue;
        }

        if (TestBoundsCollide(item, lara_item, coll->radius) && TestCollision(item, lara_item)) {
            return true;
        }
    }
    return false;
}
```

I'm not sure what you mean about the loop ending in a return? In this method, it checks all the doors nearby. If Lara collides with one, return true. Otherwise if all doors get checked and there's no collision, return false.

rr- (approved):

PR #351, opened 2022-01-18
Support DPI scaling on Windows

### LINKED ISSUE #280: OG Bug: Windows DPI scaling causes resolution issues
I have a 1440p screen, Windows 10 21H2, and I use 150% DPI scaling. on 150% the game reports my resolution as 1707x960. on 125% the game reports my resolution as 2048x1152 (you see where this is going). Selecting the correct/higher resolution doesn't make the game look sharper.

Another related issue: on 150% scaling, there is a single-pixel vertical-line to the right of the screen where I can see the desktop.

Overriding DPI scaling in compatibility settings fixes both of these.

walkawayy:
Resolves #280.

The SDL high DPI flag doesn't work yet. It's still a WIP. The progress can be tracked [here](https://github.com/libsdl-org/SDL/issues/4908).

This solution fixes the game resolution for me. I have a 1440p display with 125% scaling. Before, the game would auto detect 2048x1152 as my resolution. Now, it correctly starts at 1440p.

### DIFF
--- a/src/specific/s_shell.c
+++ b/src/specific/s_shell.c
@@ -25,6 +25,7 @@ static int m_ArgCount = 0;
 static char **m_ArgStrings = NULL;
 static bool m_Fullscreen = true;
 static SDL_Window *m_Window = NULL;
+static float m_Scaling = 0;
 
 static void S_Shell_PostWindowResize();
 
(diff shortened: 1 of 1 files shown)


src/specific/s_shell.c
```diff
@@ -133,6 +134,32 @@ int main(int argc, char **argv)
...
+                shcoreDLL, "SetProcessDpiAwareness");
+        if (SetProcessDpiAwareness) {
+            SetProcessDpiAwareness(PROCESS_PER_MONITOR_DPI_AWARE);
+
+            float ddpi;
+            if (SDL_GetDisplayDPI(0, &ddpi, NULL, NULL)
+                != -1) { // SDL_WINDOWPOS_UNDEFINED is Display 0
+                // When using HiDPI mode, set correct DPI scaling
+                m_Scaling = ddpi / 96.f; // = 0 but not needed for some reason
+            }
```
rr-, line 159: These lines are dead code? `m_Scaling` is never read.

src/specific/s_shell.c
```diff
@@ -133,6 +134,32 @@ int main(int argc, char **argv)
...
+    // Enable HiDPI mode in Windows to detect DPI scaling
+    typedef enum PROCESS_DPI_AWARENESS {
+        PROCESS_DPI_UNAWARE = 0,
+        PROCESS_SYSTEM_DPI_AWARE = 1,
+        PROCESS_PER_MONITOR_DPI_AWARE = 2
+    } PROCESS_DPI_AWARENESS;
+
+    HRESULT(WINAPI * SetProcessDpiAwareness)
+    (PROCESS_DPI_AWARENESS dpiAwareness); // Windows 8.1 and later
+    void *shcoreDLL = SDL_LoadObject("SHCORE.DLL");
```
rr-, line 146: Naming conventions :]

src/specific/s_shell.c
walkawayy: Ya it comes out as 0 for some reason. I was using it to multiply by resolution but idk. Maybe I can just delete this part and setting the DPI stuff is enough.

rr- (approved):
Does this also get rid of the dummy line near the edge of the window reported in the original ticket?

walkawayy:
Hm I never noticed that. But there might be a line. Here's the edge of my screen. Is that a DPI specific thing or a different issue? Do you have it?

image

rr-:
I don't have Windows so can't really test, only Wine that doesn't let me test such things.

walkawayy:
It looks like there is a one pixel line right? Not sure how since the log prints: `GFX_Context_SetWindowSize Window size: 2560x1440` which matches my screen resolution.

walkawayy:
I turned off my DPI scaling, and the single pixel issue looks like it's still there. Seems like a separate issue. Could it be a bug in SDL?

image

DonelBueno:
Tomb1Main_2022_01_18_06_44_45_608

I have the black lines on the lower and right borders even running at 1920x1080 with no dpi scaling, so definitely a separate issue.

rr- (approved):
LGTM,  LMK if OK to merge

walkawayy:
Ok it’s good to merge then.

PR #355, opened 2022-01-18
lava: fix lava trap not setting Lara on fire

### LINKED ISSUE #353: Bug: Lava trap does not kill Lara
In the corridor with the unreachable medi packs, after pulling switch, the lava does not kill Lara and instead pushes her away.

20220118_150412_Atlantis

I have to say I like this bug ;)

walkawayy:
Resolves #353.

### DIFF
--- a/src/game/traps/lava.c
+++ b/src/game/traps/lava.c
@@ -96,7 +96,6 @@ void LavaBurn(ITEM_INFO *item)
     int16_t height = GetHeight(floor, item->pos.x, 32000, item->pos.z);
 
     if (item->floor != height) {
-        LOG_DEBUG("item->floor: %d != %d: height", item->floor, height);
         return;
     }
 
--- a/src/game/traps/lava.c
+++ b/src/game/traps/lava.c
@@ -33,11 +33,11 @@ void SetupLavaWedge(OBJECT_INFO *obj)
 bool TestLavaFloor(ITEM_INFO *item)
 {
     if (g_Lara.water_status == LWS_CHEAT) {
-        return;
+        return false;
     }
 
     if (item->hit_points < 0) {
-        return;
+        return false;
     }
 
     int16_t room_num = item->room_number;
--- a/src/game/traps/lava.c
+++ b/src/game/traps/lava.c
@@ -30,7 +30,7 @@ void SetupLavaWedge(OBJECT_INFO *obj)
     obj->save_flags = 1;
 }
 
-void LavaBurn(ITEM_INFO *item)
+void TestLavaFloor(ITEM_INFO *item)
 {
     if (g_Lara.water_status == LWS_CHEAT) {
         return;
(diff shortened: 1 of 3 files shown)


src/game/traps/lava.c
```diff
@@ -76,8 +76,23 @@ void LavaBurn(ITEM_INFO *item)
 
     if (!on_lava) {
         return;
+    } else {
+        LavaBurn(item);
+    }
```
rr-, line 81: Please change the signature of this function to return a `bool`. We have too many `Test` functions that produce unexpected side effects.

rr- (changes requested):
Thanks for looking into this.

src/game/traps/lava.c
```diff
@@ -75,9 +75,25 @@ void LavaBurn(ITEM_INFO *item)
     } while (!(type & END_BIT));
 
     if (!on_lava) {
+        return false;
+    } else {
+        LavaBurn(item);
+        return true;
```
rr-, line 81: This still produces side effects.

What I meant is doing this:

```diff
-TestLavaFloor(g_LaraItem);
+if (TestLavaFloor(g_LaraItem) {
+    LavaBurn(g_LaraItem);
+}
```

rr- (changes requested):

src/game/traps/lava.c
walkawayy: Ah gotcha.

src/game/traps/lava.c
```diff
@@ -75,9 +75,24 @@ void LavaBurn(ITEM_INFO *item)
     } while (!(type & END_BIT));
 
     if (!on_lava) {
+        return false;
+    } else {
+        return true;
+    }
+}
```
rr-, line 82: Please make this function exit early. In other words refactor it to `return true` as soon as possible within its main loop, and `return false` at the end of the function as a fallback value.

Off-topic for educational purposes (doesn't apply after the suggestion above):

```c
if (!on_lava) {
    return false;
} else {
    return true;
}
```

This snippet is quite low quality. First, since this snippet has both `if` and `else`, and both of the branches are one-liners, there's no need for the negation - it only reduces readability. So let's flip it:

```c
if (on_lava) {
    return true;
} else {
    return false;
}
```

This is still wordy. Both of the branches evaluate to the same statement – `return x`, so we can turn it into a ternary expression. We get:

```c
return on_lava ? true : false;
```

But hopefully you can see that this is silly and can be just represented as:

```c
return on_lava;
```

However, like I said earlier, there's no need to run the entire loop – we can know the state of `on_lava` much earlier. And making the function exit as early as possible is what we call early exits. It's not an optimization technique, it's to boost readability - the `return` statements get highlighted in any decent IDE and you can easily see the control flow.


Don't forget to remove the no longer necessary `on_lava` variable after the fact :)

rr- (approved):

Issue #227, opened 2021-11-14
Original TRUB levels crash on final Atlantean Stronghold jump


Richard-L:
https://user-images.githubusercontent.com/4934209/141695425-c3f6512c-8d7c-4e3a-bf27-02a0ef556415.mp4

![image](https://user-images.githubusercontent.com/4934209/141695432-5e9d3f37-33c0-474c-9faf-853d5b823a95.png)

Save state: [saveuba.zip](https://github.com/rr-/Tomb1Main/files/7534477/saveuba.zip)

Can't test in TombATI too easily but I haven't seen this bug in any TRUB media where TombATI is used.

rr-:
I can't reproduce this issue on my virtual machine or on wine.  @oziphantom could you look into it with your debugger? You detected a stack corruption once :)

oziphantom:
nothing showing up in VS, I'm was running basically 1.3.0 + 60fps for the test.

Richard-L:
Yeah latest release.

I did replace the level's PHD from the stock EU TRUB release btw. So it's no longer the file that comes with TombATI v1.7.

Would that at all make a difference? Do I upload the PHD here?

oziphantom:
running a different level file is a massive change. which version did you change it to?

Richard-L:
The one I mentioned. PHD from the original UB release in 1998. TombATI uses a custom one with fan-made music triggers worked in. I wanted the vanilla experience.

rr-:
Unfortunately I don't have that file. Can you share it? To my knowledge UB is freeware so I don't think this is an issue.

Richard-L:
Oh good point! I forgot they had made it freeware since I'm always using the original CD. Will upload when I get home.

oziphantom:
finally got it to crash 
~~~
>	Tomb1Main.dll!T_RemovePrint(TEXTSTRING * textstring) Line 286	C
 	Tomb1Main.dll!Overlay_DrawAmmoInfo() Line 310	C
 	Tomb1Main.dll!Overlay_DrawGameInfo() Line 413	C
 	Tomb1Main.dll!DrawPhaseGame() Line 49	C
 	Tomb1Main.dll!GameLoop(int demo_mode) Line 76	C
 	Tomb1Main.dll!GF_InterpretSequence(int level_num, GAMEFLOW_LEVEL_TYPE level_type) Line 956	C
 	Tomb1Main.dll!GameMain() Line 100	C
[...]

PR #357, opened 2022-01-19
s/output refactors

rr-:
This shouldn't produce any visible changes to the game apart from perhaps resolving #227 – I added more out-of-bounds assertions and made a few arrays a bit bigger. I also simplified palette management and merged almost-identical clipper functions into one.


walkawayy:
Still crashes on that UB jump. But it sounds like that's a problem with the original level. LGTM.

walkawayy (approved):

PR #360, opened 2022-01-19
input: fix certain inputs skipping too many things

### LINKED ISSUE #359: Bug: level stats trigger FMV skips too fast
Holding the Action key for too long now skips both the level stats screen and FMVs. Regression from #341 .

rr-:
Resolves #359.

Scenarios needed to be tested:

(Escape key)

- [x] Holding the Escape key when transitioning from the game screen to the level stats screen.  
      This should trigger inventory on the frame before, and skip the level stats screen on the frame after. Hard to test and can be assumed to work correctly.
- [x] Holding the Escape key when transitioning from the game screen to an FMV.  
      This should trigger inventory on the frame before, and skip the FMV on the frame after. Hard to test and can be assumed to work correctly.
- [x] Holding the Escape key when transitioning from an FMV to the game.  
      Should skip the FMV immediately and then should not trigger the inventory ring.
- [ ] Holding the Escape key when transitioning from an FMV to the level stats screen.  
      Should skip the FMV immediately and then not skip the level stats screen until another key is pressed. (scenario not in vanilla TR1)
- [x] Holding the Escape key when transitioning from the level stats screen to an FMV.
      Should skip the level stats screen immediately and then not skip the FMV until another key is pressed.
- [x] Holding the Escape key when transitioning from the level stats screen to the game.  
      Should skip the FMV immediately and then should not trigger the inventory ring.
- [x] Holding the escape key when transitioning from a cutscene to the level stats screen.
      Should skip the cutscene immediately and then should not skip the level stats screen until another key is pressed.

(Action key)

- [x] Holding the Action key when transitioning from the game screen to an FMV.  
      This should not skip the FMV until another key is pressed.
- [x] Holding the Action key when transitioning from the game screen to the level stats screen.  
      This should not skip the level stats screen until another key is pressed.
- [x] Holding the Action key when transitioning from an FMV to the game.  
      Should skip the FMV immediately and then should make Lara immediately grab or shoot her guns.
- [ ] Holding the Action key when transitioning from an FMV to the level stats screen.  
      Should skip the FMV immediately and then not skip the level stats screen until another key is pressed. (scenario not in vanilla TR1)
- [x] Holding the Action key when transitioning from the level stats screen to an FMV.
      Should skip the level stats screen immediately and then not skip the FMV until another key is pressed.
- [x] Holding the Action key when transitioning from the level stats screen to the game.  
      Should skip the FMV immediately and then should make Lara immediately grab or shoot her guns.
- [x] Holding the escape key when transitioning from a cutscene to the level stats screen.
      Should skip the cutscene immediately and then should not skip the level stats screen until another key is pressed.

### DIFF
--- a/src/game/cinema.c
+++ b/src/game/cinema.c
@@ -83,7 +83,7 @@ bool DoCinematic(int32_t nframes)
         }
 
         Input_Update();
-        if (g_Input.option) {
+        if (g_InputDB.deselect || g_InputDB.select) {
             return true;
         }
 
--- a/src/game/control.c
+++ b/src/game/control.c
@@ -251,7 +251,7 @@ int32_t ControlPhase(int32_t nframes, GAMEFLOW_LEVEL_TYPE level_type)
             }
         }
 
-        if ((g_Input.option || g_Input.save || g_Input.load
+        if ((g_InputDB.option || g_Input.save || g_Input.load
              || g_OverlayFlag <= 0)
             && !g_Lara.death_count) {
             if (g_OverlayFlag > 0) {
--- a/src/game/game.c
+++ b/src/game/game.c
@@ -171,18 +171,8 @@ void LevelStats(int32_t level_num)
     Text_CentreH(txt, 1);
     Text_CentreV(txt, 1);
 
-    // wait till action key release
-    while (g_Input.select || g_Input.deselect) {
-        Input_Update();
-        Output_InitialisePolyList();
-        Output_CopyBufferToScreen();
-        Input_Update();
-        Text_Draw();
-        Output_DumpScreen();
-    }
-
-    // wait till action or escape key press
-    while (!g_Input.select && !g_Input.deselect) {
+    // wait till a skip key is pressed
+    do {
         if (g_ResetFlag) {
             break;
         }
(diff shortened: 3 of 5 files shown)


walkawayy:
Is there a scenario with this:
`Holding the Escape key when transitioning from an FMV to the level stats screen.
Should skip the FMV immediately and then not skip the level stats screen until another key is pressed.`? I think it only happens with cutscene -> level stat screen which I added to your list as working. Same applies for the action key one.

I think this scenario ` Holding the Escape key when transitioning from the level stats screen to an FMV.
Should skip the level stats screen immediately and then not skip the FMV until another key is pressed.` is broken. If I hold escape to skip the Lara vs Larson cutscene, the cutscene is skipped and the level stat screen appears. When I let go of escape, the level stat screen stays up. If I hold escape on the level stat screen, the FMV with Lara at Natla HQ starts, but if I release escape the FMV is skipped. Weirdly this seems to work as expected with the action key.

As a note, I can't test action key with cutscenes on this branch. Maybe that should be merged here? Or maybe we need to list out scenarios to test on the other PR.

rr-:
I'll merge it here.

rr-:
@walkawayy fixed and also included a similar fix for the credit roll pictures.

rr-:
>Holding the Escape key when transitioning from an FMV to the level stats screen. Should skip the FMV immediately and then not skip the level stats screen until another key is pressed.? I think it only happens with cutscene -> level stat screen which I added to your list as working. Same applies for the action key one.

I agree this doesn't make sense because this scenario doesn't exist, but should it exist, I believe it still would work with the new implementation :) I included it just because I was trying every possible combination and didn't think of it from this angle.

walkawayy:
Ok I think I tested every scenario and couldn't find any issues! Lgtm.

PR #361, opened 2022-01-20
Improve overlays

rr-:
This is continuation of #357 and lays the groundwork for the fade-in / fade-out effects. Rather than rendering the hacky picture buffer, we actually render the game, then a translucent quad, then the inventory.

This causes me to drop FPS.
Normally my game runs at around 12 FPS, but the inventory due to its simplicity renders at 25 FPS. With the new approach, the inventory runs at 8 FPS because it has to render both the game and the inventory ring on every frame.

Nonetheless, in my opinion it is acceptable, because I am running the game on a 4k window with a stupid emulator, which is equivalent of having a super potato machine.

### DIFF
--- a/src/game/cinema.c
+++ b/src/game/cinema.c
@@ -47,7 +47,6 @@ int32_t StopCinematic(int32_t level_num)
     Sound_StopAllSamples();
 
     g_LevelComplete = true;
-    Output_CopyScreenToBuffer();
 
     return level_num | GF_LEVEL_COMPLETE;
 }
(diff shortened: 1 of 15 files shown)


rr-:
@richard-l could you tell how fast this runs on your toaster?

[Tomb1Main-2.4-8-gd8778ba.zip](https://github.com/rr-/Tomb1Main/files/7901440/Tomb1Main-2.4-8-gd8778ba.zip)

walkawayy:
LGTM. I don't have a potato to test on though. 😉  Runs fine on my desktop.

walkawayy (approved):

rr-:
Pinging @richard-l :-]

Richard-L:
@rr- sorry only had time now :)

Reporting from potato, and the above build runs at steady 30fps. I overwrote all the cfg so it ran with your settings.

PR #368, opened 2022-01-20
output: add fade animations

rr-:
The fade effects use fixed transparency deltas and thus are FPS dependent. LMK if it's acceptable.

Testing scenarios
- [x] Starting the game. There should be no initial delay before first FMV
- [x] Starting the title screen. It should fade from black. Overlay should be drawn over the ring items.
- [x] Loading a game from the title screen. It should fade to black. Overlay should be drawn over the ring items.
- [x] Starting a demo from the title screen. It should fade to black. Overlay should be drawn over the ring items.
- [x] Starting Gym from the title screen. It should fade to black. Overlay should be drawn over the ring items.
- [x] Starting any level. It should not fade, Lara should be immediately fully visible.
- [x] Finishing a level / stats creen. It should fade to a semitransparent backdrop, then when the player exits the level stats, it should fade to black. Text should be drawn over the backdrop. Text should be completely removed as soon as fade to black happens.
- [x] Pause screen. It should fade to a semitransparent backdrop, then when the player exits the pause screen, it should fade out. Text should be drawn over the backdrop. Text should be completely removed as soon as the player exits the pause screen.
- [x] Opening the inventory. It should open immediately and fade to a semitransparent backdrop in the background.
- [x] Exiting the inventory without doing anything. It should fade to a semitransparent backdrop as soon as the player presses a key.
- [x] Exiting the inventory via using an item. It should use the item immediately and start fading out on the keypress, not after the ring is closed.
- [x] Exiting the inventory via saving the game (entered through Escape). It should start fading out right after choosing the slot, not after the ring is closed.
- [x] Exiting the inventory via saving the game (entered through F5). It should start fading out right after choosing the slot, not after the ring is closed.
- [x] Exiting the inventory via loading the game (entered through Escape). It should fade to black immediately after choosing the slot, not after the ring is closed.
- [x] Exiting the inventory via loading the game (entered through F6). It should fade to black immediately after choosing the slot, not after the ring is closed.
- [x] Exiting the inventory via loading the game (entered through dying). It should fade to black immediately after choosing the slot, not after the ring is closed.
- [x] Exiting the inventory via exiting to title (entered through Escape). It should fade to black immediately after choosing the slot, not after the ring is closed.
- [x] Exiting the inventory via exiting to title (entered through dying). It should fade to black immediately after choosing the slot, not after the ring is closed.
- [x] Exiting the inventory via saving the game in Gym. This redirects the user to starting a new game. It should fade to black immediately after the player confirms.
- [x] Displaying Eidos logo. There should be no fades.
- [x] Displaying credits. There should be no fades. It might be tackled in the future as a separate ticket.
- [x] The version string in the main menu should not be placed above the overlay
- [x] All of these should behave like it used to when the `enable_fade_effects` setting is turned off.

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,6 +250,7 @@ Not all options are turned on by default. Refer to `Tomb1Main.json5` for details
 - added support for HD FMVs
 - added fanmade 16:9 menu backgrounds
 - added ability to skip FMVs with the Action key
+- added fade effects
 - changed internal game memory limit from 3.5 MB to 16 MB
 - changed moveable limit from 256 to 10240
 - changed maximum textures from 2048 to 8192
--- a/bin/cfg/Tomb1Main.json5
+++ b/bin/cfg/Tomb1Main.json5
@@ -63,6 +63,10 @@
     // Makes the healthbar and airbar use smooth color transitions.
     "enable_smooth_bars": true,
 
+    // Makes inventory and pause screen use fade transitions for
+    // the semitransparent backdrop, similar to PS1.
+    "enable_fade_effects": true,
+
     // Changes how the healthbar is displayed. Possible values:
     // - default: show the healthbar at the start of a level, after getting
     //   hit, or while having weapons equipped; flash the healthbar if it's already
--- a/src/config.c
+++ b/src/config.c
@@ -137,6 +137,7 @@ bool Config_ReadFromJSON(const char *cfg_data)
     READ_BOOL(enable_compass_stats, true);
     READ_BOOL(enable_timer_in_inventory, true);
     READ_BOOL(enable_smooth_bars, true);
+    READ_BOOL(enable_fade_effects, true);
     READ_BOOL(fix_tihocan_secret_sound, true);
     READ_BOOL(fix_pyramid_secret_trigger, true);
     READ_BOOL(fix_secrets_killing_music, true);
--- a/src/config.h
+++ b/src/config.h
@@ -55,6 +55,7 @@ typedef struct {
     bool enable_compass_stats;
     bool enable_timer_in_inventory;
     bool enable_smooth_bars;
+    bool enable_fade_effects;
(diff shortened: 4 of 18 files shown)


walkawayy:
Very nice!

Couple of notes:
- [ ] When saving in game, the inventory fades to black and returns to the game instead of fading back into the game. Oddly, it fades back to the game correctly when saving via F5.
- [ ] Starting a new game from Lara's Home fades into game instead of black before loading the Caves FMV.
- [ ] I think the fade needs to be darker when starting a new game, loading from title, and when loading a game from the in game inventory. The last frames you can see the game still. See below for a not perfect screenshot but pretty close to get the picture:
20220120_105434_Tomb_of_Qualopec

- [ ] Not sure if this is really a bug and def not a normal use case, but if you spam screenshot during the fading, the game doesn't fade properly. Just thought I'd bring it up. See below:
20220120_105235_Tomb_of_Qualopec
.

walkawayy:
I didn't notice any issues with fade effects disabled.

In save crystal mode, I didn't notice any issues.

walkawayy:
> 

I think that's a very extreme case that I was only doing to show the game not fully fading. I don't think it's an issue but I mentioned it just in case.

walkawayy:
I think there's just one thing left to fix:

- [ ] Starting a new game from Lara's Home from the inventory works. Starting a new game using F5 fades into game instead of black before loading the Caves FMV.

rr-:
Fixed this as well :)

walkawayy:
Awesome!

walkawayy (approved):

rr-:
Since this uses #357 we'll have to wait with merging until then.

rr-:
Resolved the conflicts and fixed a bug where opening the inventory when Lara was underwater would cause the items to be drawn in blue.

PR #300, opened 2022-01-06
misc: fix memory leaks

rr-:
The worst offender was the audio samples loader that only freed first 20 samples instead of all of them, which leaked tons of raw PCM data for each level start / savegame load. The second worst offender was the gameflow loader which didn't do anything at all to free the memory.

Also added `Memory_Dup` that replaces `strdup` so that we only allocate with `Memory_Alloc` across the entire application. This has actually helped me track bad allocations (see the notes below).
Also added `Memory_FreePointer` utility that can replaces the following idiom:
```c
if (memory) {
    Memory_Free(memory);
    memory = NULL;
}
```
with
```c
Memory_FreePointer(&memory);
```

---

The following patch and script helped me analyze the leaks. It tracks every single memory operation in logs, and the script checks if the memory is still reachable. Adding these in case this ever needs to be used again. In theory the same should be doable with valgrind, but I can assure you that using valgrind on wine is not a pleasant experience.

```diff
diff --git a/src/memory.c b/src/memory.c
index c73c24cf..582fe29c 100644
--- a/src/memory.c
+++ b/src/memory.c
@@ -1,14 +1,20 @@
 #include "memory.h"
 
 #include "game/shell.h"
+#include "log.h"
 
 #include <assert.h>
 #include <stdlib.h>
 #include <string.h>
 
-void *Memory_Alloc(size_t size)
+#define DEBUG_MEMORY
+
+void *Memory_AllocImpl(size_t size, const char *file, int line, const char *func)
 {
     void *result = malloc(size);
+    #ifdef DEBUG_MEMORY
+    Log_Message(file, line, func, "Allocating memory (%d bytes): %p", size, result);
+    #endif
     if (!result) {
         Shell_ExitSystem("ERROR: Could not allocate enough memory");
     }
@@ -16,33 +22,39 @@ void *Memory_Alloc(size_t size)
     return result;
 }
 
-void *Memory_Realloc(void *memory, size_t size)
+void *Memory_ReallocImpl(void *memory, size_t size, const char *file, int line, const char *func)
 {
     void *result = realloc(memory, size);
+    #ifdef DEBUG_MEMORY
+    Log_Message(file, line, func, "Reallocating memory at %p (%d bytes): %p", memory, size, result);
+    #endif
     if (!result) {
         Shell_ExitSystem("ERROR: Could not allocate enough memory");
     }
     return result;
 }
 
-void Memory_Free(void *memory)
+void Memory_FreeImpl(void *memory, const char *file, int line, const char *func)
[...]

PR #370, opened 2022-01-20
s/audio: fix memory leak

rr-:
Also changed json.c to use our allocator routines, after all. The script from #300 reported invalid frees() because we use `Memory_FreePointer()` on data allocated by `json_write_pretty`, which is a simple `char*`. We could make a separate `json_write_free` for it but I guess it's overkill.

### DIFF
--- a/src/json.c
+++ b/src/json.c
@@ -1,6 +1,7 @@
 #include "json.h"
 
 #include "global/const.h"
+#include "memory.h"
 
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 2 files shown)


rr-:
@walkawayy ping

walkawayy:
I'm sorry my trip extended, so I am away from my desktop the rest of this week again. The code looks good to me, but I can't do any testing. I think it looks safe to merge though.

walkawayy (approved):

rr-:
No worries. Thanks for the review

PR #377, opened 2022-01-31
option_graphics: make vsync optional from glasses menu

### LINKED ISSUE #364: Feature request: make the vsync option configurable
Related: #312

walkawayy:
Resolves #364.


walkawayy:
In `context.c` during `GFX_Context_Attach` this was getting set once on startup. This happened before the runtime JSON was getting read. So I just left this on by default for startup, and if a runtime JSON is present, it will set the vsync option depending on the user's preference.

src/gfx/context.h
```diff
@@ -10,6 +10,7 @@ typedef struct GFX_Context GFX_Context;
 
 void GFX_Context_Attach(HWND hwnd);
 void GFX_Context_Detach();
+void GFX_Context_SetVsync();
```
rr-, line 13: Can we change the signature to `GFX_Context_SetVsync(bool x)` to decouple GFX from Config?

rr-:
Also let's add a line to CHANGELOG.md to signify the changes. This is a new file meant to help us prepare releases and provide historic data.

src/gfx/context.h
walkawayy: Ya idk what I was thinking when I wrote that lol. Fixed.

walkawayy:
Btw this is probably a separate change for the future, but I was thinking that these rendering struct variables should be bools
```
        uint32_t enable_perspective_filter : 1;
        uint32_t enable_bilinear_filter : 1;
        uint32_t enable_vsync : 1;
```
Would that break anything in the JSON reading or anything?

rr-:
I don't think it would break anything.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 - added fade animations
 - fixed certain inputs skipping too many things
 - fixed a memory leak in the audio sampler
+- added a vsync option
```
rr-, line 6: Let's move it a bit to the top to keep the changelog items sorted via their category.

src/gfx/context.h
```diff
@@ -10,6 +10,7 @@ typedef struct GFX_Context GFX_Context;
 
 void GFX_Context_Attach(HWND hwnd);
 void GFX_Context_Detach();
+void GFX_Context_SetVsync(bool vsync);
```
rr-, line 13: In another place we say `VSync` but here we use `Vsync`, let's make it `VSync` here as well.

rr- (approved):

PR #381, opened 2022-01-31
Fix detail levels text flashing

### LINKED ISSUE #380: All detail levels text flashes when an option is changed
If you change any option in the glasses menu, the entire screen of text flashes when the text refreshes.

Also, toggle options should use bools.

walkawayy:
option_graphics: fix options flashing with any option change
option_graphic: use bools for graphic settings

Resolves #380.

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -90,10 +90,10 @@ typedef struct {
     } input;
 
     struct {
-        bool enable_perspective_filter : true;
-        bool enable_bilinear_filter : true;
-        bool enable_vsync : true;
-        bool enable_fps_counter : true;
+        bool enable_perspective_filter;
+        bool enable_bilinear_filter;
+        bool enable_vsync;
+        bool enable_fps_counter;
         float anisotropy_filter;
     } rendering;
 
--- a/src/config.h
+++ b/src/config.h
@@ -90,10 +90,10 @@ typedef struct {
     } input;
 
     struct {
-        uint32_t enable_perspective_filter : 1;
-        uint32_t enable_bilinear_filter : 1;
-        uint32_t enable_vsync : 1;
-        uint32_t enable_fps_counter : 1;
+        bool enable_perspective_filter : true;
+        bool enable_bilinear_filter : true;
+        bool enable_vsync : true;
+        bool enable_fps_counter : true;
         float anisotropy_filter;
     } rendering;
 
--- a/src/game/option_graphics.c
+++ b/src/game/option_graphics.c
@@ -14,6 +14,7 @@
 #define TOP_Y -55
 #define ROW_HEIGHT 25
 #define ROW_WIDTH 280
+#define OPTION_LENGTH 256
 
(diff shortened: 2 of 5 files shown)


src/game/option_graphics.c
```diff
@@ -195,64 +262,68 @@ void Option_Graphics(INVENTORY_ITEM *inv_item)
...
-    if (reset) {
+    if (reset == TEXT_NUMBER_OF) {
         for (int i = 0; i < TEXT_NUMBER_OF; i++) {
             Text_Remove(m_Text[i]);
             m_Text[i] = NULL;
         }
+        m_IsTextInit = false;
+        Settings_Write();
+    } else if (reset > -1) {
+        Option_GraphicsChangeTextOption(reset);
```
rr-, line 326: Are we sure we don't need to `Settings_Write` otherwise?

src/config.h
```diff
@@ -90,10 +90,10 @@ typedef struct {
...
 
     struct {
-        uint32_t enable_perspective_filter : 1;
-        uint32_t enable_bilinear_filter : 1;
-        uint32_t enable_vsync : 1;
-        uint32_t enable_fps_counter : 1;
+        bool enable_perspective_filter : true;
+        bool enable_bilinear_filter : true;
+        bool enable_vsync : true;
+        bool enable_fps_counter : true;
```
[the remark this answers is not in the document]
walkawayy: LMAO whoops.

src/game/option_graphics.c
walkawayy: It writes the settings in both the if and else if.

src/game/option_graphics.c
```diff
@@ -114,10 +117,74 @@ static void Option_GraphicsInitText()
...
+                                                              : GS_MISC_OFF]);
+        Text_ChangeText(m_Text[TEXT_PERSPECTIVE], buf);
+        break;
+
+    case TEXT_BILINEAR:
+        sprintf(
+            buf, g_GameFlow.strings[GS_DETAIL_BILINEAR_FMT],
+            g_GameFlow.strings
+                [g_Config.rendering.enable_bilinear_filter ? GS_MISC_ON
+                                                           : GS_MISC_OFF]);
```
rr-, line 140: Man this formatting. I hate clang-format sometimes but it's 20× times better than nothing

rr- (approved):

src/game/option_graphics.c
walkawayy: No because it's the case where no option was selected. We don't need to save settings every loop right? Maybe we need to make it save in the initial case?

rr-:
We're missing the entry in changelog.

walkawayy:
> We're missing the entry in changelog.

Do you know if this is an OG bug or a T1M bug?

rr-:
Pretty sure OG bug

walkawayy:
Alright I think it's done.

rr- (approved):

Issue #250, opened 2021-11-22
OG Bug: Lara torsos on Atlantean eggs in UB's "The Hive"


Richard-L:
Near the end of the fourth UB level "The Hive", the empty Atlantean eggs show these Lara torso objects: 

![image](https://user-images.githubusercontent.com/4934209/142786688-b32e5dfb-9c43-47d7-9856-2e69766782cb.png)

I know this has been brought up before elsewhere, but I don't recall fully why this happens. I believe it may be an OG bug?

Save state if you wanted: [saveuba.zip](https://github.com/rr-/Tomb1Main/files/7577560/saveuba.zip)

oziphantom:
I believe they go over the objects limits which causes issues, but I also though rr upped the limit on T1M..

rr-:
We have more capacity for ITEM_INFO / Items, yes. I'll investigate it later

walkawayy:
WadTool shows the "mutant spawn object" is Lara's torso. I'd guess normally these are hidden by the eggs, and are supposed to disappear when the egg hatches and the mutant appears so the player never sees it. Must not be going away for some reason.

rr-:
Can someone confirm if it's an OG bug or not and if it is, how to reproduce it (apart from loading a broken save)?

Richard-L:
You can see the torso object here: https://youtu.be/hos-BgFkTpI?t=11633

I'm slightly confused by the size of the health bar in this video though. Perhaps TombATI in very low resolution, which makes it appear so big.

walkawayy:
> Can someone confirm if it's an OG bug or not and if it is, how to reproduce it (apart from loading a broken save)?

Ok I just played a fresh game on TombATI and got the bug just be playing normally. After the first few eggs in this room, no more mutants spawned. Upon a save and reload the torsos disappeared.

![AnyConv com__screenshot0000](https://user-images.githubusercontent.com/81546780/148300452-b12e9397-9063-4a43-9837-2506fc8cb8c7.jpg)

Richard-L:
Actually, I always assumed no mutants are supposed to spawn, to symbol "bad" hatchings in the natla-does-experiments sense. But perhaps that's wrong?

walkawayy:
[...]

PR #387, opened 2022-02-02
compass: add total number of pickups and killables

### LINKED ISSUE #362: Feature request: more robust level stats
Add an optional ability to display

KILLS: x **of y**
PICKUPS: x **of y**

in the level stats screen.

walkawayy:
Resolves #362.

Prepared to get slammed, but as far as I know we have to manually define lists for objects that are killable and pickupable. Not sure if I missed any.

For killables, I had to remove some enemies like the Mummy from Qualopec who can't be killed. But Atlantis has an issue where Lara can kill more than there are killables, but idk what I missed.

For pickups, the different Scions behave differently. I excluded O_SCION_ITEM3 because that is the one you have to shoot.


src/game/level.c
```diff
@@ -403,6 +429,20 @@ static bool Level_LoadItems(MYFILE *fp)
             }
 
             InitialiseItem(i);
+
+            // Calculate number of pickups in a level
+            for (int16_t j = 0; j < NUM_PICKUP_OBJ; j++) {
+                if (item->object_number == m_PickupObjs[j]) {
+                    m_LevelPickups++;
+                }
+            }
```
rr-, line 438: What if we test if the draw routine is `DrawPickupItem`?

src/game/level.c
```diff
@@ -403,6 +429,20 @@ static bool Level_LoadItems(MYFILE *fp)
...
+
+            // Calculate number of killable objects in a level
+            for (int16_t j = 0; j < NUM_KILLABLE_OBJ; j++) {
+                if (item->object_number == m_KillableObjs[j]) {
+                    m_LevelKillables++;
+                }
+            }
```
rr-, line 445: How this behaves with Pierre?

rr- (commented):
We have to introduce a bit more intricate solution around `O_PODS`, which are mutant eggs. If we have initialized items already at this point, we can check if `item->data` is set to a non-zero value (see `InitialisePod`). I think `O_ABORTION` won't be accounted for either and needs to be handled under similar rules as `O_BIG_POD`.

rr-:
@richard-l would you prefer this to be an opt-in feature or is it OK to include it as-is?

Richard-L:
@rr- I would say do it whichever way is easier to maintain and less workload for you.

The only reason I could think of to make it optional was to maintain a "console perfect" list of settings, where the game's look and behaviour is really perfectly identical to the PS1 version. A blurry line of course.

rr-:
I would say it's OK to make it non-configurable.
However, we should add a changelog entry.

src/game/level.c
walkawayy: Hmm in levels where you fight but don't kill him, it seems like the game doesn't let his HP go below half, so this will cause there to be extra unkillables for those levels. I'll need to check how the Pierre code is handled.

src/game/level.c
walkawayy: That could work I think. Covers it all xcept for `PickUpScion4Collision` I think.

rr-:
Before we go through with it, I'd like for some professional Tomb Raider player to confirm the expected numbers of reachable items and enemies in every level. Maybe Stella's walkthrough could help.

walkawayy:
Ya I agree. speedrun.com doesn't list kills or pickups.

Thanks to Rapora9 from Discord for compiling this list from [this 100% playthrough](https://www.youtube.com/watch?v=2Ea6_odQv5w). I can try and cross check it with another playthrough. Especially since Atlantis can't be seen.

```
LEVEL: KILLS, PICKUPS

Caves:      14, 7
Vilcabamba: 29, 13
Valley:     13, 16
Qualopec:   8, 8
Folly:      23, 19
Colosseum:  26, 14
Midas:      43, 23
Cistern:    34, 28
Tihocan:    17, 26
City:       14, 24
Obelisk:    16, 38
Sanctuary:  15, 29
Natla:      3, 30
Atlantis: (not visible on the recording)
Pyramid:    6, 31
```

walkawayy:
Here's a list from speedrunner Footi:
               
Level | Kills | Pickups | Secrets |
| ----------- | ----------- | ----------- | ----------- |
Caves | 14 | 7 | 3
Vilcabamba | 13  | 13 | 3
Lost Valley | 14 | 16 | 5
Qualopec | 8 | 8 | 3
Folly | 23 | 19 | 4
Colosseum | 26 | 14 | 3
Midas | 43 | 23 (22) | 3
Cistern | 34 | 28 | 3
Tihocan | 17 | 26 | 2
City | 14 | 24 | 3
Obelisk | 16 | 38 | 3
Scion | 15 | 29 | 1
Caves | 14 | 7 | 3
Caves | 14 | 7 | 3
Caves | 14 | 7 | 3
Natla's Mines     3	   30	     3
Atlantis	  32	   50        3
Great Pyramid     6        31	     3

rr-:
It's OK to count that medipack as pickupable IMO.

Richard-L:
I agree. We were just talking about it on Discord. I've also asked Footi for a list for the UB levels.

walkawayy:
> I agree. We were just talking about it on Discord. I've also asked Footi for a list for the UB levels.

Cool thanks. That number is probably messed up due to #250.

Richard-L:
Footi thankfully asked Anopob, the only guy who has done a UB max% run, for this numbers. Without guarantee:

> yeah i have, for glitchless (same as in my run):
Return to Egypt: 41 Kills, 53 Pickups
Temple of the Cat: 44 Kills, 63 Pickups
Atlantean Stronghold: 31 Kills, 63 Pickups
The Hive: 41 Kills, 60 Pickups

src/game/level.c
```diff
@@ -17,8 +17,8 @@
 
 #include <stdio.h>
 
-#define NUM_PICKUP_OBJ 28
-#define NUM_KILLABLE_OBJ 26
+#define NUM_PICKUP_OBJ 24
+#define NUM_KILLABLE_OBJ 27
```
rr-, line 21: Let's remove this in favor of `NO_ITEM` guard at the end of the list.

walkawayy: Ok. Would I change the for loops to something like?
```
            // Calculate number of killable objects in a level
            for (int16_t j = 0; j != NO_ITEM; j++) {
                if (item->object_number == m_KillableObjs[j]) {
                    m_LevelKillables++;
                }
            }
```

Also thanks for marking as draft because I forgot to do that.

walkawayy:
> We have to introduce a bit more intricate solution around `O_PODS`, which are mutant eggs. If we have initialized items already at this point, we can check if `item->data` is set to a non-zero value (see `InitialisePod`). I think `O_ABORTION` won't be accounted for either and needs to be handled under similar rules as `O_BIG_POD`.

So I did some digging on the pods stuff. The reason my code does not currently calculate the number of mutants correctly is because I do not include `O_PODS` and `O_BIG_POD` as killables (for good reason). When the level loads (where my code is counting the enemies), the game doesn't load any of the mutants spawned from pods. These enemies are spawned when a pod trigger is stepped on the pod explodes. I can't just count the pods though because the Atlantis level has the abortion pod that will never spawn an enemy because it has no trigger. Unfinished Business has similar eggs that have no trigger.

The solution might require some floor checking to find all pod triggers while making sure not to count multiple triggers for the same pod. The last case would be the broken UB Hive eggs that spawn an abortion mutant that isn't in the level files. But that seems like a level design issue like unreachable medipacks.

src/game/level.c
rr-: No, the point of having a guardian is to avoid having to specify the array size.
The loop doesn't look right either.

```c
static int16_t m_KillableObjs[] = { ... };

...

for (int j = 0; m_KillableObjs[j] != NO_ITEM; j++) {
    if (item->object_number == m_KillableObjs[j]) {
        m_LevelKillables++;
    }
}
```

Also, no need for being this specific about the loop variable type.

rr-:
```
The solution might require some floor checking to find all pod triggers while making sure not to count multiple triggers for the same pod. The last case would be the broken UB Hive eggs that have triggers and explode, but they spawn an abortion mutant that isn't in the level files. But that seems like a level design issue like unreachable medipacks.
```

I'm not sure about UB but yes, that's what I meant in my previous post about the Atlanteans. 

>We have to introduce a bit more intricate solution around O_PODS, which are mutant eggs. If we have initialized items already at this point, we can check if item->data is set to a non-zero value (see InitialisePod). I think O_ABORTION won't be accounted for either and needs to be handled under similar rules as O_BIG_POD.

walkawayy:
> 

Ya. Also not sure what to do about pickups right now. Enemies like Pierre and the mercenaries spawn items on death. Adding to the total mid level would be easy but idk if that's nice. For example, when you kill Pierre in Tihocan, the stats screen could go from 23 of 23 pickups, to 23 of 26. 

Pierre's code as reference:
```

    if (item->hit_points <= 0) {
        if (item->current_anim_state != PIERRE_DEATH) {
            item->current_anim_state = PIERRE_DEATH;
            item->anim_number =
                g_Objects[O_PIERRE].anim_index + PIERRE_DIE_ANIM;
            item->frame_number = g_Anims[item->anim_number].frame_base;
            SpawnItem(item, O_MAGNUM_ITEM);
            SpawnItem(item, O_SCION_ITEM2);
            SpawnItem(item, O_KEY_ITEM1);
        }
```

rr-:
Same for the Skate kid and the Baldie. I think it's OK to hardcode the amount of additional items with a proper comment.

IIRC Baldie drops a Shotgun. I don't remember what happens if Lara already has it in her inventory, if he drops Shotgun, Shotgun Ammo, or nothing (the worst case).

walkawayy:
> Same for the Skate kid and the Baldie. I think it's OK to hardcode the amount of additional items with a proper comment.
> 
> IIRC Baldie drops a Shotgun. I don't remember what happens if Lara already has it in her inventory – whether he drops Shotgun, Shotgun Ammo, or nothing (the worst case).

Ya all the mercenaries do. Looks like skate kid dropped the Uzis even though I already had them. I know custom games aren't the main focus, but I also didn't want to hard code extra items per level. And you can't just check if Pierre is in the level or something, since he runs away depending on the one shot trigger or not. Maybe I'll do floor parsing for this too and if those enemies are present and can die, just add their items to the total.

Richard-L:
Remember Lara's stripped of all weapons at the beginning of Natla's Mines, so all weapons will show as weapons, not ammo.

Richard-L:
@walkawayy since Natla's Mines strips you of all weapons, how did you have the UZIs already, via cheat? I don't recall there being a secret where you get them. I think you only get them from skateboard kid. There is however the boulder secret after the lava pit, where you get the shotgun.

You can see here that Footi's killing baldie with the shotgun he has from that secret. And when he goes to pick it up (pulls a switch first), it shows as the shotgun item instead of ammo, interestingly. https://youtu.be/D3qgf84dLDg?t=8353

I know that if you pick up the shotgun in Lost Valley, then the secret with it in Folly will show shotgun ammo. If you haven't picked up the Lost Valley shotgun, then it will be here instead of ammo. So perhaps do weapons only update into ammo if there's a level swap in between? The shotgun behaviour of Natla's Mines would indicate so.

src/game/level.c
```diff
@@ -711,8 +708,167 @@ bool Level_Load(int level_num)
...
+                }
+                }
+
+                if (floor & END_BIT) {
+                    break;
+                }
+            }
+        }
+    }
+}
```
rr-, line 874: Since this contains a lot of logic, let's have a new module for this, `levelstats` and move `getsecretcount` there as well.

walkawayy:
Ok I think this is ready for review, but by new module do you mean moving this to a new .c file with those other functions @rr- ?

rr-:
Yes. And reduce the amount of global state. Perhaps introduce a new module-level struct that you pass around between the functions.

src/game/level.c
```diff
@@ -671,7 +731,190 @@ bool Level_Load(int level_num)
...
+
+static bool Level_FindElement(int16_t elem, int16_t array[], int16_t size)
+{
+    for (int16_t i = 0; i < size; i++) {
+        if (array[i] == elem) {
+            return true;
+        }
+    }
+    return false;
```
rr-, line 806: I don't get the purpose of this function

src/game/level.c
```diff
@@ -34,6 +41,44 @@ static int32_t m_AnimTextureRangeCount = 0;
...
+static int32_t m_LevelKillables = 0;
+static int32_t m_LevelSecrets = 0;
+static int32_t m_UninitItemCount = 0;
+static FLOOR_INFO **m_FloorArray = NULL;
+
+typedef struct STAT_INFO {
+    int32_t drops;
+    int32_t kills;
+    int16_t cur_idx;
+    int16_t idxs[STAT_SIZE];
```
rr-, line 54: What do we need the indices for?

src/game/level.c
walkawayy: Since there can be multiple triggers for the same object, the item index is stored to make sure the same item isn’t counter multiple times. The idea was to use the array kind of like a hash map. If there’s a better way to do this let me know. My C is rusty and in Python I would have used a dict.

rr-: Maybe a linked list of found objects?

walkawayy: I guess I don't need a key and a value in a dict. I just said that because it's easy to check if a key is already in the dict to avoid adding duplicates. C doesn't have much standard library support with containers. I could put objects in a linked list, but I need to check for duplicates before insertion, or just add everything and remove duplicates after right? If you have a recommendation let me know. I'm honestly not sure how best to implement that in C.

src/game/stats.c
```diff
@@ -0,0 +1,198 @@
...
+    O_VOLE,     O_DINOSAUR,    O_RAPTOR,   O_WARRIOR1,  O_WARRIOR2,
+    O_WARRIOR3, O_CENTAUR,     O_MUMMY,    O_ABORTION,  O_DINO_WARRIOR,
+    O_FISH,     O_LARSON,      O_SKATEKID, O_COWBOY,    O_BALDY,
+    O_NATLA,    O_SCION_ITEM3, O_STATUE,   O_PODS,      O_BIG_POD,
+    NO_ITEM
+};
+
+static void Stats_CheckTriggers(FLOOR_INFO **floor_array);
+
+static void Stats_CheckTriggers(FLOOR_INFO **floor_array)
```
rr-, line 44: Looks pretty good but I don't quite don't understand why do we need to create a copy of the floor data for the stats calc?

src/game/stats.c
```diff
@@ -0,0 +1,198 @@
...
+
+int32_t Stats_GetKillables()
+{
+    return m_LevelKillables;
+}
+
+int32_t Stats_GetSecrets()
+{
+    return m_LevelSecrets;
+}
```
rr-, line 198: Missing newline

src/game/stats.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include "global/types.h"
+
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
```
rr-, line 7: I don't think we need `stdbool` and `stddef`

src/game/stats.h
```diff
@@ -0,0 +1,13 @@
...
+
+void Stats_Init();
+void Stats_CalculateStats(int32_t uninit_item_count, FLOOR_INFO **floor_array);
```
rr-, line 10: I think this can be `FLOOR_INFO*` and later on we just reduce one level of pointer dereferencing (e.g. remove `&` or change `->` to `.`)

src/game/stats.c
```diff
@@ -0,0 +1,198 @@
...
+        }
+    }
+}
+
+void Stats_Init()
+{
+    for (int i = 0; m_KillableObjs[i] != NO_ITEM; i++) {
+        m_IfKillable[m_KillableObjs[i]] = true;
+    }
+}
```
rr-, line 146: I think this needs to be ran at the beginning of stats calculation, not once per game launch?

src/game/stats.c
walkawayy: Because of the way some doors work. There was a bug where the trigger check was missing triggers in a rare instance. See the screenshot below for a trigger that is part of the bigger room but is behind a closed door. When the doors are initialized, closed doors set the floor index to 0, so this trigger was never getting checked.

image

src/game/stats.c
walkawayy: The enemies that are killable never change, so it should only need run once on game launch.

walkawayy:
I'm checking on the floor_data pointer thing.

The only two issues remaining I think are:
1. GetSecretCount still exists and used in `gameflow.c` for `g_GameFlow.levels[g_CurrentLevel].secrets = GetSecretCount();`. I tried to add secret counting into my `Stats_CheckTriggers` function. But I don't understand why Great Pyramid secret count fix, so my calculation seems to be missing one secret in that level.

2. I can't figure out why the order is flipped for the compass and the end of level stats screen. Not a big deal but I'd rather it just be consistent.

image

image

src/game/stats.h
walkawayy: I don't think that will work. I think it's fine in the current form.

rr-:
Let's make the order consistent with the end of the level stats, but in a separate ticket, not this one.

The Great Pyramid secret fix works by changing tiles that trigger a secret. It makes sure that secret triggers placed in different rooms trigger unique secrets. So room 555 and room 666 shouldn't both trigger secret 2.

The level stats need to be recalculated after applying this fix.

I'll need to think what to do with the extra allocation, as I don't like the way it's coupled with the stats now. Maybe tomorrow.

Richard-L:
@walkawayy the "OF" should be "of". Helps a lot with legibility.
image

Richard-L:
And whilst I'm comparing the two... here's a late thought: Do we spoil a level to someone if they're seeing very high max available kills/pickup numbers? It would indicate the level is huge and something wild is happening, or if max numbers are low you would know the level is more riddle based and chill re combat. I wonder if this might take away the wonder of a first time TR1 experience, as information about the level is so readily benchmarked. Should we only show the "x of y" at the level end stats?

I'm not sure myself, and I'm also not sure if this opinion is helpful........ rr I think you even asked if it makes sense to have this optional.

The secrets are more like a treasure hunt, so I can see why it's nice to know during a level.

walkawayy:
> And whilst I'm comparing the two... here's a late thought: Do we spoil a level to someone if they're seeing very high max available kills/pickup numbers? It would indicate the level is huge and something wild is happening, or if max numbers are low you would know the level is more riddle based and chill re combat. I wonder if this might take away the wonder of a first time TR1 experience, as information about the level is so readily benchmarked. Should we only show the "x of y" at the level end stats?
> 
> I'm not sure myself, and I'm also not sure if this opinion is helpful........ rr I think you even asked if it makes sense to have this optional.
> 
> The secrets are more like a treasure hunt, so I can see why it's nice to know during a level.

rr is the one who opened this issue, and the compass stats are optional via a JSON setting.

src/game/stats.c
rr-: Could we remove `Stats_Init` and replace `m_IfKillable` with something like this?

```c
static bool Stats_IsObjectKillable(int32_t obj_num)
{
    for (int i = 0; m_KillableObjs[i] != NO_ITEM; i++) {
        if (m_KillableObjs[i] == obj_num) {
            return true;
        }
    }
    return false;
}

File: src/game/stats.c [hunk 12]
```diff
@@ -0,0 +1,305 @@
...
+static void Stats_CheckTriggers(FLOOR_INFO **floor_array)
+{
+    uint32_t secrets = 0;
+
+    for (int i = 0; i < g_RoomCount; i++) {
+        ROOM_INFO *r = &g_RoomInfo[i];
+        for (int x_floor = 0; x_floor < r->x_size; x_floor++) {
+            for (int y_floor = 0; y_floor < r->y_size; y_floor++) {
+
+                if (x_floor == 0 || x_floor == r->x_size - 1) {
```
rr-, line 60: This is too nested; let's introduce a function for this.

src/game/stats.c
walkawayy: We could I guess, that's how the old method was sort of done. This just does more work since every time you check an object, you have to iterate through the `m_KillableObjs` array.

src/game/stats.c
walkawayy: Just this part?
```
                if (x_floor == 0 || x_floor == r->x_size - 1) {
                    if (y_floor == 0 || y_floor == r->y_size - 1) {
                        continue;
                    }
                }
```

rr-:
@walkawayy I would suggest the following *untested* patch. It reduces the state creep in `level.c` offloading the caching to `stats.c` where it belongs, and also postpones the stats calculation until after the TGP secret fix. LMK what you think.

```
diff --git a/src/game/control.h b/src/game/control.h
index 0553ad15..cc7c20ff 100644
--- a/src/game/control.h
+++ b/src/game/control.h
@@ -24,6 +24,4 @@ void FlipMap();
 void RemoveRoomFlipItems(ROOM_INFO *r);
 void AddRoomFlipItems(ROOM_INFO *r);
 
-int32_t GetSecretCount();
-
 bool Control_Pause();
diff --git a/src/game/control_util.c b/src/game/control_util.c
index 92b5708c..0134d66f 100644
--- a/src/game/control_util.c
+++ b/src/game/control_util.c
@@ -606,69 +606,3 @@ int32_t ClipTarget(GAME_VECTOR *start, GAME_VECTOR *target, FLOOR_INFO *floor)
 
     return 1;
 }
-
-int GetSecretCount()
-{
-    int count = 0;
-    uint32_t secrets = 0;
-
-    for (int i = 0; i < g_RoomCount; i++) {
-        ROOM_INFO *r = &g_RoomInfo[i];
-        FLOOR_INFO *floor = &r->floor[0];
-        for (int j = 0; j < r->y_size * r->x_size; j++, floor++) {
-            int k = floor->index;
-            if (!k) {
-                continue;
-            }
-
-            while (1) {
-                uint16_t floor = g_FloorData[k++];
-
-                switch (floor & DATA_TYPE) {
-                case FT_DOOR:
-                case FT_ROOF:
-                case FT_TILT:
-                    k++;
-                    break;
-
-                case FT_LAVA:
-                    break;
-
-                case FT_TRIGGER: {
-                    uint16_t trig_type = (floor & 0x3F00) >> 8;
-                    k++; // skip basic trigger stuff
-
-                    if (trig_type == TT_SWITCH || trig_type == TT_KEY
-                        || trig_type == TT_PICKUP) {
-                        k++;
-                    }
-
-                    while (1) {
-                        int16_t command = g_FloorData[k++];
-                        if (TRIG_BITS(command) == TO_CAMERA) {
-                            k++;
-                        } else if (TRIG_BITS(command) == TO_SECRET) {
-                            int16_t number = command & VALUE_BITS;
-                            if (!(secrets & (1 << number))) {
-                                secrets |= (1 << number);
-                                count++;
-                            }
-                        }
-
-                        if (command & END_BIT) {
-                            break;
-                        }
-                    }
-                    break;
-                }
-                }
-
-                if (floor & END_BIT) {
-                    break;
-                }
-            }
-        }
-    }
-
-    return count;
-}
diff --git a/src/game/game.c b/src/game/game.c
index a8eb28a5..4971ebe4 100644
--- a/src/game/game.c
+++ b/src/game/game.c
@@ -1,18 +1,19 @@
 #include "game/game.h"
 
-#include "game/inv.h"
-#include "game/settings.h"
 #include "config.h"
 #include "game/camera.h"
 #include "game/control.h"
 #include "game/draw.h"
 #include "game/gameflow.h"
 #include "game/input.h"
+#include "game/inv.h"
 #include "game/music.h"
 #include "game/savegame.h"
+#include "game/settings.h"
 #include "game/setup.h"
 #include "game/shell.h"
 #include "game/sound.h"
+#include "game/stats.h"
 #include "global/const.h"
 #include "global/vars.h"
 #include "log.h"
@@ -65,6 +66,11 @@ int32_t GameLoop(GAMEFLOW_LEVEL_TYPE level_type)
     g_OverlayFlag = 1;
     InitialiseCamera();
 
+    Stats_CalculateStats();
+    g_GameFlow.levels[g_CurrentLevel].pickups = Stats_GetPickups();
+    g_GameFlow.levels[g_CurrentLevel].kills = Stats_GetKillables();
+    g_GameFlow.levels[g_CurrentLevel].secrets = Stats_GetSecrets();
+
     bool ask_for_save = g_GameFlow.enable_save_crystals
         && level_type == GFL_NORMAL
         && g_CurrentLevel != g_GameFlow.first_level_num
diff --git a/src/game/gameflow.c b/src/game/gameflow.c
index e415d614..8d18f062 100644
--- a/src/game/gameflow.c
+++ b/src/game/gameflow.c
@@ -994,8 +994,6 @@ static void FixPyramidSecretTrigger()
         }
         global_secrets |= room_secrets;
     }
-
-    g_GameFlow.levels[g_CurrentLevel].secrets = GetSecretCount();
 }
 
 GAMEFLOW_OPTION
index 01b8345f..883ea67d 100644
--- a/src/game/level.c
+++ b/src/game/level.c
@@ -17,7 +17,6 @@
 #include "memory.h"
 
 #include <stdio.h>
-#include <string.h>
 
 static int32_t m_MeshCount = 0;
 static int32_t m_MeshPtrCount = 0;
@@ -36,8 +35,6 @@ static int32_t m_AnimTextureRangeCount = 0;
 static int32_t m_SpriteInfoCount = 0;
 static int32_t m_SpriteCount = 0;
 static int32_t m_OverlapCount = 0;
-static int32_t m_UninitItemCount = 0;
-static FLOOR_INFO **m_FloorArray = NULL;
 
 static bool Level_LoadRooms(MYFILE *fp);
 static bool Level_LoadObjects(MYFILE *fp);
@@ -152,10 +149,6 @@ static bool Level_LoadRooms(MYFILE *fp)
     File_Read(&g_RoomCount, sizeof(uint16_t), 1, fp);
     LOG_INFO("%d rooms", g_RoomCount);
 
-    // Copy for stats info
-    m_FloorArray =
-        GameBuf_Alloc(sizeof(FLOOR_INFO *) * g_RoomCount, GBUF_ROOM_FLOOR);
-
     g_RoomInfo =
         GameBuf_Alloc(sizeof(ROOM_INFO) * g_RoomCount, GBUF_ROOM_INFOS);
     int i = 0;
@@ -237,13 +230,6 @@ static bool Level_LoadRooms(MYFILE *fp)
         current_room_info->right = 0;
         current_room_info->item_number = -1;
         current_room_info->fx_number = -1;
-
-        // Save copy of floor info for stats calculation
-        m_FloorArray[i] =
-            GameBuf_Alloc(sizeof(FLOOR_INFO) * count4, GBUF_ROOM_FLOOR);
-        memcpy(
-            m_FloorArray[i], current_room_info->floor,
-            sizeof(FLOOR_INFO) * count4);
     }
 
     File_Read(&m_FloorDataSize, sizeof(uint32_t), 1, fp);
@@ -327,6 +313,8 @@ static bool Level_LoadObjects(MYFILE *fp)
         object->loaded = 1;
     }
 
+    Stats_ObserveLevelLoad();
+
     InitialiseObjects();
 
     File_Read(&m_StaticCount, sizeof(int32_t), 1, fp);
@@ -384,7 +372,6 @@ static bool Level_LoadSprites(MYFILE *fp)
 
 static bool Level_LoadItems(MYFILE *fp)
 {
-    m_UninitItemCount = 0;
     int32_t item_count = 0;
     File_Read(&item_count, sizeof(int32_t), 1, fp);
 
@@ -399,7 +386,6 @@ static bool Level_LoadItems(MYFILE *fp)
 
         g_Items = GameBuf_Alloc(sizeof(ITEM_INFO) * MAX_ITEMS, GBUF_ITEMS);
         g_LevelItemCount = item_count;
-        m_UninitItemCount = item_count;
         InitialiseItemArray(MAX_ITEMS);
 
         for (int i = 0; i < item_count; i++) {
@@ -418,6 +404,7 @@ static bool Level_LoadItems(MYFILE *fp)
                     "Level_LoadItems(): Bad Object number (%d) on Item %d",
                     item->object_number, i);
             }
+
             InitialiseItem(i);
         }
     }
@@ -687,11 +674,5 @@ bool Level_Load(int level_num)
         }
     }
 
-    Stats_CalculateStats(m_UninitItemCount, m_FloorArray);
-    g_GameFlow.levels[level_num].pickups = Stats_GetPickups();
-    g_GameFlow.levels[level_num].kills = Stats_GetKillables();
-    g_GameFlow.levels[level_num].secrets = Stats_GetSecrets();
-    Memory_FreePointer(&m_FloorArray);
-
     return ret;
 }
diff --git a/src/game/stats.c b/src/game/stats.c
index 2ee2003d..11307460 100644
--- a/src/game/stats.c
+++ b/src/game/stats.c
@@ -1,6 +1,7 @@
 #include "game/stats.h"
 
 #include "game/draw.h"
+#include "game/gamebuf.h"
 #include "game/gameflow.h"
 #include "game/input.h"
 #include "game/output.h"
@@ -19,6 +20,8 @@
 #define COWBOY_ITEMS 1
 #define BALDY_ITEMS 1
 
+static int32_t m_CachedItemCount = 0;
+static FLOOR_INFO **m_CachedFloorArray = NULL;
 static int32_t m_LevelPickups = 0;
 static int32_t m_LevelKillables = 0;
 static int32_t m_LevelSecrets = 0;
@@ -46,9 +49,9 @@ int16_t m_KillableObjs[] = {
     NO_ITEM
 };
 
-static void Stats_CheckTriggers(FLOOR_INFO **floor_array);
+static void Stats_CheckTriggers();
 
-static void Stats_CheckTriggers(FLOOR_INFO **floor_array)
+static void Stats_CheckTriggers()
 {
     uint32_t secrets = 0;
 
@@ -64,7 +67,7 @@ static void Stats_CheckTriggers(FLOOR_INFO **floor_array)
                 }
 
                 FLOOR_INFO *floor =
-                    &floor_array[i][x_floor + y_floor * r->x_size];
+                    &m_CachedFloorArray[i][x_floor + y_floor * r->x_size];
 
                 if (!floor->index) {
                     continue;
@@ -152,7 +155,19 @@ void Stats_Init()
     }
 }
 
-void Stats_CalculateStats(int32_t uninit_item_count, FLOOR_INFO **floor_array)
+void Stats_ObserveLevelLoad()
+{
+    m_CachedItemCount = g_LevelItemCount;
+    m_CachedFloorArray = GameBuf_Alloc(g_RoomCount * sizeof(FLOOR_INFO*), GBUF_ROOM_FLOOR);
+    for (int i = 0; i < g_RoomCount; i++) {
+        const ROOM_INFO *current_room_info = &g_RoomInfo[i];
+        int count = current_room_info->y_size * current_room_info->x_size;
+        m_CachedFloorArray[i] = GameBuf_Alloc(count * sizeof(FLOOR_INFO), GBUF_ROOM_FLOOR);
+        memcpy(m_CachedFloorArray[i], current_room_info->floor, count * sizeof(FLOOR_INFO) );
+    }
+}
+
+void Stats_CalculateStats()
 {
     // Clear old values
     m_LevelPickups = 0;
@@ -160,19 +175,19 @@ void Stats_CalculateStats(int32_t uninit_item_count, FLOOR_INFO **floor_array)
     m_LevelSecrets = 0;
     memset(&m_KillableItems, 0, sizeof(m_KillableItems));
 
-    if (uninit_item_count) {
-        if (uninit_item_count > MAX_ITEMS) {
+    if (m_CachedItemCount) {
+        if (m_CachedItemCount > MAX_ITEMS) {
             Shell_ExitSystem(
-                "Stats_GetPickupCount(): Too Many g_Items being Loaded!!");
+                "Stats_CalculateStats(): Too Many g_Items being Loaded!!");
             return;
         }
 
-        for (int i = 0; i < uninit_item_count; i++) {
+        for (int i = 0; i < m_CachedItemCount; i++) {
             ITEM_INFO *item = &g_Items[i];
 
             if (item->object_number < 0 || item->object_number >= O_NUMBER_OF) {
                 Shell_ExitSystemFmt(
-                    "Stats_GetPickupCount(): Bad Object number (%d) on Item %d",
+                    "Stats_CalculateStats(): Bad Object number (%d) on Item %d",
                     item->object_number, i);
             }
 
@@ -186,7 +201,7 @@ void Stats_CalculateStats(int32_t uninit_item_count, FLOOR_INFO **floor_array)
     }
 
     // Check triggers for special pickups / killables
-    Stats_CheckTriggers(floor_array);
+    Stats_CheckTriggers();
 }
 
 int32_t Stats_GetPickups()
diff --git a/src/game/stats.h b/src/game/stats.h
index a0d0ead1..b69bf0f9 100644
--- a/src/game/stats.h
+++ b/src/game/stats.h
@@ -5,7 +5,8 @@
 #include <stdint.h>
 
 void Stats_Init();
-void Stats_CalculateStats(int32_t uninit_item_count, FLOOR_INFO **floor_array);
+void Stats_ObserveLevelLoad();
+void Stats_CalculateStats();
 int32_t Stats_GetPickups();
 int32_t Stats_GetKillables();
 int32_t Stats_GetSecrets();
```

rr-:
Offtopic: we *mustn't* free gamebuf-like allocations with `Memory_*` function family. Freeing the pointer returned by `GameBuf_Alloc` will cause an invalid call to `free()`. Game buffer is a fixed-size singular, and each allocation just advanced an internal pointer within. The convenient thing about it that in contrast to `Memory_Alloc`, things allocated through game buffer get all freed in their entirety after each level ends or the game exits, so we do not have to micromanage the memory.

walkawayy:
I'll try it out and see,

walkawayy:
Wow I think that worked pretty flawlessly. I checked all pickup, kill and secret counts for all levels and they look good.

Btw, do you have a good method to check for unused includes? include what you use seems like a pain to install. With all these changes, I'm not 100% sure I removed all unused ones.

walkawayy:
Ok sorry I did rebase on latest master but forget to fetch upstream on my Github fork, so I had to do it twice.

src/game/stats.c
rr-: Yes, I'd prefer it this way, we do millions of iterations per second anyway. This is just one calculation.

src/game/stats.c
rr-: No, that won't save us much; I meant the entire body of the third inner loop.

rr-:
>Btw, do you have a good method to check for unused includes? include what you use seems like a pain to install. With all these changes, I'm not 100% sure I removed all unused ones.


Sadly I don't know any.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - fixed libavcodec-related memory leaks (#389)
 - fixed crash in custom levels that call `level_stats` after playing an FMV (#393, regression from 2.5)
 - fixed sounds playing after demo mode ends when game is minimized (#399)
+- added total pickups and kills per level to the compass and end level stats screens (#362)
```
rr-, line 10: Let's put it at the top, as it's a new feature.

src/game/game.c
```diff
@@ -8,14 +8,14 @@
 #include "game/draw.h"
 #include "game/gameflow.h"
 #include "game/input.h"
+#include "game/inv.h"
```
rr-, line 11: I don't think we need `inv`?

src/game/game.c
```diff
@@ -8,14 +8,14 @@
...
 #include "game/music.h"
-#include "game/output.h"
 #include "game/savegame.h"
-#include "game/screen.h"
+#include "game/settings.h"
```
rr-, line 14: I don't think we need `settings`?

src/game/gameflow.c
```diff
@@ -1077,7 +1076,7 @@ GameFlow_InterpretSequence(int32_t level_num, GAMEFLOW_LEVEL_TYPE level_type)
             break;
 
         case GFS_LEVEL_STATS:
-            LevelStats((int32_t)seq->data);
+            Stats_LevelEnd((int32_t)seq->data);
```
rr-, line 1079: Maybe `Stats_Show`?

src/game/level.c
```diff
@@ -383,6 +386,7 @@ static bool Level_LoadItems(MYFILE *fp)
 
         g_Items = GameBuf_Alloc(sizeof(ITEM_INFO) * MAX_ITEMS, GBUF_ITEMS);
         g_LevelItemCount = item_count;
+        // m_UninitItemCount = item_count;
```
rr-, line 389: We don't need this variable.

src/game/level.c
```diff
@@ -401,7 +405,6 @@ static bool Level_LoadItems(MYFILE *fp)
                     "Level_LoadItems(): Bad Object number (%d) on Item %d",
                     item->object_number, i);
             }
-
```
rr-, line 404: Let's keep this blank line

src/game/stats.c
```diff
@@ -0,0 +1,328 @@
...
+#define COWBOY_ITEMS 1
+#define BALDY_ITEMS 1
+
+static int32_t m_CachedItemCount = 0;
+static FLOOR_INFO **m_CachedFloorArray = NULL;
+static int32_t m_LevelPickups = 0;
+static int32_t m_LevelKillables = 0;
+static int32_t m_LevelSecrets = 0;
+static bool m_KillableItems[MAX_ITEMS] = { 0 };
+static bool m_IfKillable[O_NUMBER_OF] = { 0 };
```
rr-, line 29: Let's go with that auxiliary function instead

src/game/stats.c
```diff
@@ -0,0 +1,328 @@
...
+                           O_KEY_ITEM2,      O_KEY_ITEM3,     O_KEY_ITEM4,
+                           O_PUZZLE_ITEM1,   O_PUZZLE_ITEM2,  O_PUZZLE_ITEM3,
+                           O_PUZZLE_ITEM4,   O_GUN_ITEM,      O_SHOTGUN_ITEM,
+                           O_MAGNUM_ITEM,    O_UZI_ITEM,      O_GUN_AMMO_ITEM,
+                           O_SG_AMMO_ITEM,   O_MAG_AMMO_ITEM, O_UZI_AMMO_ITEM,
+                           O_EXPLOSIVE_ITEM, O_MEDI_ITEM,     O_BIGMEDI_ITEM,
+                           O_SCION_ITEM,     O_SCION_ITEM2,   O_LEADBAR_ITEM,
+                           NO_ITEM };
+
+// Pierre has special trigger check
```
rr-, line 41: So we don't need to check pods after all?

src/game/stats.c
```diff
@@ -0,0 +1,328 @@
...
+void Stats_Init()
+{
+    for (int i = 0; m_KillableObjs[i] != NO_ITEM; i++) {
+        m_IfKillable[m_KillableObjs[i]] = true;
+    }
+}
+
+void Stats_CalculateStats()
+{
+    // Clear old values
```
rr-, line 184: I don't think this comment is very helpful

src/game/stats.c
```diff
@@ -0,0 +1,328 @@
...
+        for (int i = 0; i < m_CachedItemCount; i++) {
+            ITEM_INFO *item = &g_Items[i];
+
+            if (item->object_number < 0 || item->object_number >= O_NUMBER_OF) {
+                Shell_ExitSystemFmt(
+                    "Stats_CalculateStats(): Bad Object number (%d) on Item %d",
+                    item->object_number, i);
+            }
+
+            // Calculate number of pickups in a level
```
rr-, line 206: I don't think this comment is very helpful

src/game/stats.c
```diff
@@ -0,0 +1,328 @@
...
+    m_LevelPickups = 0;
+    m_LevelKillables = 0;
+    m_LevelSecrets = 0;
+    memset(&m_KillableItems, 0, sizeof(m_KillableItems));
+
+    if (m_CachedItemCount) {
+        if (m_CachedItemCount > MAX_ITEMS) {
+            Shell_ExitSystem(
+                "Stats_CalculateStats(): Too Many g_Items being Loaded!!");
+            return;
```
rr-, line 194: Can we make it a non-fatal error and `LOG_ERROR` instead?

src/game/stats.c
```diff
@@ -0,0 +1,328 @@
...
+        }
+
+        for (int i = 0; i < m_CachedItemCount; i++) {
+            ITEM_INFO *item = &g_Items[i];
+
+            if (item->object_number < 0 || item->object_number >= O_NUMBER_OF) {
+                Shell_ExitSystemFmt(
+                    "Stats_CalculateStats(): Bad Object number (%d) on Item %d",
+                    item->object_number, i);
```
rr-, line 203: Can we make it a non-fatal error and `LOG_ERROR` instead?

Note that `LOG_*` already includes function name, so we only need to directly log the error itself.

rr- (changes requested):
Almost there,
although I probably would like to investigate those pods a bit more.

src/game/stats.c
walkawayy: Since we do the calculations before the objects are initialized, we check pods. This is useful because pods add mutants to the list of game items, but some pods (like the Abortion pod in Atlantis) don't have a trigger so there is no killable object associated with it.

src/game/stats.c
```diff
@@ -0,0 +1,329 @@
...
+        }
+
+        for (int i = 0; i < m_CachedItemCount; i++) {
+            ITEM_INFO *item = &g_Items[i];
+
+            if (item->object_number < 0 || item->object_number >= O_NUMBER_OF) {
+                LOG_ERROR(
+                    "Bad Object number (%d) on Item %d", item->object_number,
+                    i);
+            }
```
rr-, line 202: Here we can exit early with `continue;` to avoid a pointless run of the loop below.

rr- (approved):

src/game/stats.c
rr-: I remember some troll pods in the game that don't spawn anything, however reading the code for the pod reveals they always should spawn something. Do I remember wrong?

rr-: Also Atlantis has a big pod that we should definitely not be counting.

walkawayy: I checked all triggers and there are no pods in the base game that don't spawn anything. That only happens in UB but that's because they spawned pods with the Abortion flag when the abortion isn't in the level file hence the Lara torso bug.

The big pod won't be counted incorrectly because it doesn't have a trigger in Atlantis. I checked all the levels and the killables are correct.

rr-:
@walkawayy could you please test this patch for invalid pods? Related: #404 

```diff
diff --git a/src/game/stats.c b/src/game/stats.c
index 5bf50685..f2380c3b 100644
--- a/src/game/stats.c
+++ b/src/game/stats.c
@@ -45,8 +45,7 @@ int16_t m_KillableObjs[] = {
     O_VOLE,     O_DINOSAUR,    O_RAPTOR,   O_WARRIOR1,  O_WARRIOR2,
     O_WARRIOR3, O_CENTAUR,     O_MUMMY,    O_ABORTION,  O_DINO_WARRIOR,
     O_FISH,     O_LARSON,      O_SKATEKID, O_COWBOY,    O_BALDY,
-    O_NATLA,    O_SCION_ITEM3, O_STATUE,   O_PODS,      O_BIG_POD,
-    NO_ITEM
+    O_NATLA,    O_SCION_ITEM3, O_STATUE,   NO_ITEM
 };
 
 static void Stats_CheckTriggers();
@@ -111,20 +110,39 @@ static void Stats_CheckTriggers()
                                 }
                             } else {
                                 int16_t idx = trigger & VALUE_BITS;
+
+                                if (m_KillableItems[idx]) {
+                                    continue;
+                                }
+
                                 ITEM_INFO *item = &g_Items[idx];
 
                                 // Add Pierre pickup and kills if oneshot
                                 if (item->object_number == O_PIERRE
-                                    && trig_flags & IF_ONESHOT
-                                    && !m_KillableItems[idx]) {
+                                    && trig_flags & IF_ONESHOT) {
                                     m_KillableItems[idx] = true;
                                     m_LevelPickups += PIERRE_ITEMS;
                                     m_LevelKillables += 1;
                                 }
 
+                                // Check for only valid pods
+                                if ((item->object_number == O_PODS
+                                     || item->object_number == O_BIG_POD)
+                                    && item->data != NULL) {
+                                    int16_t bug_item_num =
+                                        *(int16_t *)item->data;
+                                    const ITEM_INFO *bug_item =
+                                        &g_Items[bug_item_num];
+                                    if (g_Objects[bug_item->object_number]
+                                            .loaded) {
+                                        m_KillableItems[idx] = true;
+                                        m_LevelKillables += 1;
+                                    }
+                                }
+
                                 // Add killable if object triggered
-                                if (Stats_IsObjectKillable(item->object_number)
-                                    && !m_KillableItems[idx]) {
+                                if (Stats_IsObjectKillable(
+                                        item->object_number)) {
                                     m_KillableItems[idx] = true;
                                     m_LevelKillables += 1;

```

walkawayy:
Yes give me a few minutes and I'll check. I was wondering if that pod change would break this count. I'll check.

walkawayy:
That patch works perfectly @rr- .

rr-:
Can I ask you to commit the pickups count table somewhere in the repository, so that we can direct the people to it when they claim we count things wrong? With the notes such as the ones you added here, (+1 unreachable medpack). May be in the README, somewhere far down.

Also I think there's still +1 enemy in the Atlantean Stronghold AFAIR compared to Stella's walkthrough. I'm not sure where it comes from, it would be nice to double check that. Apart from that, is there anything more left? :)

walkawayy:
Found the Atlantean Stronghold culprit. This centaur is triggered in this unreachable room probably for scare tactics.

image

walkawayy:
Ok I made a new section near the bottom of the README. I was going to add an item to the Q&A, but tables can't be indented in Markdown. So, I put the table and section right under the Q&A.

rr-:
Can we list all the numbers so that if we make a regression we have a single source of truth?

rr- (approved):

rr-:
Thank you.

PR #392, opened 2022-02-03
Fix memory leaks

rr-:


rr-:
libavcodec is a really intricate library. To illustrate its fragility, here's an excerpt about `avio_alloc_context`:

```
 * @param buffer Memory block for input/output operations via AVIOContext.
 *        The buffer must be allocated with av_malloc() and friends.
 *        It may be freed and replaced with a new buffer by libavformat.
 *        AVIOContext.buffer holds the buffer currently in use,
 *        which must be later freed with av_free().
```

so we have to NOT free the buffer we passed, instead we have to free something else (`result->buffer`).

At the same time, calling another function, `void avformat_close_input(AVFormatContext **ps)` which has a signature really similar to `void avformat_free_context(AVFormatContext *s)` but it doesn't suggest it frees the context itself, does just that - it calls `avformat_free_context` underneath. So freeing the context again is a (forgivable) mistake, but still, a mistake.

My point here is that when some of the calls look weird, it's mostly likely due to ffmpeg's idiosyncrasies, so before flagging them, please consult the ffmpeg documentation. I profiled the game by loading all samples 100 times in a row to amplify the leaks, reloading the save multiple times and monitoring the resident memory usage. It now stays consistent, previously it would quickly go into hundreds of wasted megabytes.

walkawayy:
I still think there's something weird happening. I reloaded a Caves save about 40 times with no issues and it doesn't look like memory usage increased which is good.

However, I level skipped through about four playthroughs and every time the memory usage jumps up after the credits end and the main menu reloads.

Memory starts around .



On playthrough 5 the memory is up to 300+ MB,
image

On title screen after playthrough 5:
image

rr-:
I'll look into it, didn't test this variant. Thanks. I suspect the music streams still need polishing.

rr-:
@walkawayy I haven't experienced such massive drops on Wine, but found another thing and it indeed was in s/picture. Could you please check now?

walkawayy:
I just skipped through 5 playthroughs, and the memory usage doesn't seem to go up every time. I think it might be fixed (no jinx please lol). Should we give a build rapora to test? It looks good to me, but they reported the issue. Up to you. On my end it looks good to approve.

rr-:
Yeah, good call, especially that a custom level is involved to which we do not have access. I'll upload a test build in the ticket.

PR #394, opened 2022-02-03
s/fmv: do not free up game memory

### LINKED ISSUE #393: Bug: game crashes if level_stats is called after play_fmv
Since there isn't any instance like that in original TR1, this only affects custom levels. In the gameflow file, if level_stats are called after play_fmv, the game crashes. This worked fine in 2.4. Might be because of the new fade effect?

rr-:
Resolves #393

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,11 @@
-## [Unreleased](https://github.com/rr-/Tomb1Main/compare/2.4...master)
+## [Unreleased](https://github.com/rr-/Tomb1Main/compare/2.5...master)
 - changed fade animations to block the main menu inventory ring like in PS1 (#379)
 - changed fade animations to be FPS-independent
 - changed fade animations to run faster in the main menu
 - fixed detail levels text flashing with any option change (#380)
 - fixed broken poses at the end of cinematics (#390)
 - fixed libavcodec-related memory leaks (#389)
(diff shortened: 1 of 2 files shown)


walkawayy:
This works for me, but I have a few suggestions. It mainly applies to custom levels and stuff, so I totally get if it's not a big priority to change.

I suggest if level stats are called after an FMV, it should use the last FMV frame as the background. For example, if I set stats to show after the Caves FMV (extreme case since there's been no level yet), it will correctly show a faded FMV frame behind the stats.

20220204_115141_Title


However, if I play the Caves FMV after the level ends, the FMV plays but the stats screen background changes back to the last game frame.

20220204_115256_Caves


My other suggestion is that levels with `exit_to_cine` don't have a `level_stats` entry in their gameflow. I couldn't fully tell why the 
`exit_to_cine` always call stats after based on the code, but maybe they shouldn't be stuck together to allow for some more customization?

If these changes are going to be a lot of a work for small features only a few might ever use, I totally get not wanting to bother. But I wanted to give some thoughts on it.

rr-:
I'd say these belong to separate feature requests, this is a bugfix of a regression, and the changes you ask for change the current behavior a great deal.

walkawayy (approved):

rr-:
(Resolved conflicts)

PR #395, opened 2022-02-03
Improve title menu fades

### LINKED ISSUE #379: Update fade behaviour
Granted this is very meticulous, and perhaps a bit too much of personal preference, but on the console versions the black fades do not animate over the ring menu animation. So as the game loads it would fade up the main menu backdrop, and then once the fade is complete animate out the ring with the music starting. Currently the animation and fade up happens at the same time.

Similarly the other way round, when a level is loaded or started, the ring menu animates in, and _then_ the fade takes place.

I understand if you'd rather have it the way you've done it as of v2.5. I just wish we could maintain console similarity for as long as possible.

Generally speaking the newly added fades are very nice.

Another thing I noticed is that the gym level has no fade out upon completion and Lara's "better take off these wet clothes" line. The game should freeze like it were a level end with the stat screen appearing, just that the stats are skipped altogether.

rr-:
Resolves #379

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## [Unreleased](https://github.com/rr-/Tomb1Main/compare/2.4...master)
-- improved fade animations in the title menu (#379)
+- changed fade animations to block the main menu inventory ring like in PS1 (#379)
 - changed fade animations to be FPS-independent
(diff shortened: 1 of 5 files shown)


rr-:
@walkawayy fixed.

walkawayy (approved):

PR #397, opened 2022-02-05
inv: fix fade ending one frame too early

rr-:
No changelog entry since it's a mistake introduced by my implementation that was not released yet.


walkawayy:
LGTM.

walkawayy (approved):

walkawayy:
@rr- whoops I noticed while testing the compass stats, there is now one black frame at the end of the inventory closing. The pause screen, title screen, etc all look fine though.

PR #400, opened 2022-02-05
game: fix title menu resuming music after demo when minimized

### LINKED ISSUE #399: Bug: title music and sounds resume after demo plays
If you minimize the game on the main menu, the demo mode will play due to inactivity. After demo mode ends, the game goes back to the main menu screen but music will play even if the game is still minimized. The game will then go into demo mode and all sounds will play while still minimized.

walkawayy:
Resolves #399.

I tested things like minimizing for multiple demos, credits, pause screens, etc and didn't notice any bugs with this solution.


rr- (approved):

rr- (changes requested):
Missing changelog

rr- (approved):

PR #403, opened 2022-02-06
inv: fix black frame after saving

rr-:
This is another regression on the unreleased branch.
8eaba246 introduced better waiting for the fade finish for the inventory, however, this revealed a problem with the waiting for the finished fade in case it carries over to the normal gameplay. Because in that fallback branch we didn't restore the camera, it would trigger black frames.


walkawayy:
Inventory looks fixed.

walkawayy (approved):

PR #404, opened 2022-02-06
ai/pod: spawn only objects that are loaded

### LINKED ISSUE #250: OG Bug: Lara torsos on Atlantean eggs in UB's "The Hive"

rr-:
#### Checklist

- [X] I have added a changelog entry about what my pull request accomplishes

#### Descripton

Fixes #250.
This makes empty pods no longer spawn Lara's hips. This is consistent with Stella's walkthrough, that clearly states that these shells are supposed to be "duds". This doesn't break savegames.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - fixed libavcodec-related memory leaks (#389)
 - fixed crash in custom levels that call `level_stats` after playing an FMV (#393, regression from 2.5)
 - fixed sounds playing after demo mode ends when game is minimized (#399)
+- fixed empty mutant shells in Unfinished Business spawning Lara's hips (#250)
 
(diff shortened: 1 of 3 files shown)


walkawayy:
Works great. Nice. My WIP bad fix screwed with save games because I was trying to replace what the pod spawned probably. Which wasn't the devs intention anyway.

walkawayy (approved):

PR #407, opened 2022-02-07
game: fix gun pickups disappearing on save load

### LINKED ISSUE #406: OG Bug: bad weapon replacements
Steps to reproduce:

1. Walk towards a gun pickup (such as the shotgun in the Lost Valley)
2. Save in slot 1. The gun is on the floor.
3. Pick it up
4. Save in slot 2
5. Load slot 2
6. Load slot 1. The gun is gone, instead it gets replaced with ammo.

The reason for this is because of the following chain of function calls:
1. `StartGame()`
2. `InitialiseLevel()`
3. `InitialiseLara()`
4. `InitialiseLaraInventory()`, this consults `START_INFO` and sees that Lara has a shotgun from the previous save
5. `GlobalItemReplace(O_SHOTGUN_ITEM, O_SG_AMMO_ITEM)` replaces Shotgun with its ammo. This is a destructive, non-reversible operation.
6. `SaveGame_ApplySaveBuffer()` this now loads the `START_INFO` and establishes that Lara has no shotgun.
7. `InitialiseLara()` to remove the shotgun from her backpack. But it's too late

A possible solution to this would be to clear `START_INFO` flags before loading a saved game. We mustn't clear these for regular playthrough otherwise Lara will lose her guns every time we load next level.

rr-:
#### Checklist

- [X] I have added a changelog entry about what my pull request accomplishes

#### Descripton

Resolves #406.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - fixed libavcodec-related memory leaks (#389)
 - fixed crash in custom levels that call `level_stats` after playing an FMV (#393, regression from 2.5)
 - fixed sounds playing after demo mode ends when game is minimized (#399)
+- fixed gun pickups disappearing in rare circumstances on save load (#406)
 - fixed empty mutant shells in Unfinished Business spawning Lara's hips (#250)
 
 ## [2.5](https://github.com/rr-/Tomb1Main/compare/2.4...2.5) - 2022-01-31
--- a/README.md
+++ b/README.md
@@ -294,6 +294,7 @@ Not all options are turned on by default. Refer to `Tomb1Main.json5` for details
 - fixed Lara switching to pistols when completing a level with other guns
 - fixed detail levels text flashing with any option change
 - fixed empty mutant shells in Unfinished Business spawning Lara's hips
+- fixed gun pickups disappearing in rare circumstances on save load (#406)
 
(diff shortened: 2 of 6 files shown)


rr-:
I'm not sure if I want these routines to stay in savegame.c or move them somewhere else hence I didn't rename them.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 - fixed libavcodec-related memory leaks (#389)
 - fixed crash in custom levels that call `level_stats` after playing an FMV (#393, regression from 2.5)
 - fixed sounds playing after demo mode ends when game is minimized (#399)
+- fixed gun pickups disappearing in rare circumstances on save load (#406)
```
walkawayy, line 12: I'll test this later tonight, but this fix should also probably go in the README too.

rr-: Good point; fixed

walkawayy:
The Lost Valley shotgun fix works. Pierre still doesn't have drops that save which I know is a separate issue. I also think that Pierre bug applies to other mercenaries. I killed the cowboy, and the magnums disappeared upon reload. Assuming that's a separate issue like Pierre, I think this is good.

walkawayy (approved):

PR #428, opened 2022-02-09
Add optional deaths counter

### LINKED ISSUE #388: Feature Request: Add death count to level statistics
Would it be possible to add player death count to the level statistics? I think it would be interesting to see this information.

I'm not sure if this should be added to #362 or be its own.

rr-:
#### Checklist

#### Descripton

Resolves #388
It works by remembering the last loaded savegame file and increasing death counter when Lara dies (so resaving without any user action, kind of counterintuitive, but it seems to work).
Dying and loading an older save (not the most recent one) should reset the death counter to the one from the past, which is actually a sensible solution.


walkawayy:
1. The "deaths" text in the compass doesn't appear until the player dies for the first time. Is that supposed to happen? I would think it would show "deaths 0">?
2. Deaths persist across levels. I also thought this would be a per level stat like the other end of level stats like secrets and kills. I think it should track deaths per level, then when a game end stats screen is added, we can total all the level deaths and show them all there.

rr-:
@walkawayy addressed

The issue was that I forgot to reset the death counter between levels. It is tied to individual levels :)

walkawayy:
I think both of those are fixed, nice! I think the last issue is the Deaths is out of order on the level end stats screen. It should be above time taken like it is in the compass.

image

rr-:
Fixed

walkawayy:
Looks good!

walkawayy (approved):

PR #434, opened 2022-02-11
wip: save effects and reload them

walkawayy:
#### Checklist

- [ ] I have read the [coding conventions](https://github.com/rr-/Tomb1Main/blob/master/CONTRIBUTING.md#coding-conventions)
- [ ] I have added a changelog entry about what my pull request accomplishes

#### Descripton

...

I'm posting a draft PR hoping someone can help with this because I can't seem to get this working. I tried different save flags for some of the effects, calling and not calling `CreateEffect` on load, saving and loading the global FX linked list variables, etc.

The waterfall splashes and lava emitters seem to save and load now. Stuff like underwater bubbles can cause crashing on load. There's also a big issue I can't fix when saving with exploding mutant body parts in mid air. The game loads with the body parts gone, but the game then crashes the next time you kill a mutant. It seems like the last mutant body part effect from the linked list calls `KillEffect` in an infinite loop upon reload, and I don't know why.

I think I'm in over my head on this one and appreciate any help if someone has time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
 ## [Unreleased](https://github.com/rr-/Tomb1Main/compare/stable...develop)
+
+## [2.10.1](https://github.com/rr-/Tomb1Main/compare/2.10...2.10.1) - 2022-07-27
 - fixed Lara being able to equip pistols in the gym level (#594)
 
 ## [2.10](https://github.com/rr-/Tomb1Main/compare/2.9.1...2.10) - 2022-07-26


rr-:
There's too much going on at the same time in this PR. Let's open a single PR that only tackles lava emitters and think about it there.

Messing with `save_flags` and other similar item settings is an absolute no-go unless we don't care about legacy saves.

walkawayy:
@rr- I have no idea why this merged... I was doing some cleaning and pushed up a simplified version of this branch. I have no idea why it says merged. It says no changes which is true because I stashed the changes to start cleanly from the renamed codebase.

EDIT: Oops I think it just says it automerged since it matches develop. I'll push a new draft PR when it's ready.

Issue #371, opened 2022-01-22
Prevent music triggers from playing again after loading a save file


aredfan:
I've noticed that music triggers can play more than once, it goes like this:

1. Lara steps on a trigger, music plays.
2. The player creates a save file afterwards.
3. The player loads the save file.
4. If Lara treads on the same trigger again, the music plays again.

This is noticable in areas where Lara have to do backtracking.

rr-:
This error cannot be fully fixed. It also manifests in a slightly different form:

1. Lara steps on a tile triggering music track A.
2. Lara goes elsewhere and steps on a tile triggering music track B (this includes collecting a secret, for instance).
3. Lara goes back to the first tile. The game shouldn't play the music track again, but it does.

The reason for this is because the game remembers which track was played recently and even if the author of the level forgot to mark given trigger as one-shot, this logic ensures that we do not play the same music track twice in a row.

To fix this, we have two solutions: the first is to find every broken trigger in each level and mark it as a one-shot. Not very sustainable or elegant in my opinion. A second, more sensible approach would be to add to the save file information about what track was last played, letting the in-game protection against playing the same track twice in a row continue doing its job even after save and reload.

walkawayy:
@rr- maybe marking every trigger as one shot is now possible? We already traverse them all to do calculate stats. Let me know what you think.

rr-:
I haven't thought about this. Won't this lead to regressions?

walkawayy:
How so? Aren't all music triggers supposed to be played only once?

rr-:
Custom levels with Lara saying "I should revisit this area later" before a flip?

Richard-L:
For custom levels, could people just create two triggers in the same place, to be used up in concession of game progression?

PR #436, opened 2022-02-12
savegame: save and load current music track and timestamp

walkawayy:
#### Checklist

#### Description
Added the current music track and timestamp to the savegame so they now persist on load.

### DIFF
--- a/src/game/music.c
+++ b/src/game/music.c
@@ -98,8 +98,6 @@ bool Music_Play(int16_t track)
     m_AudioStreamID = S_Audio_StreamSoundCreateFromFile(file_path);
     Memory_FreePointer(&file_path);
 
-    LOG_DEBUG("Playing music on sound_id %d.", m_AudioStreamID);
-
     if (m_AudioStreamID < 0) {
         LOG_ERROR("All music streams are busy");
         return false;
--- a/src/game/music.c
+++ b/src/game/music.c
@@ -98,6 +98,8 @@ bool Music_Play(int16_t track)
     m_AudioStreamID = S_Audio_StreamSoundCreateFromFile(file_path);
     Memory_FreePointer(&file_path);
 
+    LOG_DEBUG("Playing music on sound_id %d.", m_AudioStreamID);
+
     if (m_AudioStreamID < 0) {
         LOG_ERROR("All music streams are busy");
         return false;
@@ -191,4 +193,20 @@ void Music_Unpause(void)
 int16_t Music_CurrentTrack(void)
 {
     return m_Track;
+}
+
+int64_t Music_GetTimestamp(int16_t track)
+{
+    if (m_AudioStreamID < 0) {
+        return -1;
+    }
+    return S_Audio_StreamGetTimestamp(m_AudioStreamID);
+}
+
+bool Music_SeekTimestamp(int16_t track, int64_t timestamp)
+{
+    if (m_AudioStreamID < 0) {
+        return false;
+    }
+    return S_Audio_StreamSeekTimestamp(m_AudioStreamID, timestamp);
 }
--- a/src/game/savegame/savegame_bson.c
+++ b/src/game/savegame/savegame_bson.c
@@ -812,16 +812,13 @@ static bool SaveGame_BSON_LoadCurrentMusic(struct json_object_s *music_obj)
     }
 
     int16_t current_track = json_object_get_int(music_obj, "current_track", -1);
-    int32_t timestamp_arr[2];
-    timestamp_arr[0] = json_object_get_int(music_obj, "timestamp1", 0);
-    timestamp_arr[1] = json_object_get_int(music_obj, "timestamp2", 0);
-    int64_t *timestamp = (int64_t *)timestamp_arr;
+    int64_t timestamp = json_object_get_int64(music_obj, "timestamp", -1);
     if (current_track) {
         Music_Play(current_track);
-        if (!Music_SeekTimestamp(current_track, *timestamp)) {
+        if (!Music_SeekTimestamp(current_track, timestamp)) {
             LOG_WARNING(
                 "Could not load current track %d at timestamp %d.",
-                current_track, *timestamp);
+                current_track, timestamp);
         }
     }
 
@@ -1146,9 +1143,7 @@ static struct json_object_s *SaveGame_BSON_DumpCurrentMusic(void)
     if (Music_CurrentTrack()) {
         timestamp = Music_GetTimestamp(Music_CurrentTrack());
     }
-    int32_t *timestamp_arr = (int32_t *)&timestamp;
-    json_object_append_int(current_music_obj, "timestamp1", timestamp_arr[0]);
-    json_object_append_int(current_music_obj, "timestamp2", timestamp_arr[1]);
+    json_object_append_int64(current_music_obj, "timestamp", timestamp);
 
     return current_music_obj;
 }
--- a/src/game/savegame/savegame_bson.c
+++ b/src/game/savegame/savegame_bson.c
@@ -1220,14 +1220,18 @@ bool Savegame_BSON_LoadFromFile(MYFILE *fp, GAME_INFO *game_info)
         goto cleanup;
     }
 
-    int16_t load_track = json_object_get_int(root_obj, "music_track", -1);
+    int16_t music_track = json_object_get_int(root_obj, "music_track", -1);
     int32_t timestamp_arr[2];
     timestamp_arr[0] = json_object_get_int(root_obj, "music_timestamp1", -1);
     timestamp_arr[1] = json_object_get_int(root_obj, "music_timestamp2", -1);
     int64_t *music_timestamp = (int64_t *)timestamp_arr;
-    if (load_track) {
-        Music_Play(load_track);
-        Music_SeekTimestamp(load_track, *music_timestamp);
+    if (music_track) {
+        Music_Play(music_track);
+        if (!Music_SeekTimestamp(music_track, *music_timestamp)) {
+            LOG_WARNING(
+                "Could not load music track %d at timestamp %d.", music_track,
+                *music_timestamp);
+        }
     }
 
     ret = true;
--- a/src/game/savegame/savegame_bson.c
+++ b/src/game/savegame/savegame_bson.c
@@ -1222,7 +1222,7 @@ bool Savegame_BSON_LoadFromFile(MYFILE *fp, GAME_INFO *game_info)
 
     int16_t load_track = json_object_get_int(root_obj, "music_track", -1);
 
-    int32_t timestamp_arr[2];
+    int64_t timestamp_arr[2];
     timestamp_arr[0] = json_object_get_int(root_obj, "music_timestamp1", -1);
     timestamp_arr[1] = json_object_get_int(root_obj, "music_timestamp2", -1);
     LOG_DEBUG("music_timestamp1 %d", timestamp_arr[0]);
@@ -1308,7 +1308,7 @@ void Savegame_BSON_SaveToFile(MYFILE *fp, GAME_INFO *game_info)
     if (save_track) {
         save_timestamp = Music_GetTimestamp(Music_CurrentTrack());
     }
-    int *timestamp_arr = (int *)&save_timestamp;
+    int64_t *timestamp_arr = (int64_t *)&save_timestamp;
     json_object_append_int(root_obj, "music_timestamp1", timestamp_arr[0]);
     json_object_append_int(root_obj, "music_timestamp2", timestamp_arr[1]);
     LOG_DEBUG("music_timestamp1 %d", timestamp_arr[0]);
--- a/src/game/savegame/savegame_bson.c
+++ b/src/game/savegame/savegame_bson.c
@@ -1221,19 +1221,13 @@ bool Savegame_BSON_LoadFromFile(MYFILE *fp, GAME_INFO *game_info)
     }
 
     int16_t load_track = json_object_get_int(root_obj, "music_track", -1);
-
-    int64_t timestamp_arr[2];
+    int32_t timestamp_arr[2];
     timestamp_arr[0] = json_object_get_int(root_obj, "music_timestamp1", -1);
     timestamp_arr[1] = json_object_get_int(root_obj, "music_timestamp2", -1);
-    LOG_DEBUG("music_timestamp1 %d", timestamp_arr[0]);
-    LOG_DEBUG("music_timestamp2 %d", timestamp_arr[1]);
-    int64_t load_timestamp = *(int64_t *)timestamp_arr;
-    LOG_DEBUG("load_timestamp %d", load_timestamp);
-    LOG_DEBUG("load_track %d", load_track);
-
+    int64_t *music_timestamp = (int64_t *)timestamp_arr;
     if (load_track) {
         Music_Play(load_track);
-        Music_SeekTimestamp(load_track, load_timestamp);
+        Music_SeekTimestamp(load_track, *music_timestamp);
     }
 
     ret = true;
@@ -1303,19 +1297,15 @@ void Savegame_BSON_SaveToFile(MYFILE *fp, GAME_INFO *game_info)
     json_object_append_array(root_obj, "fx", SaveGame_BSON_DumpFx());
     json_object_append_object(
         root_obj, "lara", Savegame_BSON_DumpLara(&g_Lara));
-    int16_t save_track = Music_CurrentTrack();
-    json_object_append_int(root_obj, "music_track", save_track);
-    int64_t save_timestamp = 0;
-    if (save_track) {
-        save_timestamp = Music_GetTimestamp(Music_CurrentTrack());
+    int16_t music_track = Music_CurrentTrack();
+    json_object_append_int(root_obj, "music_track", music_track);
+    int64_t music_timestamp = 0;
+    if (music_track) {
+        music_timestamp = Music_GetTimestamp(Music_CurrentTrack());
     }
-    int64_t *timestamp_arr = (int64_t *)&save_timestamp;
+    int32_t *timestamp_arr = (int32_t *)&music_timestamp;
     json_object_append_int(root_obj, "music_timestamp1", timestamp_arr[0]);
     json_object_append_int(root_obj, "music_timestamp2", timestamp_arr[1]);
-    LOG_DEBUG("save_timestamp %d", save_timestamp);
-    LOG_DEBUG("music_timestamp1 %d", timestamp_arr[0]);
-    LOG_DEBUG("music_timestamp2 %d", timestamp_arr[1]);
-    LOG_DEBUG("save_track %d", save_track);
 
     struct json_value_s *root = json_value_from_object(root_obj);
     SaveGame_BSON_SaveRaw(fp, root, SAVEGAME_CURRENT_VERSION);
--- a/src/game/savegame/savegame_bson.c
+++ b/src/game/savegame/savegame_bson.c
@@ -1266,12 +1266,9 @@ bool Savegame_BSON_LoadFromFile(MYFILE *fp, GAME_INFO *game_info)
         goto cleanup;
     }
 
-    if (header.version >= VERSION_3) {
-        struct json_object_s *music_obj =
-            json_object_get_object(root_obj, "music");
-        if (!SaveGame_BSON_LoadCurrentMusic(music_obj)) {
-            goto cleanup;
-        }
+    if (!SaveGame_BSON_LoadCurrentMusic(
+            json_object_get_object(root_obj, "music"))) {
+        goto cleanup;
     }
 
     ret = true;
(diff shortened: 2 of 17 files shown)


rr-:
I don't think I'll be able to help much in the following days; let's focus on testing the other prs and revisit these two in the future.

walkawayy:
Did some code cleanup.
Bumped the savegame version to 3 (this will also include #371). The game only warns if the loading process fails.
Made this optional and enabled by default.

src/game/music.c
```diff
@@ -192,3 +194,19 @@ int16_t Music_CurrentTrack(void)
...
+
+int64_t Music_GetTimestamp(int16_t track)
+{
+    if (m_AudioStreamID < 0) {
+        return -1;
+    }
+    return S_Audio_StreamGetTimestamp(m_AudioStreamID);
+}
+
+bool Music_SeekTimestamp(int16_t track, int64_t timestamp)
```
lahm86, line 204: Maybe `int16_t track` should be removed if it's unused.

src/game/savegame/savegame_bson.c
```diff
@@ -797,6 +800,34 @@ static bool Savegame_BSON_LoadLara(
...
+    }
+
+    int16_t current_track = json_object_get_int(music_obj, "current_track", -1);
+    int32_t timestamp_arr[2];
+    timestamp_arr[0] = json_object_get_int(music_obj, "timestamp1", 0);
+    timestamp_arr[1] = json_object_get_int(music_obj, "timestamp2", 0);
+    int64_t *timestamp = (int64_t *)timestamp_arr;
+    if (current_track) {
+        Music_Play(current_track);
+        if (!Music_SeekTimestamp(current_track, *timestamp)) {
```
lahm86: I've noticed that on load, you hear a very brief moment of the beginning of the track before it jumps to the restored timestamp. It's negligible imo, but I wonder if there is a way to seek before the music is started?

walkawayy: Yes I noticed that too, but it's hard to tell since it's like a one frame thing it seems like. I'll look into it.

src/game/music.c
rr-: I agree, but it seems now to be used

src/game/savegame/savegame_bson.c
```diff
@@ -1219,6 +1266,13 @@ bool Savegame_BSON_LoadFromFile(MYFILE *fp, GAME_INFO *game_info)
         goto cleanup;
     }
 
+    if (header.version >= VERSION_3) {
+        if (!SaveGame_BSON_LoadCurrentMusic(
+                json_object_get_object(root_obj, "music"))) {
+            goto cleanup;
+        }
+    }
+
```
rr-, line 1269: I don't think it's necessary to bump the version, as your scheme is forwards compatible?  I don't remember our SG versioning policy. Anyway we could extract the JSON retrieval to a variable, `music_obj`.

src/specific/s_audio_stream.c
```diff
@@ -390,6 +394,7 @@ bool S_Audio_StreamSoundClose(int sound_id)
     stream->is_playing = false;
     stream->is_looped = false;
     stream->volume = 0.0f;
+    stream->timestamp = 0;
```
rr-: There's lots of similar code here,  I can see opportunity to extract it to a common procedure

src/specific/s_audio_stream.c
```diff
@@ -516,3 +521,41 @@ void S_Audio_StreamSoundMix(float *dst_buffer, size_t len)
...
+        || sound_id >= AUDIO_MAX_ACTIVE_STREAMS) {
+        return false;
+    }
+
+    if (m_StreamSounds[sound_id].is_playing) {
+        SDL_LockAudioDevice(g_AudioDeviceID);
+        AUDIO_STREAM_SOUND *stream = &m_StreamSounds[sound_id];
+        av_seek_frame(
+            stream->av.format_ctx, sound_id, timestamp, AVSEEK_FLAG_ANY);
+        avcodec_flush_buffers(stream->av.codec_ctx); // ?
```
rr-: ? :D

tools/config/Tomb1Main_ConfigTool/Resources/Lang/es.json
```diff
@@ -195,6 +195,10 @@
       "Description": "Permite que los sonidos del juego continúen sonando en la pantalla de inventario.",
       "Title": "Habilitar sonidos de juegos en el inventario"
     },
+    "load_current_music": {
+      "Title": "Load music track",
+      "Description": "TBD."
```
rr-: Let's remember to change this

tools/config/Tomb1Main_ConfigTool/Resources/Lang/fr.json
```diff
@@ -275,6 +275,10 @@
       "Title": "Format des screenshots",
       "Description": "Selectionner le format voulu pour les captures d'écran en jeu."
     },
+    "load_current_music": {
+      "Title": "Load music track",
+      "Description": "TBD."
```
rr-: let's remember to change this

src/specific/s_audio_stream.c
walkawayy: Lmao I tried to fix the audio clipping for a frame or two on load, but this didn't help. From what I read though, the buffer should be flushed after seeking. There doesn't seem to be the most online resources for this though, so it's hard to find help and ideas.

http://dranger.com/ffmpeg/tutorial07.html

src/game/savegame/savegame_bson.c
walkawayy: Hm maybe. I don't need it for this though? I guess the game just warns it's not found and continues loading. I'll leave the decision up to you.

```
    if (header.version >= VERSION_3) {
        if (!SaveGame_BSON_LoadCurrentMusic(
                json_object_get_object(root_obj, "music"))) {
            goto cleanup;
        }
    }
```

Also not sure what you mean by ` Anyway we could extract the JSON retrieval to a variable, music_obj`. I already have a `json_object_s music_obj`.

src/game/music.c
walkawayy: Yes originally I didn't use it. The ffmpeg docs claimed using `-1` for `stream_index` uses the default stream, but it didn't actually ever seek. I only fixed the seeking by using the `track`.

https://ffmpeg.org/doxygen/trunk/group__lavf__decoding.html#gaa23f7619d8d4ea0857065d9979c75ac8

src/game/savegame/savegame_bson.c
rr-: Something like this:

```diff
     if (header.version >= VERSION_3) {
-        if (!SaveGame_BSON_LoadCurrentMusic(
-                json_object_get_object(root_obj, "music"))) {
+        json_object_s music_obj = json_object_get_object(root_obj, "music");
+        if (!SaveGame_BSON_LoadCurrentMusic(music_obj)) {
             goto cleanup;
         }
     }
```

As to whether to bump the sg version, I'm indifferent, maybe @lahm86 has stronger opinion on this?

src/specific/s_audio_stream.c
rr-: I don't see anything about flushing the buffers in ffmpeg documentation. (They have excellent documentation in their source code.) So if this doesn't help us any way let's remove this call.

src/game/savegame/savegame_bson.c
rr-: I think that this might be caused by the flow:

- We call `Game_Start`
- Inside `Game_Start`, we call `Level_Initialise`
- `Level_Initialise` calls `Music_Play` which starts the audio track, probably in an asynchronous way
- Next up, inside `Game_Start` we call `Savegame_Load`
- This calls `Music_SeekTimestamp` as well as `Music_Play`

Could you verify if removing the call to `Music_Play` from inside `Level_Initialise` fixes the clipping issue?
If so, I suggest testing whether a fix like this would work:

```diff
diff --git a/src/game/game/game.c b/src/game/game/game.c
index 8f95ea29..c016ea69 100644
--- a/src/game/game/game.c
+++ b/src/game/game/game.c
@@ -133,6 +133,12 @@ bool Game_Start(int32_t level_num, GAMEFLOW_LEVEL_TYPE level_type)
         // reset current info to the defaults so that we do not do
         // Item_GlobalReplace in the inventory initialization routines too early
         Savegame_InitCurrentInfo();
+
+        // prevent audio clipping that occurs when level plays a music track,
+        // and later savegame routines override it while the stream is already
+        // playing in the background thread
+        Music_Pause();
+
         if (!Level_Initialise(level_num)) {
             return false;
         }
@@ -140,6 +146,8 @@ bool Game_Start(int32_t level_num, GAMEFLOW_LEVEL_TYPE level_type)
             LOG_ERROR("Failed to load save file!");
             return false;
         }
+        Music_Unpause();
+
         break;
 
     case GFL_RESTART:
```

src/game/savegame/savegame_bson.c
lahm86: No strong feelings either, but I guess if it's not essential it can be omitted. A warning and it continuing to load regardless is fine imo.

src/game/savegame/savegame_bson.c
```diff
@@ -797,6 +800,34 @@ static bool Savegame_BSON_LoadLara(
     return true;
 }
 
+static bool SaveGame_BSON_LoadCurrentMusic(struct json_object_s *music_obj)
```
lahm86, line 804: I've noticed overall that this function will always return `true`.

src/game/savegame/savegame_bson.c
walkawayy: > Something like this:

Ok I can do it this way. No other loading call does it this way FYI. What does it accomplish?

src/game/savegame/savegame_bson.c
walkawayy: Ah I was confused at first but you mean `Level_Initialise` calls `Music_PlayLooped`.

I made this change and I think it does actually fix the issue! Will definitely need some others to test as well to confirm.

tools/config/Tomb1Main_ConfigTool/Resources/Lang/es.json
walkawayy: @ajtudela would you mind translating again?

```
"Loads the music track that was playing when the game was saved."
```

src/game/savegame/savegame_bson.c
walkawayy: Hm yes but is that an issue? I didn't want to cause a savegame loading failure because of the music track only. Just a warning.

lahm86: Was thinking it might be cleaner just to make it `void`, but appreciate that goes against the common approach with the other functions.

src/specific/s_audio_stream.c
walkawayy: I was going to make this function for `S_Audio_StreamSoundInit` and for `S_Audio_StreamSoundClose`.
```
static void S_Audio_StreamSoundClear(AUDIO_STREAM_SOUND *stream)
{
    stream->is_used = false;
    stream->is_playing = false;
    stream->is_looped = false;
    stream->is_read_done = true;
    stream->volume = 0.0f;
    stream->timestamp = 0;
    stream->sdl.stream = NULL;
    stream->finish_callback = NULL;
}
```

But I'm unsure what this part is doing in `S_Audio_StreamSoundClose`:
```
    if (stream->finish_callback) {
        stream->finish_callback(sound_id, stream->finish_callback_user_data);
    }
```
Because `S_Audio_StreamSoundInit` sets this to: `stream->finish_callback = NULL;`

src/game/savegame/savegame_bson.c
lahm86: It's definitely better, I think I do still hear it extremely briefly though - or am I imagining it?

https://youtu.be/kEjJxek18Cs

src/game/savegame/savegame_bson.c
rr-: My thinking was to reduce the indentation a bit, but I'd say consistency here is more important than looks, so it's OK to leave it as-is if it follows an established way.

src/game/savegame/savegame_bson.c
rr-: I think I hear something but I'd say it's negligible and feels like a part of the melody, too.

src/specific/s_audio_stream.c
rr-: The suggestion above is OK, however, the clear routine has to happen after the stream finish callback gets executed. It's used to play looped background tracks.

src/game/savegame/savegame_bson.c
walkawayy: Yes it's hard to tell. I think it sounds pretty good now also.

rr- (approved):

src/game/savegame/savegame_bson.c
Richard-L: It sounds as though a snippet of the passport in/out SFX is repeated very briefly upon load. Is that possible?
It's super nice this issue is being worked on (!), but I do find the added noise unfortunate.

tools/config/Tomb1Main_ConfigTool/Resources/Lang/es.json
ajtudela:
````
"load_current_music": {
      "Title": "Cargar pista de sonido",
      "Description": "Carga la pista de música que se estaba reproduciendo cuando se guardó el juego".
}
````

Next time, if you don't mind, left the file "es.json" without changes. That way I can compare the file with the "en.json" using i18n Ally. I think it would be easier for boths.

Comment (walkawayy) [reply tools/config/Tomb1Main_ConfigTool/Resources/Lang/es.json]: Ok. Thanks!

File: src/specific/s_audio_stream.c (hunk 4 above)
Comment (walkawayy) [reply src/specific/s_audio_stream.c]: Ok lmk if what I pushed works for you.

File: src/game/savegame/savegame_bson.c (hunk 2 above)
Comment (walkawayy) [reply src/game/savegame/savegame_bson.c]: It's hard to tell if it's just mid instrumental or not. It sounds OK on these to me? I'm not totally sure lol.

https://streamable.com/lj45x7

Comment (Richard-L) [reply src/game/savegame/savegame_bson.c]: Hmmm. Is it possible it's different now compared to the Folly clip yesterday? It sounds changed (improved!).

I've compiled all of McCree's TR1-3 music in my cloud some years ago and have since accumulated (too) many hours listening to it. Want to say I know all of its nuances decently well. Something sounds off in your above video, but it is pretty hard to tell. Can you post it on the Discord to get more opinions?

Is there a loudness spike for a split second until it adjusts, right after load?

Other times it reminds me of a cassette player resuming playback, and needing a very brief moment to reach full pitch.

In any case, I generally love how this is behaving now. Such a nice improvement 👏

File: src/specific/s_audio_stream.c [hunk 9]
```diff
@@ -386,21 +393,16 @@ bool S_Audio_StreamSoundClose(int sound_id)
...
-    stream->is_looped = false;
-    stream->volume = 0.0f;
-    stream->timestamp = 0;
-    SDL_UnlockAudioDevice(g_AudioDeviceID);
-
     if (stream->finish_callback) {
         stream->finish_callback(sound_id, stream->finish_callback_user_data);
     }
 
+    S_Audio_StreamSoundClear(stream);
```
rr-, line 402: Haven't tested it, but IMO this will now freeze the game when the BGM is about to be replayed, and crash it after a short while with a stack call overflow error:

1. S_Audio_StreamSoundClose
2. stream->finish_callback
3. Music_StreamFinished
4. Music_PlayLooped
5. Music_StopActiveStream
6. S_Audio_StreamClose
7. stream->finish_callback (which still isn't NULLed)
8. repeat ad infinitum

Even if it works due to some internal checks inside music.c, I think it's by accident, and there's still the issue of nesting SDL_LockAudioDevice calls. We can do better.

I suggest attaching the callback to a variable like so, and then calling it afterwards:

```c
void (*finish_callback)()&(*@++__ilovec) = stream->finish_callback;
void *finish_callback_user_data = stream->finish_callback_user_data;
S_Audio_StreamSoundClear(stream);
SDL_UnlockAudioDevice(g_AudioDeviceID);
if (finish_callback) {
    finish_callback(sound_id, finish_callback_user_data);
}
```

and sprinkle 2 or 3 blank lines for extra prettiness. (This is pseudocode in case anyone wonders.)

walkawayy: This seems more convoluted to me than the original code. 😰

rr-: @lahm86 WDYT

src/game/music.c
```diff
@@ -98,6 +98,8 @@ bool Music_Play(int16_t track)
     m_AudioStreamID = S_Audio_StreamSoundCreateFromFile(file_path);
     Memory_FreePointer(&file_path);
 
+    LOG_DEBUG("Playing music on sound_id %d.", m_AudioStreamID);
```
walkawayy: TODO: delete.

src/specific/s_audio_stream.c
lahm86: The pseudocode above looks good and I think it makes sense to avoid the issue you mention (although I have tested it out and you're right, it doesn't appear to be a problem - but at least this should future-proof it).

walkawayy: Thanks for the help Lahm. Seems to work! Last question is, `finish_callback_user_data` is never set to `NULL` like `finish_callback`. Should it be? (For the record, this was not the case in the code before this PR either. I just wanted to ask while we are touching this code.)

src/game/savegame/savegame_bson.c
```diff
@@ -1106,6 +1137,22 @@ static struct json_object_s *Savegame_BSON_DumpLara(LARA_INFO *lara)
...
+    struct json_object_s *current_music_obj = json_object_new();
+    json_object_append_int(
+        current_music_obj, "current_track", Music_CurrentTrack());
+    int64_t timestamp = 0;
+    if (Music_CurrentTrack()) {
+        timestamp = Music_GetTimestamp(Music_CurrentTrack());
+    }
+    int32_t *timestamp_arr = (int32_t *)&timestamp;
+    json_object_append_int(current_music_obj, "timestamp1", timestamp_arr[0]);
+    json_object_append_int(current_music_obj, "timestamp2", timestamp_arr[1]);
```
rr-: I only noticed this now. Hah, clever. Could we make a `json_object_append_int64` though for this? JSON is plenty capable of holding arbitrary-precision numbers and the limitation of our code shouldn't be reflected in the files IMO. @lahm86 what's your opinion on this?

src/specific/s_audio_stream.c
rr-: Yeah let's clear that, too. (Inside `S_Audio_StreamSoundClear` of course.)

src/game/savegame/savegame_bson.c
walkawayy: Ok I can do that. I didn't want to rely on a generic `int` since the timestamp is 64 bit.

lahm86: I like the array approach too, but agree, having JSON do the work sounds great.

walkawayy:
Alright I added 64 bit saving / loading to the new savegame format. My testing still seems good, so let me know if this OK. I think this should be the last review before we can finally merge!

src/game/savegame/savegame_bson.c
```diff
@@ -812,16 +812,13 @@ static bool SaveGame_BSON_LoadCurrentMusic(struct json_object_s *music_obj)
...
-    int64_t *timestamp = (int64_t *)timestamp_arr;
+    int64_t timestamp = json_object_get_int64(music_obj, "timestamp", -1);
     if (current_track) {
         Music_Play(current_track);
-        if (!Music_SeekTimestamp(current_track, *timestamp)) {
+        if (!Music_SeekTimestamp(current_track, timestamp)) {
             LOG_WARNING(
                 "Could not load current track %d at timestamp %d.",
-                current_track, *timestamp);
+                current_track, timestamp);
```
rr-, line 822: Will %d work with int64_t?

rr- (approved):

src/game/savegame/savegame_bson.c
walkawayy: Here's what I kept reading on stack overflow

https://stackoverflow.com/questions/9225567/how-to-portably-print-a-int64-t-type-in-c

rr-: Yeah that's what I figured, it was used in the JSON dump/load sequences, but here in the log call we use plain `%d`. I believe this will print malformed results:

```c
#include <inttypes.h>
#include <stdint.h>
#include <stdio.h>

int main() {
    int64_t tmp = 10e10;
    printf("%d\n", tmp);
    printf("%" PRIi64 "\n", tmp);
    return 0;
}
```

```console
rr-@lightning:~❯❯❯ gcc test.c -o test && chmod +x ./test && ./test
1215752192
100000000000
```

walkawayy: Oh oops I missed that last night and my early morning brain did not read your comment correctly lol. Fixed.

PR #454, opened 2022-03-02
Various cleanups

### LINKED ISSUE #448: Issues with memory.c
```
void *Memory_Dup(const char *string)
{
    void *memory = Memory_Alloc(strlen(string) + 1);
    strcpy(memory, string);
    return memory;
}
```
A function called *Memory_Dup* that does a string copy? That is a misleading name that could easily cause somebody to call it for binary data, which would end badly. 
It also returns it as `void*`, which hints more that it's for binary data. 
I suggest making it return `char*` and adding String to the name, so its `Memory_DupString` or `Memory_DupStr` just so it's more clear on its function. 

```
void Memory_FreePointer(void *arg)
{
    void *memory;
    memcpy(&memory, arg, sizeof(void *));
    memcpy(arg, &(void *) { NULL }, sizeof(void *));
    Memory_Free(memory);
}
```
Ummm??? I think I see what you are thinking... no. 
So ` memcpy(&memory, arg, sizeof(void *));` this could just be an assign, I get that you want to "make sure you have a proper copy of it" but an assign will do that just fine. Also you don't have to clear the pointer before you do the freeing of memory, this is single threaded from the game allocations perspective so there is no thread race conditions to consider. 
Further more, `memcpy(arg, &(void *) { NULL }, sizeof(void *));` is actually utterly pointless anyway. I guess what you think it is doing is looking up the address of the param passed in from where it was called and setting that to `nullptr` automatically. It doesn't. You are setting the arg that is passed in to now be 0. which is exactly the same as just doing `arg = nullptr`, because `arg` is either a register ( in a release build ) so the memcopy of a `NULL` will just be `mov eax, 0` or some other register or in say a debug build it will be 4 bytes sitting on the stack somewhere. The original pointer is not visible to this function and you can't touch it with it, that would break the promise C makes to you. You can make it `void** arg` and then you will `Memory_Free(&arg); **arg=nullptr`, however you will then need to remember to pass &thing to this function every time. Shame ole C doesn't have pass by reference huh ;) 
Typically we made `#DEFINE` that would call free and then assign the pointer to null for us. 
It is also nice that `Memory_Free` checks for `NULL` first, however this function doesn't so when it goes to `memcpy` from 0 its going to trip the OS memory protection and give you a crash/error etc. And if it didn't by some miracle, the memcpy to it certainly will. 

as an aside and while on the topic of memcpy. 
In s_output.c :167 ( or there abouts my code may differ)
`memcpy(&vertices[1], &vertices[2], left * sizeof(GFX_3D_Vertex));`
memcpy with overlapping regions is "undefined" and crashes when built with clang on windows. For such cases where the is overlap one should use `memmove` which will detect and handle overlap correctly and make sure it copies in the right direction at all times.

rr-:
#### Checklist

#### Descripton

Resolves #448

### DIFF
--- a/scripts/analyze_memleaks.patch
+++ b/scripts/analyze_memleaks.patch
@@ -54,11 +54,11 @@ index c73c24cf..fbc704f7 100644
 +    Memory_FreeImpl(memory, file, line, func);
  }
  
--void *Memory_Dup(const char *string)
-+void *Memory_DupImpl(const char *string, const char *file, int line, const char *func)
+-char *Memory_DupStr(const char *string)
++char *Memory_DupStrImpl(const char *string, const char *file, int line, const char *func)
  {
--    void *memory = Memory_Alloc(strlen(string) + 1);
-+    void *memory = Memory_AllocImpl(strlen(string) + 1, file, line, func);
+-    char *memory = Memory_Alloc(strlen(string) + 1);
++    char *memory = Memory_AllocImpl(strlen(string) + 1, file, line, func);
      strcpy(memory, string);
      return memory;
  }
@@ -74,15 +74,15 @@ index 23c23f95..3995bada 100644
 -void *Memory_Realloc(void *memory, size_t size);
 -void Memory_Free(void *memory);
 -void Memory_FreePointer(void *memory);
--void *Memory_Dup(const char *string);
+-char *Memory_DupStr(const char *string);
 +void *Memory_AllocImpl(size_t size, const char *file, int line, const char *func);
 +void *Memory_ReallocImpl(void *memory, size_t size, const char *file, int line, const char *func);
 +void Memory_FreeImpl(void *memory, const char *file, int line, const char *func);
 +void Memory_FreePointerImpl(void *memory, const char *file, int line, const char *func);
-+void *Memory_DupImpl(const char *string, const char *file, int line, const char *func);
++char *Memory_DupStrImpl(const char *string, const char *file, int line, const char *func);
 +
 +#define Memory_Alloc(size)              (Memory_AllocImpl(size, __FILE__, __LINE__, __func__))
 +#define Memory_Realloc(memory, size)    (Memory_ReallocImpl(memory, size, __FILE__, __LINE__, __func__))
 +#define Memory_Free(memory)             (Memory_FreeImpl(memory, __FILE__, __LINE__, __func__))
 +#define Memory_FreePointer(memory)      (Memory_FreePointerImpl(memory, __FILE__, __LINE__, __func__))
-+#define Memory_Dup(string)              (Memory_DupImpl(string, __FILE__, __LINE__, __func__))
++#define Memory_DupStr(string)           (Memory_DupStrImpl(string, __FILE__, __LINE__, __func__))
--- a/scripts/analyze_memleaks.patch
+++ b/scripts/analyze_memleaks.patch
@@ -47,6 +47,7 @@ index c73c24cf..fbc704f7 100644
 -void Memory_FreePointer(void *arg)
 +void Memory_FreePointerImpl(void *arg, const char *file, int line, const char *func)
  {
+     assert(arg);
      void *memory;
      memcpy(&memory, arg, sizeof(void *));
      memcpy(arg, &(void *) { NULL }, sizeof(void *));
--- a/src/game/text.c
+++ b/src/game/text.c
@@ -120,8 +120,7 @@ void Text_ChangeText(TEXTSTRING *textstring, const char *string)
         return;
     }
     size_t length = strlen(string) + 1;
-    CLAMPG(length, TEXT_MAX_STRING_SIZE);
-    strncpy(textstring->string, string, length);
+    strncpy(textstring->string, string, TEXT_MAX_STRING_SIZE - 1);
     if (length >= TEXT_MAX_STRING_SIZE) {
         textstring->string[TEXT_MAX_STRING_SIZE - 1] = '\0';
     }
--- a/src/filesystem.c
+++ b/src/filesystem.c
@@ -60,7 +60,7 @@ char *File_GetFullPath(const char *path)
             return out;
         }
     }
-    return Memory_Dup(path);
+    return Memory_DupStr(path);
 }
 
 char *File_GuessExtension(const char *path, const char **extensions)
@@ -81,14 +81,14 @@ char *File_GuessExtension(const char *path, const char **extensions)
             }
         }
     }
-    return Memory_Dup(path);
+    return Memory_DupStr(path);
 }
 
 MYFILE *File_Open(const char *path, FILE_OPEN_MODE mode)
 {
     char *full_path = File_GetFullPath(path);
     MYFILE *file = Memory_Alloc(sizeof(MYFILE));
-    file->path = Memory_Dup(path);
+    file->path = Memory_DupStr(path);
     switch (mode) {
     case FILE_OPEN_WRITE:
         file->fp = fopen(full_path, "wb");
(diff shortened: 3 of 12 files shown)


walkawayy:
I can do some reviewing, but I'm also thinking @oziphantom should be added as a reviewer since he opened the issue?

rr-:
OK, but I expect sensible review times.

walkawayy:
The changes look fine to me. But I am not as familiar with proper memory management, and like I said I didn't open the issue. Hopefully Ozi will review if he has a chance.

walkawayy (approved):

src/game/text.c
```diff
@@ -121,7 +121,7 @@ void Text_ChangeText(TEXTSTRING *textstring, const char *string)
     }
     size_t length = strlen(string) + 1;
     CLAMPG(length, TEXT_MAX_STRING_SIZE);
-    strncpy(textstring->string, string, length);
+    memcpy(textstring->string, string, length);
```
oziphantom, line 124: this is not the correct way to solve the warning, in this case the length passed into strncpy should be the size of the buffer, not the string we want to copy. So the warning is telling you, you are passing the size of the string, not the "container size" and hence it could still buffer overflow. 
so `strncpy(textstring->string, string, TEXT_MAX_STRING_SIZE-1);` gives us protection without needing the CLAMPG, or strlen call..
The -1 is because if strncpy has to truncate it will not ensure the last character is a '/0' which means any thing that accesses the pointer will continue to walk through RAM until it hits a '/0', the allocation of the string should clear it to 0 and hence -1 will ensure there is always a '/0'.

scripts/analyze_memleaks.patch
```diff
@@ -47,18 +47,19 @@ index c73c24cf..fbc704f7 100644
 -void Memory_FreePointer(void *arg)
 +void Memory_FreePointerImpl(void *arg, const char *file, int line, const char *func)
```
oziphantom, line 48: sorry my 20hrs a week job has got a horror project that turned into a 40 + weekends. So I've not been able to look through all of FFMPEG lib code to see if they have also made this error. After OPEN_SSL anything is possible. 

This issue is made worse by the fact that gcc is a stickler for one of C's golden rules, and that is 'anything can be converted to a void* no questions asked', so it won't give you or anybody a warning about this. If you put it through g++ it will complain heavily, I'm told. These should be `void** arg`. As you are calling them with `Memory_FreePointer(&pointer);` which is correct, just the `void**` is getting converted to a `void*`.  Thus when you do `*pointer` via the `memcopy` you get a `void*` and not a `void` but the type is actually `void`. Memcpy is not type safe in anyway was so happily takes your `void` and reads `void*` number of bytes anyway because you told it that many bytes and you just so happened to pass in a `void**` so it was ok. Sadly gcc will not give you any warning about forgetting the & and doing `Memory_FreePointer(MyPointer)` will compile just fine. Even if the function is a `void**` and you pass a `void*` to it ; still no error.

From what I've looked at in FFMPEG they typically have REF forms of their objects so `AudioBufferREF` where in that is a `typedef` to `AudioBuffer*` and hence when they do `Free(AudioBufferREF* pointer)` that is actually a `Free(AudioBuffer** pointer)` which does give you a warning. For example 
```
#include <stdlib.h>
#include <string.h>

struct tempStruct {
    int i;
};

typedef struct tempStruct* MyRef;

void Memory_Free(void *memory)
{
    if (memory) {
        free(memory);
    }
}

void Memory_FreePointer(MyRef* arg)
{
    void *memory;
    memcpy(&memory, arg, sizeof(void *));
    memcpy(arg, &(void *) { NULL }, sizeof(void *));
    Memory_Free(memory);
}

void main()
{
    struct tempStruct temp = { 4 };
    MyRef pTS = &temp;
    Memory_FreePointer(pTS);
}
```
will give you 
```
<source>(29): warning C4047: 'function': 'MyRef *' differs in levels of indirection from 'MyRef'
<source>(29): warning C4024: 'Memory_FreePointer': different types for formal and actual parameter 1
````
from MSVC and
```
<source>:29:24: warning: incompatible pointer types passing 'MyRef' (aka 'struct tempStruct *') to parameter of type 'MyRef *' (aka 'struct tempStruct **'); take the address with & [-Wincompatible-pointer-types]
    Memory_FreePointer(pTS);
                       ^~~
                       &
<source>:17:32: note: passing argument to parameter 'arg' here
void Memory_FreePointer(MyRef* arg)
```
from Clang and 
```
<source>:29:24: warning: passing argument 1 of 'Memory_FreePointer' from incompatible pointer type [-Wincompatible-pointer-types]
   29 |     Memory_FreePointer(pTS);
      |                        ^~~
      |                        |
      |                        MyRef {aka struct tempStruct *}
<source>:17:32: note: expected 'struct tempStruct **' but argument is of type 'MyRef' {aka 'struct tempStruct *'}
   17 | void Memory_FreePointer(MyRef* arg)
      |                         ~~~~~~~^~~
```
from GCC. Making it much safer.  I don't think the REF typedef is really suitable for this case though, but at least having the ** so people can see that its meant to be an & if they fall victim to it, or a MACRO that applies the & for them would be a more suitable solution for this code base.

**Review by oziphantom — COMMENTED** [review]
Mostly good, just a couple of small issues to tighten it up.

File: src/game/text.c (hunk 1 above)
Comment (rr-) [reply src/game/text.c]: Fixed, thanks

File: scripts/analyze_memleaks.patch (hunk 2 above)
Comment (rr-) [reply scripts/analyze_memleaks.patch:48]: I agree that it leaves an easy opportunity to shoot oneself in the foot but I perceive the majority of C to be like this. The signature was chosen to use `void*` precisely to use C's implicit casting and avoid warnings such as:

```
char *string = NULL;
Memory_FreePointer(&string);
```

```
/home/rr-/test.c: In function ‘main’:
/home/rr-/test.c:27:24: warning: passing argument 1 of ‘Memory_FreePointer’ from incompatible pointer type [-Wincompatible-pointer-types]
   27 |     Memory_FreePointer(&string);
      |                        ^~~~~~~
      |                        |
      |                        char **
/home/rr-/test.c:16:32: note: expected ‘void **’ but argument is of type ‘char **’
   16 | void Memory_FreePointer(void **arg)
```

whereas it's the intended use of this function. I'd prefer to stay away from macros as much as possible (compare: `Memory_FreePointer(&ptr)` vs `MEMORY_FREE_PTR(ptr);`). Do we have any alternative?

Comment (oziphantom) [reply scripts/analyze_memleaks.patch:48]: I've been thinking about alternatives but then I remembered 
```
#define Memory_FreePointer(memory)      (Memory_FreePointerImpl(memory, __FILE__, __LINE__, __func__))
```
you already have a macro, that is not all uppercase, because macros don't need to be upper case by way of C.  
However an annoying problem is gcc is fine with 
```
void*  to void** fine
void** to void*  fine
char** to void*  fine
char** to void** not fine
```
is a bit of an issue, because the macro needs to become
```
#define Memory_FreePointer(memory)      (Memory_FreePointerImpl((void**)&memory, __FILE__, __LINE__, __func__))
```
which throws away all type safety as we cast. However having the functions porotype and definition have ** so if anybody looks it, they will be able understand how it is used, rather than looking at it, seeing they are doing the right thing and wondering why they have a really strange crash with it, I feel is the lesser of the two evils. 
But then also at this point you could just avoid the complicated setup, performance hit of doing a few mem copies and just 
```
#define Memory_FreePointer(memory)     {Memory_FreePointerImpl(memory, __FILE__, __LINE__, __func__); memory=NULL;}
```
and the function just takes a `void*`

rr-: This macro actually only exists in a .patch that's supposed to help trace memory leaks; normally we use `UPPER_CASE`. The broken name convention is to make the patch minimal in size.

Unfortunately, none of the alternatives look appealing to me. The proposals that add taking a reference underneath actually would confuse me more than the current approach. I'll add some documentation on this function and its intended usage.

oziphantom (approved):

PR #455, opened 2022-03-03
pod: fix pods unhatching on reload if user saves during explosion

### LINKED ISSUE #423: Bug: Boss unhatched egg save issue (The Great Pyramid)
If the player creates a save file as soon as The Great Pyramid level starts, and reloads, the egg of the Giant Atlantean remains unhatched.  This will be noticeable again when Lara returns to the scion room and looks out of the balcony.

20220209_065927_The_Great_Pyramid

Save file: [saveati.zip](https://github.com/rr-/Tomb1Main/files/8031887/saveati.zip)

walkawayy:
Resolves #423.

#### Checklist

#### Descripton

...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 - added ability to automatically walk to pickups when nearby (#18)
 - added ability to automatically walk to switches when nearby (#222)
 - added ability to turn off detailed end of the level stats (#447)
+- fixed exploded mutant pods sometimes appearing unhatched on reload (#423)
 
 ## [2.6.4](https://github.com/rr-/Tomb1Main/compare/2.6.3...2.6.4) - 2022-02-20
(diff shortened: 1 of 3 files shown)


walkawayy:
This was happening because `item->status` was not getting set as `IS_DEACTIVATED` until `AC_DEACTIVATE` was hit in `AnimateItem`. So, if the user saved while the pod explosion was not finished, on reload the pod would look unhatched but with the mutant already spawned.

rr-:
LGTM but can you get sb to test it with legacy saves also?

rr- (approved):

walkawayy:
The pod is still broken only if the user saves a legacy save when the pod was in the middle of exploding and loads it into T1M. Seems like a really rare case that will 99.9% of the time never happen.

Issue #386, opened 2022-02-02
Feature: Unicode support


rr-:
:D

oziphantom:
with a bitmap font?... nah

rr-:
Yes, with a bitmap font
ビットマップ フォント even

Richard-L:
I love it. I would think we can use OpenLara's bitmap glyphs perhaps? Should I ask?

rr-:
I don't think OpenLara uses bitmaps that are readily reusable?

https://github.com/XProger/OpenLara/blob/a4eac0afa15bf11c1c62544cd62f50f277fa2a6c/src/lang/glyph_gr.h

Richard-L:
Ah, maybe I misassumed. Was thinking of these:
https://github.com/XProger/OpenLara/blob/master/src/lang/glyph_ru.png
https://raw.githubusercontent.com/XProger/OpenLara/master/src/lang/glyph_ja.bmp
https://raw.githubusercontent.com/XProger/OpenLara/master/src/lang/glyph_gr.bmp

PR #456, opened 2022-03-05
text: add left/right arrow support and new text functions

walkawayy:
#### Checklist

#### Descripton

...

### DIFF
--- a/src/game/text.c
+++ b/src/game/text.c
@@ -13,6 +13,7 @@
 #define TEXT_BOX_OFFSET 2
 #define TEXT_MAX_STRING_SIZE 100
 #define TEXT_MAX_STRINGS 64
+#define RIGHT_ARROW_SYM 109
 
 static int16_t m_TextstringCount = 0;
 static TEXTSTRING m_TextstringTable[TEXT_MAX_STRINGS] = { 0 };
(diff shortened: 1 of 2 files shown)


src/game/text.c
```diff
@@ -127,6 +128,15 @@ void Text_ChangeText(TEXTSTRING *textstring, const char *string)
...
 }
 
+void Text_MovePos(TEXTSTRING *textstring, int16_t x, int16_t y)
+{
+    if (!textstring) {
+        return;
+    }
+    textstring->pos.x = x;
+    textstring->pos.y = y;
+}
```
rr-, line 138: Since this operation is not additive it should be named `SetPos` not `MovePos`.

src/game/text.c
```diff
@@ -336,6 +363,8 @@ static void Text_DrawText(TEXTSTRING *textstring)
         int32_t sprite_num = letter;
         if (letter >= 16) {
             sprite_num = m_TextASCIIMap[letter - 32];
+        } else if (letter == '\t') {
+            sprite_num = RIGHT_ARROW_SYM;
```
rr-, line 367: This comment is out of scope for this PR but I don't like this approach at all. This define exists only because we ran out of letters, doesn't it? That way it only perpetuates the already terrible character table hacking from Core.
```
"Ammo: %d {icon_sg_ammo}"
```
Related: #386

npel-21: Ya I couldn't find another character to use. And these changes are all just internal text changes to prepare for the arrows in the menu options, so they are in a way related.

walkawayy: Yes I ran out of characters. And these changes are all internal text changes to support arrows so they are related in a way.

rr- (approved):

rr-:
Please feel free to squash + merge if it's done :)

walkawayy:
Cool. I was just wrapping up reworking SOME of the option_passport code... Didn't do a full nuke, but it looks a lot better. And it adds arrow codes. No rush on it! Just didn't want to do a mega rework without some of your input.

rr-:
Thanks, that's the best approach - increment in small steps, discuss early :)

PR #469, opened 2022-03-09
Add final stats screen

walkawayy:
#### Checklist

#### Description
Adds a final statistics screen after the credits play.
...


walkawayy:
This is a WIP because no matter what I currently do, I get a black screen flicker trying to draw text over an image when calling `Text_Draw`. Video of flicker:

https://streamable.com/oq3dwh

bin/cfg/Tomb1Main_gameflow.json5
```diff
@@ -363,6 +363,8 @@
                 {"type": "display_picture", "picture_path": "data\\cred1.pcx", "display_time": 7.5},
                 {"type": "display_picture", "picture_path": "data\\cred2.pcx", "display_time": 7.5},
                 {"type": "display_picture", "picture_path": "data\\cred3.pcx", "display_time": 7.5},
+                // {"type": "display_picture", "picture_path": "data\\install.pcx", "display_time": 7.5},
```
rr-, line 366: Let's remove this

bin/cfg/Tomb1Main_gameflow.json5
```diff
@@ -363,6 +363,8 @@
...
+                {"type": "total_stats"},
```
rr-, line 367: Let's not hardcode the path to `install.pcx` and instead use it as an argument here similar to `GFS_DISPLAY_PICTURE`. Remember to free the memory for the argument in `gameflow.c`.

src/game/stats.c
```diff
@@ -364,3 +364,143 @@ void Stats_Show(int32_t level_num)
...
+    txt = Text_Create(0, top_y, " ");
+    Text_CentreH(txt, 1);
+    Text_CentreV(txt, 1);
+    Text_AddBackground(txt, row_width, height, 0, 0);
+    Text_AddOutline(txt, 1);
+
+    // wait till a skip key is pressed
+    do {
+        // Output_DrawEmpty();
+        // Output_FadeToTransparent(true);
```
rr-, line 495: These lines can be removed.

src/game/stats.c
```diff
@@ -364,3 +364,143 @@ void Stats_Show(int32_t level_num)
...
+        // Output_DisplayPicture("data\\install.pcx");
```
rr-, line 496: This line obviously needs to be added back. However, right now this is loading the picture from disk and decoding it into the picture buffer on every single frame. You should do this once, outside this loop, then continue to use `CopyPictureToScreen` to copy the cached picture buffer to screen within the loop.

Preliminary testing shows that this also removes the issue of black flicker.

src/game/stats.c
```diff
@@ -364,3 +364,143 @@ void Stats_Show(int32_t level_num)
...
+        Output_InitialisePolyList();
+        // Draw_DrawScene(false);
```
rr-, line 498: `DrawScene` is used to draw the game (Lara, rooms, sprites, objects) and it does not belong here.

src/game/stats.c
```diff
@@ -364,3 +364,143 @@ void Stats_Show(int32_t level_num)
...
+        Output_CopyPictureToScreen();
+        Input_Update();
+        Text_Draw();
+        Output_DumpScreen();
+    } while (!g_InputDB.select && !g_InputDB.deselect);
+
+    Text_RemoveAll();
+}
```
rr-, line 506: Missing newline.

rr- (changes requested):
I would also add a setting for this feature so that the users can disable it regardless of the gameflow file, which

- we impose on the users by expecting them to always use the newest version
- does not have backwards compatibility in case users mess something up
- users shouldn't generally tinker with it unless they're custom level builders

bin/cfg/Tomb1Main_gameflow.json5
walkawayy: Ok. I wasn't sure if we wanted to leave it up to the builders to display the stats over a picture, an in game frame, or a custscene.

walkawayy:
Ok it works well now. The user can provide a background picture. What do you think about the placement right now? Also all the credit images and this install image could use widescreen versions like the title screen.

src/game/gameflow.c
```diff
@@ -904,6 +924,7 @@ void GameFlow_Shutdown()
                     case GFS_LOOP_CINE:
                     case GFS_STOP_CINE:
                     case GFS_LEVEL_STATS:
+                    case GFS_TOTAL_STATS:
```
rr-, line 927: This case needs to be moved to `GFS_DISPLAY_PICTURE` to free the memory.

rr- (changes requested):
- I don't see a setting
- TR2 looks better especially with this graphic imo.

walkawayy:
Ok I think everything is done except getting a conclusion on this: https://github.com/rr-/Tomb1Main/issues/385#issuecomment-1062263243

I also realized for the reworked graphics options menu I also used a TR2 like text box that doesn't have the offset title bar.

walkawayy:
How does this look @Richard-L  @rr- 

20220310_122102_The_Great_Pyramid

Richard-L:
I think that's beautiful 👌

rr- (approved):

PR #471, opened 2022-03-10
gameflow: add support for unobtainable pickups and kills

### LINKED ISSUE #470: Feature request: adjust the item and kill counters for unobtainable items
There is a section in the readme that explains that the end of the level counters aren't 100% faithful due to unobtainable items or inaccessible creatures.

That's nice, but I don't think many players will read it. We can do better, which is especially important in the context of the upcoming #469.

I propose to add a delta value in the gameflow for each level, `obtainable_kills` and `obtainable_items`, that defaults to 0, but for the levels with unreachable items, it's set to adjust the values to ones that casual players will understand.

So
Midas: `"unobtainable_items": 1`
Atlantis: `"unobtainable_kills": 4`

Then we could subtract these values during the stats screens.

This way level builders will never have to hardcode their item counts, and we only make an exception to broken levels.

We could also display the total amount in brackets, as in

Midas
ITEMS: 15 (16)

Note that these numbers are completely made up, I didn't look up the actual stats at the time of writing this.

walkawayy:
Resolves #470.

#### Checklist

#### Descripton

...

### DIFF
--- a/src/game/gameflow.c
+++ b/src/game/gameflow.c
@@ -726,6 +726,12 @@ static bool GameFlow_LoadScriptLevels(struct json_object_s *obj)
             cur->water_color.override = false;
         }
 
+        cur->unobtainable.pickups =
+            json_object_get_int(jlvl_obj, "unobtainable_pickups", 0);
+
+        cur->unobtainable.kills =
+            json_object_get_int(jlvl_obj, "unobtainable_kills", 0);
+
         struct json_object_s *jlbl_strings_obj =
             json_object_get_object(jlvl_obj, "strings");
         if (!jlbl_strings_obj) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - added contextual arrows to graphics option navigation (#462)
 - added a final statistics screen (#385)
 - added music during the credits (#356)
+- added unobtainable pickups and kills stats support in the gameflow (#470)
 - fixed exploded mutant pods sometimes appearing unhatched on reload (#423)
 
 ## [2.6.4](https://github.com/rr-/Tomb1Main/compare/2.6.3...2.6.4) - 2022-02-20
--- a/README.md
+++ b/README.md
@@ -260,6 +260,7 @@ Not all options are turned on by default. Refer to `Tomb1Main.json5` for details
 - added contextual arrows to menu options
 - added a final statistics screen
 - added music during the credits
+- added unobtainable pickups and kills stats support in the gameflow
 - changed internal game memory limit from 3.5 MB to 16 MB
 - changed moveable limit from 256 to 10240
 - changed maximum textures from 2048 to 8192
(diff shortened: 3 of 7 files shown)


src/game/gameflow.c
```diff
@@ -706,6 +706,22 @@ static bool GameFlow_LoadScriptLevels(struct json_object_s *obj)
             cur->water_color.override = false;
         }
 
+        tmp_i = json_object_get_int(
+            jlvl_obj, "unobtainable_pickups", JSON_INVALID_NUMBER);
+        if (tmp_i != JSON_INVALID_NUMBER) {
+            cur->unobtainable.pickups = tmp_i;
+        } else {
+            cur->unobtainable.pickups = 0;
+        }
```
rr-, line 715:
```c
        cur->unobtainable.pickups = json_object_get_int(
            jlvl_obj, "unobtainable_pickups", 0);
```

src/game/gameflow.c
```diff
@@ -706,6 +706,22 @@ static bool GameFlow_LoadScriptLevels(struct json_object_s *obj)
...
+
+        tmp_i = json_object_get_int(
+            jlvl_obj, "unobtainable_kills", JSON_INVALID_NUMBER);
+        if (tmp_i != JSON_INVALID_NUMBER) {
+            cur->unobtainable.kills = tmp_i;
+        } else {
+            cur->unobtainable.kills = 0;
+        }
```
rr-, line 723: Ditto

rr- (commented):
LGTM apart from a minor cosmetic issue. Please consult with posix if we should display the physical number in brackets or not.

walkawayy:
I'd think no because they are called unobtainable for a reason? And the user can change or remove the unobtainable entry from the gameflow to change the number if they'd like. But I'll tag @Richard-L .

rr- (approved):
Let's wait for posix though

Richard-L:
Okay and so in the stats you always just want to show the corrected count, right? Or both in some way?

walkawayy:
I think showing just the corrected count is fine. Most players aren't glitching to get  the unobtainable items in the game. And there's only 4 total, 1 of which is completely unobtainable. If they are glitching, they are probably power user enough to adjust the gameflow file if they want to show the total number of pickups including those unobtainable items.

Richard-L:
I see. I think you're right.

Which levels have the obtainable by glitches pick-ups, and which has the entirely unobtainable one? I wonder if it would be nice to show a line at the bottom of the compass stats saying: "This level has x unobtainable pickups". Just a thought.

walkawayy:
This list is here:
https://github.com/rr-/Tomb1Main/pull/387#issuecomment-1028246879

That sounds kinda wonky for the average player.

Richard-L:
Hmm okay. I guess you're right it's best to keep it simple. That table you linked looks really nice though.

walkawayy:
I removed the stat discrepancy part of the README since that no longer applies.

rr- (approved):

walkawayy:
Resolved conflicts and gonna merge.

PR #473, opened 2022-03-10
option_graphics: offset title bar to match other menus

walkawayy:
#### Checklist

#### Descripton
Offset title bar to match other menus.
...

Pic:
20220310_141118_Title

### DIFF
--- a/src/game/option_graphics.c
+++ b/src/game/option_graphics.c
@@ -87,7 +87,7 @@ static void Option_GraphicsInitText()
 {
     char buf[OPTION_LENGTH];
 
-    m_Text[TEXT_TITLE_BORDER] = Text_Create(0, TOP_Y, " ");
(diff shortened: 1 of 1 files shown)


rr- (commented):
Missing changelog entry. I don't see any difference in this picture

walkawayy:
I don't think it really needs one since it's just a slight adjustment to the graphics menu title bar that was recently changed. It just offsets the "Select Details" at the top of the text box to be like double boxed.

rr-:
Hmm, I thought we only touched the arrows in this dialog. Can you show an image of before and after?

walkawayy:
Here's TombATI's graphics menu which is what T1M used to look like just with additional options:
image



This is the new graphics menu without the title bar offset:
image

rr-:
Hmm I had hoped for something more like this:

20220310_213935_Title
20220310_214015_Title

That said, :+1:

rr- (approved):
I agree that there is no need for a changelog entry now that I understand the context.

PR #488, opened 2022-03-16
style: make parameterless functions void

rr-:
#### Checklist

- [X] ~~I have added a changelog entry about what my pull request accomplishes~~ This is internal change.

#### Descripton

As ChocolateFan once explained on Discord, in C functions declared as parameterless in fact are free to take arbitrary number of arguments. To tell the compiler that the function is not allowed to take any arguments, we use "func(void)" idiom.

And sure enough we had two broken function signatures because of this peculiarity.


walkawayy:
LGTM. ChocolateFan also approves lol.

walkawayy (approved):

PR #492, opened 2022-03-17
Refactor object signatures

rr-:
#### Checklist

- [X] ~~I have added a changelog entry about what my pull request accomplishes~~ This is internal change that should not produce any visible changes to the users

#### Descripton

This PR refactors src/game/objects/ directory to comply with the coding conventions.

- Moved GunShot to effects/
- Moved certain gunshot routines from people.c to gunshot.c (I'll think what to do about leftovers once I get to people.c)
- Renamed SavegameCrystal and SavegameItem to SaveCrystal
- Renamed PickUp to Pickup
- Used bool where possible
- Changed function signatures to include a prefix with their module name
- Made certain private functions static


src/game/objects/switch.c
```diff
@@ -24,7 +24,7 @@ static int16_t m_Switch1Bounds[12] = {
     +10 * PHD_DEGREE,
 };
 
-static int16_t m_Switch1BoundsControlled[12] = {
+static int16_t m_Switch_BoundsControlled[12] = {
```
walkawayy, line 27: Should we add back in the `1` at the end to make sure this is clear? I know it only applies to switch1 though. Maybe we can drop the number 1 from all switch1 variables and rename `m_Switch_Bounds2` to `m_Switch_BoundsUW` etc. `UW` is much more clear than `2`.

I was thinking about something similar for Scion numbering too, but I didn't dig into that.

src/game/objects/trapdoor.c
```diff
@@ -3,16 +3,39 @@
 #include "game/control.h"
 #include "global/vars.h"
 
-void SetupTrapDoor(OBJECT_INFO *obj)
+static bool TrapDoor_OnTrapDoor(ITEM_INFO *item, int32_t x, int32_t z);
+
+static bool TrapDoor_OnTrapDoor(ITEM_INFO *item, int32_t x, int32_t z)
```
walkawayy, line 8: Maybe rename to `TrapDoor_On` or `TrapDoor_StandingOn` or something? IMO we don't need to repeat `TrapDoor` here.

src/game/people.c
```diff
@@ -2,18 +2,15 @@
...
-#include "game/effects/ricochet.h"
+#include "game/effects/gunshot.h"
 #include "game/items.h"
 #include "game/random.h"
-#include "game/sound.h"
-#include "game/sphere.h"
 #include "global/vars.h"
 
-int32_t Targetable(ITEM_INFO *item, AI_INFO *info)
+bool Targetable(ITEM_INFO *item, AI_INFO *info)
```
walkawayy, line 10: `IsTargetable` or `IsLaraTargetable`?

src/game/people.c
```diff
@@ -30,70 +27,6 @@ int32_t Targetable(ITEM_INFO *item, AI_INFO *info)
...
-        + ((Random_GetDraw() - 0x4000) * (WALL_L / 2)) / 0x7FFF;
-    pos.y = g_LaraItem->floor;
-    pos.z = g_LaraItem->pos.z
-        + ((Random_GetDraw() - 0x4000) * (WALL_L / 2)) / 0x7FFF;
-    pos.room_number = g_LaraItem->room_number;
-    Ricochet(&pos);
-    return GunShot(x, y, z, speed, y_rot, room_num);
-}
-
 int32_t ShotLara(
```
walkawayy, line 30: Should this be a bool function? Also maybe `IsLaraShot` or `IsShot` to match the `Targetable` function?

walkawayy (changes requested):
Most of it looks good. A lot of it is just internal renaming. I ran around a couple levels and didn't notice any bugs, but I definitely didn't extensively test.

src/game/objects/switch.c
rr-: Agreed about UW being much more readable. Ignored Scion pieces.

src/game/objects/trapdoor.c
rr-: Fixed

src/game/people.c
rr-: I didn't move it yet, but probably will want to move it to gunshot.c later. For now I want to keep it here.

src/game/people.c
rr-: Fixed. Also no for `Is...` because this function has side effects (see `CreatureEffect`) + I don't want to touch names from outside game/objects/ in this PR.

walkawayy (approved):

PR #495, opened 2022-03-22
Trap names

rr-:
#### Checklist

- [X] ~~I have added a changelog entry about what my pull request accomplishes~~ Internal change

#### Descripton

This pull request refactors `src/objects/traps/` to conform with the coding and naming conventions.

**Note**: I did not rename `AlterFloorHeight` because

- it's more about floor data rather than movable block, so I'm thinking we should eventually move it somewhere else 
- it's shared also with rolling block (e.g. qualopec columns room)

@walkawayy tell me also what you think about merging `dart_emitter.c` with `dart.c` similar to what we have in `lava.c` and `flame.c`. Or maybe the other way around, split the others into smaller files, one for each object type.


src/game/traps/movable_block.h
```diff
@@ -2,25 +2,17 @@
...
+void MovableBlock_Control(int16_t item_num);
+void MovableBlock_Collision(
     int16_t item_num, ITEM_INFO *lara_item, COLL_INFO *coll);
-void DrawMovableBlock(ITEM_INFO *item);
+void MovableBlock_Draw(ITEM_INFO *item);
 
-int32_t TestBlockMovable(ITEM_INFO *item, int32_t blockhite);
-int32_t TestBlockPush(ITEM_INFO *item, int32_t blokhite, uint16_t quadrant);
-int32_t TestBlockPull(ITEM_INFO *item, int32_t blokhite, uint16_t quadrant);
+bool MovableBlock_Test(ITEM_INFO *item, int32_t blockhite);
```
walkawayy, line 12: It is easy to tell what `MovableBlock_TestPush` and `MovableBlock_TestPull` are checking. But `MovableBlock_Test` is not clear IMO.

walkawayy (changes requested):
Looks good. I think you could merge the darts into one function. But if you like it separate just leave it. I don't really have a preference.

src/game/traps/movable_block.h
rr-: I'm not sure what it's testing TBH.

walkawayy: Ok so ChocolateFan says it is checking if a block can be moved by testing the square that the block is going to be in. If it has the same height as the square the block is in now. Possible suggestions:
`MovableBlock_TestDestination`
`MovableBlock_TestHeight`
` MovableBlock_TestMovement`

rr-: Applied

walkawayy (approved):

PR #503, opened 2022-03-24
Effect names

rr-:
#### Checklist

- [X] ~~I have added a changelog entry about what my pull request accomplishes~~ This is an internal change

#### Descripton

- Moved all object routines to objects/, including ai/, traps/ and effects/
- Moved things not relevant to _object_ routines outside, namely:
    - `g_EffectRoutines` members to effect_routines/
    - other effects to effects/
- Added a script to sort imports
- Sorted the imports
- Moved gunshot routines to effects/, however I kept a few routines within object/effects as Blabla_Spawn(…)

The motivation for splitting the effects and moving things to objects/ is to make clear distinction between code that is a part of the OBJECT_INFO, and other parts of the game.

The FX_ prefix breaks the convention a bit but I believe it is justified considering the alternative prefix, EffectRoutine, is quite long. All FX_ share the same signatures.

I don't really like effects/ directory because we just threw random things into it. At the same time some of these routines do not really belong to objects/effects/ – for example ExplodingDeath shouldn't be named BodyPart_Spawn, because it spawns multiple body parts, not just 1. And BodyPart_ExplodingDeath sounds arbitrary. Effect_ExplodingDeath looked OK to me. Most of things under effects/ are compatible with CreatureEffect's `spawn` argument.

---

After this pull request the only thing related to object routines naming will be the creatures, which I plan to tackle right after.


walkawayy:
This looks good to me.

walkawayy (approved):

PR #509, opened 2022-04-05
s_clock: add a cheat to increase the game speed

### LINKED ISSUE #135: Feature request: Fast forward toggle
This would be very useful for testing. Mainly to speed up block pulling and pushing.

https://user-images.githubusercontent.com/4934209/137978916-f539feea-6a67-4c4d-839d-04ac0d8ba496.mp4

^ this footage is from Duckstation. You can hotkey toggle between uncapped FPS (=speed up) or a set "turbo" level of a desired gamespeed percentage, above 100%. So perhaps in the json file you might set your turbo to 400% to push those blocks faster.

Thanks for considering.

walkawayy:
Resolves #135.

#### Checklist

#### Description
Added a cheat (tab key) to increase the game speed.
...


rr- (commented):
Almost all keys are configurable (including cheat keys), why should this be an exception?

walkawayy:
I went through the whole process of adding it as a cheat key like level skip, give items, etc with a game flow string, but I couldn't figure out how to make the mode a toggle. It only worked with the key held down, and would stop if the key was lifted.

rr-:
- Please make this key configurable like the others
- Please avoid adding `g_TurboKey` in favor of a setter and a getter in `clock.c` such as `Clock_IsTurboEnabled()` `Clock_SetTurbo(bool enable)`
- Use `InputDB` instead of `Input` to deal with the holding key issue (it will be true only on the frame the user pressed the key, letting you toggle this setting)

walkawayy:
https://streamable.com/o1kdp9

src/game/clock.c
```diff
@@ -5,6 +5,13 @@
 #include <stdio.h>
 #include <time.h>
 
+static bool m_IsTurbo = false;
+
+void Clock_SetTurbo()
+{
+    m_IsTurbo ^= true;
```
rr-, line 12: - Please call it `ToggleTurbo`
- Missing `(void)`
- This is a bool. Please use `m_IsTurbo = !m_IsTurbo`. Has no effect but not everyone knows what xor is

src/game/clock.h
```diff
@@ -3,6 +3,7 @@
 #include <stdbool.h>
 #include <stdint.h>
 
+void Clock_SetTurbo();
```
rr-, line 6: Missing `(void)`

rr- (commented):
It still isn't possible to remap this key from the game UI, is it?

Richard-L:
Would we perhaps make a new paragraph or empty line break for cheats/tools that aren't native control inputs?

rr-:
We could move look to the left column and put a blank line above cheats.

src/game/clock.c
```diff
@@ -5,6 +5,13 @@
 #include <stdio.h>
 #include <time.h>
 
+static bool m_IsTurbo = false;
+
+void Clock_ToggleTurbo()
```
rr-, line 10: Still missing `(void)`

rr- (changes requested):

Richard-L:
Not to get too complicated with this, but I'd suggest: Make two groups, each with two columns so they use up the least amount of height.

rr-:
@Richard-L that's too complex for this PR. It needs a new issue.

walkawayy:
How does this look.

Cheats disabled:
image

Cheats enabled:
image

rr-:
LGTM. IMO let's capitalize, good catch.

Richard-L:
I love it @walkawayy and good find on the capitalisation :+1:

rr- (approved):
Very neat!

walkawayy:
We will probably need a few different setups to test this unfortunately lol. I chose 1 because anything higher was unplayably fast. But I noticed the speed can depend on the system. Playing at 1440p for example the FPS counter would show around 130 FPS but 480p could show above 160 FPS. Not a big deal and it is a cheat, but I just don't know exactly how it will behave.

Richard-L:
You set the speed as a % in the JSON, and it toggles via hotkey. So if someone sets 200% the game runs twice as fast, if turbo is active, and if the hardware can perform enough.

The other option is just uncapped speed, but that all too often results in absurd results.

rr-:
A simple way to deal with this would be to not return `1` but instead `elapsed * 1.5`

walkawayy:
Problem is it needs to return an int.

rr-:
I agree with your sentiment. Also you can check for `g_Input.slow`.

src/game/clock.c
```diff
@@ -5,6 +5,26 @@
 #include <stdio.h>
 #include <time.h>
 
+#define TURBO_NONE 1
+#define TURBO_DOUBLE 2
+#define TURBO_TRIPLE 3
+
+static int16_t m_TurboSpeed = TURBO_NONE;
```
rr-, line 12: Let's use `m_TurboSpeedMul` to make it clear it's used as a multiplier.

src/game/clock.c
```diff
@@ -5,6 +5,26 @@
 #include <stdio.h>
 #include <time.h>
 
+#define TURBO_NONE 1
+#define TURBO_DOUBLE 2
+#define TURBO_TRIPLE 3
```
rr-, line 10: The code never uses `TURBO_DOUBLE`. Please replace these defines with `#define MAX_TURBO_SPEED_MUL 3`.

(Alternatively we could have a `switch` statement but it seems to be an overkill.)

src/game/clock.c
```diff
@@ -5,6 +5,26 @@
...
+    TURBO_TRIPLE,
+};
+
+void Clock_ToggleTurbo(void)
+{
+    if (m_TurboSpeed == TURBO_TRIPLE) {
+        m_TurboSpeed = TURBO_NONE;
+    } else {
+        m_TurboSpeed++;
+    }
```
rr-, line 25:
```
if (m_TurboSpeedMul >= MAX_TURBO) {
    m_TurboSpeedMul = 1;
} else {
    m_TurboSpeedMul++;
}
```

src/game/clock.h
```diff
@@ -3,6 +3,7 @@
 #include <stdbool.h>
 #include <stdint.h>
 
+void Clock_ToggleTurbo(void);
```
rr-, line 6: This function no longer Toggles things, it Cycles things. It also no longer toggles turbo "mode" (which would be OK to omit) but cycles turbo speed multipliers. So it should be named `Clock_CycleTurboSpeed` (we omit the "Mul" part here because it's an implementation detail).

rr- (changes requested):
Cosmetic remarks only

Richard-L:
@walkawayy "incremental" turbo steps on press are a nice idea!

walkawayy:
> @walkawayy "incremental" turbo steps on press are a nice idea. Just, to turn turbo off, people will always have to have the game go even faster for a brief moment before they can return to 100% speed?
> 
> Another option might be to hold down tab for turbo. So on release it returns to 100%.

You can tap the key quickly through 3x speed and it's only a few frames.

rr- (approved):

walkawayy:
Would adding that potentially mess with the shift key as the walk key?

I'm also wondering if the turbo key check should be moved out of LaraControl. Right now it can only be used during normal gameplay, but the turbo will persist elsewhere like the inventory screen. Maybe it should be moved to a place where it can always be toggled.

rr-:
Good point about this feature affecting the inventory and the inability to turn it off from there.

We handle such global keys in `S_Input_GetCurrentState` :)

walkawayy:
`S_Input_GetCurrentState` doesn't use the debounced `g_InputDB` though. Would I need to add a cooldown like `m_MedipackCoolDown`?

rr-:
Wow, you're right. Please do it in `Input_Update` instead. Can you also create a ticket for us to move game-specific things away from `s/input`?

walkawayy:
Ok if the new placement in Input_Update looks good, I think this is ready to merge. Now you can toggle the cheat on startup, in puase screen, inventory screen, etc.

rr- (approved):

PR #533, opened 2022-05-01
Configuration Option to Change the Number of Save Slots

### LINKED ISSUE #170: Feature request: increase number of save slots
Would be nice to have at least two per level I'd say.

fluxtom:
#### Checklist

#### Description

This PR adds the `maximum_save_slots` configuration option, to resolve #170.

This requires two arrays to be dynamically allocated: [m_SelectLevelStrings](https://github.com/rr-/Tomb1Main/blob/61d6789d67b046cd35de99e81b70d2bf7e7a364e/src/game/option_passport.c#L30) and [m_SavegameInfo](https://github.com/rr-/Tomb1Main/blob/61d6789d67b046cd35de99e81b70d2bf7e7a364e/src/game/savegame.c#L38), as they use the [MAX_SAVE_SLOTS](https://github.com/rr-/Tomb1Main/blob/db9c32c7f2fd41cb2fff1a5ba4fee7a8280267ac/src/global/const.h#L29) constant to initialize their size. Further, the `item_texts` field in [g_SavegameRequester](https://github.com/rr-/Tomb1Main/blob/db9c32c7f2fd41cb2fff1a5ba4fee7a8280267ac/src/game/option_passport.c#L70) must now be set at run time.

`Savegame_ScanSavedGames()` expects that the  `m_SavegameStrings` array will be initialised by the time it is called. I could not find a nice place to initalise this array, that is called before `Savegame_ScanSavedgames`; therefore, I have added the function `Option_PassportInitSavegameStrings` that is called in `Savegame_Shutdown`.

I have marked this PR as a draft because although the functionality is working correctly, I am 90% sure that I have done somethings incorrect:

- The compiler complains about this line: `g_SavegameRequester.item_texts = &m_SavegameStrings[0];` being a cast to a `char**` from a `char*`, however doing it correctly causes a segmentation fault in `Savegame_ScanSavedGames` [here](https://github.com/rr-/Tomb1Main/blob/91de66e03613f28237653e45ec2c358b9a5b0770/src/game/savegame.c#L520) on the second iteration.
- I am not confident that I am freeing the memory that I dynamically allocate correctly, I am still quite hazy on this area :p


src/game/option_passport.c
```diff
@@ -27,7 +29,7 @@ static PASSPORT_MODE m_PassportMode = PASSPORT_MODE_FLIP;
 static bool m_IsTextInit = false;
 static TEXTSTRING *m_Text[TEXT_NUMBER_OF] = { 0 };
 static char m_NewGameStrings[MAX_GAME_MODES][MAX_GAME_MODE_LENGTH] = { 0 };
-static char m_SavegameStrings[MAX_SAVE_SLOTS][MAX_LEVEL_NAME_LENGTH] = { 0 };
+static char **m_SavegameStrings;
```
fluxtom: A later commit sets this to `NULL`

Richard-L:
@fluxtom you've checked how TR2M does it, right?

src/game/option_passport.c
```diff
@@ -97,6 +99,14 @@ static void Option_PassportInitSaveRequester(int16_t page_num);
 static void Option_PassportLoadGame(void);
 static void Option_PassportSelectLevel(void);
 
+void Option_PassportInitSavegameStrings(void)
+{
+    m_SavegameStrings = Memory_Alloc(g_Config.maximum_save_slots * g_SavegameRequester.item_text_len);
+
+    // this is casting to a char** from a char*
+    g_SavegameRequester.item_texts = &m_SavegameStrings[0];
```
fluxtom, line 107: This works fine, but the compiler does not like it. I looked at other places that `item_texts` is [initalised](https://github.com/rr-/Tomb1Main/blob/91de66e03613f28237653e45ec2c358b9a5b0770/src/game/option_passport.c#L233), but doing a similar thing here causes a segmentation fault.

fluxtom:
@Richard-L didn't know that existed until now, I will have a look now...

Richard-L:
Hmm I was certain the corresponding issue had a link to the respective code snippet of TR2M, but it doesn't look like it's there.

Richard-L:
Hmm I was certain the corresponding issue had a link to the respective code snippet of TR2M, but it doesn't look like it's there.

src/game/option_passport.c
rr-: You're declaring it as a double pointer, so each member should be a pointer too, whereas it's zero-initialised. 
To circumvent this, either allocate each member separately, like this:
```c
m_SavegameStrings = Memory_Alloc(g_Config.maxium_save_slots * sizeof(char*));
for (int i = 0; i < g_Config.maxium_save_slots; i++) {
    m_SavegameStrings[i] = Memory_Alloc(g_SavegameRequester.item_text_len);
}
```
or declare these strings as `char*` then remove the cast.
Are you sure that using `item_text_len` here is OK?

src/game/savegame.c
```diff
@@ -497,7 +503,7 @@ void Savegame_ScanSavedGames(void)
     REQUEST_INFO *req = &g_SavegameRequester;
 
     req->items = 0;
-    for (int i = 0; i < MAX_SAVE_SLOTS; i++) {
+    for (int i = 0; i < g_Config.maximum_save_slots; i++) {
```
rr-, line 506: We don't need `MAX_SAVE_SLOTS` anymore do we?

src/game/savegame.c
```diff
@@ -433,13 +434,18 @@ bool Savegame_UpdateDeathCounters(int32_t slot_num, GAME_INFO *game_info)
...
 void Savegame_Shutdown(void)
 {
-    for (int i = 0; i < MAX_SAVE_SLOTS; i++) {
-        SAVEGAME_INFO *savegame_info = &m_SavegameInfo[i];
-        savegame_info->format = 0;
-        savegame_info->counter = -1;
-        savegame_info->level_num = -1;
-        Memory_FreePointer(&savegame_info->full_path);
-        Memory_FreePointer(&savegame_info->level_title);
+    if (m_SavegameInfo == NULL) {
```
rr-, line 437: Shutdown functions are about cleaning things up, freeing memory and resetting back to neutral state. They should not allocate new memory.
Please prepare a new function `Savegame_Init(void)` that does the `Memory_Alloc` call, and in `Savegame_Shutdown`, free that memory.

rr- (changes requested):
The big picture LGTM, however we need to do some cosmetic adjustments.

src/game/option_passport.c
```diff
@@ -1,5 +1,7 @@
 #include "game/option.h"
 
+#include "config.h"
+
```
walkawayy, line 4: You can remove the blank line here.

src/game/savegame.c
walkawayy: Nope. I checked, and you can remove the `#define MAX_SAVE_SLOTS 16` in `const.h`.

src/game/savegame.c
walkawayy: Agreed. Then I think you can just do:
```
void Savegame_ScanSavedGames(void)
{
    Savegame_Shutdown();
    Savegame_Init();
    ...
```

fluxtom: Doing `Savegame_Init` after `Savegame_Shutdown` seg faults, the other way around works though.

src/game/option_passport.c
fluxtom: That makes sense, still slightly in the mindset that the language will do everything for me :p

And yes, `item_text_len` probably isn't okay on second look.

src/game/savegame.c
```diff
@@ -431,9 +432,15 @@ bool Savegame_UpdateDeathCounters(int32_t slot_num, GAME_INFO *game_info)
     return ret;
 }
 
+void Savegame_Init()
+{
+    m_SavegameInfo = Memory_Alloc(sizeof(*m_SavegameInfo) * g_Config.maximum_save_slots);
```
rr-, line 437: Let's take `sizeof(SAVEGAME_INFO)`

src/game/savegame.c
```diff
@@ -431,9 +432,15 @@ bool Savegame_UpdateDeathCounters(int32_t slot_num, GAME_INFO *game_info)
     return ret;
 }
 
+void Savegame_Init()
```
rr-, line 435: Missing `(void)`

src/game/savegame.c
```diff
@@ -445,12 +452,13 @@ void Savegame_Shutdown(void)
 
 void Savegame_ScanSavedGames(void)
 {
+    Savegame_Init();
```
rr-, line 455: Please flip the order of these calls as currently they make little semantic sense (init and immediately clean up)?

src/game/savegame.c
```diff
@@ -431,9 +432,15 @@ bool Savegame_UpdateDeathCounters(int32_t slot_num, GAME_INFO *game_info)
...
+void Savegame_Init()
+{
+    m_SavegameInfo = Memory_Alloc(sizeof(*m_SavegameInfo) * g_Config.maximum_save_slots);
+    Option_PassportInitSavegameStrings();
+}
+
 void Savegame_Shutdown(void)
 {
-    for (int i = 0; i < MAX_SAVE_SLOTS; i++) {
+    for (int i = 0; i < g_Config.maximum_save_slots; i++) {
```
rr-, line 443: Missing `Memory_FreePointer(&m_SavegameInfo)`

rr-: And to make it work with the flipped init/shutdown order, we can check if `m_SavegameInfo` is not null before dereferencing it down below :)

src/game/option_passport.c
```diff
@@ -97,6 +98,16 @@ static void Option_PassportInitSaveRequester(int16_t page_num);
 static void Option_PassportLoadGame(void);
 static void Option_PassportSelectLevel(void);
 
+void Option_PassportInitSavegameStrings(void)
+{
+    m_SavegameStrings = Memory_Alloc(g_Config.maximum_save_slots * sizeof(char *));
+    for (int i = 0; i < g_Config.maximum_save_slots; i++) {
+        m_SavegameStrings[i] = Memory_Alloc(sizeof(char *));
```
rr-, line 105: That doesn't make sense – we're now allocating 4 bytes for each string. Please just use `MAX_LEVEL_NAME_LENGTH` as it used to be the case.

src/game/option_passport.c
```diff
@@ -130,6 +141,8 @@ static void Option_PassportShutdownText(void)
     Option_PassportFreeLevelStrings();
     m_PassportMode = PASSPORT_MODE_FLIP;
     m_IsTextInit = false;
+
+    Memory_FreePointer(&m_SavegameStrings);
```
rr-, line 145: Need to also deallocate the individual strings.

```c
for (int i = 0; i < g_Config.maximum_save_slots; i++) {
    Memory_FreePointer(&m_SavegameStrings[i]);
}
```

This comes before deallocating the entire array.

rr- (commented):
Minor suggestions related to memory management

src/game/option_passport.c
fluxtom: Should this not be `Memory_FreePointer(m_SavegameStrings[i]);`? Maybe I'm missing something but keeping the `&` crashes (because `m_SavegameStrings` already stores pointers, right?) and removing seems to fix the issue.

rr-: It should take `&`; alternatively, `Memory_Free(m_SavegameStrings[i]); m_SavegameStrings[i] = NULL;`.

rr-:
Continued in #574

PR #540, opened 2022-05-05
PS1 UI

### LINKED ISSUE #517: Feature: PS1 UI
Let's make TR1M even more authentic TR1 vintage shall we. A small thing, but still puzzling why the DOS version got rid of it: the purple hover effect in the UI.

PS1:

SLES-00024_2022-04-11_23-15-22

SLES-00024_2022-04-11_23-15-56

Saturn:

image

I would say the PS1 style looks best.

OpenLara has implemented it quite successfully:

level-select_openlara

walkawayy:
#### Checklist

#### Description
Adds a vertical center gradient for text boxes that is used for all main menu options to add PS1 style purple backgrounds.

Video:
https://streamable.com/q7sl3y
...


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## [Unreleased](https://github.com/rr-/Tomb1Main/compare/2.8.1...master)
 - fixed Tihocan chain block sound (#433)
+- added background color gradients to selected menu text boxes (#517)
```
rr-: Additions come before fixes

src/global/types.h
```diff
@@ -1593,6 +1594,10 @@ typedef struct TEXTSTRING {
         int32_t h;
         int32_t v;
     } scale;
+    struct {
+        RGBA8888 centre;
+        RGBA8888 edge;
+    } centre_v_gradient;
```
rr-: What does the `v` in `centre_v_gradient`  stand for?

rr- (commented):
TBH I'm not a fan of this, can we get a feature flag for this? Could be in the `Text_CentreVGradient` or `Text_DrawText`.

rr-:
Could we also get a screenshot how this looks in PS1?

walkawayy:
Ya I was thinking of just doing `if (textstring->flags.centre_v_gradient && config.menu_text_highlights)` or whatever we want to call it in `Text_DrawText`. As with all these options, on or off by default? See the linked issue for pictures.

src/global/types.h
walkawayy: Center vertical gradient. If you have a better name lmk. I wasn't sure what to call it.

rr-:
Vote on Discord IMO. (Emoji reactions work great for this sort of thing)
The colors look good. They only feel a bit off on bright passport pages but PS1 never rendered them this way so we don't have comparison.

Richard-L:
@rr- why are you not a fan? Have some screenshots in the issue I'd posted

rr-:
Because I think it looks better with one color instead of two, plus I'm used to the PC version.

Richard-L:
😆 come on that's a funny argument. 

It'll probably be optional under something like "PS1 style settings" anyway, no?

rr-:
In the proposed implementation it is not optional and that was the entire point of my comment.
I don't know what's funny about my personal preferences and my request to make it optional?

Richard-L:
Nothing, it's all good. I just came to this with the mindset from last night between walkaway and I, when we (mostly he) tried to tweak this again for hours, find the right shade of purple, bit more red, bit more blue, more/less transparent, comparison on top of white passport or different backgrounds, lots of back and forth....

So when I read you didn't like it I thought you meant parameters like that.

I agree it should be optional. I wasn't aware. There will always be people more emotionally attached to either PS1 or PC style, and it's good to cater to both groups, so long as there isn't too much code bloat entailed.

walkawayy:
Alright pushed up a JSON setting for this. Waiting for Discord poll results on whether to enable or disable this by default. It doesn't matter to me either way.

src/global/types.h
rr-: Maybe something related to `highlight background`?

rr- (approved):

Richard-L:
Discord poll result after ~7hrs
image

Interesting to see how perfectly split opinions are. Some useful debate in the chat today, leading to the desire to first bring the PS1 darker textbox and border style of the UI back. Feels like that would move past what this issue is concerned with though, but is probably linked after all and affecting the look of the focus purple.

walkawayy:
In the current release, to add text box outlines, the `Text_AddOutline` command is called, and the `Output_DrawScreenBox` call hard codes the light and dark gold colors to draw for the two boxes. This is the only outline style and is like this Windows 98 menu with two outline colors:



For the PS1 UI style, I do not need this two color outline style. The PS1 has three outline types:
1. For the header an all black outline.
2. For the entire window border, a light grey to black gradient.
3. For the highlighted user selection, a dark grey to light grey centered gradient.




I can remove the hard coded colors from `Output_DrawScreenBox`, replace the original single `Output_DrawScreenBox` call with two calls and have it only draw box at a time with the four corner colors specified. This solves the original PC outline style and styles 1 and 2 for the PS1 UI. However, the problem is I also need style 3 for the PS1 which is a centered gradient. This requires only two colors, a corner and center color. This different outline style also means I have to specify the style somehow when `Text_AddOutline` is called.

Any ideas on how to cleanly go about this? I was thinking about having the config option set a theme and just use switch statements in `text.c` to pick the colors, but I still need to be able to specify different colors at each outline call because PS1 has three different styles. Adding an if/else for every outline call based on config option could work but seems like bloat to me. If you have any ideas let me know.

What I have so far is an array like the overlay bars:
```
static RGBA8888 m_MenuColorMap[] = {
    // Purple
    { 70, 30, 107, 225 },
    ...
}
```
The outline call accepts `MENU_COLOR tl, etc` corner colors and then within `text.c` they get translated to the RGBA with:
```
static RGBA8888 Text_GetMenuColor(MENU_COLOR color)
{
    return m_MenuColorMap[color];
}
```

rr-:
Maybe you could change the function to not take any colors, and instead take the type of the border instead (header, outer box, inner box, …)? Inside you could consult the config and decide how to draw each kind of border.

src/config.c
```diff
@@ -180,6 +186,7 @@ bool Config_ReadFromJSON(const char *cfg_data)
     READ_ENUM(airbar_color, BC_BLUE, m_BarColors);
     READ_ENUM(enemy_healthbar_color, BC_GREY, m_BarColors);
     READ_ENUM(screenshot_format, SCREENSHOT_FORMAT_JPEG, m_ScreenshotFormats);
+    READ_ENUM(ui.menu_style, UI_STYLE_PC, m_UIStyle);
```
rr-: Let's use plural, `m_UIStyles`

src/config.h
```diff
@@ -108,6 +113,7 @@ typedef struct {
     struct {
         double text_scale;
         double bar_scale;
+        int8_t menu_style;
```
rr-: `UI_STYLE menu_style`

src/game/inventory/inventory.c
```diff
@@ -274,7 +274,12 @@ static void Inv_DrawItem(INVENTORY_ITEM *inv_item)
                     break;
                 case SHAPE_BOX:
                     Output_DrawScreenBox(
-                        sx + spr->x, sy + spr->y, spr->param1, spr->param2);
+                        sx + spr->x, sy + spr->y, spr->param1, spr->param2,
+                        Text_GetMenuColor(MC_GOLD_LIGHT), OUTLINE_THICKNESS);
+                    Output_DrawScreenBox(
+                        sx + spr->x - 1, sy + spr->y - 1, spr->param1,
+                        spr->param2, Text_GetMenuColor(MC_GOLD_DARK),
+                        OUTLINE_THICKNESS);
```
rr-: Please inline this logic within `DrawScreenBox` and don't draw 2 rectangles here

src/game/text.c
```diff
@@ -406,14 +444,102 @@ static void Text_DrawText(TEXTSTRING *textstring)
...
+                    sx, sy, sh, sv, Text_GetMenuColor(MC_GREY_E),
+                    Text_GetMenuColor(MC_GREY_C));
+            }
+        } else {
+            Output_DrawScreenBox(
+                sx, sy, sh, sv, Text_GetMenuColor(MC_GOLD_LIGHT),
+                OUTLINE_THICKNESS);
+            Output_DrawScreenBox(
+                sx - 1, sy - 1, sh, sv, Text_GetMenuColor(MC_GOLD_DARK),
+                OUTLINE_THICKNESS);
```
rr-: Please inline this logic within `DrawScreenBox` and don't draw 2 rectangles here

src/game/text.c
```diff
@@ -406,14 +444,102 @@ static void Text_DrawText(TEXTSTRING *textstring)
...
+                Output_DrawGradientScreenBox(
+                    sx, sy, sh, sv, Text_GetMenuColor(MC_GREY_TL),
+                    Text_GetMenuColor(MC_GREY_TR),
+                    Text_GetMenuColor(MC_GREY_BL),
+                    Text_GetMenuColor(MC_GREY_BR), OUTLINE_THICKNESS);
+            } else if (textstring->outline.style == TS_REQUESTED) {
+                Output_DrawCentreGradientScreenBox(
+                    sx, sy, sh, sv, Text_GetMenuColor(MC_GREY_E),
+                    Text_GetMenuColor(MC_GREY_C));
+            }
```
rr-: Actually please move the logic to two new functions, `Text_DrawTextBackground` and `Text_DrawTextFrame`. Each of these can do the above logic inside.

src/game/text.c
```diff
@@ -406,14 +444,102 @@ static void Text_DrawText(TEXTSTRING *textstring)
...
+                    Text_GetMenuColor(MC_PURPLE_E),
+                    Text_GetMenuColor(MC_PURPLE_C));
+                // Bottom right
+                Output_DrawScreenGradientQuad(
+                    sx + sh, sy + sv, -sh / 2, -sv / 2,
+                    Text_GetMenuColor(MC_PURPLE_E),
+                    Text_GetMenuColor(MC_PURPLE_E),
+                    Text_GetMenuColor(MC_PURPLE_E),
+                    Text_GetMenuColor(MC_PURPLE_C));
+            }
```
rr-: Consider using `for` loops where possible.
```
struct {
    int x;
    int y;
    int w;
    int h;
} coords[4] = { { … }, { … }, { … }, { … } };

for (int i = 0; i < 4; i++)  {
    Output_DrawScreenGradientQuad(
        coords[i].x,
        coords[i].y,
        coords[i].w,
        coords[i].h,
        Text_GetMenuColor(MC_PURPLE_E),
        Text_GetMenuColor(MC_PURPLE_E),
        Text_GetMenuColor(MC_PURPLE_E),
        Text_GetMenuColor(MC_PURPLE_W)
    );
}
```

src/game/option/option_compass.c
```diff
@@ -88,10 +88,11 @@ static void Option_CompassInitText(void)
...
 
-    Text_AddBackground(m_Text[TEXT_TITLE_BORDER], row_width, y - top_y, 0, 0);
-    Text_AddOutline(m_Text[TEXT_TITLE_BORDER], 1);
-    Text_AddBackground(m_Text[TEXT_TITLE], row_width - 4, 0, 0, 0);
-    Text_AddOutline(m_Text[TEXT_TITLE], 1);
+    Text_AddBackground(
+        m_Text[TEXT_TITLE_BORDER], row_width, y - top_y, 0, 0, TS_BACKGROUND);
+    Text_AddOutline(m_Text[TEXT_TITLE_BORDER], true, TS_BACKGROUND);
+    Text_AddBackground(m_Text[TEXT_TITLE], row_width - 4, 0, 0, 0, TS_HEADING);
+    Text_AddOutline(m_Text[TEXT_TITLE], true, TS_HEADING);
```
rr-, line 95: Consider deleting `Text_AddOutline` and `Text_AddBackground` altogether and instead adding `Text_SetStyle(TEXTSTRING *, TEXT_STYLE)`. Then the renderer knows what to do.

src/game/output.c
```diff
@@ -714,22 +714,90 @@ void Output_DrawScreenLine(
...
+    // Left
+    Output_DrawGradientScreenLine(
+        sx - 1, h + sy + 2, 0, (-4 - h) / 2, edge, center);
+    Output_DrawGradientScreenLine(
+        sx - 1, h + sy + 2 + (-4 - h) / 2, 0, (-3 - h) / 2, center, edge);
+    // Thicken left
+    Output_DrawGradientScreenLine(
+        sx, h + sy + 1, 0, (-2 - h) / 2, edge, center);
+    Output_DrawGradientScreenLine(
+        sx, h + sy + 1 + (-1 - h) / 2, 0, (-2 - h) / 2, center, edge);
```
rr-: This belongs to `text.c`, because it's not really reusable outside of that module.  Check other functions, if they're only called from `text.c`, please move them there.

It could use some simplification too.

If `text.c` gets too busy with the drawing code, please extract all of it, including `TextDraw`, to `text_draw.c` and adjust `tools/sort_imports` so that it looks for `text.h` and not `text_draw.h`.

rr- (changes requested):

src/game/option/option_compass.c
walkawayy: Text_AddBackground needs the x, y, width, and height. So a `Text_SetStyle` would need to take this in. Not sure we want to do that?

src/game/output.c
walkawayy: I simplified the `Output_DrawCentreGradientScreenBox` function.

walkawayy:
I think the PS1 UI is in a good state now. I've been talking about scaling the UI / using quads instead of lines with @oziphantom. Seems like it will be a pretty substantial change and probably over my head, so he might do some investigation into it when he gets a chance.

src/game/inventory/inventory.c
rr-: This comment still stands.
To my knowledge this is dead code anyway so maybe let's avoid touching it altogether in this PR.

src/game/option/option_compass.c
rr-: You're right. Let's drop it.

oziphantom:
So, the issue is Line Drawing is always at the resolution of the framebuffer and it is 1 pixel thick. Since in this game we always render at screen resolution and not the "game resolution" as you decrease the "game resolution", either via actually setting the resolution or with UI Scaling, the actual resolution doesn't change and hence you get these gaps.  This also gives the comical effect of the hud getting thinner and thinner and thinner, 3 pixels was fine on 640x480, sure it was half what it was suppose to be, but oh well, 3 pixels on 4K is silly. 
While OpenGL in various forms has different limits, it does allow you to specify a line thickness, I believe it caps out at 20 from memory. However not all APIs will allow this so it may pose porting issues, such as to Vulkan or Metal. Given those don't actually really have fixed pipeline methods it makes it worse. 
So setting the line thickness based upon the desired resolution, scaled by the UI Scale and then scaled by the actual resolution is probably an encapsulation fail as now the UI Render has to know what the actual resolution is as well as the logical resolution. So setting the line width is probably not a workable solution. As we may also run out of line width, while a 4K screen is only 12 times as wide as 320, and 5K being 16 times somebody at some point probably sooner than you think will try to run it at 8K. 
Also how do you deal with say a 1.1 scale case, as the line width might need to go 2 wide for the dark and then 1 wide for the middle, or you could start to end up with gaps due to snapping. 
To which I propose replacing them with custom built quads to help cover the needed area. The Sizes of the windows are fixed for each window so making a nice static indexed buffer for the window would be fine with it being calculated once at runtime and then every time the user modifies the resolution/scaling. This however raises a discussion point of what does this look like? 
The original is 3 pixels, of dark, light, dark but then it was also 320x200. So are they trying to make a gradient of which 3 pixels and hence 2 colours is the best you could get and, due the 256 colour limit of the graphics mode, kept the lost palette entries low, and hence when we make a 4K version we have a nice smooth gradient afforded to us by the extra pixel fidelity and the fact we render into a 24bit buffer.  Or they didn't want a gradient and instead wanted dark line, light line, dark line and we should instead make a 3 rects that follow this colour in a large block format to keep the 3 solid colours look. 
Alternatively the lines to rects transform could be done inside the shader, but as we have shader support issues with people wanting to use 1.2 and 1.3 and 1.5 etc as well as having to add the concept to the render to set and control shader state, this  would be a significant undertaking in the current engine.
I feel that such a change and fix is well beyond the scope of this PR, as it will be a significant rework of the UI rendering pipeline and code.

src/game/text.h
```diff
@@ -6,7 +6,9 @@
 #include <stdint.h>
 
 #define TEXT_HEIGHT 11
+#define OUTLINE_THICKNESS 2
```
rr-: Missing `TEXT_` prefix.
Do we need to expose it to the public? I.e., can't `text.c` be the only module that knows and controls the frame thickness?

src/game/option/option_sound.c
```diff
@@ -77,17 +77,19 @@ void Option_Sound(INVENTORY_ITEM *inv_item)
     if (g_InputDB.forward && g_OptionSelected > TEXT_OPTION_MIN) {
         Text_RemoveOutline(m_Text[g_OptionSelected]);
         Text_RemoveBackground(m_Text[g_OptionSelected]);
-        Text_AddBackground(m_Text[--g_OptionSelected], 128, 0, 0, 0);
-        Text_AddOutline(m_Text[g_OptionSelected], 1);
+        Text_AddBackground(
+            m_Text[--g_OptionSelected], 128, 0, 0, 0, TS_REQUESTED);
```
rr-: I know this is unrelated to your PR but can we move this side effect (decrement) above?

src/game/option/option_sound.c
```diff
@@ -77,17 +77,19 @@ void Option_Sound(INVENTORY_ITEM *inv_item)
...
         Text_SetPos(m_Text[TEXT_RIGHT_ARROW], 40, 0);
     }
 
     if (g_InputDB.back && g_OptionSelected < TEXT_OPTION_MAX) {
         Text_RemoveOutline(m_Text[g_OptionSelected]);
         Text_RemoveBackground(m_Text[g_OptionSelected]);
-        Text_AddBackground(m_Text[++g_OptionSelected], 128, 0, 0, 0);
-        Text_AddOutline(m_Text[g_OptionSelected], 1);
+        Text_AddBackground(
+            m_Text[++g_OptionSelected], 128, 0, 0, 0, TS_REQUESTED);
```
rr-: Same here

src/global/types.h
```diff
@@ -1555,6 +1555,28 @@ typedef struct CREATURE_INFO {
...
+    MC_BROWN_E = 3,
+    MC_GREY_C = 4,
+    MC_GREY_E = 5,
+    MC_GREY_TL = 6,
+    MC_GREY_TR = 7,
+    MC_GREY_BL = 8,
+    MC_GREY_BR = 9,
+    MC_BLACK = 10,
+    MC_GOLD_LIGHT = 11,
+    MC_GOLD_DARK = 12,
```
rr-: Please remove the magic numbers and make it an auto-enum
Please add `MC_NUMBER_OF` (I'll get to why in a sec)

src/game/text.c
```diff
@@ -55,8 +56,123 @@ static int8_t m_TextASCIIMap[95] = {
     100 /*{*/, 101 /*|*/, 102 /*}*/, 67 /*~*/
 };
 
+static RGBA8888 m_MenuColorMap[] = {
```
rr-:
```
static RGBA8888 m_MenuColorMap[MC_NUMBER_OF] = {
    {70, 30, 107, 230},   // MC_PURPLE_C
    {70, 30, 107, 0},     // MC_PURPLE_E
    {91, 46, 9, 255},     // MC_BROWN_C
    {91, 46, 9, 0},       // MC_BROWN_E
    {197, 197, 197, 255}, // MC_GREY_C
    {45, 45, 45, 255},    // MC_GREY_E
    {96, 96, 96, 255},    // MC_GREY_TL
    {32, 32, 32, 255},    // MC_GREY_TR
    {63, 63, 63, 255},    // MC_GREY_BL
    {0, 0, 0, 255},       // MC_GREY_BR
    {0, 0, 0, 255},       // MC_BLACK
    {232, 192, 112, 255}, // MC_GOLD_LIGHT
    {140, 112, 56, 255},  // MC_GOLD_DARK
};
```

Similar to what we have for other such cases + it uses `MC_NUMBER_OF`

rr-:
>To which I propose replacing them with custom built quads to help cover the needed area. 

That's what I suggested as well

>when we make a 4K version we have a nice smooth gradient afforded to us by the extra pixel fidelity and the fact we render into a 24bit buffer

IDK about the gradient, but I don't see why we shouldn't try it. The corners seem like they were trying to go for a 3d look though.

>Or they didn't want a gradient and instead wanted dark line, light line, dark line and we should instead make a 3 rects that follow this colour in a large block format to keep the 3 solid colours look.

That's what I had originally on my mind but now I'd like to see what gold gradient borders would look like.

>Alternatively the lines to rects transform could be done inside the shader, but as we have shader support issues with people wanting to use 1.2 and 1.3 and 1.5 etc as well as having to add the concept to the render to set and control shader state, this would be a significant undertaking in the current engine.
I feel that such a change and fix is well beyond the scope of this PR, as it will be a significant rework of the UI rendering pipeline and code.

Definitely agree. I'd not go into that direction.

src/game/text.h
walkawayy: It's used in `Inv_DrawItem` for `case SHAPE_BOX:` (which you mentioned above) which I honestly am not sure what that function is for. Different cases draw different shapes.

src/global/types.h
walkawayy: Ya I definitely forgot to add `MC_NUMBER_OF`. Seems like there are mixed definitions of whether they list magic numbers or not, so I wasn't sure.

Example: `INPUT_ROLE_NUMBER_OF = 23,` and `MX_NUMBER_OF` for example. It's not a big deal though.

walkawayy:
Ok made the changes except I'm not sure what to do about `Inv_DrawItem`. Last tweak I made was to draw the heading for the final stats screen last. If it's drawn before the border, the brown gradient is covered by the dark background. Now it draws properly:

image

src/global/types.h
rr-: In general we should only declare numbers for things that are defined in .PHD etc. Many of the numbers are leftovers from the decompilation era. We haven't cleared types.h yet.

(For T3M it's a different story because our code operates with yet-to-be-decompiled functions that use immutable constants.)

rr- (approved):
Approving, but let's tackle this in 2.10 since this ought to go with fixing the hairline borders on low resolutions and big screens (as a separate ticket)

src/game/text.c
```diff
@@ -55,8 +56,111 @@ static int8_t m_TextASCIIMap[95] = {
...
+    TEXT_STYLE text_style);
+static void Text_DrawTextOutline(
+    UI_STYLE ui_style, int32_t sx, int32_t sy, int32_t w, int32_t h,
+    TEXT_STYLE text_style);
+
+static void Text_DrawTextBackground(
+    UI_STYLE ui_style, int32_t sx, int32_t sy, int32_t w, int32_t h,
+    TEXT_STYLE text_style)
+{
+    if (ui_style == UI_STYLE_PS1) {
```
rr-: Let's flip the condition to save indentation

if (pc) {
blabla;
return;
}
ps1

src/game/text.c
```diff
@@ -55,8 +56,111 @@ static int8_t m_TextASCIIMap[95] = {
...
+                Text_GetMenuColor(MC_GREY_C), TEXT_OUTLINE_THICKNESS);
+        }
+    } else {
+        Output_DrawScreenBox(
+            sx, sy, w, h, Text_GetMenuColor(MC_GOLD_LIGHT),
+            TEXT_OUTLINE_THICKNESS);
+        Output_DrawScreenBox(
+            sx - 1, sy - 1, w, h, Text_GetMenuColor(MC_GOLD_DARK),
+            TEXT_OUTLINE_THICKNESS);
+    }
```
rr-, line 160: Same as above

rr- (approved):
LGTM, the low resolution thing can get addressed separately

Issue #521, opened 2022-04-15
Finish naming cleanups


rr-:
I composed a list of remaining functions that need cleanup.

For the functions we tackled so far it was pretty clear where to put them. However the remaining ones give me some trouble. @walkawayy could you please offer your opinion? Maybe Choco could share her insights too.

(I know that while this task may seem pointless to some or even counterproductive, but these renames do help immensely with locating things in the long run and mental bookkeeping, so the decision whether we'll go through with them is not debatable.)

I want to remove `draw.c` because we have tons of specializations all over the place anyway, and the presence of `draw.c` is confusing because the actual drawing happens in `output.c`.

| #  | File     | Current signature                                                              | Suggested file                   | Suggested name                   |
| -- | ---      | ---                                                                            | ---                              | ---                              |
| 1  | `draw.h` | `void Draw_PrintRoomNumStack(void);`                                           | `game/room_draw.c`               | `Room_PrintDrawStack`            |
| 2  | `draw.h` | `void DrawRooms(int16_t current_room);`                                        | `game/room_draw.c`               | `Room_DrawAllRooms`              |
| 3  | `draw.h` | `void GetRoomBounds(int16_t room_num);`                                        | `game/room_draw.c`               | `Room_GetBounds`                 |
| 4  | `draw.h` | `int32_t SetRoomBounds(int16_t *objptr, int16_t room_num, ROOM_INFO *parent);` | `game/room_draw.c`               | `Room_SetBounds`                 |
| 5  | `draw.h` | `void PrintRooms(int16_t room_number);`                                        | `game/room_draw.c`               | `Room_DrawSingleRoom`            |
| 6  | `draw.h` | `void DrawEffect(int16_t fxnum);`                                              | `game/effects.c`                 | `Effect_Draw`                    |
| 7  | `draw.h` | `void DrawSpriteItem(ITEM_INFO *item);`                                        | `game/objects.c`                 | `Object_DrawSprite`              |
| 8  | `draw.h` | `void DrawDummyItem(ITEM_INFO *item);`                                         | `game/objects.c`                 | `Object_DrawDummy`               |
| 9  | `draw.h` | `void DrawPickupItem(ITEM_INFO *item);`                                        | `game/objects.c`                 | `Object_DrawPickup`              |
| 10 | `draw.h` | `void DrawAnimatingItem(ITEM_INFO *item);`                                     | `game/objects.c`                 | `Object_DrawAnimating`           |
[...]

PR #551, opened 2022-05-12
Creature renames

rr-:
#### Checklist

#### Description

An increment for #521


src/game/creature.h
```diff
@@ -0,0 +1,27 @@
+#pragma once
+
+#include "global/const.h"
+#include "global/types.h"
+
+#define CREATURE_SHOOT_RANGE SQUARE(WALL_L * 7) // = 51380224
+#define CREATURE_SHOT_DAMAGE 50
```
walkawayy: This used to be `PEOPLE_SHOT_DAMAGE`. I don't see this new or old name used anywhere. Is it used somewhere outside the scope of this PR? Just want to make sure we didn't miss something.

src/game/creature.c
```diff
@@ -0,0 +1,695 @@
...
+
+    GAME_VECTOR target;
+    target.x = g_LaraItem->pos.x;
+    target.y = g_LaraItem->pos.y - STEP_L * 3;
+    target.z = g_LaraItem->pos.z;
+
+    return LOS_Check(&start, &target);
+}
+
+bool Creature_ShootAtLara(
```
walkawayy, line 670: Not sure about this new name. The original name better implies to me that it's checking if Lara was hit by a shot.

src/game/creature.c
```diff
@@ -0,0 +1,695 @@
...
+    }
+
+    if (item->room_number != room_num) {
+        Item_NewRoom(item_num, room_num);
+    }
+
+    return true;
+}
+
+bool Creature_IsTargetable(ITEM_INFO *item, AI_INFO *info)
```
walkawayy: What about `Creature_IsLaraTargetable`? The current name to me makes me think the function checks if an enemy creature is targetable. When it's actually checking if the creature can target Lara.

src/game/creature.h
rr-: This is unused indeed, removed

src/game/creature.c
rr-: But it also has a side effect of producing SFX.

src/game/creature.c
rr-: Wouldn't it make more sense then for this to stay near Lara? Lara_IsTargetable?

src/game/creature.c
rr-: Updated as per our conversation on Discord

src/game/creature.c
rr-: Updated as per our conversation on Discord

walkawayy (approved):
LGTM.

Issue #278, opened 2021-12-17
Feature request: Controller support


ErickASJ1977:
Greetings.
Would it be possible to update the xinput control configuration?
TR3 sidesteps are only available with keyboard.
If I use any xinput controller, left sidesteps with the left trigger does not work.
And the rest of the buttons are set as in The Angel of Darkness, not  as in the original PSX version.

rr-:
Hey, currently it is not possible. This is similar to #268 and will be tackled in #279.

rr-:
https://wiki.libsdl.org/CategoryGameController

PR #561, opened 2022-05-15
Port input code to SDL

### LINKED ISSUE #279: Feature request: port input to SDL
The reason for SDL is to reduce Windows-only dependencies.
This ticket should also make it possible to edit the controller layout.

### LINKED ISSUE #513: Move game specific logic out of s/input
Game specific logic should be moved out of s/input. It should only deal with platform specific logic if possible.

rr-:
#### Checklist

#### Description

- Switches DInput to SDL
- :warning: Removes Xbox 360 support (will need to readd this as SDL controllers #278  )
- Removes the last DirectX dependency (we still have Windows dependencies in the OpenGL context attaching)


src/global/types.h
```diff
@@ -1934,68 +1934,79 @@ typedef struct PICTURE {
...
+    INPUT_ROLE_LEVEL_SKIP_CHEAT = 15,
+    INPUT_ROLE_TURBO_CHEAT = 16,
+    INPUT_ROLE_PAUSE = 17,
+    INPUT_ROLE_CAMERA_UP = 18,
+    INPUT_ROLE_CAMERA_DOWN = 19,
+    INPUT_ROLE_CAMERA_LEFT = 20,
+    INPUT_ROLE_CAMERA_RIGHT = 21,
+    INPUT_ROLE_CAMERA_RESET = 22,
+    INPUT_ROLE_NUMBER_OF = 23,
+} INPUT_ROLE;
```
walkawayy, line 2004: Why the name change to `role` from `key`? I tried searching through some of the SDL docs and didn't see that name used unless I missed it. The name change seems strange to me since I've never heard a role to describe a key.

src/game/option/option_control.c
```diff
@@ -432,18 +398,16 @@ void Option_Control(INVENTORY_ITEM *inv_item)
         break;
 
     case 1:
-        if (!g_Input.select) {
+        if (!g_Input.any) {
```
walkawayy, line 401: Not sure what `m_KeyMode` is checking (it can be mode 0,1,2,3, but there's no enums for what these modes mean) but is this change going to break anything?

src/specific/s_input.c
```diff
@@ -1,732 +1,443 @@
...
+        case SDL_SCANCODE_APP1:               return "APP1";
+        case SDL_SCANCODE_APP2:               return "APP2";
+        case SDL_SCANCODE_AUDIOREWIND:        return "RWND";
+        case SDL_SCANCODE_AUDIOFASTFORWARD:   return "FF";
     }
-    S_Shell_SpinMessageLoop();
-
-    return true;
+    // clang-format on
+    return "????";
```
walkawayy, line 321: How does this affect the program if it falls through to this return? (I'm guessing it shouldn't on normal operation but just in case.)

walkawayy (commented):
Looks good for the most part, and my keyboard testing went well! Left a few comments asking questions I had.

src/global/types.h
rr-: We deal with multiple concepts:

- key: a physical button on a keyboard. F6, spacebar, page down.
- key code: usually means a scan code.
- scan code: a numeric value assigned by the operating system to a key. 65, 128, 211.
- key role: what the given button does within the game. jump, equip pistols, run menu…

The term "key" to me was utterly meaningless. When I saw `INPUT_KEY_JUMP` I would think that this is the scancode that is assigned to the jump button whereas it's only used as a numeric label for the game code to be able to offer custom bindings.

src/game/option/option_control.c
rr-: It's used to signify different stages of key rebinding (waiting to press select, waiting till the user releases select, waiting for a new binding, waiting for the key release, going back to the first step).

src/specific/s_input.c
rr-: It will display to the user that they bound a key to a "????" button in the controls menu, but it will still work fine, in theory. I'm saying in theory because I covered almost all SDL defines and I can't think of a button outside of this list, maybe except the Fn key which cannot be bound.

walkawayy: Ok I figured. Makes sense.

src/global/types.h
walkawayy: Ok make sense. It's like the input's action (aka role).

walkawayy (approved):

Issue #564, opened 2022-05-18
Stop unmapped keybinds having an effect on the game


Richard-L:
>>Generally didn't we have a discussion somewhere on the default buttons still having an effect unintentionally, even if the user has chosen different keybinds?
>
> This is a separate issue and needs its own discussion. Can you open a ticket for this?

_Originally posted by @rr- in https://github.com/rr-/Tomb1Main/issues/479#issuecomment-1129946114_

Keys like `ALT`, which are the default keybind for jumps, will keep having an effect on the game, even if the user has chosen "User Keys" under "Controls", and has not mapped the `ALT` key for anything. It would be nice if only actually mapped keys had an effect on the game.

This is especially noticeable when using the `ALT`+`ENTER` shortcut to toggle fullscreen mode, where Lara will definitely jump no matter what, often messing with the state you wanted to leave the game at. A workaround for now is to just pause the game prior to toggling fullscreen.

Considerations must be taken so the user doesn't softlock their control scheme which is why default keys always work currently.

rr-:
My concern here is that if the player chooses to do some silly thing such as rename all the arrows to the same key, they won't be able to revert without closing the game and editing the JSON settings file by hand.

My suggestion - when checking if the player has pressed a button with the given role (e.g. the jump button):
```
if the layout is default:
    check the default layout key
else if the corresponding user layout key is being used to map only one action (e.g. it doesn't flash in the controls menu):
    check the user layout key
else:
    check the default layout key
```

That way if the player maps up and down keys to the same button, that mapping will have no effect and they'll have to use arrows.

My other concern is that some players may want to remap Lara's movement to WSAD but still navigate the menus with the arrow keys.

Richard-L:
I would not allow mapping arrow keys to the same key. It just bricks them, as you say.

I also wouldn't separate inventory from in-game direction keys. OpenLara does this (ie in inventory you always use the standard arrow keys), and it's really quite frustrating imo. If they map to WASD then that's what the game should know as directions. Keeps things simple.

walkawayy:
I agree with Posix. Just don't let the user map two inputs to the same key and that should prevent issues I think? Also in the worst case scenario, the runtime config file can be deleted to remove the custom keybinds.

walkawayy:
Potential patches from rr:
```
diff --git a/src/specific/s_input.c b/src/specific/s_input.c
index 83c9828f..1812fcd7 100644
--- a/src/specific/s_input.c
[...]

Issue #545, opened 2022-05-08
Feature: PS1 style "Select" and "Go Back" indicators


Richard-L:
I've been debating for a while whether I should post this or not, but since we've been doing so much work with #517, I felt prompted again now. It's another small thing, but in fact good UX practice that for some reason didn't make it into TombATI: To indicate to the user what directions they can take and how, at hand of the `Select` and `Go Back` fields.

We've already improved this UX aspect quite a bit with #420 (in my opinion), and this would be the last "feature" the PS1 version has above TR1M in terms of UI behaviour.

Notice how `Select` is always showing. `Go Back` consequently only appears after an item has been selected.

https://user-images.githubusercontent.com/4934209/167318361-8caf5552-2a63-4903-a41e-f0419ae653d2.mp4

You also have it in **Select Level**, where going back simply returns to the "Load Game" page.

https://user-images.githubusercontent.com/4934209/167318643-eb3bfecd-05e7-409a-940b-257d18374c28.mp4

You then of course have it in the inventory:

https://user-images.githubusercontent.com/4934209/167318680-4891d57f-63fe-4124-9f40-bd5489735d5c.mp4

So what would we show, considering there are no nice little icons for the PS1 controller pad like there are here? I would suggest it should show whatever keybind the player has set for "Action" and "Inventory"/"Esc". So for example for the default layout, `CTRL` would show next to "Select".

FYI on the PS1's default layout the Circle is the roll button.

Richard-L:
Wanted to add some more info to this after [@annlyt](https://github.com/annlyt) posted a screen to show TR2M did in fact implement this:

![image](https://user-images.githubusercontent.com/4934209/167394482-328321dd-447a-4902-bc78-639805c7611c.png)

^He's using an XBox controller, which Arsunt chose to support and apparently created icon assets for, as well as custom textures in his setup.

I then spoke to Arsunt briefly who said in the current implementation, the `Select` and `Go Back` indicators only appear when a control pad is connected. He did seem to agree though that it would make sense to also enable the feature for keyboard users.

He quoted these lines in TR2M's code to look into for reference how he did it:

> Check for joystick presence: https://github.com/Arsunt/TR2Main/blob/5bec0d77667276d37eef4e4e23d295a06ae2a596/specific/option.cpp#L326
> 
> Further:
> https://github.com/Arsunt/TR2Main/blob/5bec0d77667276d37eef4e4e23d295a06ae2a596/specific/option.cpp#L287
[...]

PR #568, opened 2022-05-26
s/input: add SDL controller support

### LINKED ISSUE #278: Feature request: Controller support

rr-:
Resolves #278

#### Checklist

#### Description

I kept Ozi's decision to switch the action and the jump button (I'm a bit conflicted whether to keep it).
I switched Draw Guns with Look to match PS1 controller layout.
I changed D-pad's function to movement. This was dictated by seeing people map D-pad for movement on streams, and Lara herself instructing the player to use the D-pad to move her around.

Tested on my DualShock 4 controller. I have no other devices. I am not a controller player so I'm clueless whether these feel OK.

I want the game to offer sane defaults. Customization comes way later (or perhaps never, or via third party software).

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,20 +232,23 @@ Not all options are turned on by default. Refer to `Tomb1Main.json5` for details
 - added save game crystals game mode (enabled via gameflow)
 - added pause screen
 - added movable camera on W,A,S,D
-- added Xbox One Controller support
-    - Per Axis Dead Zone
-    - Left Stick = movement
-    - A = Jump/Select
-    - B = Roll/Deselect
-    - X = Action/Select
-    - Y = Look/Select
-    - LB = Walk
-    - RB = Draw Weapons
-    - Dpad Up = Draw Weapons
-    - Back = Option
-    - Start = Pause
-    - Right Stick = Camera Movement
-    - R3 = Reset Camera
+- added controller support
+    | Xbox button | DualShock button   | Action          |
+    | ----------- | ------------------ | --------------- |
+    | Left Stick  | Left stick         | Movement        |
+    | A           | X                  | Action/select   |
+    | B           | Circle             | Roll/deselect   |
+    | X           | Square             | Jump            |
+    | Y           | Triangle           | Draw weapons    |
+    | LB          | L1                 | Look            |
+    | RB          | R1                 | Walk            |
+    | LT          | L2                 | Sidestep left   |
+    | RT          | R2                 | Sidestep right  |
+    | D-pad       | D-pad              | Movement        |
+    | Back        | Share/Touchpad     | Inventory       |
+    | Start       | Options            | Pause           |
+    | Right Stick | Right stick        | Camera movement |
+    | R3          | R3                 | Reset camera    |
 - added rounded shadows (instead of the default octagon)
 - added per-level customizable water color (with customizable blue component)
 - added per-level customizable fog distance
--- a/src/specific/s_input.c
+++ b/src/specific/s_input.c
@@ -1,7 +1,10 @@
 #include "specific/s_input.h"
 
 #include "game/input.h"
+#include "log.h"
 
+#include <SDL2/SDL.h>
+#include <SDL2/SDL_gamecontroller.h>
 #include <SDL2/SDL_keyboard.h>
 #include <SDL2/SDL_scancode.h>
 #include <SDL2/SDL_stdinc.h>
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -232,20 +232,21 @@ Not all options are turned on by default. Refer to `Tomb1Main.json5` for details
...
+    | B           | Circle             | Roll/deselect   |
+    | X           | Square             | Action/select   |
+    | Y           | Triangle           | Look/select     |
+    | RB          | R1                 | Walk            |
+    | LB          | L1                 | Draw weapons    |
+    | D-pad       | D-pad              | Movement        |
+    | Back        | Share/Touchpad     | Inventory       |
+    | Start       | Options            | Pause           |
+    | Right Stick | Right stick        | Camera movement |
+    | R3          | R3                 | Reset camera    |
```
walkawayy, line 251: Should we match the PS1 controller layout? (I didn't update the Xbox column)

    | Xbox button | DualShock button   | Action          |
    | ----------- | ------------------ | --------------- |
    | Left Stick  | Left stick         | Movement        |
    | A           | Square             | Jump            |
    | B           | Circle             | Roll/deselect   |
    | X           | X                  | Action/select   |
    | LB          | Triangle           | Draw weapons    |
    | Y           | L1                 | Look            |
    | LT          | L2                 | Sidestep left   |
    | RB          | R1                 | Walk            |
    | RT          | R2                 | Sidestep right  |
    | D-pad       | D-pad              | Movement        |
    | Back        | Share/Touchpad     | Inventory       |
    | Start       | Options            | Pause           |
    | Right Stick | Right stick        | Camera movement |
    | R3          | R3                 | Reset camera    |

This could also be helped with #564 to allow proper custom control schemes.

rr-: Can you also provide which buttons act like select and which deselect?

walkawayy: I updated the table with sidesteps too. But after checking out #545 I did a little digging.

TLDR: The select / deselect button has changed a bunch depending on console, region, and game. Since the PS5 release, X is select and circle is deselect in all regions. [(source)](https://www.eurogamer.net/in-a-historic-move-playstation-5-swaps-x-and-circle-button-use-in-japan) I vote we go with that. Custom controls potentially in the future will allow that to be changed.

Long story:
X was always select (for non Asian regions). However, for deselect, it seems different buttons were used. Back in the PS1 days, I remember triangle being the primary deselect button for games. For TR1 though, they used square as deselect for some reason. TR2+ changed deselect to be triangle which matched most games for the PS1-PS2 era.

Since the PS3, X has been select and circle has been deselect. In Asian regions this was reversed, and X was deselect and circle was select. Since the PS5, there is now an all region standard: X is select and circle is deselect.

oziphantom (approved):
Looks fine, hard to tell exactly how SDL will handle all the mappings across different controllers, the documentation is vague sadly.

oziphantom:
as for switching the xbox buttons its 50/50. I'm not a person who played the original game before T1M, so I have no connection to the "playstation layout", so I made the controls in a sense that makes sense for an "Xbox controller" with xbox conventions. Rather than "I have an Xbox but I think I'm holding a PS1 controller". There is no real answer, but I guess this way it gives people a choice of "classic" and "more modern".

rr-:
Let's go with PS1 era layout walkawayy composed, release as 2.9 to get the universal controller support out there, and then possibly follow up with some sort of customization if enough people want it.

walkawayy:
Ok that sounds like a plan. Hopefully at some point we can get custom controls working properly. Then people are free to customize to their liking. We tested on PS4 and PS5 controllers, but I don't have an Xbox controller to test. When we release 2.9, I guess we will hear some feedback if Xbox controllers have problems lol.

src/specific/s_input.c
```diff
@@ -351,61 +357,134 @@ static bool S_Input_Key(INPUT_ROLE role)
...
+        linput.action       |= S_Input_JoyBtn(SDL_CONTROLLER_BUTTON_A);
+        linput.select       |= S_Input_JoyBtn(SDL_CONTROLLER_BUTTON_A);
+        linput.roll         |= S_Input_JoyBtn(SDL_CONTROLLER_BUTTON_B);
+        linput.deselect     |= S_Input_JoyBtn(SDL_CONTROLLER_BUTTON_B);
+        linput.jump         |= S_Input_JoyBtn(SDL_CONTROLLER_BUTTON_X);
+        linput.draw         |= S_Input_JoyBtn(SDL_CONTROLLER_BUTTON_Y);
+        linput.look         |= S_Input_JoyBtn(SDL_CONTROLLER_BUTTON_LEFTSHOULDER);
+        linput.slow         |= S_Input_JoyBtn(SDL_CONTROLLER_BUTTON_RIGHTSHOULDER);
+        linput.option       |= S_Input_JoyBtn(SDL_CONTROLLER_BUTTON_BACK);
+        linput.deselect     |= S_Input_JoyBtn(SDL_CONTROLLER_BUTTON_TOUCHPAD);
```
walkawayy, line 483: Is deselect supposed to be used twice here? Set to TOUCHPAD and B button.

rr-: Hmm I kept it because I felt like it but I can remove it.

walkawayy: Oh that's fine. Just wanted to make sure it wasn't an oversight. Makes sense to be able to close the inventory in game with the same button you opened it with.

rr-: On that note maybe I should add this behavior to the pause and the alt inventory button too?

walkawayy: Sure. Makes sense.

rr-: Addressed

walkawayy (approved):
LGTM

Issue #557, opened 2022-05-14
Doesn't revert to pistols at new level start


MatthTheGeek:
OG always reverted to pistols as the selected weapon when starting a new level. Tomb1Main keeps the last weapon equipped.

This is a minor issue but does cause a balance issue for speedrun/races between OG / TombATI and Tomb1Main. It is not bad for QoL for casual play tho, so a config toggle might be best.

Richard-L:
Could you post a screenshot of the title screen with your version in the bottom right corner?

Richard-L:
Hmm I was sure we'd talked about this one before. Anyhow, it seems there is indeed currently no setting to maintain OG behaviour.

PS1

https://user-images.githubusercontent.com/4934209/168423845-82d47a40-35c3-4b9a-aa4b-c1cb66793de8.mp4

TombATI

https://user-images.githubusercontent.com/4934209/168423850-0383f70b-b6e9-4bf1-abd4-545615c08711.mp4

TR1M

https://user-images.githubusercontent.com/4934209/168423855-f9389c4c-90cc-4d69-8760-244724d53368.mp4

rr-:
Is there anyone actually using Tomb1Main for speedrunning / races, or is this ticket strictly hypothetical?

MatthTheGeek:
In theory I would, I'm in a community that sometimes does TR classic races, we mostly use tombATI and I've quietly been promoting Tomb1Main there. 

We're mostly casual so that issue is not considered a big deal, but I thought it would be best to bring it up anyway.

PR #586, opened 2022-06-20
Pistol start

fluxtom:
#### Checklist

#### Description

Relevant to #557. `g_GameInfo.current[level_num]`'s fields `request_gun_type` & `gun_type` were being used to set the corresponding fields on `g_Lara`; this has been changed so that `LGT_PISTOLS` is used instead.

I hope this is the correct approach, I couldn't find any immediate issues with it. I also added it as a config option, not too sure if this is appropriate either.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 ## [Unreleased](https://github.com/rr-/Tomb1Main/compare/stable...develop)
-- ...
+- added the option to make Lara revert to pistols on new level start
 
 ## [2.9.1](https://github.com/rr-/Tomb1Main/compare/2.9...2.9.1) - 2022-06-03
 - fixed crash on centaur hatch (#579, regression from 2.9)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 ## [Unreleased](https://github.com/rr-/Tomb1Main/compare/stable...develop)
-- added the option to make Lara revert to pistols on new level start
+- added the option to make Lara revert to pistols on new level start (#557)
 
 ## [2.9.1](https://github.com/rr-/Tomb1Main/compare/2.9...2.9.1) - 2022-06-03
 - fixed crash on centaur hatch (#579, regression from 2.9)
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,5 @@
 ## [Unreleased](https://github.com/rr-/Tomb1Main/compare/stable...develop)
-- ...
+- added the option to make lara revert to pistols on new level start
```
rr-: I'd capitalize Lara

rr- (approved):
LGTM

rr-:
(When merging please squash and apply proper commit message conventions)

walkawayy:
Looks good! Have you tested equipping a shotgun, saving, and loading the savegame with this option set to true? Just want to make sure it doesn't change her gun to pistols in every scenario. It should only switch to pistols when beginning a new level.

fluxtom:
> Looks good! Have you tested equipping a shotgun, saving, and loading the savegame with this option set to true? Just want to make sure it doesn't change her gun to pistols in every scenario. It should only switch to pistols when beginning a new level.

Yeah that's been tested and works.

> (When merging please squash and apply proper commit message conventions)

I have royally messed up the commits on this branch doing trying to do this. If you prefer, I can close this PR and open another with nicer commits...

Also with the commit message conventions, I see them now in the commit history. I had only checked the CONTRIBUTING.md file which doesn't talk about them, as far as I can see. Maybe it would be worth adding a small note about them for people like me that don't check the commit history?

rr-:
>I have royally messed up the commits on this branch doing trying to do this. If you prefer, I can close this PR and open another with nicer commits...

No, that actually would be worse, it's fine as it is. The maintainers have the option to squash the commits when merging anyway.

>Also with the commit message conventions, I see them now in the commit history. I had only checked the CONTRIBUTING.md file which doesn't talk about them, as far as I can see. Maybe it would be worth adding a small note about them for people like me that don't check the commit history?

Great point, we definitely should have added it. I'll see to it

@walkawayy could you do some light testing on this? Thanks

walkawayy:
I was able to test and it looks good to me. I tried finishing the level with different weapons, save + load, restart level, and select level. All seem to work fine. Approved.

walkawayy (approved):

PR #588, opened 2022-06-29
passport: add the story so far option to view cutscenes and FMVs

### LINKED ISSUE #201: Feature request: Story so far...
A feature added to PS1 TR2 where it will simply play back all FMVs and cutscenes so far unlocked in concession.
SLUS-00437_2021-11-09_00-51-03

Alternatively FMVs and cutscenes could simply be added to the "Select Level" #197 table perhaps.

walkawayy:
Resolves #201.

#### Checklist

#### Description
Added the "story so far..." option in the select level menu to view cutscenes and FMVs from the current savegame.

Video: https://streamable.com/4zcrye
...


rr-:
The placement feels very much like an afterthought but I can't think of a better place

rr- (approved):
Code looks OK
Purists are gonna have many comments regarding the ui though

walkawayy:
None of the PC versions had this feature since the PS1 placed it at the bottom of the Load Game list and only supported one savegame. This goes back to the new menu for Select Level debate since we support multiple savegames. I'm not sure where else we would put it, but here's a few other suggestions:
1. Make story viewing global by having only one option for it, and view story scenes up to the furthest savegame. But I'm not sure what menu it would belong to in that case.
2. Move Story So Far to the top of the Select Level menu instead of the bottom.
3. Leave it at the bottom of Select Level, but remove the Level Locked entries so it's under the current level.
4. Leave as is.

LMK what you would prefer or if you have another idea. I know I said I wasn't working on anything, but I forgot I started this like two months ago, and I just had to do some code cleanup.

walkawayy:
BTW here is the location in TR2 taken from the issue page. Their Load Game menu was basically our Select Level menu.

image

rr-:
IDK, personally I have no opinion. Let's give it some time for people to speak up here, and if there's no feedback, let's merge.

Richard-L:
> 3. Leave it at the bottom of Select Level, but remove the Level Locked entries so it's under the current level.

^ since this mimics console behaviour, that's what I would do. Will likely also sit best with "purists" since that's what they'll know.

walkawayy:
> > 3. Leave it at the bottom of Select Level, but remove the Level Locked entries so it's under the current level.
> 
> ^ since this mimics console behaviour, that's what I would do. Will likely also sit best with "purists" since that's what they'll know.

So I'll go with this solution to hide locked levels? Just want to confirm. Maybe I should solicit some more Discord feedback too. Update: added a Discord vote to get some more opinions.

walkawayy:
I think this is done. 100% of Discord voted to remove the `Level Locked` entries. So I removed that and made `Story so far...` the bottom option. I had to adjust some logic for the requestor selection. Then, I fixed a bug where cutscenes/FMVs that played at the end of a level were playing. For example, story from a Great Pyramid played the final cutscene. Not it plays up until a `start_game` is read from the gameflow.

Here's a pic of the new menu:
image

src/game/gameflow.c
```diff
@@ -1299,3 +1299,114 @@ GameFlow_InterpretSequence(int32_t level_num, GAMEFLOW_LEVEL_TYPE level_type)
...
+            break;
+        case GFS_PLAY_SYNCED_AUDIO:
+            Music_Play((int32_t)seq->data);
+            break;
+
+        case GFS_MESH_SWAP: {
+            GAMEFLOW_MESH_SWAP_DATA *swap_data = seq->data;
+            int16_t *temp;
+
+            temp = g_Meshes
```
rr-: We can inline the declaration and definition
`int16_t *temp = blabla;`

src/game/gameflow.c
```diff
@@ -1299,3 +1299,114 @@ GameFlow_InterpretSequence(int32_t level_num, GAMEFLOW_LEVEL_TYPE level_type)
 
     return ret;
 }
+
+GAMEFLOW_OPTION
+GameFlow_StorySoFar(
+    int32_t level_num, GAMEFLOW_LEVEL_TYPE level_type, int32_t savegame_level)
```
rr-: I'm pretty sure `level_type` parameter is not needed here, there's no way to run this function on a saved game

src/game/gameflow.c
```diff
@@ -1299,3 +1299,114 @@ GameFlow_InterpretSequence(int32_t level_num, GAMEFLOW_LEVEL_TYPE level_type)
...
+
+        case GFS_START_GAME:
+            if (level_num == savegame_level) {
+                return GF_EXIT_TO_TITLE;
+            }
+            seq++;
+            continue;
+
+        case GFS_START_CINE:
+            if (level_type != GFL_SAVED) {
```
rr-: Which means these checks are not necessary either

src/game/gameflow.c
```diff
@@ -1299,3 +1299,114 @@ GameFlow_InterpretSequence(int32_t level_num, GAMEFLOW_LEVEL_TYPE level_type)
...
+        case GFS_FIX_PYRAMID_SECRET_TRIGGER:
+            seq++;
+            continue;
+
+        case GFS_START_GAME:
+            if (level_num == savegame_level) {
+                return GF_EXIT_TO_TITLE;
+            }
+            seq++;
+            continue;
```
rr-: What's this check about? Why not just skip it?

src/game/gameflow.c
```diff
@@ -1299,3 +1299,114 @@ GameFlow_InterpretSequence(int32_t level_num, GAMEFLOW_LEVEL_TYPE level_type)
 
     return ret;
 }
+
+GAMEFLOW_OPTION
+GameFlow_StorySoFar(
```
rr-: This function return value is not used. Let's make it a void.

src/game/gameflow.c
```diff
@@ -1299,3 +1299,114 @@ GameFlow_InterpretSequence(int32_t level_num, GAMEFLOW_LEVEL_TYPE level_type)
...
+            break;
+
+        case GFS_PLAY_FMV:
+            if (level_type != GFL_SAVED) {
+                FMV_Play((char *)seq->data);
+            }
+            break;
+
+        case GFS_EXIT_TO_TITLE:
+            return GF_EXIT_TO_TITLE;
```
rr-, line 1348: Since the return value is not used let's make these `break;`.

src/game/gameflow.c
```diff
@@ -1299,3 +1299,114 @@ GameFlow_InterpretSequence(int32_t level_num, GAMEFLOW_LEVEL_TYPE level_type)
...
+        case GFS_STOP_GAME:
+        case GFS_LEVEL_STATS:
+        case GFS_TOTAL_STATS:
+        case GFS_DISPLAY_PICTURE:
+        case GFS_GIVE_ITEM:
+        case GFS_REMOVE_GUNS:
+        case GFS_REMOVE_SCIONS:
+        case GFS_FIX_PYRAMID_SECRET_TRIGGER:
+            seq++;
+            continue;
```
rr-: Since there's `seq++` after switch, this can be simplified to `break;` (here and in all other switch cases).

src/game/inventory/inventory.c
```diff
@@ -935,6 +945,11 @@ int32_t Inv_Display(int inv_mode)
                 && g_GameInfo.passport_mode == PASSPORT_MODE_SELECT_LEVEL) {
                 // page 1: select level
                 return GF_SELECT_GAME | g_GameInfo.select_level_num;
+            } else if (
+                g_GameInfo.passport_page == PASSPORT_PAGE_1
+                && g_GameInfo.passport_mode == PASSPORT_MODE_STORY_SO_FAR) {
+                // page 1: story so far
+                return GF_STORY_SO_FAR | g_GameInfo.current_save_slot;
```
rr-, line 952: Why is our code so shitty :(

src/game/savegame/savegame.c
```diff
@@ -614,3 +616,30 @@ bool Savegame_RestartAvailable(int32_t slot_num)
...
+
+    int32_t gf_option = GF_START_GAME | g_GameFlow.first_level_num;
+
+    bool loop_continue = true;
+    while (loop_continue) {
+        int32_t gf_direction = gf_option & ~((1 << 6) - 1);
+        int32_t gf_param = gf_option & ((1 << 6) - 1);
+
+        gf_option =
+            GameFlow_StorySoFar(gf_param, GFL_NORMAL, savegame_info->level_num);
```
rr-: Unnecessary assignment

src/game/savegame/savegame.c
```diff
@@ -614,3 +616,30 @@ bool Savegame_RestartAvailable(int32_t slot_num)
...
     return savegame_info->features.restart;
 }
+
+GAMEFLOW_OPTION Savegame_PlayAvailableStory(int32_t slot_num)
+{
+    SAVEGAME_INFO *savegame_info = &m_SavegameInfo[slot_num];
+
+    int32_t gf_option = GF_START_GAME | g_GameFlow.first_level_num;
+
+    bool loop_continue = true;
```
rr-, line 626: I have a feeling that this entire function body can be replaced with
```
SAVEGAME_INFO *savegame_info = &m_SaveGameInfo[slot_num];
GameFlow_StorySoFar(g_GameFlow.first_level_num, GFL_NORMAL, savegame_info->level_num);
return GF_EXIT_TO_TITLE;
```

rr- (changes requested):

src/game/gameflow.c
walkawayy: I added this check:
```
            if (level_num == savegame_level) {
                return GF_EXIT_TO_TITLE;
            }
```
because levels like the Great Pyramid were playing the ending FMV when doing the story so far. This check makes it so the function reads the current level but stops at the point of `start_game`. So the story so far will play any cutscenes/FMVs that occur before the level (like City of Khamoon has a pre level FMV), but will not play anything that comes after `start_game`.

src/game/gameflow.c
walkawayy: The return value `gf_option` is used in `Savegame_PlayAvailableStory` for `gf_direction ` and `gf_param `:
```
    while (loop_continue) {
        int32_t gf_direction = gf_option & ~((1 << 6) - 1);
        int32_t gf_param = gf_option & ((1 << 6) - 1);

        gf_option = GameFlow_StorySoFar(gf_param, savegame_info->level_num);
        ...
```

src/game/inventory/inventory.c
walkawayy: Ya this inventory code is kind of a mess but I did not want to refactor it at the same time as this feature lol. It works for now.

src/game/savegame/savegame.c
walkawayy: See above about return value. I think it is necessary.

src/game/savegame/savegame.c
walkawayy: No because of things like:
```
        case GFS_EXIT_TO_LEVEL:
            return GF_START_GAME | ((int32_t)seq->data & ((1 << 6) - **1));**
```

Using your code snippet, only the Caves FMV plays, then it returns to title.

rr-: OK fair, I understand it now. However `loop_continue` is not needed.

walkawayy: Yes I'll just use `while(1)`. I was trying to keep it close to `Shell_Main` but that does:
```
        case GF_EXIT_GAME:
            loop_continue = false;
            break;
```
The `loop_continue` there is just visual anyway.

rr- (approved):

PR #592, opened 2022-07-19
add new readme showcase images

walkawayy:
#### Checklist

#### Description
Adds new showcase images and a .webp for a draw distance comparison. LMK if you like the shots or not and feel free to resize/rename/whatever.
...


rr- (commented):
Thanks. Let's make the enemy health bars twice as big to make them stand out more and also demonstrate UI scaling while at it.

Richard-L:
fov-anim
Tried myself at a FOV GIF. 5.12mb. I never quite trust Photoshop's GIF file size optimisation abilities.

Like this I would see it for draw distance too.

rr-:
Yeah we'd be using a WEBP file anyway rather than a GIF due to precisely what you said about the filesizes.

rr- (approved):

PR #593, opened 2022-07-19
game: add an option to enable inventory buffering

### LINKED ISSUE #591: Allow speedrun "buffering" in TR1M
Explained by @Rapora9 on Discord:
> Btw, if tomb1main wants to appeal for speedrunners, it should support "inventory buffering". This means that if you
> 
> - open the inventory
> - press a button to close the inventory
> - hold the button to open the inventory
> -> inventory should open immediately again after closing (2 frames from that iirc)
> 
> Currently if you hold the button to open the inventory again, it won't open. I guess this is because of some checks to make things not skip too quickly or something? You can try tombati to see how it works there.
> 
Footi the other day complained it wasn't working in TR1M: https://www.twitch.tv/videos/1531467943?t=03h48m18s

Could be worth considering.

Is it at all to do with the fades?

walkawayy:
Resolves #591.

#### Checklist

#### Description
Adds the option to enable inventory buffering.

https://streamable.com/2pibcc
...


rr- (commented):
Can we rename it to `enable_buffering` and also restore f2/f3/f4 buffering too while we're at it?

walkawayy:
Good idea. I'm trying to track down why the F2-F4 keys used to stutter on press. Tried searching repo history for all words forms of stutter, lag, freeze, slow, etc to no luck.

EDIT: Troye also mentioned speedrunners will never migrate so maybe this isn't worth it.

rr-:
>How to find the code

Just search for F2-F4. The commit that patched this is 4692d9b1.

>Doubts whether it'll be ever used

I think it's best to ask Footi directly as he's the biggest in the scene. If he'd run should we be 100% compatible, I'm OK with all backwards compatibility options. If not, let's just forget about it.

Some rationale why T1M should be allowed: people are now using TombATI, and TombATI is a fanmade mod that in itself  introduces F2-F4 buffering as well as the end of level Action key freeze. Yet, the scene is OK with these changes from the OG. Under such circumstances I don't know why it would be  wrong to use an open source variant that maintains these things.

Richard-L:
I do agree.

Not sure I would ask Footi. If anything maybe Rapora could speak to him in Finnish.

I'd ask Tombeaux first. He's very active in TR1 currently and has a better attitude.

Another one where I'd expect at least some constructive dialogue would be Eycore.

Btw I wouldn't make this buffering optional. I would just make it behave like TombATI by default. Don't see why it would need to be configurable, other than runs wanting to forbid buffering at some point in the future.

rr- (approved):
The changes LGTM.
F2-F4 should not by default do the thing they used to.
But I don't see why we should not let buffer with the inventory by default, so if that can be done, it'd be neat.

That being said I don't think it makes sense to land this PR until there's interest from actual speedrunners. I'd appreciate if someone could get their collective opinion on this ideally on their discord.

Richard-L:
> I'd appreciate if someone could get their collective opinion on this ideally on their discord.

Let's get the release with the installer out first so those of them who are new to TR1M have a higher chance of conversion, at least what regards interest in the project.

Richard-L:
Interim result:
The reaction isn't as unfavourable as expected, and there's a mild thought for allowing TR1M for extension runs (pacifist, no damage, pistols only, etc).
But there's of course scepticism as TR1M is seen as a "mod" which usually aren't allowed.
TombATI was allowed per exception because of the weight of Dosbox. Some seem to see the further improvement of TR1M as too small to legitimise another exception, as far as speedrunning goes.
For casual playthroughs it seems TR1M is quite accepted as a norm among them. Perhaps the attention of Sabatu's level has helped here.

Richard-L:
After some more exchange:
Speedrunners are already unhappy about TombATI as it's against a "purist" point of view, but since Dosbox meant so much trouble, and TombATI looked like just a wrapper giving access to nothing but better rendering, they made an exception for it. To grant another exception for TR1M to what's already a suboptimal situation would need strong reasons.
I tried to list some of the features and benefits TR1M would bring: all the config aimed at runs, FMV skip ESC key, more savestates and resurrection through fly cheat during practice, etc. They did seem to like them and there was a certain indecision what to do. I guess it didn't help that Troye casually came in to say if anything then TR2M should be accepted before TR1M....
One of their admins then explained another issue besides the purist concern would be game verification. Ironically, the fact that classic TR had been so underdeveloped prior to TR1M, nothing on the PC looked like it. This means they can comfortably know and control that TombATI is used in speedruns. With something as modifiable as TR1M, it would be impossible to control what game settings someone has during a run. Obviously there are solutions to mitigate this, but it means an extra layer of effort for both sides in a situation where they're already hesitant.
Curiously someone even suggested to reach out to Steam/GOG, to include TR1M, in a naive thinking this might establish an "official" version that would then be more readily accepted.

To sum up:
I think some do like TR1M, and it wasn't me in the beginning who brought up the question on its potential inclusion for speedrun leaderboards. But there are a couple of question marks for them that feel like roadblocks which currently nobody wants to quite tackle.

If it doesn't mean too much trouble or dirty code, I would add buffering to TR1M (as per this PR), so that at least a door for speedrunners can stay open, and they have less reason to dismiss the project.

src/game/game/game.c
```diff
@@ -70,8 +70,9 @@ static int32_t Game_Control(int32_t nframes, GAMEFLOW_LEVEL_TYPE level_type)
             }
         }
 
-        if ((g_InputDB.option || g_Input.save || g_Input.load
-             || g_OverlayFlag <= 0)
+        if ((g_Config.enable_buffering ? g_Input.option
+                                       : g_InputDB.option || g_Input.save
```
rr-, line 76: Let's wrap the ternary operator with `()`, since now it's not evident if it's meant to say `(a ? b : c) || d || e` or `a ? b : (c || d || e)`.

Richard-L:
There's been some renewed interest in this from a speedrunner called Stelf: https://www.speedrun.com/user/stelf

Walkaway kindly put out a test build and Stelf gave it a try:
- With `"enable_buffering": true` F5 and F6 quick save/load are broken. With it set back to `false` it immediately works again.
- So far this suits what's called "inventory buffering" (=holding ESC to advance 1 frame before going into inventory, etc). Ideally they'd also have what's called "frame buffering", as explained here: https://youtu.be/jrZDx0dKc_8 | Another use case: https://youtu.be/9fouplVeEDQ
Walkaway I'm not sure you'd intended for this to already be included? So far it doesn't seem to be working.

PR #598, opened 2022-07-28
fixed underwater currents breaking in rare cases

### LINKED ISSUE #127: Underwater currents sometimes stop working
When you start the level and fall into the stream, it sucks you down the waterfall as is expected. After you've picked up the three cogs in the level and make your way back to put them in, this happened:

TombATI:

https://user-images.githubusercontent.com/4934209/137548483-d09b6e1b-6e82-4d6e-8d83-ac0a3a5f127d.mp4

TR1Main:

https://user-images.githubusercontent.com/4934209/137548586-585ea41d-9256-4480-9f4c-4352a2d18885.mp4

Tested the full scenario of selecting the cogs in both games and then falling into the stream. If the cogs are actually related or not I don't know of course. In TR1Main, if you save whilst you're swimming in the non-working stream, and load, it'll suddenly suck you down again, as if it was reset.

walkawayy:
Resolves #127.

#### Checklist

#### Description
Fixed underwater currents breaking in rare cases. This solves two water current issues:
1. Saving and loading can break a current in rare situations. Can occur in all of the OG game's currents.
2. Lara entering a second water current before reaching the first current's target breaks the second water current. Doesn't occur in OG levels.

...

### DIFF
--- a/src/game/lara/lara_control.c
+++ b/src/game/lara/lara_control.c
@@ -12,6 +12,7 @@
 #include "game/lara/lara_col.h"
 #include "game/lara/lara_look.h"
 #include "game/lara/lara_state.h"
+#include "game/lot.h"
 #include "game/objects/general/door.h"
 #include "game/room.h"
 #include "game/sound.h"
@@ -290,6 +291,8 @@ void Lara_HandleSurface(ITEM_INFO *item, COLL_INFO *coll)
 
     if (g_Lara.current_active && g_Lara.water_status != LWS_CHEAT) {
         Lara_WaterCurrent(coll);
+    } else {
+        LOT_ClearLOT(&g_Lara.LOT);
     }
 
     Lara_Animate(item);
@@ -356,6 +359,8 @@ void Lara_HandleUnderwater(ITEM_INFO *item, COLL_INFO *coll)
 
     if (g_Lara.current_active && g_Lara.water_status != LWS_CHEAT) {
         Lara_WaterCurrent(coll);
+    } else {
+        LOT_ClearLOT(&g_Lara.LOT);
     }
 
     Lara_Animate(item);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ## [Unreleased](https://github.com/rr-/Tomb1Main/compare/stable...develop)
+- fixed underwater currents breaking in rare cases (#127)
 
 ## [2.10](https://github.com/rr-/Tomb1Main/compare/2.9.1...2.10) - 2022-07-26
 - added a .NET-based installer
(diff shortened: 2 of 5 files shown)


src/game/lara/lara_control.c
```diff
@@ -356,6 +359,8 @@ void Lara_HandleUnderwater(ITEM_INFO *item, COLL_INFO *coll)
 
     if (g_Lara.current_active && g_Lara.water_status != LWS_CHEAT) {
         Lara_WaterCurrent(coll);
+    } else {
+        LOT_ClearLOT(&g_Lara.LOT);
```
rr-, line 363: Clearing it on every frame sounds like too much TBH. Can we be smarter about it?

walkawayy: The thing is `g_Lara.current_active = 0;` happens every frame too in `Lara_WaterCurrent` which is why I did it this way. I'll try to think of something.

Trxyebeep: walkawayy has asked about the fix over on discord, and for what its worth we've determined that this is what the later games that do not use LOT for currents do too, clear out all related data every frame

rr-: I was thinking of clearing the LOT like this
```
if (g_Lara.current_active) {
    LOT_ClearLOT(&g_Lara.LOT);
    g_Lara.current_active = 0;
}
```
but IDK, if Troye says this fix is compatible with Tomb 2+ then I'm OK with the current variant.

walkawayy: Where would that snippet be called? @Trxyebeep said he thinks the fix I implemented is fine.

rr-: I meant to call it this way where currently the only assignment `g_Lara.current_active = 0;` is living, but I've no idea if it's even gonna work.

walkawayy: This doesn't seem to work. Lara is slightly affected by current but it's jumpy and she can beat it easily by swimming.

https://streamable.com/8s2qpi


If you aren't comfortable with the `LOT_ClearLOT` I added, I can take them out. Those changes only fix a bug not present in the original games where Lara enters current 1 and travels to sink 2. Then, jumping into current 2 will never work. The main bug was an issue with wrongly restoring the LOT on load where it now clears.

oziphantom: that looks as if its getting current for a frame and then not the next. i.e Clear, set , clear , set ...

walkawayy: Was there a consensus on this? Is the current solution acceptable or do we want to keep exploring alternatives?

rr-: Since my approach doesn't work, and besides my early optimization concerns, there are no problem with your solution, let's get this merged too.

rr- (approved):

PR #601, opened 2022-08-01
savegame: added FX, lava emitters, flame emitters, and waterfalls

### LINKED ISSUE #418: OG Bug: flame, lava and water emitters state is not saved: return to default after loading
Related to #277.

Flame and lava emitters as well as waterfall mist fall back to their default state after loading a save.

Examples:

- player saves at the middle of the flame obstacle in Palace Midas and then loads the save: the flames turn ON immediately because they're set ON at the beginning of the level.

- player saves and loads after getting the last secret in The Great Pyramid: the flame and lava emitters in the bottom of the previous room turn OFF because they're OFF by default, only turning on when the player triggers them a couple rooms before.

- player saves and loads when by the cogs in Lost Valley. The waterfall mist objects turn OFF because they're OFF by default, turned on when player goes over a trigger before the bridge.

Hopefully the status of these objects could be saved and restored properly.

walkawayy:
Resolves #418.
Creates new savegame version.

#### Checklist

#### Description
Added graphics effects, lava emitters, flame emitters, and waterfalls to the savegame so they now persist on load. Creates new savegame version 2. Everything seems to be working, but marking this as draft for now. All FX, emitters, and waterfalls now save, but I'm not sure if I missed any other items that should be saved too. I looked into adding darts, but they are above the `g_LevelItemCount` so they get killed. I also thought about incrementing and decrementing as darts are created and killed similar to the mutant pods. But, the number of darts fired changes depending on how far away the wall is from the emitter.
...

### DIFF
--- a/src/game/objects/traps/lava.c
+++ b/src/game/objects/traps/lava.c
@@ -1,5 +1,6 @@
 #include "game/objects/traps/lava.h"
 
+#include "config.h"
 #include "game/effects.h"
 #include "game/items.h"
 #include "game/lara.h"
@@ -126,7 +127,9 @@ void LavaEmitter_Setup(OBJECT_INFO *obj)
     obj->control = LavaEmitter_Control;
     obj->draw_routine = Object_DrawDummyItem;
     obj->collision = Object_Collision;
-    obj->save_flags = 1;
+    if (g_Config.enable_enhanced_saves) {
+        obj->save_flags = 1;
+    }
 }
 
 void LavaEmitter_Control(int16_t item_num)
--- a/src/game/objects/traps/lava.c
+++ b/src/game/objects/traps/lava.c
@@ -127,9 +127,7 @@ void LavaEmitter_Setup(OBJECT_INFO *obj)
     obj->control = LavaEmitter_Control;
     obj->draw_routine = Object_DrawDummyItem;
     obj->collision = Object_Collision;
-    if (g_Config.enable_enhanced_saves) {
-        obj->save_flags = 1;
-    }
+    obj->save_flags = 1;
 }
 
 void LavaEmitter_Control(int16_t item_num)
--- a/src/game/objects/traps/lava.c
+++ b/src/game/objects/traps/lava.c
@@ -126,6 +126,7 @@ void LavaEmitter_Setup(OBJECT_INFO *obj)
     obj->control = LavaEmitter_Control;
     obj->draw_routine = Object_DrawDummyItem;
     obj->collision = Object_Collision;
+    obj->save_flags = 1;
 }
 
 void LavaEmitter_Control(int16_t item_num)
--- a/src/game/objects/traps/lava.c
+++ b/src/game/objects/traps/lava.c
@@ -1,6 +1,5 @@
 #include "game/objects/traps/lava.h"
 
-#include "config.h"
 #include "game/effects.h"
 #include "game/items.h"
 #include "game/lara.h"
--- a/src/game/savegame/savegame_bson.c
+++ b/src/game/savegame/savegame_bson.c
@@ -1,6 +1,7 @@
 #include "game/savegame/savegame_bson.h"
 
 #include "config.h"
+#include "game/effects.h"
 #include "game/gameflow.h"
 #include "game/inventory.h"
 #include "game/items.h"
@@ -49,6 +50,7 @@ static bool Savegame_BSON_LoadInventory(struct json_object_s *inv_obj);
 static bool Savegame_BSON_LoadFlipmaps(struct json_object_s *flipmap_obj);
 static bool Savegame_BSON_LoadCameras(struct json_array_s *cameras_arr);
 static bool Savegame_BSON_LoadItems(struct json_array_s *items_arr);
+static bool SaveGame_BSON_LoadFx(struct json_array_s *fx_arr);
 static bool Savegame_BSON_LoadArm(struct json_object_s *arm_obj, LARA_ARM *arm);
 static bool Savegame_BSON_LoadAmmo(
     struct json_object_s *ammo_obj, AMMO_INFO *ammo);
@@ -62,6 +64,7 @@ static struct json_object_s *Savegame_BSON_DumpInventory(void);
 static struct json_object_s *Savegame_BSON_DumpFlipmaps(void);
 static struct json_array_s *Savegame_BSON_DumpCameras(void);
 static struct json_array_s *Savegame_BSON_DumpItems(void);
+static struct json_array_s *SaveGame_BSON_DumpFx(void);
 static struct json_object_s *Savegame_BSON_DumpArm(LARA_ARM *arm);
 static struct json_object_s *Savegame_BSON_DumpAmmo(AMMO_INFO *ammo);
 static struct json_object_s *Savegame_BSON_DumpLOT(LOT_INFO *lot);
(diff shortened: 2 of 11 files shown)


rr- (commented):
1. I'm pretty sure fiddling with `obj->save_flags` this will break legacy savegames in Atlantis.
2. Lava emitters are a thing in speedruns; can we have a `g_Config.enable_enhanced_saves`? (Naming forward compatible with concepts such as saving music tracks)

src/game/savegame/savegame_bson.c
```diff
@@ -575,18 +575,20 @@ static bool SaveGame_BSON_LoadFx(struct json_array_s *fx_arr)
...
+                fx->pos.x = x;
+                fx->pos.y = y;
+                fx->pos.z = z;
+                fx->object_number = object_number;
+                fx->speed = speed;
+                fx->fall_speed = fall_speed;
+                fx->frame_number = frame_number;
+                fx->counter = counter;
+                fx->shade = shade;
+            }
```
rr-: Shouldn't this entire function exit early if the setting is disabled?

src/game/objects/traps/lava.c
```diff
@@ -126,7 +127,9 @@ void LavaEmitter_Setup(OBJECT_INFO *obj)
     obj->control = LavaEmitter_Control;
     obj->draw_routine = Object_DrawDummyItem;
     obj->collision = Object_Collision;
-    obj->save_flags = 1;
+    if (g_Config.enable_enhanced_saves) {
+        obj->save_flags = 1;
+    }
```
rr-: That's not good for the legacy files because now whether they'll work or not will be controlled be a JSON setting.
The code should be amended to always disregard save_flags for these objects, similar to how we have some hardcoded checks for the Bacon Lara. It's OK to set save_flags to 1 here unconditionally.

src/game/savegame/savegame_bson.c
walkawayy: My idea was to save the FX regardless if the user enabled it in the JSON in case they ever want to turn it on. I wasn't sure if I could return early from this function because I don't completely understand how the savegame load works. So, I read all the FX data to be safe but just don't create the effect. I can change it if that's not the case.

rr-: Yeah that seems to be not necessary unless for validation sake.

src/game/objects/traps/lava.c
walkawayy: By disregarding save flags, what exactly do you mean?

My thought was in `Savegame_Legacy_SaveToFile`, can I change `if (obj->save_flags)` to `if (obj->save_flags && item->object_number != O_LAVA_EMITTER && item->object_number != O_FLAME_EMITTER && item->object_number != O_WATERFALL) {`? A little ugly but it's a hardcoded check like you said.

rr-: Yes, that's exactly it, but it needs testing too. Basically make a save in original TombATI in The Great Pyramid and see if loading it in T1M with the proposed changes goes through or crashes the game.

walkawayy: Gonna pose a multi part question sorry. Savegames are complicated to mess with and I want to make sure I don't break anything. : /

1. I made a Great Pyramid save near a lava emitter in TombATI. I tried to load it into T1M's 2.10.2 release and the game crashed. I added the `(obj->save_flags && item->object_number !=` check for `O_LAVA_EMITTER`, `O_FLAME_EMITTER`, and `O_WATERFALL` in `Savegame_Legacy_LoadFromFile`. That fixed the loading crash and the items behaved like TombATI where they do not restore on load. I also added those checks to `Savegame_Legacy_SaveToFile` in case someone ever saved legacy so those items that now have save flags do not save that way in legacy format. Does this sound right?

2. In `Savegame_Legacy_NeedsBaconLaraFix`, there's an `if (obj->save_flags)` check in that function, so I probably also need to add the ` item->object_number !=` checks there too for the emitters?

3. And lastly, for BSON saves, do I need to change `Savegame_BSON_LoadItems`'s `if (obj->save_flags) {` check similarly to the legacy version where it ignores certain objects unless the save is a newer version?

rr-: 1. Yes
2. Yes
3. I don't think you need to repeat this hardcoding of object ids, just make sure to only load the flags if the relevant bson keys were present in the document.

src/game/savegame/savegame_bson.c
```diff
@@ -539,6 +539,10 @@ static bool Savegame_BSON_LoadItems(struct json_array_s *items_arr)
 
 static bool SaveGame_BSON_LoadFx(struct json_array_s *fx_arr)
 {
+    if (!g_Config.enable_enhanced_saves) {
+        return true;
+    }
+
     if (!fx_arr) {
         LOG_ERROR("Malformed save: invalid or missing fx array");
```
rr-, line 547: I think this will cause previous T1M saves to break.  I'd `return true` instead.
Edit: never mind, you solved this more cleanly with the `version` field.

src/game/savegame/savegame_bson.c
```diff
@@ -883,6 +946,36 @@ static struct json_array_s *Savegame_BSON_DumpItems(void)
     return items_arr;
 }
 
+static struct json_array_s *SaveGame_BSON_DumpFx(void)
+{
+    struct json_array_s *fx_arr = json_array_new();
+
+    int16_t linknum = g_NextFxActive;
```
rr-: It's OK to declare this within the `for`.

src/game/savegame/savegame_bson.c
```diff
@@ -883,6 +946,36 @@ static struct json_array_s *Savegame_BSON_DumpItems(void)
...
+    int16_t num_fx = 0;
```
rr-: This variable is not used anywhere.

src/game/savegame/savegame_legacy.c
```diff
@@ -126,7 +126,10 @@ static bool Savegame_Legacy_NeedsBaconLaraFix(char *buffer)
         if (obj->save_hitpoints) {
             Savegame_Legacy_Read(&tmp_item.hit_points, sizeof(int16_t));
         }
-        if (obj->save_flags) {
+        if (obj->save_flags
+            && (item->object_number != O_LAVA_EMITTER
```
rr-: No need for the additional braces

src/game/savegame/savegame_legacy.c
```diff
@@ -706,7 +711,10 @@ void Savegame_Legacy_SaveToFile(MYFILE *fp, GAME_INFO *game_info)
             Savegame_Legacy_Write(&item->hit_points, sizeof(int16_t));
         }
 
-        if (obj->save_flags) {
+        if (obj->save_flags
+            && (item->object_number != O_LAVA_EMITTER
```
rr-: No need for the additional braces

src/game/objects/traps/lava.c
walkawayy: I think all three of these are solved. The only one I'm not 100% sure on is `3`. But, I created a save with and without save_flags for the new objects. Then I logged the item loading for all save_flags and nothing in the load process seemed to change. I'm not sure how exactly but it looks to already be doing what you suggested.

walkawayy:
I think this is ready for review, but I'm going to ask if anyone can help test to make sure there are no bugs. I don't trust myself as a thorough tester lol.

EDIT: Posix is gonna try and take a look.

rr- (approved):
LGTM

Richard-L:
Did some testing and cannot find fault with it. (We found some different concerns which can be addressed as separate issues - not here). For now all of these effects seem to be correctly restored upon load which is great. Again, really nice work @walkawayy :pray:

walkawayy:
One of the only things left are darts which do not restore on load. But they are spawned and created as items. So they are different than FX and the single emitters. We could hold off on this PR merge and I could try to work on them again (didn't have much luck before), or kick them down the line to a future issue.

Richard-L:
Would seem like a separate issue to me but needless to say you need to decide this.

One thing I did notice just now: This savestate cannot be loaded in the test build of this PR: [save_tr1_04.zip](https://github.com/rr-/Tomb1Main/files/9376974/save_tr1_04.zip)
I believe I had created it during tests. It's in Atlantis in the big room with the "flesh bridge" inside and lava pits to the left and right. Upon load the main menu resets.

walkawayy:
> Would seem like a separate issue to me but needless to say you need to decide this.
> 
> One thing I did notice just now: This savestate cannot be loaded in the test build of this PR: [save_tr1_04.zip](https://github.com/rr-/Tomb1Main/files/9376974/save_tr1_04.zip) I believe I had created it during tests. It's in Atlantis in the big room with the "flesh bridge" inside and lava pits to the left and right. Upon load the main menu resets.

What's happening is the savegame version says it was made in my test build (version 2), but the FX info array isn't present in the save so loading gives an error `Malformed save: invalid or missing fx array`. I made a save in the beginning of Vilcabamba where no FX are present to make sure there wasn't some loading bug when loading a save where no FX were saved. But that save loaded fine.

The save says the initial version is version 0. So it was started on a T1M version from over 3 months ago before select level. That still shouldn't matter since the save's current version is 2 and has to have been made on the new build meaning it would have the FX array. We could make this a non critical error since restoring FX aren't really required, but that seems like a bandaid solution. I'm confused how this happened. There must be some obscure bug?

rr-:
Let's tackle the darts as a separate issue, this PR is a good groundwork for it and good enough to warrant merging on its own.

re: Version 2 saves not having fx array and crashing the game – We need a consistent way to reproduce this save with the current build and the last stable build; otherwise I wouldn't worry about it. I have a feeling this savestate is not possible to achieve with the last build.

walkawayy:
I haven't been able to reproduce this but I also don't want to merge potentially buggy saves since that would be a huge issue. However, it seems Posix doesn't know when that save was made and how to reproduce, so I'm not sure how this would even get fixed unless it happens again.

rr-:
Worst case, we'll react with hotfixes. Let's get this merged.

PR #612, opened 2022-08-22
Refactor gameflow inventory removals

### LINKED ISSUE #605: remove_scions in gameflow causes Lara to draw weapons
If `remove_scions` is in the gameflow without `remove_guns`, Lara will always draw her weapons when the level starts, regardless of whether or not she finished the previous level with them drawn or holstered. This can be reproduced by adding `remove_scions` to any level in the default gameflow, or by removing `remove_guns` from Natla's Mines.

This was found during implementation of #599 - additional comments are [here](https://github.com/rr-/Tomb1Main/issues/599#issuecomment-1205629833).

### LINKED ISSUE #611: Gameflow option remove_guns can prevent weapon pickups
See #605.

- Lara starts level 1, which has `remove_guns` in the gameflow.
- She picks up the Uzis in this level.
- Lara starts level 2, which also has `remove_guns` in the gameflow.
- Level 2 also has an Uzi weapon pickup but the game changes this to Uzi ammo.

So Lara is left unarmed with no guns to pickup. Again, I think it's one of those things that doesn't affect OG because I don't believe `Item_GlobalReplace` is called against pistols/pistol ammo.

walkawayy:
#### Checklist

#### Description
Fixed gameflow option remove_guns preventing weapon pickups in rare situations.
Fixed gameflow option remove_scions causing Lara to equip weapons even if she has none.

Matches TR2Main's implementation. https://github.com/Arsunt/TR2Main/search?q=GF_RemoveWeapons
...

### DIFF
--- a/src/game/game/game.c
+++ b/src/game/game/game.c
@@ -144,10 +144,10 @@ bool Game_Start(int32_t level_num, GAMEFLOW_LEVEL_TYPE level_type)
         break;
 
     case GFL_RESTART:
-        Savegame_ResetCurrentInfo(level_num);
         if (level_num <= g_GameFlow.first_level_num) {
             Savegame_InitCurrentInfo();
         } else {
+            Savegame_ResetCurrentInfo(level_num);
             // Use previous level's ending info to start current level.
             Savegame_CarryCurrentInfoToNextLevel(level_num - 1, level_num);
             Savegame_ApplyLogicToCurrentInfo(level_num);
@@ -159,16 +159,13 @@ bool Game_Start(int32_t level_num, GAMEFLOW_LEVEL_TYPE level_type)
         break;
 
     case GFL_SELECT:
-        // reset current info to the defaults so that we do not do
-        // Item_GlobalReplace in the inventory initialization routines too early
         Savegame_InitCurrentInfo();
-        Savegame_LoadOnlyResumeInfo(g_GameInfo.current_save_slot, &g_GameInfo);
-        for (int i = level_num; i < g_GameFlow.level_count; i++) {
-            Savegame_ResetCurrentInfo(i);
-        }
-        if (level_num <= g_GameFlow.first_level_num) {
-            Savegame_InitCurrentInfo();
-        } else {
+        if (level_num > g_GameFlow.first_level_num) {
+            Savegame_LoadOnlyResumeInfo(
+                g_GameInfo.current_save_slot, &g_GameInfo);
+            for (int i = level_num; i < g_GameFlow.level_count; i++) {
+                Savegame_ResetCurrentInfo(i);
+            }
             // Use previous level's ending info to start current level.
             Savegame_CarryCurrentInfoToNextLevel(level_num - 1, level_num);
             Savegame_ApplyLogicToCurrentInfo(level_num);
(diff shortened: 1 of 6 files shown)


src/game/gameflow.c
```diff
@@ -1249,20 +1252,13 @@ GameFlow_InterpretSequence(int32_t level_num, GAMEFLOW_LEVEL_TYPE level_type)
...
-                Lara_InitialiseInventory(level_num);
+                g_GameInfo.remove_guns = true;
             }
             break;
 
         case GFS_REMOVE_SCIONS:
             if (level_type != GFL_SAVED) {
-                g_GameInfo.current[level_num].num_scions = 0;
-                Lara_InitialiseInventory(level_num);
+                g_GameInfo.remove_scions = true;
```
rr-, line 1261: Ha, smart.

src/game/lara/lara.c
```diff
@@ -513,6 +513,20 @@ void Lara_InitialiseInventory(int32_t level_num)
     RESUME_INFO *resume = &g_GameInfo.current[level_num];
 
     g_Lara.pistols.ammo = 1000;
+
+    if (g_GameInfo.remove_guns == true) {
```
rr-: Let's drop the `== true` part, it's silly :D

src/game/game/game.c
```diff
@@ -159,16 +159,13 @@ bool Game_Start(int32_t level_num, GAMEFLOW_LEVEL_TYPE level_type)
...
-        }
-        if (level_num <= g_GameFlow.first_level_num) {
-            Savegame_InitCurrentInfo();
-        } else {
+        if (level_num > g_GameFlow.first_level_num) {
+            Savegame_LoadOnlyResumeInfo(
+                g_GameInfo.current_save_slot, &g_GameInfo);
+            for (int i = level_num; i < g_GameFlow.level_count; i++) {
+                Savegame_ResetCurrentInfo(i);
+            }
```
rr-, line 168: Do I understand it right that this change technically shouldn't do anything and is cosmetic, since `Savegame_InitCurrentInfo` executed the second time for levels <= the first level would completely override the results of `Savegame_LoadOnlyResumeInfo` and `Savegame_ResetCurrentInfo`?

src/game/gameflow.c
walkawayy: Yes haha much simpler. I tried to avoid refactoring / adding a global, but this seemed like the best option and matches TR2. And it avoids the weird double / triple inventory initialization, so it's better overall IMO.

src/game/lara/lara.c
walkawayy: I always do this oof lol. Copy/paste things.

src/game/game/game.c
walkawayy: Yes cosmetic. Basically the comment was copied and pasted from the `GFL_SAVED` case and wasn't reflecting what was actually happening. In the select level / restart cases, the level is loaded after the current info was restored (since those processes don't call an inventory init like the load game process does).

And yes you are spot on with the second part of your sentence. I did some testing with `remove_guns` and `remove_scions` to make sure that Lara started the level with the proper items even after a restart, level select, save after picking up a weapon, etc.

rr- (approved):

PR #613, opened 2022-08-22
option_passport: fix save crystal mode saving in the first slot

### LINKED ISSUE #607: Save crystals won't save
When I disable them, the game saves normally.
When enabled:
- I can interact with the save crystal.
- It opens the journal with the "save" page.
- I press ctrl/enter to open the save menu.
- The journal closes and it unpauses the game without saving at all.

Although, it does make a save file when I check the save folder. It's just that I cannot load it.

Also at some point it just crashed when I pressed ctrl to open the save menu (I wasn't able to reproduce this, it only happened once)

walkawayy:
Resolves #607.

#### Checklist

#### Description
Fixed save crystal mode always saving in the first slot.
...


rr- (approved):

rr-:
Since it took us 4 months to find out about this bug, let's release this as a part of normal release rather than a hotfix.

PR #615, opened 2022-08-24
UI scale fix

### LINKED ISSUE #559: Incorrect UI scaling
With higher UI scaling, the text box borders do not increase in size. They remain a constant two pixel size at all scaling levels. Should they increase to possibly 4 pixels at 1.5+scaling and 6 pixels at 2 scaling? There may also need to be a change in the spacing so the borders do not overlap. This is also possibly related to #466.

Here's the highest UI scaling option of 2 and there's still a 2 pixel border size.
20220514_112454_Title

At low resolutions, the UI lines are also wrongly drawn.

oziphantom:
#### Checklist

#### Description

...
This converts the line draw hack for the GUI into a proper triangle based solution. Both PC and PS1 UI layout. 
This also handle the half sizes on higher resolution better now you should get something that is half as thick on both and not the 2/1 thickness of before but it depends on your resolution. 
It is basically only designed to work on 16:9 and would go wrong on my 4:5 screen but it is also fine on my 16:10. This is because I have to scale based upon the 640x480 original and I chose the Y as the scaling determinate as the "9" is very common while the 16 can get ultra wide these days and most people would be likely to run this on a X:9 monitor and I'm the only freak who runs 5:4 monitors in 4:5 orientation and X:10 are sadly rare these days and any body running a 3:2 monitor is too damn lucky and they can lump it, but 3:2 is 15:10 which given how well it works on a 16:10 ( I don't notice anything off on mine ) should probably actually work better. Sadly I don't have a 3:2 to test on.

Along the way I found some other issues, namely for some reason the code uses Triangle Fans but the Triangle Strip function was called, and the Triangle Strip function actually dealt with a Fan. So I've renamed the Strips to Fan and added more Strip functions. 
I would suggest as another PR we adjust the Fan creation logic we have to be make Strips, as Fan support has now been removed from modern versions of GL and DX, and we are making custom geometry with a whole 2 triangles of which it is trivial to change and will simplify any porting done in the future. Unless I missed some internal level data calling the Fan functions somewhere.


src/game/output.c
```diff
@@ -723,20 +723,13 @@ void Output_DrawScreenLine(
...
-        // Right
-        Output_DrawScreenLine(w + sx + 1 + i, sy - i, 0, h + 1 + (i * 2), col);
-        // Left
-        Output_DrawScreenLine(sx - i, h + sy + 1 + i, 0, (-i * 2) - 1 - h, col);
-        // Bottom
-        Output_DrawScreenLine(
-            w + sx + 1 + i, h + sy + i, -w - 1 - (i * 2), 0, col);
-    }
+    int32_t sx, int32_t sy, int32_t w, int32_t h, RGBA8888 colDark,
+    RGBA8888 colLight, int32_t thickness)
```
rr-, line 727: The argument names do not adhere to the coding convention
@walkawayy have you tested this?

src/game/output.h
```diff
@@ -65,8 +65,8 @@ void Output_DrawScreenGradientQuad(
 void Output_DrawScreenLine(
     int32_t sx, int32_t sy, int32_t w, int32_t h, RGBA8888 col);
 void Output_DrawScreenBox(
-    int32_t sx, int32_t sy, int32_t w, int32_t h, RGBA8888 col,
-    int32_t thickness);
+    int32_t sx, int32_t sy, int32_t w, int32_t h, RGBA8888 colDark,
+    RGBA8888 colLight, int32_t thickness);
```
rr-, line 69: The argument names do not adhere to the naming conventions

src/gfx/3d/vertex_stream.c
```diff
@@ -93,6 +94,30 @@ bool GFX_3D_VertexStream_PushPrimStrip(
...
+        LOG_ERROR("Unsupported prim type: %d", vertex_stream->prim_type);
+        return false;
+    }
+
+    if (count <= 2) {
+        for (int i = 0; i < count; i++) {
+            GFX_3D_VertexStream_PushVertex(vertex_stream, &vertices[i]);
+        }
+    } else {
+        // convert Fan to raw triangles
```
rr-: Let's not capitalize Fan out of nowhere

.gitignore
```diff
@@ -27,4 +27,5 @@ Release/
 *.sln
 *.vcxproj
 *.filters
+libs/
```
rr-: I don't think we use this folder?

src/game/output.c
walkawayy: Yes I did some testing on this and it looks really good. I'll post some screenshots later. It works well even at low resolutions.

walkawayy:
Here's some sample screenshots. Also works well with the PS1 UI style.

Lowest res:
20220902_153359_Title


1440p:
20220902_153337_Title

rr-:
Does it work with the UI scaling too?

walkawayy:
Yep scaling looks good.

20220902_215108_Title


20220902_215113_Title

rr- (approved):

rr-:
Missing:

"[x ] I have added a changelog entry about what my pull request accomplishes, or it is an internal change"

There's no changelog entry, and this is a change that the user sees so it's not internal.

PR #617, opened 2022-08-31
savegame: fix flame emitters in enhanced saves.

### LINKED ISSUE #616: T1M bug: enhanced saves cause triggered flame problem
The following demo shows a problem in Palace Midas with the fire room. If you save before entering the room (or crossing the pad that turns off the flames), fail the jumps and decide to reload your save, the trigger to put out the flames no longer works.

https://streamable.com/0ifi2x

If you repeat this process with `enable_enhanced_saves` switched off, the issue does not occur.

Found by Nabu1908 during randomizer testing (above video is a repeated test with OG level data).

walkawayy:
Resolves #616.

#### Checklist

#### Description
No changelog entry since this fixes a bug with enhanced saves which aren't released yet. The issue is the flame emitter works by using item->data which is normally only saved and loaded for intelligent creatures which the flame emitter isn't. So basically on load, the flame emitter is working normally. But if you save while the flame is on, the restored flame FX is never killed so there's a constant flame always on (2 flames active basically with 1 never turning off). This PR saves the item->data and restores it in load.
...

### DIFF
--- a/src/game/savegame/savegame_bson.c
+++ b/src/game/savegame/savegame_bson.c
@@ -536,7 +536,7 @@ static bool Savegame_BSON_LoadItems(struct json_array_s *items_arr)
                 && g_Config.enable_enhanced_saves) {
                 int32_t flame_num =
                     json_object_get_int(item_obj, "flame_num", flame_num);
-                item->data = (void *)(flame_num);
+                item->data = (void *)flame_num;
             }
         }
     }
@@ -945,7 +945,7 @@ static struct json_array_s *Savegame_BSON_DumpItems(void)
             }
 
             if (item->object_number == O_FLAME_EMITTER) {
-                int32_t flame_num = ((int32_t)item->data);
+                int32_t flame_num = (int32_t)item->data;
                 json_object_append_int(item_obj, "flame_num", flame_num);
             }
         }
--- a/src/game/savegame/savegame_bson.c
+++ b/src/game/savegame/savegame_bson.c
@@ -531,6 +531,13 @@ static bool Savegame_BSON_LoadItems(struct json_array_s *items_arr)
             } else if (obj->intelligent) {
                 item->data = NULL;
             }
+
+            if (item->object_number == O_FLAME_EMITTER
+                && g_Config.enable_enhanced_saves) {
+                int16_t flame_num =
+                    json_object_get_int(item_obj, "flame_num", flame_num);
+                item->data = (void *)(flame_num + 0);
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/game/savegame/savegame_bson.c
```diff
@@ -531,6 +531,13 @@ static bool Savegame_BSON_LoadItems(struct json_array_s *items_arr)
             } else if (obj->intelligent) {
                 item->data = NULL;
             }
+
+            if (item->object_number == O_FLAME_EMITTER
+                && g_Config.enable_enhanced_saves) {
+                int16_t flame_num =
+                    json_object_get_int(item_obj, "flame_num", flame_num);
+                item->data = (void *)(flame_num + 0);
```
rr-: Why `+ 0`?

walkawayy: Oh right I meant to ask about this. I get a compiler warning without the addition that I don't remember atm. I tried to use the same formula as in [flame.c](https://github.com/rr-/Tomb1Main/blob/eeb3f4d1b2ce9e554614b6100d6175dd22bf9edd/src/game/objects/traps/flame.c#L114). That has a `+1` in it which avoids the warning. I'll post the warning later.

Is this solution ok btw? It's a little hacky during the load process, but the way `item->data` is used for this trap is unique.

rr-: The compiler warning is probably about implicit type casting from int16_t to void*. Try (void*)flame_num.

walkawayy: Here's the warning:
```
../src/game/savegame/savegame_bson.c: In function ‘Savegame_BSON_LoadItems’:
../src/game/savegame/savegame_bson.c:539:30: warning: cast to pointer from integer of different size [-Wint-to-pointer-cast]
  539 |                 item->data = (void*)flame_num;
      |                              ^
```

rr-: Ah OK in that case just make it an `int`.

src/game/savegame/savegame_bson.c
```diff
@@ -945,7 +945,7 @@ static struct json_array_s *Savegame_BSON_DumpItems(void)
             }
 
             if (item->object_number == O_FLAME_EMITTER) {
-                int16_t flame_num = ((int16_t)(size_t)item->data);
+                int32_t flame_num = ((int32_t)item->data);
```
rr-: No need for the braces, `(int32_t)item->data` will do

src/game/savegame/savegame_bson.c
```diff
@@ -534,9 +534,9 @@ static bool Savegame_BSON_LoadItems(struct json_array_s *items_arr)
 
             if (item->object_number == O_FLAME_EMITTER
                 && g_Config.enable_enhanced_saves) {
-                int16_t flame_num =
+                int32_t flame_num =
                     json_object_get_int(item_obj, "flame_num", flame_num);
-                item->data = (void *)(flame_num + 0);
+                item->data = (void *)(flame_num);
```
rr-: No need for the braces, `(void *)flame_num` will do

rr- (approved):

PR #621, opened 2022-09-16
savegame: fix incorrect number of FX check during loading.

### LINKED ISSUE #620: Some saves can no longer be loaded
With enhanced saves on, the following issue was found if you save/load at various points through Atlantis. The last save will no longer load. When enhanced saves are off, this issue does not appear.

https://streamable.com/zmlmkb (issue occurs from 02:40)

The following is in Tomb1Main.log

```
../src/game/savegame/savegame_bson.c 562 SaveGame_BSON_LoadFx Malformed save: expected a max of 99 fx, got 99
../src/game/game/game.c 141 Game_Start Failed to load save file!
```

Save files:
[Saves.zip](https://github.com/rr-/Tomb1Main/files/9538607/Saves.zip)

Found by NaBu1908 during rando testing - above is original level file data.

walkawayy:
Resolves #620.

#### Checklist

#### Description
No changelog because enhanced saves haven't been released yet. Fixes loading saves with max number of FX. Future idea: maybe the max number of FX can be raised from 100.
...

### DIFF
--- a/src/game/savegame/savegame_bson.c
+++ b/src/game/savegame/savegame_bson.c
@@ -556,10 +556,10 @@ static bool SaveGame_BSON_LoadFx(struct json_array_s *fx_arr)
     }
 
     if ((signed)fx_arr->length >= NUM_EFFECTS) {
-        LOG_ERROR(
-            "Malformed save: expected a max of %d fx, got %d", NUM_EFFECTS - 1,
-            fx_arr->length);
-        return false;
+        LOG_WARNING(
+            "Malformed save: expected a max of %d fx, got %d. fx over the "
+            "maximum will not be created.",
+            NUM_EFFECTS - 1, fx_arr->length);
     }
 
     for (int i = 0; i < (signed)fx_arr->length; i++) {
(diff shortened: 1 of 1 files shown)


rr- (approved):
Personally `>= NUM_EFFECTS)` would be more readable to me, less mental wrestling.

WDYT about making this a warning instead and discarding new effects?
WDYT about expanding the `NUM_EFFECTS` limit by +50%?

Can you see if the code that appends ingame effects to the fx array also checks `NUM_EFFECTS` and doesn't do a buffer overflow?

walkawayy:
Ya I'll look more into this. Since the next release isn't happening soon, I have more time. I asked TEN developers about their FX limit and they said they raised it to 1024 just as an FYI. I also agree the -1 check is weird so I should change it. It's a copy and paste thing from a loop in the effects file.

walkawayy:
Ok I did some testing with bubbles. In `FX_Bubbles` I changed `count` to 1000. Then I raised the `NUM_EFFECTS` to 6000 and compared it to the default value of 100. The game does have protections in place and I didn't get any overflows / crashes. Here's screenshots:

100 limit:
20220916_144257_Natlas_Mines

6000 limit:
20220916_144218_Natlas_Mines

walkawayy:
Seems like we can up the limit but I'll make a new issue / PR for that since it's unrelated to this fix.

rr- (approved):

PR #622, opened 2022-09-16
savegame: highlight latest save at game start

### LINKED ISSUE #618: Highlight newest save when game starts
It would be nice when I open TR1Main for the first time and game would suggest me (highlight) literally last created savefile just like in TR2/TR2Gold instead of this one that is on the top of Lara's passport. Newest TR1Main can have more number of savefiles than stantard game so if may become an annoying issue if you will have to scroll down to the bottom of save files every time you launch a game after quit.

walkawayy:
#### Checklist

#### Description
Changed passport to highlight latest save at game start.

Brings back [this code](https://github.com/rr-/Tomb1Main/pull/534/files) except stores it in a new variable to only use it at game start.
...

### DIFF
--- a/src/game/savegame/savegame.c
+++ b/src/game/savegame/savegame.c
@@ -39,6 +39,7 @@ typedef struct SAVEGAME_STRATEGY {
     bool (*update_death_counters)(MYFILE *fp, GAME_INFO *game_info);
 } SAVEGAME_STRATEGY;
 
+static uint16_t m_NewestSlot = 0;
 static SAVEGAME_INFO *m_SavegameInfo = NULL;
 
 static const SAVEGAME_STRATEGY m_Strategies[] = {
@@ -419,8 +420,6 @@ bool Savegame_Save(int32_t slot_num, GAME_INFO *game_info)
         sprintf(
             &req->item_texts[req->item_text_len * slot_num], "%s %d",
             g_GameFlow.levels[g_CurrentLevel].level_title, g_SaveCounter);
-        g_SavedGamesCount++;
-        g_SaveCounter++;
     }
 
     Savegame_ScanSavedGames();
@@ -541,6 +540,10 @@ void Savegame_ScanSavedGames(void)
             sprintf(
                 &req->item_texts[req->items * req->item_text_len], "%s %d",
                 savegame_info->level_title, savegame_info->counter);
+
+            if (savegame_info->counter == g_SaveCounter) {
+                m_NewestSlot = i;
+            }
         } else {
             req->item_flags[req->items] |= RIF_BLOCKED;
             sprintf(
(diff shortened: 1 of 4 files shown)


src/game/savegame/savegame.c
```diff
@@ -419,8 +420,6 @@ bool Savegame_Save(int32_t slot_num, GAME_INFO *game_info)
         sprintf(
             &req->item_texts[req->item_text_len * slot_num], "%s %d",
             g_GameFlow.levels[g_CurrentLevel].level_title, g_SaveCounter);
-        g_SavedGamesCount++;
-        g_SaveCounter++;
```
rr-, line 423: Why are these removed, won't they cause the sg counter to stay stuck?

rr-:
As far as I understand you made this new function so that it only gets executed upon game launch, and not fall back to the maximum save counter each time the user opens the load game screen right? (So if they loaded slot 3, it should stay at slot 3 when opening the passport again, even if slot 5 has the latest save)

src/game/savegame/savegame.c
walkawayy: Sorry I forgot to leave a comment on this. Basically these increments are worthless because these counters are set to 0 and recalculated in the `Savegame_ScanSavedGames` function which is called at the end of the `Savegame_Save` function. All they did was cause me confusion lol.

walkawayy:
> As far as I understand you made this new function so that it only gets executed upon game launch, and not fall back to the maximum save counter each time the user opens the load game screen right? (So if they loaded slot 3, it should stay at slot 3 when opening the passport again, even if slot 5 has the latest save)

Ya exactly. Hopefully it's not too weird the way I set the requestor in the shell.

rr- (commented):
Can we move the requestor setter to `ScanSavedGames`?
Since this is called multiple times throughout the course of the game, we could initially set the `m_NewestSlot = -1` and then inside `ScanSavedGames` something like

```c
if (m_NewestSlot == -1) {
    m_NewestSlot = Savegame_GetNewestSlot()
    g_SavegameRequester.requested = m_NewestSlot;
}
```

It is possible that we don't even need `m_NewestSlot` for this kind of initialization and we can surmise whether it's the first scan from some other variables.

walkawayy:
> Can we move the requestor setter to `ScanSavedGames`? Since this is called multiple times throughout the course of the game, we could initially set the `m_NewestSlot = -1` and then inside `ScanSavedGames` something like
> 
> ```c
> if (m_NewestSlot == -1) {
>     m_NewestSlot = Savegame_GetNewestSlot()
>     g_SavegameRequester.requested = m_NewestSlot;
> }
> ```
> 
> It is possible that we don't even need `m_NewestSlot` for this kind of initialization and we can surmise whether it's the first scan from some other variables.

What about just moving the `g_SavegameRequester.requested = m_NewestSlot` to a `Savegame_HighlightNewestSlot` that still occurs during the `Shell_Init`? Not sure there's another variable we can use to detect initialization.

rr- (approved):

PR #624, opened 2022-09-17
room: fix Lara loading inside movable block if stacked near portal

### LINKED ISSUE #619: Stacked pushblocks can cause void issue
If the game is saved when Lara is standing on top of a pushblock, which itself is on top of another pushblock, when that save is reloaded Lara voids. If an earlier save, where she is not standing on the block, is loaded, she can only get on top of the block after it is pushed/pulled. This isn't an issue with single blocks, only when they are stacked.

https://streamable.com/za1zsz

Found by LORDmeSH1NE during rando testing. The above is original level data and this has been checked with enhanced saves both on and off.

walkawayy:
Resolves #619.

#### Checklist

#### Description
Fixed Lara loading inside a movable block if she's on a stack near a room portal.
...


rr-:
I don't quite understand this fix as I am not familiar with the floor data. Can you describe the change in detail, the circumstances of this `if` check and why this helps solve the issue? Thanks

walkawayy:
> I don't quite understand this fix as I am not familiar with the floor data. Can you describe the change in detail, the circumstances of this `if` check and why this helps solve the issue? Thanks

This [line](https://github.com/rr-/Tomb1Main/blob/9a8157fff6e89af1ac7286b1dae3fdef621f2341/src/game/room.c#L597) was basically saying since the floor (top of the stacked block) was equal to the ceiling (room portal), that spot isn't standable. However in this case it is standable because the portal is a sky room which means it's open to the room above. There's no actual ceiling there. I talked with ChocolateFan about it since she knows the floor data stuff much more than me.

rr-:
I'm afraid this may break some corner case. Can we test it with the following scenarios, that require building a custom level:

1. Have a pushblock standing on a normal floor, above which there's a ceiling within the same room. Move it away. Place Lara on the tile where the pushblock was. Save and load. The pushblock should be moved away, and Lara shouldn't void.
2. Have a pushblock standing on a normal floor, above which there's a ceiling connected through a portal. Move it away. Place Lara on the tile where the pushblock was. Save and load. The pushblock should be moved away, and Lara shouldn't void.
3. Have a pushblock standing on a ceiling from a room underneath, above which there's a ceiling within the same room. Move it away. Place Lara on the tile where the pushblock was. Save and load. The pushblock should be moved away, and Lara shouldn't void.
4. Have a pushblock standing on a ceiling from a room underneath, above which there's a ceiling connected through a portal. Move it away. Place Lara on the tile where the pushblock was. Save and load. The pushblock should be moved away, and Lara shouldn't void.
5. Have a pushblock standing on a normal floor, above which there's empty space within the same room. Stand on top of it. Save and load. Lara should still be standing on top of the pushblock.
6. Have a pushblock standing on a normal floor, above which there's empty space connected through a portal. Stand on top of it. Save and load. Lara should still be standing on top of the pushblock.
7. Have a pushblock standing on a ceiling from a room underneath, above which there's empty space within the same room. Stand on top of it. Save and load. Lara should still be standing on top of the pushblock.
8. Have a pushblock standing on a ceiling from a room underneath, above which there's empty space connected through a portal. Stand on top of it. Save and load. Lara should still be standing on top of the pushblock.
9. Have a pushblock standing on top of a pushblock within the same room, above which there's a ceiling within the same room. Move it away. Place Lara on the tile where the pushblock was. Save and load. The pushblock should be moved away, and Lara shouldn't void.
10. Have a pushblock standing on top of a pushblock within the same room, above which there's a ceiling connected through a portal. Move it away. Place Lara on the tile where the pushblock was. Save and load. The pushblock should be moved away, and Lara shouldn't void.
11. Have a pushblock standing on top of a pushblock located in a room underneath, above which there's a ceiling within the same room. Move it away. Place Lara on the tile where the pushblock was. Save and load. The pushblock should be moved away, and Lara shouldn't void.
12. Have a pushblock standing on top of a pushblock located in a room underneath, above which there's a ceiling connected through a portal. Move it away. Place Lara on the tile where the pushblock was. Save and load. The pushblock should be moved away, and Lara shouldn't void.
13. Have a pushblock standing on top of a pushblock within the same room, above which there's empty space within the same room. Stand on top of it. Save and load. Lara should still be standing on top of the pushblock.
14. Have a pushblock standing on top of a pushblock within the same room, above which there's empty space connected through a portal. Stand on top of it. Save and load. Lara should still be standing on top of the pushblock.
15. Have a pushblock standing on top of a pushblock located in a room underneath, above which there's empty space within the same room. Stand on top of it. Save and load. Lara should still be standing on top of the pushblock.
16. Have a pushblock standing on top of a pushblock located in a room underneath, above which there's empty space connected through a portal. Stand on top of it. Save and load. Lara should still be standing on top of the pushblock.

As a table:

| Scenario number | Standing on another pushblock | Standable on? | Portal right below pushblock? | Portal right above pushblock? |
| --- | --- | --- | --- | --- |
| Scenario 1 | - | - | - | - |
| Scenario 2 | - | - | - | x |
| Scenario 3 | - | - | x | - |
| Scenario 4 | - | - | x | x |
| Scenario 5 | - | x | - | - |
| Scenario 6 | - | x | - | x |
| Scenario 7 | - | x | x | - |
| Scenario 8 | - | x | x | x |
| Scenario 9 | x | - | - | - |
| Scenario 10 | x | - | - | x |
| Scenario 11 | x | - | x | - |
| Scenario 12 | x | - | x | x |
| Scenario 13 | x | x | - | - |
| Scenario 14 | x | x | - | x |
| Scenario 15 | x | x | x | - |
| Scenario 16 | x | x | x | x |

In the "not standable on" scenarios we check if pushing the block away and placing Lara in its original place, saving and loading doesn't cause Lara to void.
In the "standable on" scenarios we check if standing on top of the pushblock, saving and loading causes Lara to spawn in the proper place.
In all scenarios we should check if the pushblock(s) can(not) be moved as expected.

walkawayy:
Ya it's tedious but we probably should do these test cases. Apparently there are issues on at least TR2 and TR3. 

https://streamable.com/75sla4

https://streamable.com/d3y489

rr-:
@walkawayy bump

walkawayy:
> @walkawayy bump

I need to solicit some help on building the test scenarios in a level. I tried learning Tomb Editor for a little but I was not getting very far haha.

Rapora9:
@walkawayy @rr- 
[gym-to-level6.zip](https://github.com/rr-/Tomb1Main/files/9901317/gym-to-level6.zip)
[level7a-to-level10c.zip](https://github.com/rr-/Tomb1Main/files/9901321/level7a-to-level10c.zip)

Here are the test levels. Gym is scenario 16, otherwise they come from level 1 = scenario 1 to level 10c = scenario 15. If gym doesn't work, rename it to level1 or something when testing that level. Also use level skip cheat to move forward in levels.

**Is everything set up properly?**

In every scenario (11, 12, 15, 16) where there is 2 pushables on top of each other so that they are in different rooms, you cannot push the upper pushable or it will fall and you get all kind of strange things. I don't know what causes this. Maybe I did something wrong when building the levels? Or maybe my T1M version is broken?

I decided to share this anyway now, so that I can fix any other stuff that is not done correctly in these test levels.

walkawayy:
Thanks so much! I'll test when I get time. Also @rr- could you clarify the phrase `Have a pushblock standing on a ceiling from a room underneath`? Rapora was wondering if this is the setup you meant by that.

image

rr-:
>Rapora was wondering if this is the setup you meant by that.

Yes that's exactly it.

Rapora9:
About levels/scenarios 11, 12, 15, 16. I was told that the blocks cannot be in different rooms. So these might not be possible in TR1. Is there any level with a situation like this? I thought the beginning of City of Khamun would be, but this is all same room.

Näyttökuva (165)

rr-:
> I was told that the blocks cannot be in different rooms. So these might not be possible in TR1. 

I don't know of any such level, but even if it's not possible to have with the current engine state, it doesn't mean it cannot be implemented. Having a test level that places these objects in two rooms would help us with it :)

Rapora9:
> I don't know of any such level, but even if it's not possible to have with the current engine state, it doesn't mean it cannot be implemented. Having a test level that places these objects in two rooms would help us with it :)

I see :) That would be a nice update for future levels. As always, I hope the test levels are useful.

walkawayy:
Ok all scenarios passed except three.

11 and 12: push block falls through the one below it if Lara tries to move it. Video: https://streamable.com/ecb7od

16: Lara can't jump on or grab the push block on top. If she moves it, it falls through the one below it like scenarios 11 and 12.

rr-:
@walkawayy if you're able to easily fix this, please go for it, otherwise I'm OK with creating a separate issue for it.

walkawayy:
> @walkawayy if you're able to easily fix this, please go for it, otherwise I'm OK with creating a separate issue for it.

Facepalm. I went to debug this issue and all the scenarios are working as intended... I think I forgot to rebuild when I switched branches and accidentally tested on `develop`. On this PR branch, all the scenarios work properly. When I switch back to `develop`, the scenarios I talked about above are broken. So, I think the current PR fix solves all the known movable block issues.

The only thing I noticed is some weird clipping in some scenarios when the block is near room portals. The block still functions properly, but it can look invisible at certain angles. But that's a whole separate camera issue from the OG game.

20221101_002030_Obelisk_of_Khamoon

20221101_002254_Caves

rr- (approved):

PR #628, opened 2022-09-23
bridge: avoid moving bridges sitting on vertical portals

### LINKED ISSUE #627: Misaligned bridge in City of Khamoon
In relation to #523, the flat bridge entity in City of Khamoon (item 82) is incorrectly moved up in position. It seems to be targeted as a bridge that is sitting on a floor (like those in Lost Valley), but in this case there is a room portal below so its position should remain as-is.
This is the only case outside of Lost Valley where a bridge is targeted for repositioning.

20220923_142440_City_of_Khamoon

image

lahm86:
Resolves #627.

#### Checklist

#### Description

The nature of `Room_GetFloor` means that if the `y` value is exactly that of the floor's `floor` value and there is a portal beneath, the function will move down into that room and find the floor sector from there (line 269 onwards in `room.c`). So when seeking the sector that matches a bridge's location, we now pass the `y` value of the *top* of the bridge so that `Room_GetFloor` will not traverse the portal, and so will return the sector from the same room.
An alternative would be to check if `room_num != item->room_number` after having called `Room_GetFloor`, but I decided against that as it may not work properly if items have been defined with incorrect room numbers (although as I understand the original issue was to target an OG bug in Lost Valley so this itself may not be relevant).
For testing, I added logging to `Bridge_FixEmbeddedPosition` before amending the code and then loaded each level. Only the 3 items in Lost Valley and 1 in Khamoon were targeted.
After the code change, I repeated the test and only the Lost Valley items were then targeted, and their resultant Y values remained consistent. I checked the positions in-game as well.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - added gameflow option remove_ammo to remove all shotgun, magnum and uzi ammo from the inventory on level start (#599)
 - added gameflow option remove_medipacks to remove all medi packs from the inventory on level start (#599)
 - improved the UI frame drawing, it will now look consistent across all resolutions and no longer have gaps between the lines
+- fixed bridge item in City of Khamoon being incorrectly raised (#627)
 
 ## [2.10.3](https://github.com/rr-/Tomb1Main/compare/2.10.2...2.10.3) - 2022-09-15
 - fixed save crystal mode always saving in the first slot (#607, regression from 2.8)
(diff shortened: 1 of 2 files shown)


rr-:
Great fix. The original change was about getting the bridges in Lost Valley not to embed in the floor near the trex spawn. I know they don't change their y pos but just to be safe, can you see if this still works?

walkawayy:
Fix LGTM. Double checked with ChocolateFan in Discord.

walkawayy (approved):

lahm86:
> Great fix. The original change was about getting the bridges in Lost Valley not to embed in the floor near the trex spawn. I know they don't change their y pos but just to be safe, can you see if this still works?

Yes, sorry I should have included some visuals of this. Everything looks good in Lost Valley, and I've included Khamoon below as well.

https://streamable.com/twh97f

rr- (approved):

Issue #633, opened 2022-09-30
Config tool


rr-:
To my surprise we don't seem to have had logged an issue for this, but that's the goal of 2.11.

In this ticket, we want to improve the user experience around tinkering with the game configuration.

- Merge config JSON and runtime config JSON
- Create a GUI tool to edit the configuration

Technical details:

- GUI uses C# and WPF like our other GUI project, the installer
- We use GitHub Actions to automate building and shipping of the configuration tool
- We should merge/move `src/game/settings.c` to `src/config.c`
- Hold one global struct for all the settings
- Game reads the settings at the start, updates the struct upon certain player actions (changing game volume, rebinding the keys), and writes the struct back into a JSON file
- Same with the config tool
- Data races are tolerable (e.g. if the user tweaked the settings using the config tool while the game was running, it's OK if adjusting the music volume ingame discards these changes)
- Losing the comments in the config JSON is OK, the users are not expected to be editing the file any longer
- We still should pretty-print the configuration file for the tech-savvy
- Backwards compatibility is not important, but we need to warn the users about it with big flashy warnings when the 2.11 release is complete
- We need to pay attention to the GitHub Action new version release script, that it packages the new configuration correctly
- We need to pay attention to the installer, that it understands the files it needs to ship upon fresh installation and an update
- No new settings or features as part of this ticket

 @lahm86 has volunteered to create a tool for us :heart:

Richard-L:
Oh wow, thank you sooooo much @lahm86 🙏

Other than an issue we had these two discussions:

Mainly: https://github.com/rr-/Tomb1Main/discussions/402

Also: https://github.com/rr-/Tomb1Main/discussions/247

walkawayy:
Done and released in 2.11.

PR #637, opened 2022-10-17
config: merge settings and add a .NET config tool

lahm86:
Part of #633.

#### Checklist

#### Description

The config tool allows editing `Tomb1Main.json5` and internal changes mean that all settings that are set inside the game are also now stored in this file. The main changes to the game to support this move are detailed below.

* All settings previously loaded/saved in `settings.c` have been moved to `config.c`.
* `Config_Init` function has been added to set specific settings at the same stage as they were being set in `Settings_Read`. This is primarily `GFX_Context_SetVSync`, as calling this during `Config_Read` causes a crash because `Output_Init` has not been called at that stage.
* Some `bool` settings have been inverted in order to maintain readability and consistency in the config tool GUI.
* `Config_Write` replaces `Settings_Write` throughout, and this writes *all* global settings to file.

The comments in the default config file have been updated in line with the GUI; this may be redundant, but it maintains consistency for those who may not wish to use the config tool.


bin/cfg/Tomb1Main.json5
```diff
@@ -220,19 +220,19 @@
...
+    // Makes Lara revert to pistols at the start of each level (in line with TombATI behavior).
     "revert_to_pistols": false,
 
-    // Enhances savegames so graphic effects, waterfall mist, flame emitters, and
+    // Enhances savegames so that graphic effects, waterfall mist, flame emitters, and
     // more are saved instead of disappearing on load.
     "enable_enhanced_saves": true,
 
-    // Sound effects are randomly, slightly pitched to vary the game sounds.
+    // Allows sound effects to be randomly, slightly pitched to vary the game sounds.
```
rr-: Do we need the comments in this file considering that it's designed to get written back to, and all the comments will get discarded?

Actually, do we even need to ship this file, considering that 
- the user is expected to use the config tool
- the game should do just fine without it, using the defaults for each setting
- the game will create the config file for the power users upon first launch

?

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## [Unreleased](https://github.com/rr-/Tomb1Main/compare/stable...develop)
 - added graphics effects, lava emitters, flame emitters, and waterfalls to the savegame so they now persist on load (#418)
 - added an option to turn off sound effect pitching (#625)
+- added a .NET-based configuration tool (#633)
```
rr-: I'd move it to the top, as it's the crown of this release.

src/config.c
```diff
@@ -230,3 +303,115 @@ bool Config_Read(void)
...
+    struct json_value_s *root = json_value_from_object(root_obj);
+    char *data = json_write_pretty(root, "  ", "\n", &size);
+    json_value_free(root);
+
+    File_Write(data, sizeof(char), size - 1, fp);
+    File_Close(fp);
+    Memory_FreePointer(&data);
+
+    return true;
+}
```
rr-: Missing end of the line character

src/config.h
```diff
@@ -130,3 +130,5 @@ extern CONFIG g_Config;
 
 bool Config_ReadFromJSON(const char *json);
 bool Config_Read(void);
+void Config_Init(void);
+bool Config_Write(void);
```
rr-: Missing end of the line character

tools/config/Tomb1Main_ConfigTool/Models/Specification/Specification.cs
```diff
@@ -0,0 +1,44 @@
...
+public class Specification
+{
+    public Dictionary<string, List<EnumOption>> Enums { get; private set; }
+    public List<Category> CategorisedProperties { get; private set; }
+    public List<BaseProperty> Properties { get; private set; }
+
+    public Specification(string sourceData)
+    {
+        JObject data = JObject.Parse(sourceData);
+        Dictionary<string, List<string>> enumTypes = JsonConvert.DeserializeObject<Dictionary<string, List<string>>>(data[nameof(Enums)].ToString());
```
rr-: The lines seem to be a bit long, can you run it through some autoformatter? (Optional, best C# autoformatters are proprietary AFAIR)

bin/cfg/Tomb1Main.json5
lahm86: I'd be in favour of removing the file altogether - it'd be one less place to have to define future settings. The only issue I can think of is if you install the game and launch the config tool before running the game - in this case it won't generate the file by default, you would have to open a file manually. But this could easily be changed so that it does generate it. Or the expectation could be that the user must launch the game once before running the config tool?

rr-: I'd rather have the config tool work ootb without requiring the user to launch the game (and use good defaults).

src/config.c
lahm86: Resolved in [4ee0f5a](https://github.com/rr-/Tomb1Main/pull/637/commits/4ee0f5a3487067af292d756b533979a5b9ffd017)

src/config.h
lahm86: Resolved in [4ee0f5a](https://github.com/rr-/Tomb1Main/pull/637/commits/4ee0f5a3487067af292d756b533979a5b9ffd017)

CHANGELOG.md
lahm86: Resolved in [b27bc7c](https://github.com/rr-/Tomb1Main/pull/637/commits/b27bc7c9fb0f28786dbbb609a64194a4238523ff)

tools/config/Tomb1Main_ConfigTool/Models/Specification/Specification.cs
lahm86: Updated in [87bf6ae](https://github.com/rr-/Tomb1Main/pull/637/commits/87bf6ae5de281eb53de6382941738394d8b1b2f6).
This should be a bit more readable now. Unfortunately the second `JsonConvert.DeserializeObject` call has to remain as it requires the custom property converter, whereas the enums can be parsed more simply with `JObject`.

bin/cfg/Tomb1Main.json5
lahm86: The default configuration file is now removed. The config tool will launch normally if it can't locate the default file, and it will have all properties initialised to their defaults (as defined in `specification.json` in the project resources). If you make a change in the tool and save, it will create the file. See [b27bc7c](https://github.com/rr-/Tomb1Main/pull/637/commits/b27bc7c9fb0f28786dbbb609a64194a4238523ff).

I had to make a small change in the game as without the default config file it wouldn't load. I believe the issue was that the screen resolution wasn't being set. When it couldn't find the config file, it was jumping to the `cleanup` label, so while the default settings were being loaded, it was skipping over the `g_AvailableResolutions` initialisation.
See [4ee0f5a](https://github.com/rr-/Tomb1Main/pull/637/commits/4ee0f5a3487067af292d756b533979a5b9ffd017).

I adjusted the log message here to become a warning rather than an error, and it will now pass `"{}"` to `Config_ReadFromJSON` instead of an empty string to avoid a further error message in that function. The warning message line breaches the 80 column rule so `make lint` shifted it over 3 lines. The wording can perhaps be amended here if this is not ideal.

rr- (approved):
LGTM. Now we need to integrate with GH Actions. My suggestion is to merge this PR and have me create a follow-up PR that integrates the release GH Action. What do you think?

lahm86:
> LGTM. Now we need to integrate with GH Actions. My suggestion is to merge this PR and have me create a follow-up PR that integrates the release GH Action. What do you think?

That sounds good to me, thanks!

PR #640, opened 2022-10-21
Audio convert fix

### LINKED ISSUE #631: Sounds click and pop if they stop prematurely
What I mean is, for example, here in level 3 

https://youtu.be/byJ5to20C6I?t=1751

when Lara starts cog mechanism game uses sfx which plays and stops several times in a row, and in that automated fix version it works smoothly enough, but in tomb1main it clicks every time it replays. Same thing happens with other sounds as well, like when boulder rolls. I guess there needs to be implemented feature like earlyStopFadeOutTime Indirectsound has

oziphantom:
#### Checklist

#### Description
Tiny fix to the audio conversion function. I don't think it is worth mentioning in the changelog.

There was a report about clipping noise in the rolling ball sfx. So I started to investigate after you inquired if I had any ideas. I did find something amiss, and this code is to rectify it. However the noise at the end of the ball persists. Upon checking TR1 Dos from Steam, it is an "original bug" . The crackle is from the sample not being long enough and not ending on a 0. This is present in the Steam level files.  Although maybe they were not talking about it, and were hearing the missing 32 samples in the middle. 

Another point of note, 

```
sample->num_samples =
        working_buffer_size / sizeof(swr.dst_channels) / swr.dst_channels;
```
getting the size of the swr.dst_channels is wrong because 
a.) you want the format size not the size of the channel
b.) you want the size it would take, not how big the variable that holds said format specifier is. 
Luckily you are converting to float and holding the value in an int when only compiled in x86 and not x64 and hence it magically worked. I have added the correct code for you.


rr-:
Do I understand it correctly that this is more of a cosmetic fix so that the code is more correct, and there's no audible effect in the game? E.g. #631 still persists even with this change, because of the audio files distributed in the Steam version?

oziphantom:
there is a code fix in this, the original sample code was wrong, and was missing samples in some cases, I didn't test with cogs, but rather the rolling ball which I noticed had clicking issues and I though was mentioned in the github issue I looked at. It was also easy to isolate, the cogs are annoying as the river ambient sounds are also playing. However the rolling ball audio issues where also in the original. 
For the cogs in does perform a `SOUND_MODE_RESTART` command which due to the frame timed nature of the event and things being resampled, will probably cause such events to become out of sync giving the issue we have. I know of no way to fix this other than patch the level data to put the reset on the cog animation latter in the animation, however this still has the issue that it will fire "1 animation loop amount of time" away from each other, so you would also have to slow down the animation to match the resampled time. 
Alternatively we could ignore the restart and instead fire off a new instance of the sound and then mark the current playing one for a rapid fade out. It would still sound a little odd due to the slight overlap but wouldn't have the nasty pop it has now. 
Or possibly we work out the sync point and then add some FSM logic to the audio, so that it resets again on the next point it should, which in this case is probably just "when it finishes", so a counter for "number of times to play" might suffice.

walkawayy:
Can this be merged? It doesn't fix the sound FX popping, but it does fix another issue with the audio process.

rr- (approved):

PR #642, opened 2022-10-23
Add additional control schemes and refactor

### LINKED ISSUE #636: Increase number of custom control schemes
Currently only one custom control scheme is allowed. Most modern games support more. Also, if the user wants to use a controller to play, that takes up the one custom scheme. More custom schemes would allow a custom keyboard and controller scheme, multiple people sharing a system to have their own scheme, etc.

This is also in preparation for #564 and will require some code refactors.

walkawayy:
#### Checklist

#### Description
Added 2 additional custom control schemes (default + 3 custom schemes total). Added customizable keybinds. Refactored `option_control.c` to support this feature and prepare for #564.

- [X] Refactor `option_control.c`.
- [X] Add custom keybinds for drawing each weapon, small medipack, large medipack, quick save, quick load, FPS, and bilinear filter.
- [X] Add 2 additional custom control schemes for a total of 3 custom schemes.
- [X] Remove the two column control menu and change to a scrolling list to fit all the new keybinds.
- [x] Lock controls when navigating the controls option to prevent medipack usage, toggling the FPS, toggling bilinear filter, etc.
- [x] Bind action and select to the same key.

Future PRs:
- Add the ability to unbind non essential keys (such as camera controls, FPS, etc).
- Add a reset to defaults button for all keys in the current scheme.
- Don't fallback to default keys (example, if user rebinds `alt`, the game shouldn't do the custom key and also still jump)
- Test remapping a controller.

...


rr-:
@walkawayy can I ask for a video demo what this PR accomplishes, so that I have a better understanding of what I'm reviewing? Thanks.

walkawayy:
Yes sure, sorry. Here's a video of this PR's new control schemes + additional custom keybinds. There's now 3 custom control schemes + the default scheme (total of 4 control schemes).

https://streamable.com/5rmfic

I'm trying to take inspiration from an accessible modern game (The Last of Us): https://www.youtube.com/watch?v=qbkRptgqQg8

My future plans include removing the default key always functioning issue (#564). To solve the problem of the user accidentally breaking a control scheme, there will be a reset to defaults button. I also want to add a similar button to unbind a singular key. For example, the user might want to unbind the camera movement keys. Also, I probably need to make only certain keys unbindable (the user shouldn't be able to unbind move forward or jump for example). Here's a super early WIP demo of a reset to defaults button concept.

https://streamable.com/258yux

src/config.c
```diff
@@ -259,14 +261,19 @@ bool Config_ReadFromJSON(const char *cfg_data)
...
-        json_object_get_array(root_obj, "layout_sdl");
-    for (INPUT_ROLE role = 0; role < INPUT_ROLE_NUMBER_OF; role++) {
-        INPUT_SCANCODE scancode =
-            Input_GetAssignedScancode(INPUT_LAYOUT_USER, role);
-        scancode = json_array_get_int(layout_arr, role, scancode);
-        Input_AssignScancode(INPUT_LAYOUT_USER, role, scancode);
+    char *layout_name = Memory_Alloc(CONTROL_SCHEME_SIZE);
+    for (INPUT_LAYOUT layout = INPUT_LAYOUT_CUSTOM_1;
+         layout < INPUT_LAYOUT_NUMBER_OF; layout++) {
+        sprintf(layout_name, "%s_%d", "layout_sdl", layout);
```
rr-: Let's inline the literal `layout_sdl` to `"layout_sdl_%d"`

src/config.c
```diff
@@ -259,14 +261,19 @@ bool Config_ReadFromJSON(const char *cfg_data)
...
     CLAMP(g_Config.ui.bar_scale, MIN_UI_SCALE, MAX_UI_SCALE);
 
-    struct json_array_s *layout_arr =
-        json_object_get_array(root_obj, "layout_sdl");
-    for (INPUT_ROLE role = 0; role < INPUT_ROLE_NUMBER_OF; role++) {
-        INPUT_SCANCODE scancode =
-            Input_GetAssignedScancode(INPUT_LAYOUT_USER, role);
-        scancode = json_array_get_int(layout_arr, role, scancode);
-        Input_AssignScancode(INPUT_LAYOUT_USER, role, scancode);
+    char *layout_name = Memory_Alloc(CONTROL_SCHEME_SIZE);
```
rr-: No need to allocate it dynamically, just allocate it on stack.
`char layout_name[20];`
It's also OK to hardcode `CONTROL_SCHEME_SIZE` in this case, and I like to round the array sizes to increments of 10 hence `20` not `13`.

src/config.c
```diff
@@ -259,14 +261,19 @@ bool Config_ReadFromJSON(const char *cfg_data)
...
+        sprintf(layout_name, "%s_%d", "layout_sdl", layout);
+        struct json_array_s *layout_arr =
+            json_object_get_array(root_obj, layout_name);
+        for (INPUT_ROLE role = 0; role < INPUT_ROLE_NUMBER_OF; role++) {
+            INPUT_SCANCODE scancode = Input_GetAssignedScancode(layout, role);
+            scancode = json_array_get_int(layout_arr, role, scancode);
+            Input_AssignScancode(layout, role, scancode);
+        }
     }
+    Memory_FreePointer(&layout_name);
```
rr-: If you allocate it on stack, this is no longer necessary.

src/config.c
```diff
@@ -398,12 +405,18 @@ bool Config_Write(void)
...
+        struct json_array_s *layout_arr = json_array_new();
+        sprintf(layout_name, "%s_%d", "layout_sdl", layout);
+        for (INPUT_ROLE role = 0; role < INPUT_ROLE_NUMBER_OF; role++) {
+            json_array_append_int(
+                layout_arr, Input_GetAssignedScancode(layout, role));
+        }
+        json_object_append_array(root_obj, layout_name, layout_arr);
     }
-    json_object_append_array(root_obj, "layout_sdl", layout_arr);
+    Memory_FreePointer(&layout_name);
```
rr-: Similar remarks as above

src/game/option/option_control.c
```diff
@@ -234,16 +282,36 @@ static void Option_FlashConflicts(void)
...
         Text_Flash(
-            m_TextB[item->option],
+            m_TextKeyNames[item->option],
             g_Config.input.layout != INPUT_LAYOUT_DEFAULT && item->option != -1
                 && Input_IsKeyConflictedWithUser(item->option),
             20);
     }
 }
 
+static void Option_ControlChangeScheme()
```
rr-: Argument-less functions take `(void)` in C

src/game/option/option_control.c
```diff
@@ -234,16 +282,36 @@ static void Option_FlashConflicts(void)
...
+        g_InputDB.left && g_Config.input.layout == INPUT_LAYOUT_DEFAULT) {
+        g_Config.input.layout = INPUT_LAYOUT_NUMBER_OF - 1;
+    } else if (
+        g_InputDB.right && g_Config.input.layout < INPUT_LAYOUT_NUMBER_OF - 1) {
+        g_Config.input.layout += 1;
+    } else if (
+        g_InputDB.right
+        && g_Config.input.layout == INPUT_LAYOUT_NUMBER_OF - 1) {
+        g_Config.input.layout = INPUT_LAYOUT_DEFAULT;
+    }
```
rr-: This block can be simplified to
```c
if (g_InputDB.left || g_InputDB.right) {
    g_Config.input.layout += g_InputDB.left ? -1 : 0;
    g_Config.input.layout += g_InputDB.right ? 1 : 0;
    g_Config.input.layout += INPUT_LAYOUT_NUMBER_OF;
    g_Config.input.layout %= INPUT_LAYOUT_NUMBER_OF; 
}
```

src/game/option/option_control.c
```diff
@@ -12,22 +12,45 @@
...
-#define TOP_Y -60
+#define TOP_Y -120
 #define BORDER 4
 #define HEADER_HEIGHT 25
 #define ROW_HEIGHT 17
 
+#define KM_BROWSE 0
+#define KM_BROWSEKEYUP 1
+#define KM_CHANGE 2
+#define KM_CHANGEKEYUP 3
```
rr-: Maybe an enum?

rr- (commented):
- How does this look at lower resolutions?
- It's still not possible to rebind certain keys (F2 and F3).
- I'd remove the F11 cheat (not in this change request).
- In the future I'd like this configuration feature to support game controller and make it possible to natively rebind pad buttons.

Otherwise, LGTM.

Richard-L:
Very beautiful @walkawayy 🙏
Even more like the PS1 with the three schemes.

Is it worth considering to make "custom camera" an optional feature, and if turned off hide these binds here entirely? Would help debloat it a bit.

I also wonder up to which UI scaling this menu will still not overflow.

src/config.c
walkawayy: Speaking of this, I realized I named the new controls `Custom Scheme X` since I was looking at The Last of Us's naming conventions when I made this. But the current code mostly uses `layout`. What name do you want to stick with? I'll go through and refer to everything as `layout` or `scheme` depending on what you prefer.

walkawayy:
It looks ok at the lowest resolution of 640x480.

image


However, it's definitely an issue with increased UI text scale even at higher resolutions.

image

walkawayy:
To fix the UI scaling issue, I think we'd have to write an entirely new controls menu for text scaling over `1.2`. It would most likely have to function similarly to the passport requestor where the user can scroll through the options.

rr-:
I'm OK with this as it's a more scalable solution, should we need any more (!) keys.
(With the ability to bind them however we want, we can add more interesting cheats.)

I believe this will significantly simplify the controls menu code, too, since we won't have to deal with four columns of text and instead just two, one of which is interactive.

It could also pave the way towards in-game configuration of things.

src/config.c
rr-: I'd rather not change things too much so I'd stick with layout. It's OK to deviate from this just for the labels, though.

Also since we're dropping backwards compatibility here anyway, I don't think it makes sense to keep the `_sdl` suffix anymore. It was used to tell the SDL input mappings apart from the old DirectInput-based mappings – I didn't want the numeric values that were valid for DirectInput to be used for SDL configuration when people upgraded their games, because it'd be a complete chaos, so I introduced a new configuration key for the new way. Not only the DirectInput code is long gone so we don't need to worry about people upgrading versions, these new custom layout options will be new, too, so they don't need this trick.

walkawayy:
I rewrote the controls menu to use a single column, scrolling menu. I had to make the text scaling .8 to make sure it could all fit on the screen at all resolutions and at all UI scaling. Hopefully the code isn't that much of a nightmare.

Video:
https://streamable.com/f4zezn

Screenshot at 480p resolution at max 2x UI scaling:
20221106_201645_Title

Richard-L:
That's sweet walkaway, but a certain break of course with vanilla TR1. Worth it here for what it enables.

Two thoughts:
- if this were CSS you would make it so the 1, 2, n column layout changes dynamically based on available viewport width and what font scaling is selected. You'd want to allow users with a very large monitor to benefit from it and see more without having to scroll.
- it feels as though the gap between input and assigned key is quite large now. Had you tried a right align on the left side? Means even more visual departure from vanilla, but could reduce eye travel when reading.

rr-:
Can we flip the columns with each other and align the assigned inputs to the right?

walkawayy:
Flipped the columns and aligned the assigned inputs to the right. Looks much nicer. I also no longer have to scale the controls text down to .8 with this change. I also just did a few more code cleanup things. Here's a pic:

image


Here's a few questions I have:
1. The perspective toggle is still hard coded to F4. This option currently doesn't do anything, so I left it alone. Maybe this is just something to note in case we ever add functionality to that key in the future. Or I can add it to the list of custom keys now.
2. The `select` input is hard coded to `SDL_SCANCODE_RETURN`. Some of the UI accepts `action` and `select` but some things like the requestor only accept `select`. Should we map `action` and `select` to the same key? It looks like the controller does this to `SDL_CONTROLLER_BUTTON_A`, so I think that should work. `option` and `deselect` are mapped to the same key already.
3. When remapping a control, the button's functionality still happens. For example, assigning draw weapons to `F2` will also toggle the FPS counter during the remapping process. IMO the key shouldn't perform its function during the remapping.

@rr- lmk what you think. I also added some tasks to the PR. Though I'm not sure if all of them are suited for this PR. I just wanted to list out possible plans for the future and get your input.

rr-:
1. Let's leave F4 hardcoded for now.
2. I think that's OK. Can you show examples of UI that doesn't accept action, but accepts select?
3. I agree, it shouldn't. There should be very few scenarios for this though, as most of the keybindings apply to the game, not the UIs.

The tasks look great, but I'm not sure if you want to tackle all of them in this PR?

walkawayy:
In `input.c`, the function `Input_Update` actually already does `g_Input.select |= g_Input.action;`. So, pressing `return` is hard coded to always perform the `select` functionality. But, the user can set `action` to any key and it will always act like `select`. This means there's some actions like selecting an inventory object like the passport that accept `select` so `return` will always perform that. LMK if you want me to do anything about it. The current problem with this is the `return` key can never be customizable and will always conflict if the user wants to rebind it.

I think the three possible solutions are to bind `select` and `action` to the same `INPUT_ROLE_ACTION`, make a separate keybind (not on controller?) for the `select` role, or we could remove `action` or `select` entirely and use one everywhere but that's a massive change. My vote is to bind them to the same key, since the game kind of does that already. The issue with this would be that the user can't navigate the main menu, skip cutscenes, etc with the `return` key anymore, so I'm not sure that would be well received. : /

To finalize this PR, I think we should decide on what to do with the `return` key's hard bind to `select`. And we should remove the key press activating its functionality during the remapping process. Then, some of the other tasks I listed can be solved in future PRs. Maybe the next PR can be adding an unbind a non required role button and a reset to defaults button. Then the next one can be to remove the fallback to default keys. Finally, I will investigate the stretch goal to see if combo button keybinds are possible (ex: Triangle + Up dpad to draw shotgun) and would make a new PR for that.

rr-:
Let's bind them to the same key.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## [Unreleased](https://github.com/rr-/Tomb1Main/compare/stable...develop)
 - fixed small cracks in the UI borders for PS1-style menus (#643)
 - fixed Lara loading inside a movable block if she's on a stack near a room portal (#619)
+- added additional custom control schemes (#636)
```
rr-: Let's move this to the top.

src/config.c
```diff
@@ -17,8 +17,10 @@
 #include "specific/s_shell.h"
 #include "util.h"
 
+#include <stdio.h>
 #include <string.h>
 
+#define CONTROL_LAYOUT_FMT "layout_%d"
```
rr-: I think this define is a bit overzealous since we don't reuse other hardcoded setting names either between the read/write routines.

src/game/option/option_control.c
```diff
@@ -15,419 +15,545 @@
...
-        70, TOP_Y - BORDER + (HEADER_HEIGHT + BORDER - ROW_HEIGHT) / 2, "\201");
-    Text_CentreH(m_TextArrowRight, 1);
-    Text_CentreV(m_TextArrowRight, 1);
+    int32_t visible_lines = 0;
+    int32_t height = Screen_GetResHeight();
+    if (height <= 480) {
+        visible_lines = 10;
+    } else {
+        visible_lines = 12;
+    }
```
rr-: I think these lines can be safely removed considering how we overwrite `visible_lines` just below.

src/game/option/option_control.c
```diff
@@ -15,419 +15,545 @@
...
+                ? m_Text[TEXT_TITLE]
+                : m_ControlMenu.role_texts[m_ControlMenu.cur_row],
+            0, 0, 0, 0, TS_REQUESTED);
+        Text_AddOutline(
+            m_ControlMenu.cur_row == KC_TITLE
+                ? m_Text[TEXT_TITLE]
+                : m_ControlMenu.role_texts[m_ControlMenu.cur_row],
+            true, TS_REQUESTED);
+        break;
+    case KM_BROWSEKEYUP:
```
rr-, line 343: Let's separate consecutive case blocks with empty lines.

rr- (approved):
The code looks good to me apart from minor comments.

walkawayy:
I addressed your changes.

Lmk what you think of [61c0578](https://github.com/rr-/Tomb1Main/pull/642/commits/61c057878155d567be8bf39636950bf169f2c8d1). This locks some of the input while on the controls screen so the buttons don't trigger during remapping. Before, if the user pressed buttons like small medipack, toggle FPS, toggle bilinear filter, etc... the game would activate those effects while on the controls screen.

I also allowed scrolling of the default controls since they aren't all visible at once anymore. The user can scroll but can't remap.

rr-:
I think that `Option_ControlIsLocked` can read the internal state, `KC` or `KM` and return true only if it's in the key assignment state. WDYT?

walkawayy:
> I think that `Option_ControlIsLocked` can read the internal state, `KC` or `KM` and return true only if it's in the key assignment state. WDYT?

That makes sense. To me toggling FPS should work unless we are in the assignment stage, but stuff like quick medipack shouldn't on the controls screen.

Maybe we should move the quick weapon and quick medi input checks out of `Input_Update`? That way I can make the change you want, but also Lara won't use medipacks or equip guns during the controls screen. What are your thoughts?

rr-:
I don't understand the question; the implementation of Input_Update is fine as it is. I meant that it may not be necessary to introduce `m_ControlLock` for `Option_ControlIsLocked` to work and instead it can `return m_KeyMode == KM_CHANGE` or similar, and this change shouldn't affect the game's behavior?

In any case I don't want to further decouple things from `Input_Update`. The logic that it currently performs should stay where it is.

walkawayy:
> I don't understand the question; the implementation of Input_Update is fine as it is. I meant that it may not be necessary to introduce `m_ControlLock` for `Option_ControlIsLocked` to work and instead it can `return m_KeyMode == KM_CHANGE` or similar, and this change shouldn't affect the game's behavior?
> 
> In any case I don't want to further decouple things from `Input_Update`. The logic that it currently performs should stay where it is.

Ok. My point was to say that checking `return m_KeyMode == KM_CHANGE` means the game would still perform actions like use small medipack if the hotkey is pressed on the controls screen which seems odd to me. At least it would stop that from happening during the key rebind step. Want me to make the change to your suggestion?

My point was to say that `m_ControlLock` would stop keys like Turbo speed, toggle FPS, small medipack, etc from working at all on the controls screen because they all get checked in `Input_Update`. Your `return m_KeyMode == KM_CHANGE` method would only stop them from working during the single rebind step, but they would work while browsing the controls menu. It just depends which way we want it.

rr-:
I see, I didn't think of that. We could also reuse the mode to introduce a new state, `KM_INACTIVE`, set it to inactive upon closing the controls menu, and have `IsLocked` return `m_KeyMode != KM_INACTIVE`.

PR #651, opened 2022-11-21
shell: fix a game crash if the action button is held down

### LINKED ISSUE #646: Holding down select during game exit crashes the game
Holding the `select` key during `Shell_Shutdown` spins a call to `Input_Update()` in a while loop. I tested holding it after selecting exit game from the passport, and the game got stuck in a black screen and had a Windows crash error. Is there a reason the game does this at all and doesn't just do a shutdown? I checked and this seems like an OG bug (at least a TombATI bug).

walkawayy:
Resolves #646.

#### Checklist

#### Description
Fixed a game crash on shutdown if the action button is held down.
...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 - added additional custom control schemes (#636)
 - fixed small cracks in the UI borders for PS1-style menus (#643)
 - fixed Lara loading inside a movable block if she's on a stack near a room portal (#619)
+- fixed a game crash on shutdown if the action button is held down (#646)
 
 ## [2.11](https://github.com/rr-/Tomb1Main/compare/2.10.3...2.11) - 2022-10-19
 - added a .NET-based configuration tool (#633)
--- a/README.md
+++ b/README.md
@@ -273,6 +273,7 @@ Not all options are turned on by default. Refer to `Tomb1Main_ConfigTool.exe` fo
 - fixed bridges at floor level appearing under the floor
 - fixed underwater currents breaking in rare cases
 - fixed Lara loading inside a movable block if she's on a stack near a room portal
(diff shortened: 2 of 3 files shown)


rr- (approved):

rr- (commented):
Would be good to mention the version that broke this.

walkawayy:
It's an OG bug. The same problem exists in TombATI.

https://github.com/rr-/Tomb1Main/commit/44edb5a51020da23a85564cd6d03227b285d9aa9

rr-:
I know the code is there looking similar, but I'm pretty sure it used not to freeze the game.

walkawayy:
I'll test again later tonight to confirm, but I did test TombATI when I opened the issue. I held down `Control` as I exited the game and it spikes your CPU usage and the game crashes. I also couldn't find `WinVidSpinMessageLoop` when searching Github since it seems that function is gone, so I didn't look into what that was doing.

walkawayy:
Finally tracked it down... this is what `WinSpinMessageLoop` was originally doing from https://github.com/rr-/Tomb1Main/commit/5849a1ba46287762f2197360481ac12e2f57cb34. I have no idea what the code is doing however.
```
int32_t WinSpinMessageLoop()
{
    MSG msg;
    while (PeekMessageA(&msg, 0, 0, 0, PM_NOREMOVE)) {
        if (!GetMessageA(&msg, 0, 0, 0)) {
            TerminateGame(0);
        }

        TranslateMessage(&msg);
        DispatchMessageA(&msg);
    }

    int32_t time_ms = timeGetTime();
    int32_t old_ticks = Ticks;
    Ticks = time_ms * 2 * 30 / 1000;
    return Ticks - old_ticks;
}
```

PR #655, opened 2022-12-04
save_crystal: fixed to be single use and added collision

### LINKED ISSUE #654: Save crystals are reusable and have no object collision
On original game (Saturn version but probably also PS1) the crystals were unique usage: once saved, the selected crystal is gone forever, forcing the player to wisely choose when to save.

walkawayy:
Resolves #654.

#### Checklist

#### Description
Fixed save crystals so they are single use and added object collision.

Here's a video. The first interaction I deselected which is why the crystal stayed. I also check if the save counter increases because deselected without saving and saving both return `GF_NOP`, so the `Inv_Display` is not really usable. And I didn't want to introduce a new gameflow option and risk breaking other things. I also removed the puzzle key sound effect that was being used since the PS1 version has no sound effect.

https://streamable.com/v0cshe
...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - fixed Lara loading inside a movable block if she's on a stack near a room portal (#619)
 - fixed a game crash on shutdown if the action button is held down (#646)
 - fixed the compass and new game menus at high text scaling (#648)
+- fixed save crystals so they are single use and added object collision (#654)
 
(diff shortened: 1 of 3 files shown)


walkawayy:
Hm wait. The save crystal is brought back upon reload.

walkawayy:
And fixed. Have to deactivate it before the save occurs, and restore it if the user didn't save.

rr- (commented):
Let's add an information to the changelog that the sound did get removed.

rr- (approved):

PR #658, opened 2022-12-08
Add reset and unbind options to the controls menu

walkawayy:
#### Checklist

#### Description
Added the ability to unbind unessential keys and added the ability to reset control schemes to default. There's a lot of changes in there so lmk what you think. Also shoutout to Ozi for helping debug the depth buffer issue. The progress bar needs to be drawn with the other text so it shows up in front of the darkened background.

PC UI video:
https://streamable.com/rd49ex

PS1 UI video:
https://streamable.com/vp7rye
...


src/game/overlay.c
```diff
@@ -213,7 +206,14 @@ static void Overlay_BarDraw(BAR_INFO *bar_info)
...
     int32_t y = 0;
-    Overlay_BarGetLocation(bar_info->location, width, height, &x, &y);
+    if (bar_info->location == BL_CUSTOM) {
+        width = bar_info->custom_width;
+        height = bar_info->custom_height;
+        x = bar_info->custom_x;
+        y = bar_info->custom_y;
+    } else {
+        Overlay_BarGetLocation(bar_info->location, width, height, &x, &y);
+    }
```
rr-: Can we change `Overlay_BarGetLocation` to also make width and height writable, and move the `BL_CUSTOM` case inside?

rr- (commented):
Very good!

src/game/option/option_control.c
```diff
@@ -16,23 +17,17 @@
...
-} LAYOUT_NUM_GS_MAP;
-
-static const LAYOUT_NUM_GS_MAP m_LayoutMap[] = {
-    { INPUT_LAYOUT_DEFAULT, GS_CONTROL_DEFAULT_KEYS },
-    { INPUT_LAYOUT_CUSTOM_1, GS_CONTROL_CUSTOM_1 },
-    { INPUT_LAYOUT_CUSTOM_2, GS_CONTROL_CUSTOM_2 },
-    { INPUT_LAYOUT_CUSTOM_3, GS_CONTROL_CUSTOM_3 },
-};
+#define COLOR_STEPS 5
+#define RESET_ALL_KEY "R"
```
walkawayy, line 26: Is this ok to use / will it translate to foreign language keyboards if we ever get to supporting them?

src/game/option/option_control.c
```diff
@@ -426,6 +486,94 @@ static void Option_ControlChangeLayout(void)
...
+    if (txt->flags.centre_v) {
+        y += Screen_GetResHeightDownscaled() / 2;
+    } else if (txt->flags.bottom) {
+        y += Screen_GetResHeightDownscaled();
+    }
+
+    int32_t percent = (timer * 100) / (FRAMES_PER_SECOND * BUTTON_HOLD_TIME);
+    CLAMP(percent, 0, 100);
+    Text_AddProgressBar(
+        txt, width, height, x, y, percent, g_Config.ui.menu_style);
```
walkawayy, line 511: Do I need to add a check like:
```
if (percent) {
    Text_AddProgressBar(
        txt, width, height, x, y, percent, g_Config.ui.menu_style);
} else 
    Text_RemoveProgressBar();
}
```
? I think the way it's done now, the progress bar is always made which is why in `text.c` there's a double check for: `if (textstring->flags.progress_bar && textstring->progress_bar.value) {`.

src/game/overlay.c
```diff
@@ -80,21 +80,14 @@ static RGBA8888 m_ColorBarMap[][COLOR_STEPS] = {
       { 210, 130, 160, 255 },
       { 165, 100, 120, 255 },
       { 120, 60, 70, 255 } },
+    // purple
+    { { 52, 22, 80, 255 },
+      { 70, 30, 107, 255 },
+      { 52, 22, 80, 255 },
+      { 39, 17, 60, 255 },
+      { 26, 11, 40, 255 } },
```
walkawayy, line 88: Just making a note that in the future we might as well add support for another bar color in the config tool.

Btw, to determine the purple colors, I used the PS1 menu color as the brightest highlight and then used [this site](https://www.w3schools.com/colors/colors_picker.asp?colorhex=461e6b) to get darker percentages.

walkawayy (commented):
Tagged a few code snippets I wanted to ask specific questions about.

src/game/overlay.c
walkawayy: LMK if the change I made for this is OK. Since I pass `BAR_INFO` now.

src/game/option/option_control.c
rr-: No need for now, might be a subject to change in the future if we need more progress bars that need to be displayed at 0%.

src/game/option/option_control.c
rr-: I wouldn't worry about these since we don't even support Unicode at the moment.

rr- (approved):

PR #662, opened 2022-12-09
input: stop the default controls from functioning when unbound

### LINKED ISSUE #564: Stop unmapped keybinds having an effect on the game

walkawayy:
Resolves #564.

#### Checklist

#### Description
Stopped the default controls from functioning when the user unbound them.
...

### DIFF
--- a/src/game/input.c
+++ b/src/game/input.c
@@ -18,7 +18,6 @@ INPUT_STATE g_InputDB = { 0 };
 INPUT_STATE g_OldInputDB = { 0 };
 
 static bool m_KeyConflictWithUser[INPUT_ROLE_NUMBER_OF] = { false };
-static bool m_KeyConflictWithDefault[INPUT_ROLE_NUMBER_OF] = { false };
 static int32_t m_HoldBack = 0;
 static int32_t m_HoldForward = 0;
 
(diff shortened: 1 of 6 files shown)


src/game/input.c
```diff
@@ -63,16 +58,11 @@ void Input_CheckConflicts(INPUT_LAYOUT layout_num)
             if (role1 == role2) {
                 continue;
             }
-
             INPUT_SCANCODE scancode2_user =
                 Input_GetAssignedScancode(layout_num, role2);
-
```
rr-, line 69: `scancode1_user` and `scancode2_user` can be renamed to just `scancode1` and `scancode2` now.

rr- (approved):
Please squash when merging and use appropriate commit message :)

PR #674, opened 2022-12-18
option_control: added customizable controller support

### LINKED ISSUE #659: Allow customizable controller inputs
Cannot change the joypad buttons layout.

Ingame, pressing a button on a selected function like "Roll", will trigger the new input waiting, but then only pressing a keyboard key will be accepted. Pressing a joypad button at the new input waiting does not do anything.

Also, even tried to edit manually a layout to invert pads keys (in my case, I needed to invert 93 and 91 for Roll<-->Draw Weapon) from the file Tomb1Main.json5, but again, it does not make any changes.

Got the problem on main but still also on current develop branch.

Tried also my joypad in both XInput and DInput modes.

walkawayy:
Resolves #659.

#### Checklist

#### Description
Added customizable controller support with multiple control schemes.
...


src/global/types.h
```diff
@@ -2094,4 +2105,10 @@ typedef enum INPUT_LAYOUT {
     INPUT_LAYOUT_NUMBER_OF,
 } INPUT_LAYOUT;
 
+typedef enum BUTTON_TYPE {
+    BUTTON = 0,
+    AXIS = 1,
```
rr-: Please use BT_ prefix

src/config.c
```diff
@@ -272,7 +275,36 @@ bool Config_ReadFromJSON(const char *cfg_data)
...
+            int16_t type = Input_GetAssignedButtonType(layout, role);
+            type = json_array_get_int(cntlr_arr, i, type);
+            i++;
+
+            int16_t bind = Input_GetAssignedBind(layout, role);
+            bind = json_array_get_int(cntlr_arr, i, bind);
+            i++;
+
+            int16_t axs_dir = Input_GetAssignedAxisDir(layout, role);
+            axs_dir = json_array_get_int(cntlr_arr, i, axs_dir);
```
rr-: Please rename to `axis_dir`

src/specific/s_input.c
```diff
@@ -590,27 +1020,127 @@ INPUT_SCANCODE S_Input_GetAssignedScancode(
     return m_Layout[layout_num][role];
 }
 
+int16_t S_Input_GetUniqueBind(INPUT_LAYOUT layout_num, INPUT_ROLE role)
+{
+    CONTROLLER_MAP assigned = m_ControllerLayout[layout_num][role];
+    if (assigned.type == BUTTON) {
+        return m_ControllerLayout[layout_num][role].bind.btn;
+    } else {
```
rr-: Unnecessary `else`, can either be merged to `else if` down below or removed altogether

rr- (commented):
Can we rename `axs` to `axis` everywhere?

src/game/option/option.c
```diff
@@ -38,7 +41,11 @@ void Option_DoInventory(INVENTORY_ITEM *inv_item)
         break;
 
     case O_CONTROL_OPTION:
-        Option_Control(inv_item);
+        if (m_ControlMode == CM_PICK) {
+            m_ControlMode = Option_ControlPick(inv_item);
+        } else {
+            m_ControlMode = Option_Control(inv_item, m_ControlMode);
```
rr-, line 47: Can you describe what is the difference between these?

src/specific/s_input.c
```diff
@@ -15,6 +15,17 @@
 
 #define KEY_DOWN(a) (m_KeyboardState[(a)])
 
+typedef union BIND_TYPE {
+    SDL_GameControllerButton btn;
```
rr-: And I'd name this `button`

src/specific/s_input.c
```diff
@@ -56,19 +67,19 @@ static INPUT_SCANCODE m_Layout[INPUT_LAYOUT_NUMBER_OF][INPUT_ROLE_NUMBER_OF] = {
...
+        SDL_SCANCODE_UP,         // INPUT_ROLE_UP
+        SDL_SCANCODE_DOWN,       // INPUT_ROLE_DOWN
+        SDL_SCANCODE_LEFT,       // INPUT_ROLE_LEFT
+        SDL_SCANCODE_RIGHT,      // INPUT_ROLE_RIGHT
+        SDL_SCANCODE_DELETE,     // INPUT_ROLE_STEP_L
+        SDL_SCANCODE_PAGEDOWN,   // INPUT_ROLE_STEP_R
+        SDL_SCANCODE_RSHIFT,     // INPUT_ROLE_SLOW
+        SDL_SCANCODE_RALT,       // INPUT_ROLE_JUMP
+        SDL_SCANCODE_RCTRL,      // INPUT_ROLE_ACTION
+        SDL_SCANCODE_SPACE,      // INPUT_ROLE_DRAW
```
rr-, line 79: Why do you change the second layout to match the default layout?
I remember these values were taken from TombATI.

src/specific/s_input.c
```diff
@@ -165,16 +176,170 @@ static INPUT_SCANCODE m_Layout[INPUT_LAYOUT_NUMBER_OF][INPUT_ROLE_NUMBER_OF] = {
...
+    { BUTTON, {SDL_CONTROLLER_BUTTON_INVALID}, 0 },        // INPUT_ROLE_EQUIP_UZIS
+    { BUTTON, {SDL_CONTROLLER_BUTTON_INVALID}, 0 },        // INPUT_ROLE_USE_SMALL_MEDI
+    { BUTTON, {SDL_CONTROLLER_BUTTON_INVALID}, 0 },        // INPUT_ROLE_USE_BIG_MEDI
+    { BUTTON, {SDL_CONTROLLER_BUTTON_INVALID}, 0 },        // INPUT_ROLE_SAVE
+    { BUTTON, {SDL_CONTROLLER_BUTTON_INVALID}, 0 },        // INPUT_ROLE_LOAD
+    { BUTTON, {SDL_CONTROLLER_BUTTON_INVALID}, 0 },        // INPUT_ROLE_FPS
+    { BUTTON, {SDL_CONTROLLER_BUTTON_INVALID}, 0 },        // INPUT_ROLE_BILINEAR
+    }
+        // clang-format on
+    };
```
rr-, line 326: Deindent by 4 spaces

src/specific/s_input.c
```diff
@@ -165,16 +176,170 @@ static INPUT_SCANCODE m_Layout[INPUT_LAYOUT_NUMBER_OF][INPUT_ROLE_NUMBER_OF] = {
...
+    { BUTTON, {SDL_CONTROLLER_BUTTON_INVALID}, 0 },        // INPUT_ROLE_EQUIP_MAGNUMS
+    { BUTTON, {SDL_CONTROLLER_BUTTON_INVALID}, 0 },        // INPUT_ROLE_EQUIP_UZIS
+    { BUTTON, {SDL_CONTROLLER_BUTTON_INVALID}, 0 },        // INPUT_ROLE_USE_SMALL_MEDI
+    { BUTTON, {SDL_CONTROLLER_BUTTON_INVALID}, 0 },        // INPUT_ROLE_USE_BIG_MEDI
+    { BUTTON, {SDL_CONTROLLER_BUTTON_INVALID}, 0 },        // INPUT_ROLE_SAVE
+    { BUTTON, {SDL_CONTROLLER_BUTTON_INVALID}, 0 },        // INPUT_ROLE_LOAD
+    { BUTTON, {SDL_CONTROLLER_BUTTON_INVALID}, 0 },        // INPUT_ROLE_FPS
+    { BUTTON, {SDL_CONTROLLER_BUTTON_INVALID}, 0 },        // INPUT_ROLE_BILINEAR
+    }
+        // clang-format on
```
rr-, line 325: Deindent by 4 spaces

src/specific/s_input.c
```diff
@@ -165,16 +176,170 @@ static INPUT_SCANCODE m_Layout[INPUT_LAYOUT_NUMBER_OF][INPUT_ROLE_NUMBER_OF] = {
     // clang-format on
 };
 
+static CONTROLLER_MAP
+    m_ControllerLayout[INPUT_LAYOUT_NUMBER_OF][INPUT_ROLE_NUMBER_OF] = {
+        // clang-format off
```
rr-, line 181: Deindent by 4 spaces

src/specific/s_input.c
```diff
@@ -165,16 +176,170 @@ static INPUT_SCANCODE m_Layout[INPUT_LAYOUT_NUMBER_OF][INPUT_ROLE_NUMBER_OF] = {
...
+static const char *m_ControllerName = NULL;
+static SDL_GameControllerType m_ControllerType = SDL_CONTROLLER_TYPE_UNKNOWN;
 
 static const char *S_Input_GetScancodeName(INPUT_SCANCODE scancode);
 static bool S_Input_KbdKey(INPUT_ROLE role, INPUT_LAYOUT layout);
 static bool S_Input_Key(INPUT_ROLE role, INPUT_LAYOUT layout_num);
-static bool S_Input_JoyBtn(SDL_GameControllerButton button);
-static int16_t S_Input_JoyAxis(SDL_GameControllerAxis axis);
+static bool S_Input_JoyBtn(SDL_GameControllerButton btn);
+static int16_t S_Input_JoyAxis(SDL_GameControllerAxis axs);
```
rr-: `button`, `axis`

src/specific/s_input.c
```diff
@@ -418,9 +583,181 @@ static const char *S_Input_GetScancodeName(INPUT_SCANCODE scancode)
...
+                case SDL_CONTROLLER_BUTTON_MISC1:         return "MIC"; /* Xbox Series X share button, PS5 microphone button, Nintendo Switch Pro capture button, Amazon Luna microphone button */
+                case SDL_CONTROLLER_BUTTON_PADDLE1:       return "PADDLE 1"; /* Xbox Elite paddle P1 (upper left, facing the back) */
+                case SDL_CONTROLLER_BUTTON_PADDLE2:       return "PADDLE 2"; /* Xbox Elite paddle P3 (upper right, facing the back) */
+                case SDL_CONTROLLER_BUTTON_PADDLE3:       return "PADDLE 3"; /* Xbox Elite paddle P2 (lower left, facing the back) */
+                case SDL_CONTROLLER_BUTTON_PADDLE4:       return "PADDLE 4"; /* Xbox Elite paddle P4 (lower right, facing the back) */
+                case SDL_CONTROLLER_BUTTON_TOUCHPAD:      return "TOUCHPAD"; /* PS4/PS5 touchpad button */
+                case SDL_CONTROLLER_BUTTON_MAX:           return "";
+            }
+            break;
+        case SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_PRO:
```
rr-, line 622: Since these cases are quite big, let's put a blank line after each `break;`

src/specific/s_input.c
```diff
@@ -549,35 +890,124 @@ INPUT_STATE S_Input_GetCurrentState(INPUT_LAYOUT layout_num)
...
+                linput.load                   |= btn_state;
+                break;
+            case INPUT_ROLE_FPS:
+                linput.toggle_fps_counter     |= btn_state;
+                break;
+            case INPUT_ROLE_BILINEAR:
+                linput.toggle_bilinear_filter |= btn_state;
+                break;
+            default:
+                break;
```
rr-: Can we turn this into a function?
`static INPUT_STATE S_Input_Toggle(INPUT_STATE state, INPUT_ROLE role, bool state);`

src/specific/s_input.c
```diff
@@ -622,3 +1152,14 @@ bool S_Input_CheckKeypress(const char *key_name)
     }
     return false;
 }
+
+bool S_Input_CheckButtonPress(const char *btn_name)
+{
+    SDL_GameControllerButton btn =
+        SDL_GameControllerGetButtonFromString(btn_name);
+
+    if (S_Input_JoyBtn(btn)) {
```
rr-: `return S_Input_JoyBtn(btn)`

src/global/types.h
walkawayy: Oof good call. Idk how I forgot this prefix. That's why I used `axs` instead of axis. I wanted to avoid overlapping the same name.

src/specific/s_input.c
walkawayy: I changed every control scheme to start as the default controls because we don't have separate menu controls. If these different controls stay in the code, when the user scrolls to this control scheme in the menu, the arrow keys would no longer work. The up/down/left/right would now be mapped to the number pad which seemed unintuitive to me since the user would suddenly no longer be able to navigate the menus with the default arrow keys.

src/specific/s_input.c
walkawayy: Clang format kept doing this to me for some reason. But I'll look into it.

src/game/option/option.c
walkawayy: Huh somehow I committed without the new files. Added them now.

`Option_ControlPick` is this screen where you pick keyboard or controller:
image

`Option_Control` is the normal controls screen.
image

src/specific/s_input.c
walkawayy: Seems like the indent happens because
```
static CONTROLLER_MAP
    m_ControllerLayout[INPUT_LAYOUT_NUMBER_OF][INPUT_ROLE_NUMBER_OF] = {
        // clang-format off
```
the `m_ControllerLayout` is already indented onto a second line. If I shorten this line, `// clang-format off` does not get indented. Not sure there's a way around this without shortening variable names. Or I could move the clang off/on to outside of the entire function? Not sure we want that though.

rr-: In that case, can we indent the inner block to match the outer indentation level?

rr-:
@walkawayy I don't think you've pushed the changes

walkawayy:
Oops pushed and indented.

src/specific/s_input.c
```diff
@@ -418,9 +583,186 @@ static const char *S_Input_GetScancodeName(INPUT_SCANCODE scancode)
...
+                case SDL_CONTROLLER_AXIS_LEFTY:           return axis_dir == -1 ? "L ANALOG UP" : "L ANALOG DOWN";
+                case SDL_CONTROLLER_AXIS_RIGHTX:          return axis_dir == -1 ? "R ANALOG LEFT" : "R ANALOG RIGHT";
+                case SDL_CONTROLLER_AXIS_RIGHTY:          return axis_dir == -1 ? "R ANALOG UP" : "R ANALOG DOWN";
+                case SDL_CONTROLLER_AXIS_TRIGGERLEFT:     return "L TRIGGER";
+                case SDL_CONTROLLER_AXIS_TRIGGERRIGHT:    return "R TRIGGER";
+                case SDL_CONTROLLER_AXIS_MAX:             return "";
+            }
+            break;
+    }
+    // clang-format on
```
rr-, line 763: Let's put a blank below this line as well

src/specific/s_input.c
```diff
@@ -418,9 +583,186 @@ static const char *S_Input_GetScancodeName(INPUT_SCANCODE scancode)
...
+                case SDL_CONTROLLER_BUTTON_PADDLE1:       return "PADDLE 1"; /* Xbox Elite paddle P1 (upper left, facing the back) */
+                case SDL_CONTROLLER_BUTTON_PADDLE2:       return "PADDLE 2"; /* Xbox Elite paddle P3 (upper right, facing the back) */
+                case SDL_CONTROLLER_BUTTON_PADDLE3:       return "PADDLE 3"; /* Xbox Elite paddle P2 (lower left, facing the back) */
+                case SDL_CONTROLLER_BUTTON_PADDLE4:       return "PADDLE 4"; /* Xbox Elite paddle P4 (lower right, facing the back) */
+                case SDL_CONTROLLER_BUTTON_TOUCHPAD:      return "TOUCHPAD"; /* PS4/PS5 touchpad button */
+                case SDL_CONTROLLER_BUTTON_MAX:           return "";
+            }
+            break;
+    }
+    // clang-format on
```
rr-, line 709: Let's put a blank below this line as well

src/specific/s_input.c
```diff
@@ -467,6 +809,130 @@ static int16_t S_Input_JoyAxis(SDL_GameControllerAxis axis)
     return 0;
 }
 
+static INPUT_STATE S_Input_GetControllerState(
+    INPUT_STATE state, INPUT_LAYOUT cntlr_layout_num)
+{
+    // clang-format off
```
rr-: No need to disable formatting in that big of a scope.

src/specific/s_input.c
```diff
@@ -467,6 +809,130 @@ static int16_t S_Input_JoyAxis(SDL_GameControllerAxis axis)
...
+        int16_t btn_state = 0;
+
+        if (assigned.type == BT_BUTTON) {
+            btn_state = S_Input_JoyBtn(assigned.bind.button);
+        } else {
+            btn_state = S_Input_JoyAxis(assigned.bind.axis) == assigned.axis_dir;
+        }
+
+        switch(role) {
+        case INPUT_ROLE_UP:
```
rr-: I'd prefer to see this compressed to individual lines and safe-guarded with clang-format off. Otherwise we don't need to disable the formatting.

src/specific/s_input.h
```diff
@@ -2,22 +2,48 @@
...
+void S_Input_ResetControllerToDefault(INPUT_LAYOUT layout_num);
+
+bool S_Input_ReadAndAssignKey(
+    CONTROL_MODE mode, INPUT_LAYOUT layout_num, INPUT_ROLE role);
+
+const char *S_Input_GetKeyName(
+    CONTROL_MODE mode, INPUT_LAYOUT layout_num, INPUT_ROLE role);
+
+const char *S_Input_GetButtonNameFromString(
+    INPUT_LAYOUT layout_num, const char *btn_name);
```
rr-: This still uses `btn_name`

src/game/option/option_control_pick.c
```diff
@@ -0,0 +1,116 @@
...
+#include "game/music.h"
+#include "game/option/option_sound.h"
+#include "game/sound.h"
+#include "game/text.h"
+#include "global/vars.h"
+
+#include <stdio.h>
+
+#define MIN_VOLUME 0
+#define MAX_VOLUME 10
```
rr-: These defines can be removed

src/game/option/option_control_pick.c
```diff
@@ -0,0 +1,116 @@
...
+
+static void Option_ControlPickShutdownText(void)
+{
+    for (int i = 0; i < TEXT_NUMBER_OF; i++) {
+        Text_Remove(m_Text[i]);
+        m_Text[i] = NULL;
+    }
+}
+
+CONTROL_MODE Option_ControlPick(INVENTORY_ITEM *inv_item)
```
rr-: `inv_item` is not used and as such this can take no arguments (remember to use `void`, not empty braces)

walkawayy: Ok. I kept it in there since the other inventory functions all take it even though only the option_passport uses it.

src/specific/s_input.h
walkawayy: Should I rename `btn` to `button` everywhere like I did `ax`s to `axis`? I originally used `btn` because names like `CONTROL_RESET_DEFAULTS_BTN` and `GS_CONTROL_RESET_DEFAULTS_BTN` are very long already. There are a few remaining names I can change if we want.

src/specific/s_input.c
walkawayy: Lmk if the new `S_Input_GetControllerState` looks better. I think it does.

src/game/option/option_control.c
```diff
@@ -611,16 +682,18 @@ void Option_Control(INVENTORY_ITEM *inv_item)
...
+            return CM_PICK;
         }
 
         if ((g_InputDB.left || g_InputDB.right)
             && m_ControlMenu.cur_role == KC_TITLE) {
-            Option_ControlChangeLayout();
+            layout_num = Option_ControlChangeLayout(mode);
         }
 
         if (g_InputDB.select) {
```
walkawayy, line 697: One thing I've noticed is the user can press action on the keyboard or controller to enter the remap mode no matter if they are on the keyboard or controller page. They can't actually remap a key with a button and vice versa because of this check:
`if (Input_ReadAndAssignKey(mode, layout_num, m_ControlMenu.cur_role)) {`

but, the game will be stuck waiting for an input on the appropriate device. Not a huge deal but something I should probably fix. I'm not sure how to differentiate where the input game from however since `S_Input_GetCurrentState` checks keyboard and controller input. Any ideas?

src/specific/s_input.h
rr-: No just standalone `btn`  → `button`.

src/game/option/option_control.c
rr-: I don't think this is an issue.

src/specific/s_input.c
```diff
@@ -467,6 +813,73 @@ static int16_t S_Input_JoyAxis(SDL_GameControllerAxis axis)
...
+        state.equip_uzis             |= S_Input_GetBindState(INPUT_ROLE_EQUIP_UZIS, cntlr_layout_num);
+        state.use_small_medi         |= S_Input_GetBindState(INPUT_ROLE_USE_SMALL_MEDI, cntlr_layout_num);
+        state.use_big_medi           |= S_Input_GetBindState(INPUT_ROLE_USE_BIG_MEDI, cntlr_layout_num);
+        state.save                   |= S_Input_GetBindState(INPUT_ROLE_SAVE, cntlr_layout_num);
+        state.load                   |= S_Input_GetBindState(INPUT_ROLE_LOAD, cntlr_layout_num);
+        state.toggle_fps_counter     |= S_Input_GetBindState(INPUT_ROLE_FPS, cntlr_layout_num);
+        state.toggle_bilinear_filter |= S_Input_GetBindState(INPUT_ROLE_BILINEAR, cntlr_layout_num);
+        // clang-format on
+    }
+    return state;
```
rr-, line 870: Let's put a blank line before the `return`

rr- (approved):

src/specific/s_input.c
rr-: It does with one tiny remark. In general after big block like this:
```
keyword {
    many continuous lines
    many continuous lines
}
```
I like to have a single blank line. How many? I'd say at least 5 but this is completely subjective. The reason in general is to cater for IDEs that have "jump to the next blank line" as well as for aesthetics.

walkawayy:
Ok I have one more bug I noticed that I am going to fix. The controller text continues to flash even if two roles no longer conflict. That does not happen with the keyboard controls.

Also since we are breaking old custom controller schemes with this update anyway, @rr- can I reorganize the list of controls? Here's the default starting order for example. I think important roles like action, jump, roll, etc should be higher in the list and more optional roles like equip pistols should come later in the list with the cheats coming last.

image

rr-:
> Ok I have one more bug I noticed that I am going to fix. The controller text continues to flash even if two roles no longer conflict. That does not happen with the keyboard controls.
> 
> Also since we are breaking old custom controller schemes with this update anyway, @rr- can I reorganize the list of controls? Here's the default starting order for example. I think important roles like action, jump, roll, etc should be higher in the list and more optional roles like equip pistols should come later in the list with the cheats coming last.
> 
> image

Yes, sure.

walkawayy:
Alright I think this is good to go. I reordered the controls to make more sense and fixed the text width on the square button so it matches the other Playstation face buttons.

image

walkawayy:
Symbols look pretty good now. Not perfect since they are low quality in the level files and not created evenly, but until we can inject our own data, this should work.

image

PR #675, opened 2022-12-19
lara: fix dying anim when landing on low health

byblo:
fixed rare bug when Lara was dying without playing the crash animation when her resulting health = 0

As it is illustrated on this video: https://www.youtube.com/watch?v=mteF9NCoBG8

#### Checklist


src/game/lara/lara_misc.c
```diff
@@ -579,7 +579,7 @@ bool Lara_LandedBad(ITEM_INFO *item, COLL_INFO *coll)
             / (DAMAGE_LENGTH * DAMAGE_LENGTH);
     }
 
-    if (item->hit_points < 0) {
+    if (item->hit_points <= 0) {
```
rr-: Let's simplify it further like this: `return item->hit_points <= 0`.
Additionally please adjust the commit message to conform with our guidelines, in this case I suggest:
`lara: fix dying anim when landing on low health`

Other than that, IDK, I think this is a fun bug and I'd leave it as-is (0.1% chance to see a secret sequence vs just a regular crunch) but that's just me.

byblo:
Indeed it is no big deal and adds some suspense after a risky jump.
Feel free to modify or to close this.

src/game/lara/lara_misc.c
walkawayy: Here's how it was done in [TR3](https://github.com/Trxyebeep/tomb3/blob/f06647a8d474cd6c8e3ffb4440d5cfd933816f3f/tomb3/game/lara.cpp#L241).

I'd probably follow that which matches rr's suggestion and classify this is an OG bug. It's harmless though so it's up to rr.

byblo:
If it is original bug then this fix has to be optional and disabled by default or rejected. I thought it was exclusive to Tomb1Main.

rr- (approved):

Issue #667, opened 2022-12-13
Demo level support


carlo-bramini:
I'm wondering if it would be possible to add a reply to this question to "Q&A" section of [README.md](https://github.com/rr-/Tomb1Main#readme).

It would be useful to write few instructions for knowing how to make them working, if the engine is able to run (or it may be able to do it in the future if it does not yet).
I think that it would be very useful to be able to run them, for testing/evaluating the engine without having the full game.

oziphantom:
the demo data is stored in the level data, the engine just looks at each level in turn to see if it has a demo data segment in it. I don't know of any tools that would insert it for you though.

carlo-bramini:
> the demo data is stored in the level data, the engine just looks at each level in turn to see if it has a demo data segment in it. I don't know of any tools that would insert it for you though.

Actually, I was not talking about the demo embedded inside the level data, but the playable demos of the first Tomb Raider.
The playable demo allows to play the level of the city of Vilcabamba.
I have seen two versions are available, one for plain PC and one for 3DFX.

There is also a demo version of the Unfinished Business extension. Two versions are available, one for PC and one for Mac (don't know if there are differences between them into the data files).

Those demos are also freely and legally downloadable from many places, that's why they could be useful to play with the engine even if you don't own the full game.

oziphantom:
oh, make a custom Tomb1Main_gameflow.json5 file and then in the 
```
 // level 1
        {
            "title": "Caves",
            "file": "data\\level1.phd", <- change these to be your demo phd files. 1 per level for as many as you have.
            "type": "normal",
            "music": 57,
            "sequence": [
                {"type": "play_fmv", "fmv_path": "fmv\\snow.avi"},
                {"type": "start_game"},
                {"type": "loop_game"},
                {"type": "stop_game"},
                {"type": "level_stats", "level_id": 1},
                {"type": "exit_to_level", "level_id": 2},
            ],
            "strings": {},
        },
```

Richard-L:
Hey Carlo, I also assumed you meant the demonstrations that play out after idling in the main menu.

I agree with you the official demos would be lovely to have supported by the installer, just because as you say people without the game can immediately try TR1M without having to pirate or enter credit card details to whoever distributes the games digitally.

[...]

PR #676, opened 2022-12-19
Fix support for 64-bit

carlo-bramini:
#### Checklist

#### Description

When running Tomb1Main compiled as a 64-bit application, I got an error.
This happens because `ANIM_STRUCT` has a `frame_ptr` item which is a pointer and so it varies its size if it's compiled for an addressing depth larger than 32-bit. I fixed the issue by adding a `frame_ofs` item, used for indexing `g_AnimFrames[]`, and by filling all the items of `ANIM_STRUCT` by reading them one by one, in the same manner it has been already done into function `Level_LoadRooms()`.

I tested Tomb1Main with the demo levels (see issue #667) under Windows, compiled as i686 (32-bit) and x64 (64-bit) and the engine worked fine on both.

### DIFF
--- a/src/game/level.c
+++ b/src/game/level.c
@@ -274,7 +274,23 @@ static bool Level_LoadObjects(MYFILE *fp)
     File_Read(&m_AnimCount, sizeof(int32_t), 1, fp);
     LOG_INFO("%d anims", m_AnimCount);
     g_Anims = GameBuf_Alloc(sizeof(ANIM_STRUCT) * m_AnimCount, GBUF_ANIMS);
-    File_Read(g_Anims, sizeof(ANIM_STRUCT), m_AnimCount, fp);
+    for (int i = 0; i < m_AnimCount; i++) {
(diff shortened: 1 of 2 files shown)


oziphantom:
test save and loading. And then check saving in 64bit and loading in 32 bit and vice versa.

rr- (approved):
LGTM.

rr-:
@walkawayy could you please confirm Ozi's testing scenarios?

carlo-bramini:
I tested load/save state exchanges between 32-bit and 64-bit versions and I have not problems, at least with the limited amount of data provided by the demo version.

walkawayy:
> @walkawayy could you please confirm Ozi's testing scenarios?

Ok I'll try some testing when I get a chance.

walkawayy:
Looks good to me. I tried saving and loading during various animations like freefall, normal jump, damocles sword, etc and it worked fine. I also loaded saves made before this change and they loaded fine.

walkawayy (approved):

PR #677, opened 2022-12-19
gameplay: optional Difficulty Selector based on a 'damages to lara multiplier'

byblo:
Adds a Difficulty Selector.

For instance, if a bear attacks Lara on Very Hard, the final result will be: -200 * 4.75 = -950 HP (integer) versus -200HP on normal.

By default, the Difficulty Selector is enabled but the default difficulty/damages remains normal: Tomb1Main.json5, "enable_difficulty" : true
Tomb1Main.json5, "damages_to_lara_multiplier" : 1.000000

Current values are: (maybe polishing is required later after few plays)
    0.35 = Very Easy (a bear can do 7% of damage to Lara in one regular hit. A wolf 1.7%)
    0.65 = Easy (a bear can do 13% of damage to Lara in one regular hit. A wolf 3.2%)
    1.00 = Normal (a bear can do 20% of damage to Lara in one regular hit. A wolf 5%)
    2.25 = Hard (a bear can do 45% of damage to Lara in one regular hit. A wolf 11%)
    4.75 = Very Hard (a bear can do 95% of damage to Lara in one regular hit. A wolf 24%)
    9.50 = Extreme (a bear can do 190% of damage to Lara in one regular hit. A wolf 48%)
   20.00 = Impossible (A simple dart will kill Lara!)

You can see some illustratives videos here: https://www.youtube.com/watch?v=huVGMTvEGtQ and https://www.youtube.com/watch?v=n-c3LVRT2iI Real gameplay on Extreme difficulty here (skip Caves): https://www.youtube.com/watch?v=jhEjTOA141w

Note that the core of this mod is very simple and takes very few lines of code, and is mainly located at the files: lara.c, lara_misc.c + some specific. It does modify any damage Lara is getting from the game. Currently tested the main enemies, some traps, getting shot, falling and drowning damages. Everything else is secondary, like read/write into config, showing stats, difficulty selector, and located into various *.c and *.h files.

Note also that the selected difficulty does NOT modify the savefiles. For various reasons, maybe in future it would be more interesting to save and load also the selected difficulty from a savefile.

#### Checklist


rr-:
While this is a well crafted change and a cool mod, I'm not sure if it should make its way into the mainline T1M version. I would say we already have a similar function in the configurator tool, where you get to set Lara's starting hp. It's not 100% the same, but combined with disabling medpacks, it can offer the extra challenge (esp. the no damage challenge). My main concern is that our menus already feel cluttered with Load Game + Select Level, and the UI for this proposal adds additional layer of complexity to it which makes the game feel even more like a mod rather than an improved OG. I invite others to share their thoughts on this.

Technically speaking, I don't like the introduction of the `g_OldLaraHitPoints` global, I'd either move it to `g_Lara` or make it a local in `lara.c` and reset it in one of the Lara object initialization routines.

byblo:
I put the Difficulty Selector at the passport since it seems to be the most useful and logical place.

I do not see the user leaving the game and running the configuration tool each time he wants to switch to another difficulty. (or a new title's rotating item like 'General Options' may be required)

Also made a difficulty.c file to avoid filling and to keep clear the option_passport.c file with too many stuffs that may confuse others contributors.

I also avoided to add some new popup like the 'Select Mode' (New Game, New Game+, ect...) and went instead for a new menu that uses unused up and down keys at the passport selection pages.

Please remember it is 100% optional with the bool enable_difficulty.
Maybe I can just disable the whole mod by default but I do want to keep the Difficulty Selector at the passport pages like it is currently (EDIT: or at least, it has to be accesable from the game itself).

About  the g_OldLaraHitPoints, indeed it can be probably put in lara.c instead

byblo:
Do anyone knows what is that run code linters error about?

It says: "Please run `make imports` and commit the changes."

But when I do that it just gives me some incomprehensible error: `No such file or directory: 'include-what-you-see'`

It also does that with the current 2.11 and develop branches so I assume it is not a problem?

I got no problem compiling any of the sources, but that 'make imports' command always returns that error and I have no clue what it is about.

rr-:
Above the suggested fix, you can see a diff of what the linter complains about. You can apply it by hand or install IWYU on your local machine and commit that.

byblo:
Ok I think I got it. Installed IWYU and it seems working.
But then I will first move and rename the g_OldLaraHitPoints then run again that IWYU tool in last.
Expect some commit update in the next days, the time to write and test again the mod.
Thank you for the help.

Richard-L:
Not a fan of this personally. Feels a bit like bloat to me honestly.

walkawayy:
I don't think a difficulty mode is a bad idea. rr did mention in the past about a possibility of a pacifist mode to play through the game without worrying about combat. That would take many more changes to the AI but this could lay the groundwork for it.

I think 7 difficulties is too many personally. So right off the bat I'd consider doing:
1. Story / Easy
2. Normal
3. Hard
4. Extreme
5. (maybe pacifist mode in the future)

I think the current extreme mode would be impossible to complete if one shots kills you. I don't think bosses like Larson, Pierre, Cowboy, etc are beatable without taking a single hit?

I also don't think the fall damage / underwater damage should be messed with since that can break level design unfortunately.

Most modern games allow you to change difficulty on the fly, so having an option for it it in game makes sense. But I agree with rr that it looks a little clunky where it is. I'm not sure I have a better idea at the moment though. It's hard to add menu options since we have limited menu objects where we can put stuff.

byblo:
About the amount of difficulties:
--------------------------------
Actually, I also do not like to have too many difficulties around. I like simple and clear things and I also would want to stick with: "Easy / Normal / Hard / Very Hard" difficulties if it was possible.

But I think they should stay since I am sure some people may want soon or later to finely adapt a difficulty around their favorite one, including myself, for whatever the reason: playing a custom map that is easier/harder than usual, playing tired/excited, ect...).
For instance, personaly, I am pretty sure I will never be able to resist to try to finish the game at the Impossible difficulty if I can manage already to finish both main TR and TRUB in Extreme with the saving crystals enabled.

Also, after the first scrolling to discover all the difficulties availables, players will probably stick around their favorite and barely push anymore the UP/DOWN button anymore to change it.

Thus, all those difficulties seems to be a good compromise between clarity and to fill the needs for everyone at any occasion.
Please remember also that each difficulty is already more or less the double from the previous one.
Reducing their number and thus making too much differences between the difficulties could also be a problem?



About the shooting at lara damages:
----------------------------------
Totally agree with you, the multiplicators cannot be applied directly to the 'Shooting at Lara' type of damage - unlike the animals or the traps - they are usualy spammed at a fast rate and from a distance, which can kill her in less than 2 second for the hardests ones. I discovered this with pain while testing the mod :)
They still doable, as I managed to pass them until Midas level, but barely and not fun at all, and are certainly not matching the other damaging types.
This is why I (probably) already fixed this issue by using the multiplicator's squared root for that kind of damages (still need to test this with more advanced enemies)
Thus, for instance for the Impossible difficulty which is the hardest one around, the multiplicator for Larson damage is now the following: -50(normal diff.) * 4.47(damage mult.'s squared root) = -223 HP, against the previous version which was: -50 * 20 = -1000 HP! Thus you can now eat like 5 bullets at full HP before dying if not healing.
After fixing this issue, the Extreme difficulty felt suddently even too easy now at the shooting phases: Larson is doing now -154 HP against 475 previously and I stopped dying agaisnt them... This said, still need to see how the more advanced shooters will hurt Lara.

Also I see that Baldy can do 150 of damages, which will be at max -150 * 4.47 = -670 per shot for Impossible and -462 for Extreme. I'll figure out while playing the game if it is possible to deal with such crazy damages and to finish the game. Also Natla is worrying me since I do not remember the fight agaisnt her.



About the falling damage and water:
----------------------------------
The air timer remains the same between the difficulties.
Lara still have the exact same time to stay under the water before starting to drown.
The only difference here are the damages inflicted to Lara once she starts drowning, that will be greater(faster to die), forcing you maybe to use faster your remaining medkits until you reach the surface.
But AFAIK, I do not think it is a problem.

For the falling damage, if you can finish the main games TR and TRUB in Impossible (jumps with full HP), again I do not think this is a concern. If not, then I will fix it somehow.
About the community levels, it is not possible to match them all. In case of problem with those, then the player is still able to reduce the difficulty to pass the forced hurting jump.



Concerns about the Very Easy and Easy modes:
-------------------------------------------
This said, my main concern is about those ones instead. What if a player is supposed to die by a trap, enemy or a prohibed zone that would inflict exactly a -1000 HP of damage? On those difficulties, he could potentially breaks the game and to start walking on the bottom of a community map...
About the official maps, I did not checked everything yet, but AFAIK the lethal traps like the spikes for instance are forcing the Lara's health to -1 by default if you fall onto them.
This make me hesitating to simply remove the difficulties below the Normal one. What do you think?



About the ability to change the difficulty while playing
--------------------------------------------------------
Adding such ability would requires to be accessed from some option window.
Actually, on the previous version, it was naturaly doable. But I noticed that going to the passport (that is dedicated to new/load/save/exit) to change the current difficulty really felt like a nonsense to me. Hence the "DIFFICULTY_SELECT_CONFIRM" late addition.



About the difficulty settings availables already: (rr first post)
--------------------------------------------------------
By the way, I also wanted to add that I saw the others difficulty variables like the start_lara_hitpoints, items removal, the disable_healing_between_levels, the enable_save_crystals, ect...
If I am very happy to see them (the crystals are mandatory for me, and starting a level without healing feels more natural and is great), I can tell that their philosophy is different from this mod, thus probably complementary.
Starting a level with limited health or without any medkits around must be great experiences that I will certainly do next.
But being ambushed by a lion behind a pillar and to die suddendly even at full HP is just priceless to me. Same for cruely dying after a bad jump as punishment where the normal game would simply remove -10% to Lara's health...



About the bloat feeling:
-----------------------------
Can you say more precisely what feels wrong with it?
Indeed the current concept at the moment is not perfect, specially since it does not take in account the others difficulty variables I mentionned above. I was thinking to maybe integrate them later somehow and to use the Difficulty Selector at the passport pages as presets changer instead of currently simple damages_on_lara_multiplier indicators. But this will requires a LOT more of thinking and coding. I am not sure I got the energy or the will to do that. I still thinking the best solution would be a new window or item at the Options screen, something silimar to Gameplay or whatever the name, where the player could change some of those gameplay variables.

There is also the problem with the stats where a player can load a save near the level's exit, while selecting a difficulty an thus the stats screen will show that difficulty, which is somehow a nonsense .

Idealy, I would like to link a difficulty to a save, and also to allow a difficulty change only when:
- starting a new game
- restarting the level after a death
- loading a level and to start from the beginning
- from some option screen

walkawayy:
Ok I'll type my thoughts out on various parts of this.

I think removing easy mode would be a mistake. Tomb Raider has always been a hard game and easier combat encounters can allow players to focus on the story, puzzles, and platforming. Instant kill traps still need to kill Lara though since that could break level design. What if Lara falls into a pit of spikes and survives but it's too tall to jump out of? That would create a new softlock. The same idea goes for fall damage. I think reduced fall damage on easy can work but only if the freefall animation still causes Lara to die. I know we can't predict all custom games design, but we should try to stick to some of the game's original core mechanics for builders to design around. We also don't want to break the original game's level design. We have cheat mode with options like the fly cheat if the user wants to intentionally break the level design and fly around invincible.

Regarding the amount of difficulties. The current amount could prove a lot for the user to decide on. There's currently four variations of the hardest difficulty for example. That's a lot to decide between and there's currently no way to tell what the differences are between them. Newer games usually have a separate difficulty selection screen. It shows all the difficulties with a short explanation of the changes. IMO this difficulty selection screen would need to appear after the `Select Mode` screen where the user picks new game, new game plus, etc. Here's an example of a modern game's difficulty selection screen:
image

To go along with that selection screen during the new game process, there should be a way to change it in game. If the user picks a mode too hard or too easy, they should be able to change it. Right now we don't really have a good spot for this in game. Maybe we could add the Natla's PDA to the inventory and make it a game options screen? Just an idea. The one downside is that it's not present in the title screen level file but it's present in all other level files.
image

To preserve the difficulty, I think we'd have to bump the savegame version and store the current difficulty. That would not be too much trouble.

Lastly, rr mentioned the bloated UI because as you can see here on 1.5x text scaling, the UI is crowded. The difficulty selection you added interferes with the passport menu at high scaling. That's why I think a separate screen would be best with the default difficulty being normal.
image


This is a big feature to add, so I think we need to properly plan if we want to add it, how we should add it, design mockups, etc to make sure it's polished. And the final say is down to rr since that's the repo owner at the end of the day. So let's see what rr thinks.

byblo:
About a popup or a new screen to choose the difficulties:
--------------------------------------------------------
I do not want to sound obtuse and I appreciate your suggestions but I got to disagree with this one.

The advantage to the current Difficulty Selector being below some of the passport pages is that it is always visible, thus being both informative and usable directly without interfering with the user and the vanilla version. (since UP/DOWN keys were not used anyway)

At the opposite, having an extra Difficulty Selector popup showing up every single time that you want to 'load a game/restart a level/load a level', just to ask you to comfirm if you want to keep playing the same difficulty as usual or to change it, then to have to press a button again before accessing the savegames list, would be rapidely very irritating. For those reasons, I really want to avoid a new popup or a new screen as it may get on the nerve of the player.


About the comparison with modern games
--------------------------------------
Tomb Raider is an 1996 game, and I would like to stick around the original spirit of that time, when things were simple and effective, when there was no UI and hints all around the screen like modern games does. So I do not want to add some extra hint text around, specially on the title screen.
And indeed, the current Difficulty Selector placement looks a bit invasive but it seems to me the best compromise between efficacity and weird looking.

Same for the names. With some exceptions like Doom, all my old games on 16bits and 32bits consoles were using very simple and effective names: easy, normal, hard... Everything back in the days was put on the substance before the form.
Maybe explaining the difficulties in detail into the readme.txt would be enough?



About the easier difficulties:
-----------------------------
Yes currently I keept the Easy and Very Easy difficulties. I just do not want to remove the falling damage for easier mode yet, since it may break the homogeneity of the difficulties as it must remains for harder modes.
But if someone reports later a problem about falling on easier difficulties or anything it still could be fixed later. About Lara falling on pits, she gets -1HP by the game in any case. Or were you talking about custom pits from a community level?



About the UI interferences:
--------------------------
Thank you for the report. I did not noticed that overlapping texts problem.
This will be fixed tomorrow by replacing the current "minimized" text effect with simply hidding it.
I still could keep it when there is enough room by checking both the UI text size and the screen resolution, but that may create two differents behaviors so better avoid such solution.



About a new option screen:
--------------------------
Being able to change many variables without leaving the game would benefit the game indeed.

rr-:
Let's go with the PDA item and start shifting with the settings towards the screen inside. It'd work similar to the walkman icon that lets you change music and effects volume and is a scalable way to control many aspects of the game.
Granted, we do not have a way to inject the items yet, so either we ship our own TITLE.PHD (for both T1M and UB) or wait till @lahm86  figures out the magic for us.

What should be done now – in the current form I won't be merging this UI, even if it works with 1.5× scaling. Eventually we'll have a dedicated item to change ingame settings. So please move the config to the configurator tool and remove the UI for now.

And also like @walkawayy  said, the fall damage logic is acting funny – on easy we'll lose, say, 400 hitpoints at maximum permissible height, while a few pixels later Lara will crunch.

tools/config/Tomb1Main_ConfigTool/Resources/Lang/en.json
```diff
@@ -130,6 +130,10 @@
       "Title": "Lara's starting health",
       "Description": "Sets Lara's health value for the beginning of each level."
     },
+    "enable_difficulty": {
+      "Title": "Difficulty Selector",
```
lahm86: Could we change to sentence case so to match the rest of the settings? And the same applies for the French translation.

byblo:
> Let's go with the PDA item and start shifting with the settings towards the screen inside. It'd work similar to the walkman icon that lets you change music and effects volume and is a scalable way to control many aspects of the game. Granted, we do not have a way to inject the items yet, so either we ship our own TITLE.PHD (for both T1M and UB) or wait till @lahm86 figures out the magic for us.

About a new option item, what about hardcoding the 3D model (and its texture), instead of external loading. BTW this could also be done for the Lara's braid. But it is just suggestions.



> What should be done now – in the current form I won't be merging this UI, even if it works with 1.5× scaling. Eventually we'll have a dedicated item to change ingame settings. So please move the config to the configurator tool and remove the UI for now.

I really need to think about this. I am still unsure why it is so harmful to have it at the passport page. Is it only an cosmetic issue, having all those `Difficulty_Select();` lines in option_passport.c, or is it more conceptual?

If necessary, I could eventually modify the Difficulty Selector at passport to have it optional and disabled by default?

Please note that it is important to me.
I have no problem to have it accessable from the configuration tool or from a future Extras or Options item (where it could be changed on-the-fly), but I cannot remove it entirely from the passport and want to keep it for "advanced" players like me who may need it there. I do not see myself closing the game and to run the config tool or to go to in the future to the options screen every single time I need to play or to debug another difficulty.

See my uploaded videos at youtube how fast it is to preview but also to change only if necessary a difficulty from the passport right before loading a level, and this, without interfering with the original game design: A player select "Load Game" with left key, then press a button to get the gamesaves list like the original game. Pressing up/down being 100% optional and without the need of some extra confirmation like another popup or going to the option screen first.

Edit: I forgot to add that it could also be useful to add to the Difficulty Selector list some more presets/challenges instead of only damages_multiplier, like "No Medkits", "No Items", or "Lara 25% HP" (at 1.0x original normal damages obviously) This is what I would like to add next. (GS_DIFFICULTY_NM, GS_DIFFICULTY_NI, GS_DIFFICULTY_HP50, ect...)

> And also like @walkawayy said, the fall damage logic is acting funny – on easy we'll lose, say, 400 hitpoints at maximum permissible height, while a few pixels later Lara will crunch.

The tiny difference of height that decides if Lara will be barely hurt or will crash and to die seems to be from the original game: https://www.youtube.com/watch?v=3_PuWzY6B-E Notice how Lara crashing from a certain height, even by hanging first (00:00). But if then you fall from just one lower level (-2 blocks), she barely loses 10-15% (00:43).

The same seems to happen with the easier difficulties.

I forgot to to record this on video but had it tested later, but if Lara jumps from the same lower level (thus the Lara's maximum head altitude before falling is almost the same as hanging from the upper level) she will lose around -25% at maximum, versus crashing and dying from the above level! It is weird behavior but part of the original game.

BTW notice also the falling bug #675. It seems to happen often since that falling height is linked to the standard blocks sizes. To ensure it is not related to this difficulty mod, I also tested this also on 2.11 and the result was the same. Thus, should I eventually enable again the #675 fix and make it optional by default?


> Could we change to sentence case so to match the rest of the settings? And the same applies for the French translation.

Sure. Will be done ASAP for the next commit.

byblo:
I am thinking to make some changes: As walkawayy said, there are too many difficulties around which I agreed, but wanted to keep the choice to the user.

But after few games with different difficulties, there are few differences between them, despite the huge damages differences.
Also, I found a paradox on Extreme diffculty: Since Lara is dying more often without having the opportunity to heal, she will cross the level with more and more medkits, compared the (new one 3.8x) Very Hard difficulty, where a Lion willl hurt Lara near to death, forcing her to use a big medkit. As you can see, the philosophies between Very Hard and Extreme are differents. Will all those medkits be useful later in the endgame? We will see.

Also, I made the Impossible difficulty just by frustration to the ridiculously low damages the arrows are doing to Lara. I mean, each trap in this game is lethal to Lara, while arrows are like mosquitos: -5% ...

Thus, I think the new amount of difficulties will be something like :
- Easy (former Very Easy, 0.35x),
- Normal (no change obviously)
- Hard (old updated Very Hard 3.8x)
- Extreme (no change)

Also, I want to multiply now the very little arrow's damage by the damage multiplier squared. Thus, new arrows will be:
- Easy: -50 * .35 * .35 = -6
- Normal -50 * 1 * 1 = -50 (no change obviously)
- Hard -50 * 3.8 * 3.8 = -722
- Extreme -50 * 9.5 * 9.5 = -4512 (100% lethal)

This would be the second exception to the damage multiplicator, the first one being the shootings at Lara.

And about the user choice, anyway he still able to use his own multiplicator by entering manually any value between 0.334 to 20.0

I also thought to submit a new mod where the arrows would be poisoneus instead, and where Lara would have no other choice than keep going forward while her HP going down during a certain time, forcing her to heal from time to time, but maybe this is too heavy change from the original game xD

EDIT: All 4 final difficulties damages compares: https://www.youtube.com/watch?v=7x7Vw0h-hQA

rr-:
Hey @byblo  

>I cannot remove it entirely from the passport 

I understand what you're saying, but we won't be accepting the PR in its current form. We can revisit it once we'll have developed a proper in-game settings menu.

byblo:
Yes. Lets see that later.

PR #684, opened 2023-01-01
src/game/music.c: Unused Music_Stop call in Music_Play

carlo-bramini:
#### Checklist

#### Description

I run the engine with the demo files and some custom audio music tracks and I discovered a strange bug into `src/game/music.c`. Here:
https://github.com/rr-/Tomb1Main/blob/fcdf33adf0f634c10fdf3e768f7e11c0a5c1b9b9/src/game/music.c#L73 there is this piece of code:

```
    if (track == 0) {
        Music_Stop();
        return false;
    }
```

but it seems to be never called because previously at this line: https://github.com/rr-/Tomb1Main/blob/fcdf33adf0f634c10fdf3e768f7e11c0a5c1b9b9/src/game/music.c#L65 there is this code:

```
    if (track <= 1) {
        return false;
    }
```

which returns to the caller without doing anything. In my opinion, the block of code at line 65 must be deleted and line 73 should be changed as into this PR.

### DIFF
--- a/src/game/music.c
+++ b/src/game/music.c
@@ -62,7 +62,7 @@ bool Music_Play(int16_t track)
         return false;
     }
 
-    if (track <= 1) {
+    if (track <= MX_UNUSED_1) {
         return false;
     }
 
@@ -70,12 +70,7 @@ bool Music_Play(int16_t track)
         return Sound_Effect(SFX_SECRET, NULL, SPM_ALWAYS);
     }
 
-    if (track == 0) {
-        Music_Stop();
-        return false;
-    }
-
-    if (track == 5) {
+    if (track == MX_CAVES_AMBIENT) {
         return false;
     }
 
--- a/src/game/objects/creatures/natla.c
+++ b/src/game/objects/creatures/natla.c
@@ -159,7 +159,7 @@ void Natla_Control(int16_t item_num)
                 natla->flags = 0;
                 timer = 0;
                 item->hit_points = NATLA_NEAR_DEATH;
-                Music_Play(54);
+                Music_Play(MX_NATLA_SPEECH);
             } else {
                 item->hit_points = DONT_TARGET;
             }
--- a/src/game/objects/creatures/skate_kid.c
+++ b/src/game/objects/creatures/skate_kid.c
@@ -91,7 +91,7 @@ void SkateKid_Control(int16_t item_num)
         angle = Creature_Turn(item, SKATE_KID_SKATE_TURN);
 
         if (item->hit_points < 120 && Music_CurrentTrack() != 56) {
-            Music_Play(56);
+            Music_Play(MX_SKATEKID_SPEECH);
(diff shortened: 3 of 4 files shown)


rr-:
The change looks fine, however, what's the impact on the game?
In my opinion the risk is that there are music triggers with the parameter <= 1 in the OG that used to do nothing and will now stop the music instead. It is possible that no such triggers exist or my understanding of the code is misguided, though.

carlo-bramini:
> The change looks fine, however, what's the impact on the game? In my opinion the risk is that there are music triggers with the parameter <= 1 in the OG that used to do nothing and will now stop the music instead. It is possible that no such triggers exist or my understanding of the code is misguided, though.

Yes, you are right, I discovered it just because I was playing with the script file.

walkawayy:
Music tracks 0 and 1 are not used.
https://github.com/rr-/Tomb1Main/blob/d2acd2bb8c9d87a267e0f95f1142352c56f38c93/src/global/types.h#L480

IMO the change below makes more sense. Just remove the `if (track == 0) {` block since it's never used?
```
    if (track <= 1) {
        return false;
    }

    if (g_Config.fix_secrets_killing_music && track == MX_SECRET) {
        return Sound_Effect(SFX_SECRET, NULL, SPM_ALWAYS);
    }

    if (track == 5) {
        return false;
    }
```

Also the PR name is confusing because `Music_Stop` is called in other places, just not in `Music_Play`.

rr-:
Since we're touching this function anyway, let's also replace `5` with a proper `MX_` constant.

src/game/music.c
```diff
@@ -62,20 +62,16 @@ bool Music_Play(int16_t track)
...
-    if (track <= 1) {
-        return false;
-    }
-
     if (g_Config.fix_secrets_killing_music && track == MX_SECRET) {
         return Sound_Effect(SFX_SECRET, NULL, SPM_ALWAYS);
     }
 
-    if (track == 0) {
+    if (track <= 1) {
```
walkawayy: Can we change this to `MX_UNUSED_1` as well?

Also I think rr's original point about the possibility of a music trigger with track ID 0 or 1 existing in a level is still valid. In this new code, a trigger for 0 or 1 will now stop any music being played rather than just ignoring the unused music trigger constants. If we remove this block (since it was always unused) and just use the
```
if (track <= 1) {
        return false;
}
```
above, I think that would be a better solution?

(Also I updated the PR name since it was confusing IMO. LMK if this name works better.)

walkawayy (changes requested):

src/game/music.c
walkawayy: @carlo-bramini what do you think of my suggestion?

walkawayy (approved):
New changes LGTM. I'll wait for rr to final approve though.

rr- (approved):

PR #689, opened 2023-01-05
s/audio: Get rid of lock/unlock after opening audio device

carlo-bramini:
#### Checklist

#### Description

According to:

https://wiki.libsdl.org/SDL2/SDL_OpenAudioDevice

a device is always in paused state after `SDL_OpenAudioDevice()`, so in my opinion it is safe to initialize `m_WorkingBuffer[]` and then start the playback. This will avoid the need to use lock/unlock stuff.

### DIFF
--- a/src/specific/s_audio.c
+++ b/src/specific/s_audio.c
@@ -60,15 +60,13 @@ bool S_Audio_Init(void)
         return false;
     }
 
-    SDL_PauseAudioDevice(g_AudioDeviceID, 0);
-
     m_WorkingSilence = desired.silence;
     m_WorkingBufferSize = desired.samples * desired.channels
         * SDL_AUDIO_BITSIZE(desired.format) / 8;
 
-    SDL_LockAudioDevice(g_AudioDeviceID);
     m_WorkingBuffer = Memory_Alloc(m_WorkingBufferSize);
-    SDL_UnlockAudioDevice(g_AudioDeviceID);
+
+    SDL_PauseAudioDevice(g_AudioDeviceID, 0);
 
     return true;
 }


walkawayy:
The doc page you linked also says `The audio callback runs in a separate thread in most cases; you can prevent race conditions between your callback and other threads without fully pausing playback with [SDL_LockAudioDevice](https://wiki.libsdl.org/SDL2/SDL_LockAudioDevice)(). For more information about the callback, see [SDL_AudioSpec](https://wiki.libsdl.org/SDL2/SDL_AudioSpec).`

So maybe it's safer to just leave the locks in?

carlo-bramini:
> The doc page you linked also says `The audio callback runs in a separate thread in most cases; you can prevent race conditions between your callback and other threads without fully pausing playback with [SDL_LockAudioDevice](https://wiki.libsdl.org/SDL2/SDL_LockAudioDevice)(). For more information about the callback, see [SDL_AudioSpec](https://wiki.libsdl.org/SDL2/SDL_AudioSpec).`

Yes, that's right.
But, as far as I know, until the device is paused the callback for filling new samples is not called.
The current code also looks a bit dangerous to me: the `SDL_PauseAudioDevice()` activates the device, but it is technically possible that `S_Audio_MixerCallback()` is called when `m_WorkingBuffer[]` is still `NULL`, before that the lock/unlock functions are called, at least during the `S_Audio_Init()`.

rr- (approved):
LGTM but please test this:

- Start the game. The title menu music should be playing
- Start a new game. Do not skip the first FMV. The FMV should be audible.
- Run around Caves. The background music and sounds should be audible.

Once someone confirms this is working we can merge it.

walkawayy:
> LGTM but please test this:
> 
>     * Start the game. The title menu music should be playing
> 
>     * Start a new game. Do not skip the first FMV. The FMV should be audible.
> 
>     * Run around Caves. The background music and sounds should be audible.
> 
> 
> Once someone confirms this is working we can merge it.

This worked for me.

Issue #445, opened 2022-02-19
Screenshot bugs during FMVs


Richard-L:
It looks like you can crash the game if you hit the PRINT key for a screenshot during the FMVs before the title screen. Tried both `.RPL` and `.FMV`:

![image](https://user-images.githubusercontent.com/4934209/154802438-7eee8902-29a9-469b-83be-05e61fe2a4a2.png)

Curiously during a later FMV (I tried "LIFT"), the PRINT screen will fail first but the game stays alive, and consequent PRINT hits will take screenshots as expected.
![20220219_141339_Cut_Scene_1](https://user-images.githubusercontent.com/4934209/154802464-d2b3ea1b-bbfe-4d10-b9c6-5b43758e60ef.png)

rr-:
Works for me

![20220219_213339_Level_-1](https://user-images.githubusercontent.com/1045476/154819913-67521fdd-24dd-4e94-b764-f7c21f6ec5c9.jpg)
![20220219_213339_Level_-1_2](https://user-images.githubusercontent.com/1045476/154819914-cb42f88a-3051-4d99-9eb5-4cebf4ca5f8d.jpg)
![20220219_213340_Level_-1](https://user-images.githubusercontent.com/1045476/154819915-d9452ba4-21f7-4738-a9ff-1babaa54a1a5.jpg)
![20220219_213340_Level_-1_2](https://user-images.githubusercontent.com/1045476/154819916-c5fad791-bef7-4102-9799-a04c3fc617af.jpg)
![20220219_213341_Level_-1](https://user-images.githubusercontent.com/1045476/154819917-e69858d6-6b97-4dd2-822d-5e0f715cda30.jpg)
![20220219_213341_Level_-1_2](https://user-images.githubusercontent.com/1045476/154819919-7f4cb317-2354-43c4-82b0-77c046c859e1.jpg)
![20220219_213341_Level_-1_3](https://user-images.githubusercontent.com/1045476/154819920-c9a21480-32c8-4ed7-ab17-f12915d5e277.jpg)
![20220219_213341_Level_-1_4](https://user-images.githubusercontent.com/1045476/154819921-d9c449a4-dfff-471e-a124-11654b979030.jpg)
![20220219_213341_Level_-1_5](https://user-images.githubusercontent.com/1045476/154819923-a6fc8b66-ddec-4198-900b-8d269b5a1c3c.jpg)
[...]

PR #694, opened 2023-01-10
shell: use the return value of snprintf() to save an strlen()

carlo-bramini:
#### Checklist

#### Description

While looking into issue #445, I noticed that it can be possible to avoid an `strlen()` by using the return value of the `snprintf()`. This patch implements this tiny improvement.

### DIFF
--- a/src/game/shell.c
+++ b/src/game/shell.c
@@ -41,14 +41,11 @@ static char *Shell_GetScreenshotName(void)
 {
     // Get level title of unknown length
     char level_title[100];
-    snprintf(
-        level_title, LEVEL_TITLE_SIZE, "%s",
-        g_GameFlow.levels[g_CurrentLevel].level_title);
 
-    // Trim level titles that are too long
-    if (strlen(level_title) >= LEVEL_TITLE_SIZE) {
-        level_title[LEVEL_TITLE_SIZE] = '\0';
-    }
+    strncpy(
+        level_title, g_GameFlow.levels[g_CurrentLevel].level_title,
+        LEVEL_TITLE_SIZE - 1);
+    level_title[LEVEL_TITLE_SIZE] = '\0';
 
     // Prepare level title for screenshot
     char *check = level_title;


rr- (approved):

src/game/shell.c
```diff
@@ -41,12 +41,12 @@ static char *Shell_GetScreenshotName(void)
 {
     // Get level title of unknown length
     char level_title[100];
-    snprintf(
+    size_t out_size = snprintf(
```
walkawayy: It's possible for snprintf to return a negative value, so I'm not sure size_t will work? Isn't size_t always unsigned and snprintf returns an int?

https://cplusplus.com/reference/cstdio/snprintf/

src/game/shell.c
```diff
@@ -41,12 +41,12 @@ static char *Shell_GetScreenshotName(void)
...
         level_title, LEVEL_TITLE_SIZE, "%s",
         g_GameFlow.levels[g_CurrentLevel].level_title);
 
     // Trim level titles that are too long
-    if (strlen(level_title) >= LEVEL_TITLE_SIZE) {
+    if (out_size >= LEVEL_TITLE_SIZE) {
```
walkawayy:
```
If the resulting string would be longer than n-1 characters, the remaining characters
are discarded and not stored, but counted for the value returned by the function.

A terminating null character is automatically appended after the content written.
```
So I don't think we need this check or trim at all? snprintf should already do that.

walkawayy (changes requested):

src/game/shell.c
rr-: I double checked and we don't need the manual null terminator.

src/game/shell.c
carlo-bramini: I suspect that `out_size` was made as `size_t` because it was used as parameter to `Memory_Alloc()`.

src/game/shell.c
carlo-bramini: It depends.
In the past, on Windows, `snprintf()` was historically made available as a macro to `_snprintf()`, which doesn't add a terminating null character if the size exceedes. Recent updates have a compliant `snprintf()` implementation, but older does not.

walkawayy: I went googling down this rabbit hole. What a mess lol. Like you said, it does seem like since Visual Studio 2015, `snprintf` is properly cross platform. Source from[ MS docs](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/snprintf-snprintf-snprintf-l-snwprintf-snwprintf-l?redirectedfrom=MSDN&view=msvc-170): `Beginning with the UCRT in Visual Studio 2015 and Windows 10, snprintf is no longer identical to _snprintf. The snprintf function behavior is now C99 standard conformant.`. I'd say it's safe to use since VS 2015 has been available 8 years now.


Here's an open source implementation from ffmpeg we could potentially use if we really need older than VS 2015 support:
https://github.com/FFmpeg/FFmpeg/blob/master/compat/msvcrt/snprintf.c

src/game/shell.c
walkawayy: Can we we make `out_size` and `int` and error check if the return value is not `-1` for both `snprintf` calls? Then maybe cast to `size_t` in the `Memory_Alloc` call?

Also this might be an @rr- question, but maybe we can update this PR to solve #445 while we are messing with this code?

rr-: IMO unrelated change, I'd touch that in a separate PR / commit.

src/game/shell.c
rr-: These nuances and compilation gotchas are chiefly the reason why I'm against officially supporting non-Docker builds. While I might lean towards supporting a modern version of VS, going out of our way with `#ifdefs` or choosing not to use the standard library in order to support VS2015 or any other ancient build stacks is a no-no from me. (I'm OK with supporting old client software such as Windows XP which is a different matter altogether)

Which is why I'd dismiss the `_snprintf` insanity and said we don't need to handle null terminators by hand.

walkawayy: Agreed. Which is why I think it's fine to support `snprintf` as defined [here](https://cplusplus.com/reference/cstdio/snprintf/).

carlo-bramini: Perhaps, another alternative could be also to use `SDL_snprintf()`, since the SDL has a dedicated function if the system library does not provide an implementation compliant to the standard.

https://github.com/libsdl-org/SDL/blob/9b861d2ea4fbae0718a0c1ac37397ed5ba2a5b67/src/stdlib/SDL_string.c#L1429

rr-: That's an option but personally I'm against coupling with SDL more than necessary.

carlo-bramini: At this point, after removing the test on the final NUL character, `out_size` can be removed from the first `snprintf()` and restored at its previous place. In my opinion, `snprintf()` is not expected to return `-1` here since there are no file writes and the format parameter does not include encoding conversions.

carlo-bramini: > That's an option but personally I'm against coupling with SDL more than necessary.

Or it could be possible to do this, as it has been done into `filesystem.c` and `text.c`:

```
strncpy(level_title, g_GameFlow.levels[g_CurrentLevel].level_title, LEVEL_TITLE_SIZE-1);
level_title[LEVEL_TITLE_SIZE] = '\0';
```

which solves the trouble about the possibility of returning an error code at the root, but not issue #445 yet.

rr-: `strcpy` looks most straightforward TBH.
BTW I think this discussion grew unproportionally big in regards to the impact of this change.

walkawayy (approved):
Changes LGTM and I confirmed it works with some testing. After this we can fix the FMV screenshot bug.

rr- (approved):

PR #695, opened 2023-01-11
s/shell: Add support for older FFmpeg (support for Windows XP)

carlo-bramini:
#### Checklist

#### Description

I have been able to run the engine also on Windows XP.
The last version of FFmpeg able to work on this OS is v3.4.12 "Cantor".
However, before version 4.x.x, FFmpeg required to call `av_register_all()` as first thing, otherwise it won't work anything. On newer versions, this function has been deprecated and no longer available. So, I would like to suggest to add this tiny fix, that calls `av_register_all()` if the selected libraries will need it.

### DIFF
--- a/src/specific/s_shell.c
+++ b/src/specific/s_shell.c
@@ -9,6 +9,8 @@
 #include "log.h"
 #include "memory.h"
 
+#include <libavcodec/avcodec.h>
+
 #define SDL_MAIN_HANDLED
 
 #ifdef _WIN32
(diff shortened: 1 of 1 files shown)


walkawayy (approved):
I admit I don't know the most about ffmpeg, but this looks ok to me.

rr- (approved):
@walkawayy can you check if this builds with our Docker image? And if it does merge? Thanks!

walkawayy:
Worked fine for me on Windows 11.

PR #711, opened 2023-01-17
rat: fix dead voles looking alive when a room's water drains

### LINKED ISSUE #687: Rats' death position is broken after killing them in the water and draining the water out.
Rats' death position is broken after killing them in the water and draining the water out. 
This wasn't the case in the OG version. They seem like they are still alive.
20221229_190338_The_Cistern

walkawayy:
Resolves #687.

#### Checklist

#### Description
Fixed dead water rats looking alive when the room's water is drained. Broken since 2aa66a0ce3f54d298f89f2ae619089467fd01fff.

The game used to do a hard coded check for the level number. If it was Cistern, it would make the dead vole active again as a rat, so it would animate through it's death animation. With that removed, the vole becomes a rat but stays in its death animation's base frame which looks alive.

image

With this change, the hard coded check is still removed. But, the end frame is used so the rat looks dead.
image

This hard coded level check and animation was not used for the crocodile because it's death animation's base frame looks dead.

image

Let me know what you think of this fix. (Btw something interesting I discovered is the dead vole uses `Item_Animate` while dead so the dead item isn't deactivated in `Creature_Animate`.)
...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - fixed save crystal collision pushing Lara through walls (#682)
 - fixed passport animation when deselecting the passport (#703)
 - fixed inconsistent wording in config tool health and air color options (#705)
+- fixed dead water rats looking alive when the room's water is drained (#687, regression from 0.12.2)
 
 ## [2.12.1](https://github.com/rr-/Tomb1Main/compare/2.12...2.12.1) - 2023-01-16
 - fixed crash when using enhanced saves in levels with flame emitters (#693)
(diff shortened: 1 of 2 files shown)


rr- (commented):
The linked commit only removes comments. Are you sure you didn't mean a2be7014 which introduces the `#if 0` `#endif`? 

Other than that I remember that check and it's good to finally know what it was for.

What I believe should happen is that the swimming rats (voles) are kept alive once a room is drained and get converted to regular rats. And regular rats should start swimming once a room is flooded. I believe this used to work only in Cistern but I don't see why it shouldn't be working in all levels like this.

I think the current version will just kill off the rats once a room is flooded. Is this correct?

walkawayy:
Ya looks like that commit you found is the one that commented it out. Currently the rats behave as expected except when they die in the water and they lower to the ground when the water drains. But, they  appear alive after the drain because they are in the wrong frame of their death animation. The original code and the code I changed make it so they are at the end of their death animation and look dead.

rr- (approved):
Thanks for the explanation, LGTM

PR #713, opened 2023-01-18
sound: do triggered flip effects even there is no sound

### LINKED ISSUE #583: Trigger flip effects don't work if sound devices are disabled
This is a bit esoteric, but if all sound output devices are disabled in Windows, the lights never turn on at the beginning of Atlantis.

walkawayy:
Resolves #583.

#### Checklist

#### Description
Fixed triggered flip effects not working if there are no sound devices.
...

### DIFF
--- a/src/game/effects.c
+++ b/src/game/effects.c
@@ -1,6 +1,7 @@
 #include "game/effects.h"
 
 #include "game/output.h"
+#include "game/room.h"
 #include "global/const.h"
 #include "global/types.h"
 #include "global/vars.h"
@@ -144,10 +145,10 @@ void Effect_Draw(int16_t fxnum)
     }
 }
 
-void Effect_DoRoutine(int32_t effect_num)
+void Effect_DoRoutine()
 {
     // XXX: Some of the FX routines rely on the item to be not null!
-    if (effect_num != -1) {
-        g_EffectRoutines[effect_num](NULL);
+    if (g_FlipEffect != -1) {
+        g_EffectRoutines[g_FlipEffect](NULL);
     }
 }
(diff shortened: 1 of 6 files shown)


src/game/effects.c
```diff
@@ -143,3 +143,11 @@ void Effect_Draw(int16_t fxnum)
         Matrix_Pop();
     }
 }
+
+void Effect_DoRoutine(int32_t effect_num)
```
rr-: I'd remove this argument and name the function `Effect_RunActiveFlipEffect` or something. In general I'd say that the API around `g_FlipEffect` is poorly designed and we rely in many places on this singleton, so it's OK to do it here. Refactoring the effects to allow multiple routines to run at the same time would require us to do heavier refactors, and perhaps introduce something in the spirit of `g_Items` – a lot of code for doubtful benefit as I don't think this would ever be leveraged in the OG.

src/game/effects.h
```diff
@@ -13,3 +13,4 @@ int16_t Effect_Create(int16_t room_num);
 void Effect_Kill(int16_t fx_num);
 void Effect_NewRoom(int16_t fx_num, int16_t room_num);
 void Effect_Draw(int16_t fxnum);
+void Effect_RunActiveFlipEffect();
```
rr-: Since this function takes no arguments it should be marked as `(void)`.

src/game/effects.c
```diff
@@ -143,3 +144,11 @@ void Effect_Draw(int16_t fxnum)
         Matrix_Pop();
     }
 }
+
+void Effect_RunActiveFlipEffect()
```
rr-: Here as well

rr- (approved):

PR #718, opened 2023-01-25
s/input: remove redundant S_Input_KbdKey()

carlo-bramini:
#### Checklist

#### Description

It is not clear to me the link between `S_Input_KbdKey()` and `S_Input_Key()`. They are both static and `S_Input_Key()` just calls `S_Input_KbdKey()` with the same parameters. This looks a redundant stuff. In this patch, I deleted `S_Input_Key()` and I renamed `S_Input_KbdKey()`, which requires the minimum number of changes.


rr- (approved):

walkawayy (approved):
Oops good catch. Yes I was originally going to use the second function for something and ended up changing my mind while rewriting things for the controls rework. Change looks good.

PR #722, opened 2023-01-26
Add support for demo levels.

carlo-bramini:
#### Checklist

#### Description

This patch adds support for playing the demo levels.
Details at issue #667.


walkawayy:
Gonna link the demos posted in the issue: https://allgamestaff.altervista.org/download-tomb-raider/tr-1/

carlo-bramini:
The demo levels are also available at the official page of Core Design:

https://core-design.com/goodies_tr1_downloads.html

Richard-L:
@carlo-bramini are you planning to support all the different demos? I didn't know about the one bundled with terracide. It's nice because it includes the gym and both the first two peru levels, incl their FMVs.
I'd love a windows binary if you had to playtest with the demo files a bit.

vvs-:
> The demo levels are also available at the official page of Core Design:
> 
> https://core-design.com/goodies_tr1_downloads.html

FWIW, this is not the official site: https://core-design.com/about.html

rr-:
@carlo-bramini good work on figuring out the demo file structure, however I have a question – how can it work without initializing the game palette?

Additionally, a couple of comments on the chosen design:

- For a better sustainability – can you please change the `int8_t demo` variable to a new enum `LEVEL_TYPE`? That way we pave a way for a potential support of Saturn levels in the future and also for the bit addressed in the next point.
- In case there is more than a single PC demo I'd lean towards not merging this until we support all the PC demo variants.
- I'd remove all the comments from the gameflow file, similar to how there are no comments in the UB variant.
- I'd replace the new CLI option with a `--gameflow NAME` that would look for the gameflow file in `cfg/Tomb1Main_gameflow_{NAME}.json` or `.json5`.

carlo-bramini:
> @carlo-bramini good work on figuring out the demo file structure, however I have a question – how can it work without initializing the game palette?

Excuse me, it is not clear to me what you mean here...

carlo-bramini:
> For a better sustainability – can you please change the `int8_t demo` variable to a new enum `LEVEL_TYPE`? That way we pave a way for a potential support of Saturn levels in the future and also for the bit addressed in the next point.

Are you suggesting about adding a new item into `GAMEFLOW_LEVEL_TYPE` into `src\global\types.h`, is it right?
So, for the full game, this will mean that since each level has already a property called `"type"` and set to `normal/gym/cutscene/current`, we will set it for example to `"demo_level_pc"`, if I understood correctly.

EDIT: I did a first change to the patch according this suggestion, is this what you were thinking?

carlo-bramini:
> FWIW, this is not the official site: https://core-design.com/about.html

Ouch, my mistake, I have been cheated by the URL, which identical to the original one, now probably defunct and replaced:
https://en.wikipedia.org/wiki/Core_Design

walkawayy:
I think we are good to close this now @rr-

Issue #396, opened 2022-02-04
OG bug: Pierre's dropped items disappear after reload


rr-:
Food for thought for #277 
I'm surprised nobody reported this earlier.

Applies to all dynamically spawned items.

walkawayy:
Hm. It seems like the game tried to do something for that and maybe it's just broken in `savegame.c`
```
        if (item->object_number == O_PIERRE && item->hit_points <= 0
            && (item->flags & IF_ONESHOT)) {
            if (Inv_RequestItem(O_SCION_ITEM) == 1) {
                SpawnItem(item, O_MAGNUM_ITEM);
                SpawnItem(item, O_SCION_ITEM2);
                SpawnItem(item, O_KEY_ITEM1);
            }
            g_MusicTrackFlags[MX_PIERRE_SPEECH] |= IF_ONESHOT;
        }
```

rr-:
Don't tackle this yet @walkawayy I'll be doing heavy refactors to the savegame system and we'll end up having tons of conflicts otherwise.

rr-:
It's now free to take, as we merged the big stuff in. It'd be nice to include in 2.6 but I have no clue what might be the reason why it's not working hence I haven't added it.

walkawayy:
One option for why it was bugged:
I couldn't figure out how to correctly get the legacy save system to save and load to see why it's broken in the OG game (the requestor was legacy but saving didn't work). The OG game saves flags like this:
```
            uint16_t flags = item->flags + item->active + (item->status << 1)
                + (item->gravity_status << 3) + (item->collidable << 4);
```
while the new system saves variables like `gravity_status` separately. So that's  The flags were saved wrong somehow.

The second reason it could be bugged:
I am just testing this on the new BSON save games for now. Something odd is going on with Scions.

1. I grabbed the one in Qualopec giving me 1 Scion in inventory.
2. I killed Pierre and he drops one Scion.
3. I saved without picking it up.
4. I reload the save and I have 3 Scions.

This seems to happen when Pierre is loaded which is odd. The game records a Scion piece as `O_SCION_ITEM`, `O_SCION_ITEM2`, and `O_SCION_OPTION` and then the `if (Inv_RequestItem(O_SCION_ITEM) == 1)` fails because it returns 3 and not 1. I think T1M then has an additional bug that is tripling it in the inventory screen too. So maybe the OG bug would be fixed if this tripling is fixed? Here's a log snippet:

```
[...]

PR #724, opened 2023-01-28
data: create injection framework and braid object

### LINKED ISSUE #27: Feature request: insert the braid mesh in every level
Related to #8. Currently, the hair mesh is only available in the Lost Valley level. Requires input from @Arsunt on how to structure the softpatch JSON structure for levels.

lahm86:
Resolves #27.

#### Checklist

#### Description

Added a method to allow data injection at runtime. This is initially targeted at making a braid model available in every level.

The framework currently supports full model injection using a custom binary file format, which is essentially an extremely stripped back level file, with a few additions. These files have been generated externally, but happy to open discussion about where best to place this tool - whether it's part of the T1M toolset or if it remains external.

The bin files have types, and these types are intended to be associated with T1M config options. So while the gameflow has the injection command for the braid, this is skipped when the option itself is disabled.

For the provided braid, some additional mesh edits are performed to make it fit better on Lara's model.

Every original level has been tested, along with all UB levels. In addition, this was also tested on every level in Sabatu's TR1 as a TRLE test.


rr-:
Thank you for taking ownership of this. I skimmed through the initial changes to `level.c` and one concern that just occurred to me is about the savegame files, that they might crash the game in the case when the player:

- has the braid disabled, starts a new game
- kills an entity that spawns an item drop (Pierre in Tihocan for example)
- exits the game, enables the braid
- loads the game

Can someone confirm if that's the case?

Regarding the tool location, my initial instinct is to create a GitHub organization and migrate all of our Tomb Raider stuff there. But I'd prefer if that included TOMB3-5, TR2Main, TEN, Rando etc and getting this many projects and people to align seems daunting. I'd like to hear others' opinions on this.

lahm86:
> Thank you for taking ownership of this. I skimmed through the initial changes to `level.c` and one concern that just occurred to me is about the savegame files, that they might crash the game in the case when the player:
> 
> * has the braid disabled, starts a new game
> * kills an entity that spawns an item drop (Pierre in Tihocan for example)
> * exits the game, enables the braid
> * loads the game
> 
> Can someone confirm if that's the case?
> 
> Regarding the tool location, my initial instinct is to create a GitHub organization and migrate all of our Tomb Raider stuff there. But I'd prefer if that included TOMB3-5, TR2Main, TEN, Rando etc and getting this many projects and people to align seems daunting. I'd like to hear others' opinions on this.

I can test the save/load scenario tomorrow. Is the braid added to the global item list, so your concern is that the reload would cause a mismatch with the previous "last item"? I know this happens if entity types are changed between loads but as we're not (currently at least) injecting new entities I think it might be OK. Will test and get back to you.

A GitHub organization would be very useful imo. There is a lot of knowledge and tools out there so having it collated would be very helpful. Agreed though that it'd be a lot of work to put together.

lahm86:
~~There is an issue with drops, but it's not related to the injection work as it's present on the current release. If you kill Pierre but don't pick up his items, then save and reload, the items do not re-spawn nor are they in your inventory. Sounds related to #396.~~
Ignore this - I had level skipped past ToQ so didn't have the scion. By playing through properly, this issue doesn't appear.

But in any case, there is no crash with the scenario above where you save without the braid, enable the braid, then reload that save, and equally vice-versa.

bin/cfg/Tomb1Main_gameflow.json5
```diff
@@ -55,6 +55,11 @@
             // music track ID
             "music": 0,
 
+            // data injection file paths
+            "injections": [
+                "data\\braid.bin"
```
rr-: Can we change `\` to `/` and rebase against newest `develop`?

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## [Unreleased](https://github.com/rr-/Tomb1Main/compare/stable...develop)
 - added support for .ogg, .mp3 and .wav formats for audio tracks (#688)
 - added the mummy to the level kill stats if Lara touches it and it falls (#701)
+- added the ability to inject data into levels, with Lara's braid being the initial focus (#27)
```
rr-: Can we move it to the first place as it's the killer feature of the next release?

src/game/inject.c
```diff
@@ -0,0 +1,721 @@
...
+        object->mesh_index += level_info->mesh_ptr_count;
+        File_Read(&object->bone_index, sizeof(int32_t), 1, fp);
+        object->bone_index += level_info->anim_bone_count;
+
+        File_Read(&tmp, sizeof(int32_t), 1, fp);
+        object->frame_base =
+            &g_AnimFrames[(tmp + level_info->anim_frame_count * 2) / 2];
+        File_Read(&object->anim_index, sizeof(int16_t), 1, fp);
+        object->anim_index += level_info->anim_count;
+        object->loaded = 1;
```
rr-, line 414: A lot of this code seems similar to the level loader in level.c with occassional follow-up calls to `Inject_` static functions.
Could we maybe share these functions between these two modules, favoring `level.c` and making the `inject` more lightweight? Would it make the code prettier, or is my suggestion ill-formed?

src/game/inject.c
```diff
@@ -0,0 +1,721 @@
...
+#include "util.h"
+
+#include <stddef.h>
+
+#define BIN_VERSION 1
+
+typedef enum INJECTION_TYPE {
+    INJ_GENERAL = 0,
+    INJ_BRAID = 1,
+} INJECTION_TYPE;
```
walkawayy, line 21: Will all new injections require a new `INJECTION_TYPE`? Or is `INJ_BRAID` needed because it links to `g_Config.enable_braid` and most injections will be `INJ_GENERAL`? Just curious regarding future injections.

src/game/inject.c
lahm86: I agree, there is a lot of the same kind of code, but my gut feeling was to keep it as separated as possible, mainly because for example on line 408 above, we need to be careful that we're using the right frame index base when injecting. I feel that it might end up needing a lot of additional arguments passed around. Also, thinking ahead, when we come to properly injecting such things as animations into already existing objects (as opposed to injecting the full object set) we'll need to be careful to keep frames, meshes etc contiguous with respect to the objects, so again that kind of thing would need additional information passed around if it were handled in a shared function.
It might be good to come back to this once we have a fuller set of injection types supported maybe? Then we'll know exactly what can be shared out. But if you feel that's not optimal for now, I can work through it and see how the shared function approach will look.

src/game/inject.c
lahm86: Yes, `INJ_GENERAL` would probably be targeted at default injections, if there are any like this in the future (like any data "bugs" we'd always want to fix). For others, there is flexibility where we can create further general types, or have super-specific ones. So there might be a general `INJ_TEXTURE_FIX` type in the future, linked to `g_Config.fix_og_texture_issues` for example, and for this there could be a `caves_texture_fixes.bin`, `vilcabamba_texture_fixes.bin` etc injection files. Or there might be a very specific texture fix in a level that we may not want to bundle with the general one, so that would have its own type and config option (texture fixes maybe aren't the best example but hopefully you know what I mean).
Happy to talk about a different approach to linking the injections to config options though if we feel this is a bit clunky.

src/game/inject.c
rr-: Hmm, I agree that this approach leaves us with more flexibility especially that we know more things are planned. Let's try to remember to revisit this later once we have more data. Perhaps a comment `// TODO: consider refactoring once we have more injection use cases`?

src/game/inject.c
rr-: I'm OK with this the way it is at the moment, I think it's quite clean and better than anything I was able to think of ATM.

src/game/inject.c
lahm86: Ok, that sounds like a good plan. I've added this comment for the time being.

lahm86:
While going through `inject.c`, I spotted an optimisation issue, which I've put a fix in for [here](https://github.com/rr-/Tomb1Main/pull/724/commits/3925abec74273bf0083c5321565cc8b529a68951). `Input_RemapRGB` could potentially have been called several thousands of times in a single injection, and repeatedly for the same colour. Now a LUT of `source colour index => level colour index` is created in the first instance and the subsequent functions use this instead of wastedly re-mapping the same colours time after time.

walkawayy (approved):
The changes LGTM.

src/game/game/game_cutscene.c
```diff
@@ -69,6 +78,36 @@ int32_t Game_Cutscene_Start(int32_t level_num)
...
+
+    GAME_OBJECT_ID lara_type = g_GameFlow.levels[level_num].lara_type;
+    if (lara_type == O_LARA) {
+        return;
+    }
+
+    for (int i = 0; i < g_LevelItemCount; i++) {
+        if (g_Items[i].object_number != lara_type) {
+            continue;
+        }
```
rr-, line 97: Can we change it to
```c
ITEM_INFO *lara = NULL;
for (int i = 0; i < g_LevelItemCount; i++) {
    if (g_Items[i].object_number == lara_type) {
        lara = &g_Items[i];
        break;
    }
}
if (!lara) {
    return;
}

do_stuff(lara);
```

src/game/lara/lara_hair.c
```diff
@@ -104,7 +112,8 @@ void Lara_Hair_Control(bool in_cutscene)
 
     // hips
     Matrix_Push();
-    objptr = g_Lara.mesh_ptrs[LM_HIPS];
+    objptr = in_cutscene ? g_Meshes[object->mesh_index + LM_HIPS]
+                         : g_Lara.mesh_ptrs[LM_HIPS];
```
rr-: Would it be possible to always use `object` and never refer to `g_Lara` directly, or at most once? I'd like to keep the `if (in_cutscene)` check to minimum or ideally do without it altogether.

lahm86: > Would it be possible to always use `object` and never refer to `g_Lara` directly, or at most once? I'd like to keep the `if (in_cutscene)` check to minimum or ideally do without it altogether.

Yes, no problem to make a `mesh_base` of sorts, then it's a single check at the beginning for the Lara type. We do still need `g_Lara.mesh_ptrs` to track normal in-game mesh swaps, so the check would set `mesh_base` to either these pointers or the relevant position in `g_Meshes`.

We could also change `in_cutscene` to a local variable in the function - or even a module variable - and remove it as an argument - so `in_cutscene = m_LaraType != O_LARA` ? It's still needed further down for the water level checks, but at least it would mean all calls to `Lara_Hair_Control` would be the same.

rr-: >We could also change in_cutscene to a local variable in the function - or even a module variable - and remove it as an argument - so in_cutscene = m_LaraType != O_LARA ? It's still needed further down for the water level checks, but at least it would mean all calls to Lara_Hair_Control would be the same.

Yes I'd prefer it to be this way.

src/game/game/game_cutscene.c
lahm86: Slight variation on the proposed change, as we need to pass the item number to `Lara_InitialiseLoad` rather than the item pointer. But regardless, the loop is now exited swiftly.

src/game/lara/lara_hair.c
```diff
@@ -48,9 +50,10 @@ void Lara_Hair_Initialise(void)
 void Lara_Hair_SetLaraType(GAME_OBJECT_ID lara_type)
 {
     m_LaraType = lara_type;
+    m_InCutscene = m_LaraType != O_LARA;
```
rr-: After seeing it in action, let's turn it into a normal variable within `Lara_Hair_Control`. We should keep as little state global as possible.

rr- (approved):

Issue #748, opened 2023-03-07
Textboxes drawn over text


Opaque79:
Since 2.13, the text boxes are being drawn above the text obscuring it. Any ideas?
![Tomb1Main 2023-03-07 12-27-19-11](https://user-images.githubusercontent.com/127208702/223422364-99f3335f-9979-4a94-8fb2-4f2a0c7be994.png)

The passport textures also glitch while it's opening too
https://i.imgur.com/abYCCk2.png

aredfan:
My guess is you'll need to overwrite the gameflow files with the latest ones, see if that fixes your issue.

Opaque79:
Unfortunately, I've already tried that and it didn't work. I will try again though.

EDIT: No joy unfortunately, even with no changes made to the cfg files

walkawayy:
Maybe we need to try and figure out what exact hardware is having issues. I have an AMD 7700x CPU and an Nvidia 1070 GPU with no issue.

Could anyone having this problem please post their specs?

oziphantom:
probably something to do with the drivers not following the same sorting order when we draw so it draws the polygons in different orders because everything is same Z, its technically legal. My guess at the moment.

Opaque79:
My specs are:
CPU - AMD Ryzen 7 2700X
GPU - AMD Radeon RX 5700 XT

I could roll back to 2.12.1 and lose the PSX Uzi sound

walkawayy:
So far the issue seems limited to AMD GPUs. From Discord, there was another report where an AMD Ryzen 7 3750H with Radeon Mobile Gfx had the issue. Switching to the Nvidia GeForce GTX 1650 dual GPU fixed it.

Opaque79:
Switching GPU's isn't really practical so it seems my only option so far is to go back to 2.12.1 and lose the PSX Uzi sound.

It only affects text boxes, the rest of the game is fine btw

walkawayy:
> Switching GPU's isn't really practical so it seems my only option so far is to go back to 2.12.1 and lose the PSX Uzi sound.
> 
> It only affects text boxes, the rest of the game is fine btw

Of course. That user was on a laptop with a dual GPU. Just trying to gather data to find out why it's broken. I agree with @oziphantom about a Z Buffer issue. But I don't think it's new driver related because T1M 2.12 works fine. @carlo-bramini do you think one of the changes you made could have caused this?

Opaque79:
[...]

PR #752, opened 2023-03-09
savegame: fix effect crash

### LINKED ISSUE #746: Crash when loading a save with spaz_effect_count set
If the game is saved while Lara is in the following animation and then reloaded, a crash occurs. This is present in TombATI as well as T1M (found by Radgryd during randomizer testing).

20230304_144736_Atlantis_6

Saving during Lara's normal animation doesn't lead to a crash.
20230304_144737_Atlantis

The cause I think is that while `spaz_effect_count` and `spaz_effect` are [saved](https://github.com/LostArtefacts/Tomb1Main/blob/develop/src/game/savegame/savegame_bson.c#L1065), only `spaz_effect_count` is [restored](https://github.com/LostArtefacts/Tomb1Main/blob/develop/src/game/savegame/savegame_bson.c#L712).

The crash seems to occur below as `g_Lara.spaz_effect_count` is used in the condition, but `g_Lara.spaz_effect` is assumed to be set.

https://github.com/LostArtefacts/Tomb1Main/blob/develop/src/game/lara/lara_control.c#L158

lahm86:
Resolves #746.

#### Checklist

#### Description

This will restore the explosion effect on load that puts Lara into the state in question, provided that enhanced saves are enabled (otherwise there won't be an effect to reference). With enhanced saves off, and for legacy saves, the guard in `lara_control.c` avoids the null pointer issue previously seen.

A short demo of the fix in action can be seen at https://www.youtube.com/watch?v=7H3evEL6Qgk - both the effect and Lara's stance are restored on load.


rr-:
I think this should become 2.13.2 shouldn't it?

lahm86:
> I think this should become 2.13.2 shouldn't it?

Sounds good, although I think it's a very rare bug as you need to have saved on exactly the right frame. Maybe a bug fix release along with the fix for #748 once that's in and verified?

walkawayy:
> > I think this should become 2.13.2 shouldn't it?
> 
> Sounds good, although I think it's a very rare bug as you need to have saved on exactly the right frame. Maybe a bug fix release along with the fix for #748 once that's in and verified?

I agree this is a super rare bug that's been in place since the original game. But since we need a hotfix for the AMD GPU depth buffer issue, we can probably release this fix alongside it in 2.13.2.

walkawayy (approved):

rr-:
Cherry-picked by hand to the `stable` branch.

PR #765, opened 2023-03-13
shell: fix a crash when taking a screenshot of an opening FMV

### LINKED ISSUE #445: Screenshot bugs during FMVs

walkawayy:
Resolves #445.

#### Checklist

#### Description
Fixed a crash when taking a screenshot of an opening FMV. If `g_CurrentLevel` isn't initialized yet, I just gave the level name for the screenshot as `Intro` like `20230313_160009_Intro.jpg`. FMVs don't have level names / titles like levels and cutscenes, so they use whatever `g_CurrentLevel` was set to previously. For example, the elevator FMV is called `20230313_152329_Cut_Scene_1.jpg` since the cutscene with Larson after Qualopec plays right before. If we want to use the correct FMV filenames in the screenshot names, there would need to be a larger rework, but I wanted to mention it just in case that's something we want.
...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - fixed texture issues in the Cowboy, Kold and Skateboard Kid models (#744)
 - fixed the savegame requestor arrow's position with a large number of savegames and long level titles (#756)
 - fixed empty holsters when starting a level with the shotgun equipped (#749)
+- fixed a crash when taking a screenshot of an opening FMV (#445)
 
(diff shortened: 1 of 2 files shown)


lahm86 (approved):
Looks like a good workaround 👍

rr- (approved):

PR #768, opened 2023-03-14
lara: improve braid control

### LINKED ISSUE #761: Feature request: improve braid control
Currently, the control for Lara's braid can result in some jagged behaviour. This is most noticeable when picking up an item, using a lever, pushing/pulling a moveable block, and at the beginning of the Tomb of Tihocan cutscene. @Trxyebeep pointed out that this is because the OG hair control in the classics does not take interpolation into account and has kindly offered that we can take and adapt the code from tomb3 to resolve this.

Another issue is that the braid can fall through the floor, e.g. if doing a handstand, or at the end of the Atlantis cutscene when Lara reaches Torso's platform. Floor detection should be added, but note that in cutscenes the room that Lara is in will need to be calculated on the fly, as `g_LaraItem->room_number` does not update in this situation.

lahm86:
Resolves #761.

#### Checklist

#### Description
These improvements have been translated from tomb3 (thanks to @Trxyebeep for assistance with this).

Animation comparison (before and after):
https://youtu.be/kVBBDyNjZ7w

Floor detection:
image

The change in `matrix.c` warranted extensive testing as this function is used by so many different animations. I completed a full run of the game, interacting with all moveable objects and checking their behaviour. Originally, I had copied the final `else` clause from [tomb3](https://github.com/Trxyebeep/tomb3/blob/master/tomb3/game/draw.cpp#L236), but this was causing issues with animations with higher framerates, such as Lara starting to push a block, or crocodiles swimming. Reverting to the original else clause here resolved each of these issues.

There is one outstanding issue with the braid as far as I can tell: when Lara turns to gold in Midas, the braid meshes do not. I will raise a separate issue for this as it's more data related, and it will need some thought on the best way to resolve it.

### DIFF
--- a/src/math/matrix.c
+++ b/src/math/matrix.c
@@ -338,19 +338,32 @@ void Matrix_Interpolate(void)
     MATRIX *mptr = g_MatrixPtr;
     MATRIX *iptr = m_IMMatrixPtr;
 
-    if (m_IMRate == 2) {
-        mptr->_00 = (mptr->_00 + iptr->_00) / 2;
-        mptr->_01 = (mptr->_01 + iptr->_01) / 2;
-        mptr->_02 = (mptr->_02 + iptr->_02) / 2;
-        mptr->_03 = (mptr->_03 + iptr->_03) / 2;
-        mptr->_10 = (mptr->_10 + iptr->_10) / 2;
-        mptr->_11 = (mptr->_11 + iptr->_11) / 2;
-        mptr->_12 = (mptr->_12 + iptr->_12) / 2;
-        mptr->_13 = (mptr->_13 + iptr->_13) / 2;
-        mptr->_20 = (mptr->_20 + iptr->_20) / 2;
-        mptr->_21 = (mptr->_21 + iptr->_21) / 2;
-        mptr->_22 = (mptr->_22 + iptr->_22) / 2;
-        mptr->_23 = (mptr->_23 + iptr->_23) / 2;
+    if (m_IMRate == 2 || (m_IMFrac == 2 && m_IMRate == 4)) {
+        mptr->_00 += (iptr->_00 - mptr->_00) >> 1;
+        mptr->_01 += (iptr->_01 - mptr->_01) >> 1;
+        mptr->_02 += (iptr->_02 - mptr->_02) >> 1;
+        mptr->_03 += (iptr->_03 - mptr->_03) >> 1;
+        mptr->_10 += (iptr->_10 - mptr->_10) >> 1;
+        mptr->_11 += (iptr->_11 - mptr->_11) >> 1;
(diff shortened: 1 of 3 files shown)


src/math/matrix.c
```diff
@@ -338,19 +338,32 @@ void Matrix_Interpolate(void)
...
+        mptr->_02 += (iptr->_02 - mptr->_02) >> 2;
+        mptr->_03 += (iptr->_03 - mptr->_03) >> 2;
+        mptr->_10 += (iptr->_10 - mptr->_10) >> 2;
+        mptr->_11 += (iptr->_11 - mptr->_11) >> 2;
+        mptr->_12 += (iptr->_12 - mptr->_12) >> 2;
+        mptr->_13 += (iptr->_13 - mptr->_13) >> 2;
+        mptr->_20 += (iptr->_20 - mptr->_20) >> 2;
+        mptr->_21 += (iptr->_21 - mptr->_21) >> 2;
+        mptr->_22 += (iptr->_22 - mptr->_22) >> 2;
+        mptr->_23 += (iptr->_23 - mptr->_23) >> 2;
```
rr-, line 366: I don't quite get these changes, are they necessary? Because on the first glance they seem like a hand-rolled optimization of (x*m_IMFrac/m_IMRate) operation, however the conditions seem to be too inclusive.

rr- (approved):
LGTM Although I do agree that the matrix interpolator seems sketchy.

src/math/matrix.c
lahm86: This is the result without these changes.

https://www.youtube.com/watch?v=9UaecY3xYPU

I believe it resolves an overflow issue (but I don't fully understand it I have to admit); I don't know if this affects anything in the current release. Perhaps we should make a separate function just for hair interpolation to play it safe?

rr-: Nah it's good, since you tested it and that's what Tomb3 uses anyway.

walkawayy (approved):
It looks really good! The hair segments still look like they can rotate even while Lara stands still., and they clip through her neck on the push block animation. Obviously these are minor details, but I checked out this [TR2 playthrough](https://www.youtube.com/watch?v=c63Y2f0MZ24). The braid does the exact same thing, so I guess I never paid enough attention. I think this is good to go.

lahm86:
> It looks really good! The hair segments still look like they can rotate even while Lara stands still., and they clip through her neck on the push block animation. Obviously these are minor details, but I checked out this [TR2 playthrough](https://www.youtube.com/watch?v=c63Y2f0MZ24). The braid does the exact same thing, so I guess I never paid enough attention. I think this is good to go.

Yeah, I think it behaves better in TR4/5 where it's properly jointed. Maybe something for the future :)

PR #770, opened 2023-03-14
lara: turn braid to gold with Midas touch

### LINKED ISSUE #769: Lara's braid does not turn to gold
When Lara turns to gold, her braid does not. This will need additional mesh injection and suitable mesh swapping.

20230314_111954_Palace_Midas

lahm86:
Resolves #769.

#### Checklist

#### Description

The new bin file contains a copy of the standard braid meshes, but these are textured as coloured triangles/rectangles using the same palette entry as golden Lara. We then detect when Lara's head turns to gold, and then shift the hair mesh index.

The other change removes the same issue we had in the Atlantis cutscene, where the mesh spheres interfere with the hair position. The first half of the clip below highlights that issue, and the second half is the final outcome achieved in this update.

https://www.youtube.com/watch?v=-rYRmP178TQ


walkawayy (commented):
Nice, looks good.

walkawayy (approved):

src/game/lara/lara_hair.c
```diff
@@ -74,7 +74,9 @@ void Lara_Hair_Control(void)
 
     in_cutscene = m_LaraType != O_LARA;
     object = &g_Objects[m_LaraType];
-    mesh_base = in_cutscene ? &g_Meshes[object->mesh_index] : g_Lara.mesh_ptrs;
+    mesh_base = (in_cutscene || (g_Lara.mesh_effects & (1 << LM_HIPS)))
+        ? &g_Meshes[object->mesh_index]
+        : g_Lara.mesh_ptrs;
```
rr-: I don't quite understand this change

lahm86: When Lara's hips turn to gold, the collision sphere changes, which makes the braid stand to attention.
https://github.com/LostArtefacts/Tomb1Main/blob/develop/src/game/lara/lara_state.c#L686

image

This was the same problem we had in the Atlantis cutscene with the pistol animation in that level. So we tell it in these cases to use the original Lara meshes and not those from model `O_LARA_EXTRA` (so, ignore whatever has been put into `g_Lara`'s mesh pointers and use what was defined for the model originally).
image

`g_Lara.mesh_effects` is only used with this Midas routine, so it seemed the easiest way to detect when she is turned to gold rather than introducing a new variable or function call somewhere between this function and hair control.

walkawayy: In that case maybe you also need to test the T-Rex death. That also uses the `O_LARA_EXTRA`. And `TORSO_ATTACK3` also uses it.

lahm86: Good point. I wonder if it might actually be better just to always use the original meshes instead of having this check. I'll do some testing tomorrow.

rr-: I'd vote for original meshes.
Not sure but maybe the fact that it refers to swapped meshes is related to vehicles from the later engines? IDK though if they involve mesh swaps.

lahm86: It works fine using the original meshes at all times, so I've committed that update. I've checked each of these cases:

* Gym (mesh swap)
* Regular outfit (with each gun type)
* Mauled by T-Rex (mesh swap)
* Collecting scion in Qualopec (O_LARA_EXTRA animation, but no mesh swap)
* Qualopec cutscene
* Midas touch (progressive mesh swap)
* Tihocan cutscene
* Collecting scion in Sanctuary (O_LARA_EXTRA animation, but no mesh swap)
* Atlantis cutscene
* Torso death (O_LARA_EXTRA animation, but no mesh swap)

rr- (approved):

PR #776, opened 2023-03-23
overlay: change bars to better match PS1 and fix scaling options

### LINKED ISSUE #698: Reinstate classic health and air bar style
Apologies in advance for posting something so nitpicky and annoyingly "purist", but I still cant help notice this after all this time.

If…
```json5
"enable_smooth_bars" : false,
```
…could I ask to have the authentic PS1 health and air bar style be reinstated?

When TR1M came out I thought this was just a case of having to find the right bar scaling, but I could never reproduce the same classic look.

I looked into it a bit more, and the differences are of course rather marginal:
- bigger grey border
- no bevel effect in the border
- bigger black padding inside the border
- the 5 colours making the gradient are slightly different to what TR1M uses, and each in equal height

Health Bar

Air Bar

It just feels to me that if someone chooses to not use the smooth bars, they should nail the classic look.

Thanks for considering.

walkawayy:
Resolves #698.

#### Checklist

#### Description
Changed the health, air, and enemy bars to better match the PS1 version. Fixed the health, air, and enemy bars from being affected by the text scaling option,
...

### DIFF
--- a/src/game/overlay.c
+++ b/src/game/overlay.c
@@ -401,13 +401,7 @@ void Overlay_DrawAmmoInfo(void)
     const int32_t screen_margin_h = 24;
     const int32_t screen_margin_v = 18;
 
-    // Scale ammo string properly if text scaling used.
-    double scale_ammo_to_bar = g_Config.ui.bar_scale;
-    if (g_Config.ui.text_scale > 1) {
-        scale_ammo_to_bar /= g_Config.ui.text_scale;
-    } else if (g_Config.ui.text_scale < 1) {
-        scale_ammo_to_bar *= 1 / g_Config.ui.text_scale;
-    }
+    double scale_ammo_to_bar = g_Config.ui.bar_scale / g_Config.ui.text_scale;
 
     char ammostring[80] = "";
 
--- a/src/game/overlay.c
+++ b/src/game/overlay.c
@@ -27,17 +27,17 @@ static int32_t m_OldGameTimer = 0;
 
 static RGBA8888 m_ColorBarMap[][COLOR_STEPS] = {
     // gold
-    { { 112, 92, 44, 255 },
-      { 164, 120, 72, 255 },
-      { 112, 92, 44, 255 },
-      { 88, 68, 0, 255 },
-      { 80, 48, 20, 255 } },
+    { { 124, 94, 37, 255 },
+      { 161, 131, 60, 255 },
+      { 124, 94, 37, 255 },
+      { 100, 70, 19, 255 },
+      { 76, 46, 2, 255 } },
     // blue
-    { { 100, 116, 100, 255 },
-      { 92, 160, 156, 255 },
-      { 100, 116, 100, 255 },
-      { 76, 80, 76, 255 },
-      { 48, 48, 48, 255 } },
+    { { 61, 113, 123, 255 },
+      { 101, 146, 154, 255 },
+      { 61, 113, 123, 255 },
+      { 31, 93, 107, 255 },
+      { 0, 74, 91, 255 } },
     // grey
     { { 88, 100, 88, 255 },
       { 116, 132, 116, 255 },
@@ -169,9 +169,9 @@ static void Overlay_BarBlink(BAR_INFO *bar_info)
 static void Overlay_BarGetLocation(
     BAR_INFO *bar_info, int32_t *width, int32_t *height, int32_t *x, int32_t *y)
 {
-    const int32_t screen_margin_h = 20;
+    const int32_t screen_margin_h = 25;
     const int32_t screen_margin_v = 18;
-    const int32_t bar_spacing = 8;
+    const int32_t bar_spacing = 16;
 
     if (bar_info->location == BL_CUSTOM) {
         *width = bar_info->custom_width;
@@ -187,18 +187,17 @@ static void Overlay_BarGetLocation(
     } else if (
         bar_info->location == BL_TOP_RIGHT
         || bar_info->location == BL_BOTTOM_RIGHT) {
-        *x = Screen_GetResWidthDownscaled() - *width * g_Config.ui.bar_scale
-            - screen_margin_h;
+        *x = Screen_GetResWidthDownscaledBar() - *width - screen_margin_h;
     } else {
-        *x = (Screen_GetResWidthDownscaled() - *width) / 2;
+        *x = (Screen_GetResWidthDownscaledBar() - *width) / 2;
     }
 
     if (bar_info->location == BL_TOP_LEFT || bar_info->location == BL_TOP_CENTER
         || bar_info->location == BL_TOP_RIGHT) {
         *y = screen_margin_v + m_BarOffsetY[bar_info->location];
     } else {
-        *y = Screen_GetResHeightDownscaled() - *height * g_Config.ui.bar_scale
-            - screen_margin_v - m_BarOffsetY[bar_info->location];
+        *y = Screen_GetResHeightDownscaledBar() - *height - screen_margin_v
+            - m_BarOffsetY[bar_info->location];
     }
 
     m_BarOffsetY[bar_info->location] += *height + bar_spacing;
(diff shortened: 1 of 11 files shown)


src/game/screen.c
```diff
@@ -96,6 +106,21 @@ int32_t Screen_GetRenderScale(int32_t unit)
     return MIN(scale_x, scale_y);
 }
 
+int32_t Screen_GetRenderScaleBar(int32_t unit)
+{
+    int32_t baseWidth = 640;
+    int32_t baseHeight = 480;
```
rr-: Can we fix this naming convention while we're at it? In both functions.

src/game/screen.c
```diff
@@ -96,6 +106,21 @@ int32_t Screen_GetRenderScale(int32_t unit)
...
+    int32_t scale_x = Screen_GetResWidth() > baseWidth
+        ? ((double)Screen_GetResWidth() * unit * g_Config.ui.bar_scale)
+            / baseWidth
+        : unit * g_Config.ui.bar_scale;
+    int32_t scale_y = Screen_GetResHeight() > baseHeight
+        ? ((double)Screen_GetResHeight() * unit * g_Config.ui.bar_scale)
+            / baseHeight
+        : unit * g_Config.ui.bar_scale;
+    return MIN(scale_x, scale_y);
+}
```
rr-, line 114: These two functions are practically identical. I suggest to refactor things a bit:

```c
int32_t Screen_GetRenderScaleBase(
    int32_t unit, int32_t base_width, int32_t base_height, double factor
) {
￼    int32_t scale_x = Screen_GetResWidth() > base_width
￼        ? ((double)Screen_GetResWidth() * unit * factor)
￼            / base_width
￼        : unit * factor;
￼    int32_t scale_y = Screen_GetResHeight() > base_height
￼        ? ((double)Screen_GetResHeight() * unit * factor)
￼            / base_height
￼        : unit * factor;
￼    return MIN(scale_x, scale_y);
}

int32_t Screen_GetRenderScaleText(int32_t unit) {
    return Screen_GetRenderScaleBase(unit, 640, 480, g_Config.ui.text_scale);
}

int32_t Screen_GetRenderScaleBar(int32_t unit) {
    return Screen_GetRenderScaleBase(unit, 640, 480, g_Config.ui.bar_scale);
}
```

and replace `Screen_GetRenderScale` usages with `Screen_GetRenderScaleText`.

src/game/screen.h
```diff
@@ -11,13 +11,16 @@ int32_t Screen_GetResIdx(void);
 int32_t Screen_GetResWidth(void);
 int32_t Screen_GetResHeight(void);
 int32_t Screen_GetResWidthDownscaled(void);
+int32_t Screen_GetResWidthDownscaledBar(void);
```
rr-, line 15: If you agree with my other suggestion we should rename `Screen_GetResWidthDownscaled` to `Screen_GetResWidthDownscaledText` as well for consistency.

walkawayy: Ok I'll refactor a little. I didn't want to do it yet in case you didn't want to.

walkawayy:
Here's a video showing some different combinations of text/bar scaling. I think 2.0 scaling is definitely ridiculous lol. Lmk if we wanna cap it at 1.5 or so.

https://www.youtube.com/watch?v=o_bxKSqZp30

And here's an example showing the behavior will work with the enemy health bar also in the top right.

https://www.youtube.com/watch?v=RiIP47ESseg

walkawayy:
Here's a test build. Called it v3.1 since it's just some code refactoring.

[Tomb1Main Classic Bars v3-1.zip](https://github.com/LostArtefacts/Tomb1Main/files/11052957/Tomb1Main.Classic.Bars.v3-1.zip)

Richard-L:
Very nice tests.

It looks like the gap between bar and ammo count+icon depends on bar scale. Had this better been independent perhaps? You see in your Text x2, Bar x.5 example, the ammo icon touches the bottom of the bar because no gap is left.
The reverse also happens in the other extreme (Text x.5, Bar x2), where then there is too much gap under the bar.

How is the gap calculated atm?

The enemy bar in the same spot is quite misleading if you ask me, as it looks so similar to the air bar in this particular colour, but I guess the user might set this. Almost makes me think if the config tool would ideally ask about bar positions in concession, where health and air come first, and then enemy bars cannot be placed in the same spot to avoid this confusion.

One other thing I notice: if you have weapons equipped and jump into water, the health bar immediately transforms into an air bar.

https://user-images.githubusercontent.com/4934209/227374017-41dc9306-62b8-4e32-9982-abb2f23f4501.mp4

Would this speak in favour of fixed separate positions for both bars for added clarity?

walkawayy:
The bars and text have completely different scaling options. I tried to make it so mixing and matching them dynamically yields good results. Which I think it does.

`The enemy bar in the same spot is quite misleading if you ask me, as it looks so similar to the air bar in this particular colour, but I guess the user might set this.` Yes I only tested that to make sure nothing overlaps incorrectly. The user can configure that, but I'd guess almost no one will. And it's definitely not a default setting.

I don't think the video you posted is any reason to change anything. Having fixed positions for the bars just wastes screen real estate IMO. The default colors are different enough to avoid confusion.

rr-:
Let's cap the bar scales at 1.5.

The gap is negligible because nobody's gonna play with this combination, and it'll be gone after we cap the scale at 1.5, so this can be dismissed.

PS1 fixed placement stems from lazy coding and is just aesthetically unpleasing. The bars are distinguishble enough with the colors like walkawayy said. The bar placement is to remain dynamic.

walkawayy:
Lmk if this is ok. Or I can make separate `MIN_TEXT_SCALE .5f` and `MIN_BAR_SCALE .5f` constants.

src/game/overlay.c
```diff
@@ -400,12 +395,20 @@ void Overlay_RemoveAmmoText(void)
...
+    const int32_t screen_margin_h = 24;
     const int32_t screen_margin_v = 18;
 
+    // Scale ammo string properly if text scaling used.
+    double scale_ammo_to_bar = g_Config.ui.bar_scale;
+    if (g_Config.ui.text_scale > 1) {
+        scale_ammo_to_bar /= g_Config.ui.text_scale;
+    } else if (g_Config.ui.text_scale < 1) {
+        scale_ammo_to_bar *= 1 / g_Config.ui.text_scale;
+    }
```
rr-: Wait, how are these two operations different? (Is `1` getting promoted to a double `1.0`?)

src/global/const.h
```diff
@@ -11,7 +11,8 @@
 #define PHD_135 (PHD_45 * 3) // = 24576 = 0x6000
 
 #define MIN_UI_SCALE 0.5f
-#define MAX_UI_SCALE 2.0f
+#define MAX_TEXT_SCALE 2.0f
+#define MAX_BAR_SCALE 1.5f
```
rr-: Let's make these a `double` by removing the `f `suffix.

src/game/overlay.c
walkawayy: Oops you are right these are the same. I was testing some different formulas lol. I'll do:
```
// Scale ammo string properly if text scaling used.
double scale_ammo_to_bar = g_Config.ui.bar_scale / g_Config.ui.text_scale;
```

src/global/const.h
walkawayy: Maybe they should be floats and I should change all the use cases to float:
```
    struct {
        double text_scale;
        double bar_scale;
        UI_STYLE menu_style;
    } ui;
```
?

That would match:
```
#define MIN_BRIGHTNESS 0.1f
#define MAX_BRIGHTNESS 2.0f
#define DEFAULT_BRIGHTNESS 1.0f
```

And they are read in `config.c` as floats:
```
    READ_FLOAT(ui.text_scale, DEFAULT_UI_SCALE);
    CLAMP(g_Config.ui.text_scale, MIN_UI_SCALE, MAX_TEXT_SCALE);
```

Also lmk if I should make separate `MIN_TEXT_SCALE .5f` and `MIN_BAR_SCALE .5f` constants.

rr-: Let's leave it as double. `READ_FLOAT` is actually ill-formed because it returns doubles:

```c
#define READ_FLOAT(opt, default_value)
    READ_PRIMITIVE(json_object_get_double, opt, default_value)
```

but let's not touch that here.

As for the constants, at first I liked the current way, but you're right, it'll be more consistent to have separate constants for the minimum values so let's go with your suggestion.

rr- (approved):

PR #778, opened 2023-03-23
overlay: fix equipped weapon's ammo showing on inventory screen

### LINKED ISSUE #777: Equipped ammo text is shown in the inventory screen
20230322_212856_Colosseum

walkawayy:
Resolves #777.

#### Checklist

#### Description
Fixed the equipped weapon's ammo showing on the inventory screen.

I removed some unused instances of `g_OverlayFlag` since that was just confusing things. That flag was no longer used for `Overlay_DrawGameInfo` because `Game_DrawScene(bool draw_overlay)` takes the `draw_overlay` flag to determine whether the overlay is drawn. The pause screen also is self contained in its own mini loop and doesn't need `g_OverlayFlag`.

Let me know if this is OK or if you want to fix it another way. `Overlay_DrawFPSInfo` isn't called every loop, so it can no longer be used to do the removal. Also `Overlay_OnAmmoTextRemoval` and `Overlay_OnFPSTextRemoval` seem redundant so let me know if we want to keep those. No other text seems to make use of `on_remove`. I couldn't tell from when exactly this is a regression. Seems to have happened during one of the refactors.
...

### DIFF
--- a/src/game/game/game_draw.c
+++ b/src/game/game/game_draw.c
@@ -20,12 +20,17 @@ int32_t Game_ProcessFrame(void)
     return g_Camera.number_frames;
 }
 
+#include "log.h"
+
 void Game_DrawScene(bool draw_overlay)
 {
     if (g_Objects[O_LARA].loaded) {
         Room_DrawAllRooms(g_Camera.pos.room_number);
+        LOG_DEBUG("draw_overlay: %d", draw_overlay);
         if (draw_overlay) {
             Overlay_DrawGameInfo();
+        } else {
+            Overlay_RemoveAmmoText();
         }
     } else {
         // cinematic scene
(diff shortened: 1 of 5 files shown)


src/game/game/game_draw.c
```diff
@@ -20,12 +20,17 @@ int32_t Game_ProcessFrame(void)
     return g_Camera.number_frames;
 }
 
+#include "log.h"
+
 void Game_DrawScene(bool draw_overlay)
 {
     if (g_Objects[O_LARA].loaded) {
         Room_DrawAllRooms(g_Camera.pos.room_number);
+        LOG_DEBUG("draw_overlay: %d", draw_overlay);
```
rr-: Please remove debug logs

rr- (commented):
LGTM. I'm slightly worried about a potential race if the user presses the key to display the fps counter on the same frame as opening the pause/inventory screen.

walkawayy:
I spammed a bunch of Escape/F2 at the same time for about two minutes without triggering anything.

rr- (approved):

PR #779, opened 2023-03-24
dart: added dart emitters to the savegame

### LINKED ISSUE #774: Dart emitter status isn't stored in saved games
Hello, i'm currently trying to build a TR1 level using Tomb1Main and i'm facing a problem.

Dart traps status aren't saved. 
The traps are active on level load, then i'm using a pickup trigger to deactivate. They do that correctly, but when i save and load they are back to on.

I know i can use "timer" on the trigger to set a "distance" for the trap to be active, but it would be best if it could work without it.
Is it possible to fix this?

walkawayy:
Resolves #774.

#### Checklist

#### Description
Added dart emitters to the savegame.
...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## [Unreleased](https://github.com/rr-/Tomb1Main/compare/stable...develop)
 - added Spanish localization to the config tool
 - added an option to launch Unfinished Business from the config tool (#739)
+- added dart emitters to the savegame (#774)
 - fixed Larson's gun textures in Tomb of Qualopec to match the cutscene and Sanctuary of the Scion (#737)
 - fixed texture issues in the Cowboy, Kold and Skateboard Kid models (#744)
 - fixed the savegame requestor arrow's position with a large number of savegames and long level titles (#756)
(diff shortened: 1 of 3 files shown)


rr- (commented):
Can we create a static function in `savegame_legacy.c` for the flags checker, seeing how we repeat the same list of IDs 3 times in a row? Ideally with a comment why this is needed.

Something such as

```c
static bool Savegame_Legacy_ItemHasSaveFlags(ITEM_INFO *item) {
    // T1M savegame files are enhanced to store more information by having
    // changed the save_flags bit for certain item types. However, legacy
    // TombATI saves do not contain the information that's associated with
    // these flags for these enhanced items, and the way they are structured,
    // whether this information exists or not, cannot be figured out from the
    // save file alone. So the object IDs that got changed are listed here
    // to make sure the legacy savegame reader doesn't try to reach out for
    // information that's not there.
    return (
        obj->save_flags
        && item->object_number != O_LAVA_EMITTER
        && item->object_number != O_FLAME_EMITTER
        && item->object_number != O_WATERFALL
        && item->object_number != O_SCION_ITEM
        && item->object_number != O_DART_EMITTER
    );
}
```

lahm86 (approved):
Looks great to me. I had a play around with the level @Rapora9 provided and it's all working fine.

rr- (approved):

PR #783, opened 2023-03-25
music: fix partial masked triggers killing the ambient track

### LINKED ISSUE #763: Music stopped by triggers with incomplete masks
In St. Francis' Folly, when using the keys at the end of the level (room 42), the current music track is stopped. This is because these triggers are masked, and the intended behaviour is for track 15 to be played when the door opens (although it doesn't because the heavy trigger under the boulder is missing the music entry, but that is a separate issue - see #755). The code currently stops the music if all bits are not set in the mask.

This also happens in Temple of the Cat room 12, 32 and 43. In these cases, the intended track does play, but once that finishes, the ambient track does not restart.

image

lahm86:
Resolves #763.

#### Checklist

#### Description

This resolves the ambient track being stopped in three places in the original game. The areas in St. Francis' Folly and Temple of the Cat are detailed in the issue; the third area is in Tomb of Qualopec, when passing through room 25 heading back towards room 24.

image

The antipad here is intended to stop track 17 from playing again on exit, but I believe this is actually redundant as there is handling already to avoid repeated tracks (no other track is triggered in this area); this should perhaps be addressed properly with floor data manipulation (#755), but for now at least the ambient track is restored on exiting this area.

A potential side effect of this could be in custom levels where it may be desirable to kill all music, including the ambient track, using a trigger. I'm not sure if we wish to retain a method to achieve this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - fixed Lara's braid not turning to gold during the Midas touch animation (#769)
 - fixed the equipped weapon's ammo showing on the inventory screen (#777)
 - fixed the health, air, and enemy bars from being affected by the text scaling option (#698)
+- fixed music triggers with partial masks killing the ambient track (#763)
 - improved the control of Lara's braid to result in smoother animation and to detect floor collision (#761)
 - increased the number of effects from 100 to 1000 (#623)
 
--- a/src/game/music.c
+++ b/src/game/music.c
@@ -145,6 +145,19 @@ void Music_Stop(void)
     Music_StopActiveStream();
 }
 
+void Music_StopTrack(int16_t track)
+{
+    if (track != Music_CurrentTrack()) {
+        return;
+    }
+
+    Music_StopActiveStream();
+
+    if (m_TrackLooped >= 0) {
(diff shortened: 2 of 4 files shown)


rr- (approved):
LGTM although I'm in no position to comment about level building, so this might enjoy some more consultation with the people.

walkawayy:
> LGTM although I'm in no position to comment about level building, so this might enjoy some more consultation with the people.

Same thoughts here.

walkawayy (approved):

lahm86:
We could do this, which would allow builders to use an antipad set with music track 0, and that would be interpreted as "kill all music". This may be better handled as a new issue as it's technically a feature, the other method (if it was ever used) is more of an exploit of a bug I feel.

```C
static void Room_TriggerMusicTrack(int16_t track, int16_t flags, int16_t type)
{
    if (track == MX_UNUSED_0 && type == TT_ANTIPAD) {
        Music_Stop();
        return;
    }

    if (track <= MX_UNUSED_1 || track >= MAX_CD_TRACKS) {
        return;
    }
    ...
```

rr-:
Yeah this sounds good.

PR #789, opened 2023-03-27
music: retain active track during enemy speeches

### LINKED ISSUE #762: Feature request: don't kill music when enemies talk
Similar to #17, it would be nice to provide an option to prevent enemy speeches killing the current music track. Each of these tracks has a corresponding SFX entry in the relevant level file which could be used.

* MX_BALDY_SPEECH
* MX_COWBOY_SPEECH
* MX_LARSON_SPEECH
* MX_NATLA_SPEECH
* MX_PIERRE_SPEECH
* MX_SKATEKID_SPEECH

lahm86:
Resolves #762.

#### Checklist

#### Description

Similar to the fix for secrets, enemy speeches will no longer stop the active track. Skateboard Kid needed a little extra work otherwise his speech plays endlessly. The flag implemented here isn't restored if you save after he has started his speech and then reload, but this is because of how the flags are reset on load.

https://github.com/LostArtefacts/Tomb1Main/blob/develop/src/game/savegame/savegame.c#L96

The situation is the same without this option enabled, as the standard music track would also be reset on load.

Short video of the feature in action:
https://www.youtube.com/watch?v=ODfV82X7jmo

### DIFF
--- a/src/game/music.c
+++ b/src/game/music.c
@@ -82,6 +82,12 @@ bool Music_Play(int16_t track)
         return Sound_Effect(SFX_SECRET, NULL, SPM_ALWAYS);
     }
 
+    if (g_Config.fix_speeches_killing_music && track >= MX_BALDY_SPEECH
+        && track <= MX_SKATEKID_SPEECH) {
+        return Sound_Effect(
+            SFX_BALDY_SPEECH + track - MX_BALDY_SPEECH, NULL, SPM_ALWAYS);
+    }
+
     if (track == MX_CAVES_AMBIENT) {
         return false;
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - added an option to launch Unfinished Business from the config tool (#739)
 - added dart emitters to the savegame (#774)
 - added the ability for level builders to stop all music via triggers (#785)
+- added an option to prevent enemy speeches stopping the current music track (#762)
 - changed the health, air, and enemy bars to better match the PS1 version (#698)
 - fixed Larson's gun textures in Tomb of Qualopec to match the cutscene and Sanctuary of the Scion (#737)
 - fixed texture issues in the Cowboy, Kold and Skateboard Kid models (#744)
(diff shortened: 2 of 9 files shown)


src/game/music.c
```diff
@@ -82,6 +82,12 @@ bool Music_Play(int16_t track)
         return Sound_Effect(SFX_SECRET, NULL, SPM_ALWAYS);
     }
 
+    if (g_Config.fix_speeches_killing_music && track >= MX_BALDY_SPEECH
+        && track <= MX_SKATEKID_SPEECH) {
+        return Sound_Effect(
+            SFX_BALDY_SPEECH + track - MX_BALDY_SPEECH, NULL, SPM_ALWAYS);
+    }
```
rr-, line 89: Sanity check: won't this cause an audible degradation of sound fidelity?

lahm86: It does, but to my ear it isn't too bad, especially with the background music playing. Here is a comparison, the first half being the normal mode, the second half using SFX.

https://user-images.githubusercontent.com/33758420/228029212-07f2d082-ceca-4da7-954d-9f5a8bc86b88.mp4

It's perhaps a caveat of having this option enabled, unless there is perhaps a better way to implement this.

walkawayy: Yes the secret killing music fix also has the problem. Is there a way for the audio handler to play music types (more a question for rr probably)?

walkawayy:
Do we have a way to determine how long a sound is? I wonder if we could lower the music/ambient volume while the speech plays, and then raise it back when it ends so the player can hear the audio more clearly.

lahm86:
> Do we have a way to determine how long a sound is? I wonder if we could lower the music/ambient volume while the speech plays, and then raise it back when it ends so the player can hear the audio more clearly.

Could possibly interpret the WAV SFX header. But in the YouTube video demo above, I had the main audio volume at 8 and SFX at 5, just to emphasise that the background was still playing (notably for Larson and Baldy as the BGM here is quite soft). Here it is with both set to 8:

https://youtu.be/mLTlOyIyNfE

src/game/music.c
rr-: To my ears this sounds almost indistinguishable, I really have to look out for the difference, even on headphones.
Please note that I'm hard of hearing so I'm not the best person to judge this.

walkawayy: I'm listening in quality headphones, and they sound extremely similar to me. (Granted, over Bluetooth)

rr-:
OK so in my personal opinion this is good enough because the player can opt to disable this, and I'd leave it as-is.

rr- (approved):

walkawayy (approved):

PR #790, opened 2023-03-28
data: convert Pyramid secret fix to injection format

### LINKED ISSUE #788: Convert the Great Pyramid secret fix into injection format
With floor data injection in the planning (#755), the first step should be converting the current fix for the Great Pyramid secret that does not register into data injection format. This will involve modifying the current config option to become generic for such floor data fixes.

lahm86:
Resolves #788.

#### Checklist

#### Description

This is a start to floor data injection, with the ability to target and replace trigger parameters. Some forward planning has been done to the injection header format to cater for floor data insertion in the future; the change in `inject.h` is the cause of the previous 11 bin files having changed. The only additional bin file here is `pyramid_fd.bin`, which now contains the fix for the third secret in this level.

The removal of the `fix_pyramid_secret` gameflow sequence means that players will need to be sure to overwrite their gameflow files on the next release. I have tested the fix on the DOSBox version of the level, which has the issue, and the TombATI version, which is already fixed; equally I tested disabling the option. The secret count is correct in each instance, and those that should register do so.

The wording in the config tool remains as-is for now although the underlying setting has become generic; this will be updated in the next floor data update.


rr- (approved):

walkawayy (approved):
LGTM.

From your description, it sounds like different floor data fixes will be optional? That's probably for the best so people can mix and match floor data fixes they want.

lahm86:
> LGTM.
> 
> From your description, it sounds like different floor data fixes will be optional? That's probably for the best so people can mix and match floor data fixes they want.

Yes, so things like the missing PSX enemy in Khamoon would be a separate option, but others that are bugs/oversights in the data would be grouped under this new option.

PR #802, opened 2023-04-10
lara: add jump-twist animations

### LINKED ISSUE #88: Feature request: jump twists from TR2+

lahm86:
Resolves #88.

#### Checklist

#### Description

This adds an option to enable jump-twists and swan dive somersaults to Lara's animation set. This involved changing the way animations are switched to, in order to allow Lara's base animation index to be non-zero. A follow-up task will be raised to copy this idiom to all other objects for consistency.

This has been tested with old and new save files with the config option enabled and disabled. In the rare case where a save is made during a jump-twist, and the option is then disabled, Lara will gracefully complete that animation on load but will not perform any more until the option is re-enabled.

The state change control in `lara_state.c` has been largely adopted from tomb3, so thanks to @Trxyebeep for assistance here.

One other point of note is that the inclusion of the empty `Lara_State_Twist` function is to ensure that the size of `g_LaraStateRoutines` remains consistent with the `LARA_STATE` enum membership. Otherwise the likes of [this](https://github.com/LostArtefacts/Tomb1Main/blob/develop/src/game/lara/lara_control.c#L204) could result in unknown behaviour.


rr- (approved):
:heart:

This looks great.

Can we also add another follow-up task, to clean up the `AF_` constants? First it'd be good to see which of the `AF_` constants actually need to be defined – I think we could safely replace most of these with -1 as they'll be equal to `g_Anims[item->anim_number].frame_base` anyway. Another thing is that the prefix is inconsistent with `LS_` and `LA_` and should be `LAF_` or similar.

lahm86:
@rr- I have added the Spanish translation (thank you @ajtudela) and I also noticed I had missed a few anim switch calls for Lara. These are in the latest commit - https://github.com/LostArtefacts/Tomb1Main/pull/802/commits/0c745abad8ff68bff1ac5132f1ce03dab935eb08. I've tested each of these scenarios and it's all working fine.

I'll raise an issue once this is closed to address tidying up the `AF_` constants.

rr- (approved):
There's an `int16_t` var declaration that declares two symbols at once, I'd prefer if we declared each variable separately.

lahm86:
> There's an `int16_t` var declaration that declares two symbols at once, I'd prefer if we declared each variable separately.

Just to confirm, this is what you are referring to? https://github.com/LostArtefacts/Tomb1Main/pull/802/commits/149f06028a8c2f25ee2fb0137df34ffec76a01f5

rr-:
> > There's an `int16_t` var declaration that declares two symbols at once, I'd prefer if we declared each variable separately.
> 
> Just to confirm, this is what you are referring to? [149f060](https://github.com/LostArtefacts/Tomb1Main/commit/149f06028a8c2f25ee2fb0137df34ffec76a01f5)

Yes :)

PR #809, opened 2023-04-18
lara: refactor animation enums

### LINKED ISSUE #806: Refactor: update usage of LARA_ANIMATION_FRAME
* Remove all `LARA_ANIMATION_FRAME` values that are not required i.e.:
  * where they are not referenced within the codebase
  * where their values match the `frame_base` of the relevant animation anyway, hence making them redundant
* Rename all remaining enum values to the format `LAF_...`

lahm86:
Resolves #806.

#### Checklist

#### Description

This removes the unnecessary `LARA_ANIMATION_FRAME` values as detailed in the table [here](https://github.com/LostArtefacts/Tomb1Main/issues/806#issuecomment-1509872668) - i.e. all those that had the same value as the frame base of the relevant animation. The `Item_SwitchToAnim` calls have each been updated with `-1` as the frame value to allow the animation's natural frame base to be used. In addition, any that were unreferenced in the codebase have been removed, such as `AF_GRABLEDGE_OLD`.

`LARA_SHOTGUN_ANIMATION_FRAME` and `LARA_GUN_ANIMATION_FRAME` values have been renamed so there is consistency between these, `LARA_ANIMATION_FRAME` and `LARA_ANIMATION`. 

I also removed the `LARA_ANIMATION` values that weren't referenced anywhere in the codebase.

Each scenario has been tested, including situations that don't occur in OG (such as Lara swimming out of a water room with a non-water room directly below).


rr- (approved):
LGTM although I have a question, can these removed `LA_` animations be useful in any way to the builders if we don't ever use them in the code?

walkawayy (approved):

lahm86:
> LGTM although I have a question, can these removed `LA_` animations be useful in any way to the builders if we don't ever use them in the code?

They are mostly used already e.g. 96 is in a dispatch such as below - I think the difference with these is that there is no special treatment needed for them in the code. The game just changes the state based on input and interprets what to do from the data.

image

I suppose in the future some request may need them reinstated, for example if we wanted some event to happen on a certain animation and frame that otherwise wouldn't be possible with the current change/dispatch/command mechanism. Maybe it would be better to leave all the animations defined (and add the others so it's complete from 0 - 159) for clarity/reference of what Lara can do rather than relying on checking level data?

walkawayy:
I thought about doing that at one point in the past, but it's a big undertaking for a hypothetical scenario imo. My idea was to try and match WAD Tool closely if possible.

PR #810, opened 2023-04-18
lara: add a cheat to explode Lara like in TR2 and TR3

### LINKED ISSUE #793: Add exploding Lara cheat from TR2 / TR3
Suggested by a friend using Tomb1Main.
"It's a shame there isn't an exploding Lara cheat in TR1"

Considering Tomb1Main has numerous keyboard shortcut cheats, it would be a funny novelty for a key to make Lara explode into meshes.

walkawayy:
Resolves #793.

#### Checklist

#### Description
We can't use the exploding SFX like in TR2 / TR3 because it's not in all the level files, so I made Lara use her long fall scream instead. I looked at tomb3 for reference on how to implement this. I used the swan dive at the end of the level skip / all weapon cheat since the other games required flares (TR2) or a non pistol gun in hand (TR3) while performing the same cheat. Seemed best to make this its own thing.

https://streamable.com/kcwh6t
...


rr-:
The original issue was about offering a keyboard shortcut for this. Let's ask around what people think of it. IMO this feature is borderline too unorthodox. 

The sounds shouldn't be an issue with the injection framework

walkawayy:
> The original issue was about offering a keyboard shortcut for this. Let's ask around what people think of it. IMO this feature is borderline too unorthodox. 
> 
> 
> 
> The sounds shouldn't be an issue with the injection framework

Ok. I figured do it similar to TR2. https://youtu.be/aUWlsrCE-Xw

Richard-L:
Imo this isn't worth adding. But since it's walkaway having some fun adding the infamous nude cheat, I didn't want to voice anything before you asked for opinions.

walkawayy:
I looked into it because someone opened an issue for it and it seemed quick to add.

lahm86:
It's a fun addition imo :)

rr-:
Sounds good. Could we inject the explosion sound tho?

walkawayy:
Ya Lahm already messaged me it. I'll add it to the PR.

Richard-L:
You didn't like any of the atlanteans explosion sounds here?

rr-:
I said nothing of the sort. Thing is, it's not available in half of the levels.
As far as I know, missing sprites and sounds is the  only reason why early mummies explode silently and we should make sure it's kept this way.

walkawayy:
Ok I pushed an update. Lahm kindly gave me a new .bin with the explosion SFX. I added to the `SOUND_EFFECT_ID`s with `SFX_EXPLOSION_CHEAT = 206`, so we don't mess with the early mummy explosions.

https://streamable.com/19l9wj

rr- (approved):

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## [Unreleased](https://github.com/rr-/Tomb1Main/compare/stable...develop) - ××××-××-××
 - added an option to enable TR2+ jump-twist and somersault animations (#88)
 - added the ability to unbind the sidestep left and sidestep right keys (#766)
+- added a cheat to explode Lara like in TR2 and TR3 (#793)
```
rr-, line 4: Actually can we also add it to the readme to the list of features?

lahm86 (approved):

PR #816, opened 2023-04-25
lara_look: add inverted look option

walkawayy:
Resolves #700.

#### Checklist

- [X] I have read the [coding conventions](https://github.com/rr-/Tomb1Main/blob/master/CONTRIBUTING.md#coding-conventions)
- [X] I have added a changelog entry about what my pull request accomplishes, or it is an internal change

#### Description
Adds a camera option to the controls menu. Adds an inverted look option to that new menu.
...

PR #817, opened 2023-04-25
lara_look: refactor all looks to use the correct functions

walkawayy:
#### Checklist

#### Description
Refactor all looks to use the generic functions. Prep for inverted look #816. No changelog entry because it's an internal change.
...

### DIFF
--- a/src/game/lara/lara_look.c
+++ b/src/game/lara/lara_look.c
@@ -63,7 +63,7 @@ void Lara_LookLeftRightSurf(void)
 
 void Lara_LookUpDown(void)
 {
-    Lara_LookUpDownBase(MIN_HEAD_TILT_LOOK, MAX_HEAD_TILT_LOOK, HEAD_TURN);
+    Lara_LookUpDownBase(MIN_HEAD_TILT_LOOK, MAX_HEAD_TILT_LOOK, HEAD_TURN / 2);
 }
 
 void Lara_LookUpDownSurf(void)
--- a/src/game/lara/lara_look.c
+++ b/src/game/lara/lara_look.c
@@ -37,12 +37,12 @@ static void Lara_LookUpDownBase(
     if (g_Input.forward) {
         g_Input.forward = 0;
         if (g_Lara.head_x_rot > min_head_tilt) {
-            g_Lara.head_x_rot -= head_turn / 2;
+            g_Lara.head_x_rot -= head_turn;
         }
     } else if (g_Input.back) {
         g_Input.back = 0;
         if (g_Lara.head_x_rot < max_head_tilt) {
-            g_Lara.head_x_rot += head_turn / 2;
+            g_Lara.head_x_rot += head_turn;
         }
     }
     if (g_Lara.gun_status != LGS_HANDS_BUSY) {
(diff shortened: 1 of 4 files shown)


src/game/lara/lara_state.c
```diff
@@ -934,20 +922,9 @@ void Lara_State_SurfTread(ITEM_INFO *item, COLL_INFO *coll)
...
 
     if (g_Input.look) {
-        g_Camera.type = CAM_LOOK;
-        if (g_Input.left && g_Lara.head_y_rot > -MAX_HEAD_ROTATION_SURF) {
-            g_Lara.head_y_rot -= HEAD_TURN_SURF;
-        } else if (
-            g_Input.right && g_Lara.head_y_rot < MAX_HEAD_ROTATION_SURF) {
-            g_Lara.head_y_rot += HEAD_TURN_SURF;
-        }
+        Lara_LookLeftRight(MAX_HEAD_ROTATION_SURF, HEAD_TURN_SURF);
```
walkawayy: Ah yes I see `g_Lara.head_y_rot -= HEAD_TURN_SURF;` vs `g_Lara.head_y_rot -= HEAD_TURN / 2;`. I'll fix.

(BTW, some of the state names are confusing / might be wrong? `Lara_State_SurfTread` is when she treads water at the surface. `Lara_State_Tread` is when she treads in place underwater. Ok seems like surf means tread. But I think adding `UW` to all the underwater stuff would help clarify. `Lara_State_Tread` to me means treading on the surface and `Lara_State_SurfTread` would mean underwater treading. But it's backwards from that. Wait does surf mean surface? ⚰️I was thinking it meant surf like riding a wave or something lmao.)

walkawayy:
I added separate look functions for when Lara treads on the surface.

rr- (commented):
To avoid using `MAX_HEAD_ROTATION` / `HEAD_TURN` everywhere and reduce introduced code duplication, could we name `Lara_LookLeftRight` `Lara_LookLeftRightBase` and have two functions, `Lara_LookLeftRightSurface` and `Lara_LookLeftRightUW` that pass proper constants to `Lara_LookLeftRightBase`? The division by 2 should happen in the wrapper functions and not in the base function.

(and the same for `Lara_LookUpDown()`.)

walkawayy:
Lmk what you think of the latest push. I added `Lara_LookLeftRightBase` and `Lara_LookUpDownBase`. I only added `Lara_LookLeftRightSurface` and `Lara_LookUpDownSurface` as special wrappers because Lara treading on the surface is the only unique look case as far as I can tell. Lara out of water and Lara under the water share the same look calls, so I didn't make specialized `UW` calls.

rr- (commented):
You left the division by 2 in the base function for the head turn – it should be moved to `LeftRight(void)`, otherwise Lara moves her head too slowly on the surface.

Upon seeing it I think naming the function `Surf` and not `Surface` would be better for now, not to introduce new nomenclature in the code.

walkawayy:
Ah sorry about that. I should not commit when tired. I think it should be good now? Sorry for the confusion.

src/game/lara/lara_look.c
```diff
@@ -5,44 +5,73 @@
...
+
+static void Lara_LookLeftRightBase(int16_t max_head_rot, int16_t head_turn)
 {
     g_Camera.type = CAM_LOOK;
     if (g_Input.left) {
         g_Input.left = 0;
-        if (g_Lara.head_y_rot > -MAX_HEAD_ROTATION) {
-            g_Lara.head_y_rot -= HEAD_TURN / 2;
+        if (g_Lara.head_y_rot > -max_head_rot) {
+            g_Lara.head_y_rot -= head_turn / 2;
```
rr-: Please remove `/ 2` here.

rr- (changes requested):

rr- (approved):
LGTM. Final question – does Lara turn her torso around when she's in the STOP state and has her pistols equipped? Did she use to prior to this patch?

(When merging, please squash the commits.)

walkawayy:
Seems ok to me?

TombATI:
image

T1M Look Refactor:
20230430_205738_Caves

rr-:
Let's merge

PR #824, opened 2023-05-06
input: add menu specific controls

### LINKED ISSUE #814: Feature request: bring back Enter to main menu bindings
Not being able to select stuff in the main menu with Enter is really confusing and disruptive. I also had reports of players relying on having the ability to use arrows AND WSAD bound at the same time.

One way out of this that I see is to maybe turning https://github.com/LostArtefacts/Tomb1Main/issues/564 to be optional?
Another would be to allow multiple bindings but that's way more effort.

walkawayy:
Resolves #814.

#### Checklist

#### Description
Added menu specific controls. Arrow keys, enter, and escape for keyboard. Dpad, action, and option buttons for controller.


walkawayy:
Opening as draft for now. Removed select and deselect in favor of menu_confirm and menu_back.

Current issues:
- [x] Up, down, left, and right flash conflicts due to menu up, menu down, menu left, and menu right.
  - Solved by removing `INPUT_ROLE_MENU_`.
- [ ] Mapping the Run role to the down arrow key breaks menu navigation. Both down and up arrow keys navigate up.
  - [x] Won't fix for now.
- [ ] Mapping the Left role to the right arrow key and the Right role to the left arrow key breaks menu navigation. Both left and right arrow keys navigate right.
  - [x] Won't fix for now.
- [ ] Allow holding up/down to scroll the controls faster (similar to the faster passport scrolling)? Could mess with things.

Test scenarios:
- [x] Keyboard.
- [x] Controller.
- [x] Cause conflicts between game and menu controls like mapping the down role to the up arrow key.
- [ ] Testers?

Future:
1. Do we need a custom bind confirmation prompt since we have fallback menu controls now?
2. Regarding people expecting the WASD and arrow keys to work for Lara, I don't know of any other games that allow this. And it would make a giant mess. I guess since it's legacy like you said, maybe we need to make a config option for this. Seems messy.

rr-:
>Up, down, left, and right flash conflicts due to menu up, menu down, menu left, and menu right.

I think that my suggestion to add `INPUT_MENU_LEFT` etc was too eager and changing just `INPUT_STATE` to have the new members would suffice.

>Do we need a custom bind confirmation prompt since we have fallback menu controls now?

I don't understand the question

>Regarding people expecting the WASD and arrow keys to work for Lara, I don't know of any other games that allow this. And it would make a giant mess. I guess since it's legacy like you said, maybe we need to make a config option for this. Seems messy.

Yes that's what I suggested in the linked issue, to have a config option whether to enable or disable falling back to stock keys, because first people complained it still defaults, now they complain that it doesn't default anymore, and both parties seem to be equally annoyed.

walkawayy:
> > Do we need a custom bind confirmation prompt since we have fallback menu controls now?
> 
> I don't understand the question
> 

That question was regarding this complaint you posted about in the issue:
> The complaint that rebinding arrows to WSAD stops menu from working with arrow keys as soon as the user rebinds
> the first key, without even waiting for the user to exit the menu or anything.
> I'd hardcode the arrows to always work within menus (keyboard only - controller should be fine as-is).
> At the same time I'd make the assignment to happen only after the user confirms their keybindings, not immediately.

rr-:
I still don't understand what kind of confirmation you have in mind – a popup with "Apply changes Y|N"? I only meant exiting the controls dialog.

walkawayy:
> I still don't understand what kind of confirmation you have in mind – a popup with "Apply changes Y|N"? I only meant exiting the controls dialog.

If we wanted to do this, we'd need to rework a decent chunk of code to add this I think. We'd need to add some kind of pending keybinding structure that is changed in the controls menu instead of `m_Layout` or `m_ControllerLayout`. Then when the user backs out of that menu, add a confirm/deny changes popup. If they confirm, set the current layout's controls based on the pending controls. If we want to do this, I think it's prob best for a future PR.

rr-:
Yes such changes are definitely for the future and I don't think the value to effort ratio is looking too good here.

walkawayy:
> > Up, down, left, and right flash conflicts due to menu up, menu down, menu left, and menu right.
> 
> I think that my suggestion to add `INPUT_MENU_LEFT` etc was too eager and changing just `INPUT_STATE` to have the new members would suffice.

How would I go about changing this if we got rid of `INPUT_ROLE_MENU_UP`? 
`linput.menu_up = S_Input_Key(INPUT_ROLE_MENU_UP, layout_num);`

Bring back the hard coded checks like this? I guess that is what we are basically doing with `INPUT_ROLE_MENU_` hard coded, so it's not really a big change.
`linput.menu_up = KEY_DOWN(SDL_SCANCODE_UP);`

rr-:
Yes, that's OK, so long as SDL stuff is confined to `s_input.c`.

walkawayy:
It requires the user to do do something not smart, but mapping the Run role to the down arrow key breaks menu navigation. Both down and up arrow keys navigate up. Mapping the Left role to the right arrow key and the Right role to the left arrow key breaks menu navigation. Both left and right arrow keys navigate right.

The user can also break their Action role by mapping Action to the escape key while the escape key is still mapped to Inventory. Pressing escape will just go back and forth between the Customize Controls keyboard or controller menu and the Custom Scheme remapping screen.

rr-:
Frankly I'd let that happen and handle complaints once we get them. WDYT?

walkawayy:
> Frankly I'd let that happen and handle complaints once we get them. WDYT?

I agree. We have the hold to reset controls button if needed too. Worst case, the config file can be deleted.

src/specific/s_input.c
```diff
@@ -666,7 +666,7 @@ static const char *S_Input_GetButtonName(SDL_GameControllerButton button)
                 case SDL_CONTROLLER_BUTTON_DPAD_DOWN:     return "\202 ";
                 case SDL_CONTROLLER_BUTTON_DPAD_LEFT:     return "\200 ";
                 case SDL_CONTROLLER_BUTTON_DPAD_RIGHT:    return "\201 ";
-                case SDL_CONTROLLER_BUTTON_MISC1:         return "SHARE"; /* Xbox Series X share button, PS5 microphone button, Nintendo Switch Pro capture button, Amazon Luna microphone button */
+                case SDL_CONTROLLER_BUTTON_MISC1:         return "SHARE";    /* Xbox Series X share button */
```
rr-, line 669: Why the reduced comments? Did the linter say mean things about the line length? Where is he, I'll have a word with him…

src/specific/s_input.c
```diff
@@ -858,6 +856,18 @@ static INPUT_STATE S_Input_GetControllerState(
...
+        state.menu_down              |= S_Input_JoyBtn(SDL_CONTROLLER_BUTTON_DPAD_DOWN);
+        state.menu_left              |= S_Input_JoyBtn(SDL_CONTROLLER_BUTTON_DPAD_LEFT);
+        state.menu_right             |= S_Input_JoyBtn(SDL_CONTROLLER_BUTTON_DPAD_RIGHT);
+        if (m_ControllerType == SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_PRO || m_ControllerType == SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_JOYCON_PAIR) {
+            state.menu_confirm           |= S_Input_JoyBtn(SDL_CONTROLLER_BUTTON_B);
+            state.menu_back              |= S_Input_JoyBtn(SDL_CONTROLLER_BUTTON_A);
+        } else {
+            state.menu_confirm           |= S_Input_JoyBtn(SDL_CONTROLLER_BUTTON_A);
+            state.menu_back              |= S_Input_JoyBtn(SDL_CONTROLLER_BUTTON_B);
+        }
```
rr-: To be honest I'd forgo this condition in favor of the old code:

```c
        state.menu_confirm            |= S_Input_GetBindState(INPUT_ROLE_ACTION, cntlr_layout_num);
        state.menu_back               |= S_Input_GetBindState(INPUT_ROLE_OPTION, cntlr_layout_num);
```

or even remove all `menu` assignments altogether from the controller code, as it gets augmented in `Input_GetState` anyway.

Asking for @lahm86 's opinion on this.

src/specific/s_input.c
walkawayy: It's in a `// clang-format off` block so no linter complaints. But, the line was really long, so I trimmed each console platform's comment to be relevant to that block. I left the full comment in the `default:` case. The comment itself is taken directly from SDL's codebase.

src/specific/s_input.c
lahm86: The old style seems easier to read - I assume it achieves the same result?

walkawayy: Ya I can revert this for controller since we don't really need it like we do for keyboard.

walkawayy: Oh btw the reason this condition was there was because the switch uses different confirm buttons compared to Xbox / PS. But best to be consistent.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - added an inverted look camera option (#700)
 - added a camera speed option for the manual camera (#815)
 - added an option to fix original texture issues (#826)
+- added menu specific controls (#814)
```
rr-: Let's be more visible about this and mention that this brings arrows and enter support back. Ideally mention the version that took this away.

src/specific/s_input.c
rr-: OK I see, makes sense.

src/specific/s_input.c
```diff
@@ -858,6 +856,7 @@ static INPUT_STATE S_Input_GetControllerState(
         state.load                   |= S_Input_GetBindState(INPUT_ROLE_LOAD, cntlr_layout_num);
         state.toggle_fps_counter     |= S_Input_GetBindState(INPUT_ROLE_FPS, cntlr_layout_num);
         state.toggle_bilinear_filter |= S_Input_GetBindState(INPUT_ROLE_BILINEAR, cntlr_layout_num);
+
```
rr-: I'd say this line is pointless

rr- (approved):

rr- (approved):

PR #831, opened 2023-05-11
data: fix room mesh oversights

### LINKED ISSUE #826: OG bug: room mesh oversights
This follows from discussion #736.

Implement injection functionality to fix each of the following room mesh issues.

1. Incorrect texture on a face
2. Misaligned vertices on a face leading to gaps
3. Missing face
4. Wrongly rotated face

See https://github.com/LostArtefacts/Tomb1Main/discussions/736#discussioncomment-5853608 for the complete list of issues identified. This should be used to generate the necessary injection files.

lahm86:
Resolves #826.

#### Checklist

#### Description

This resolves the texture issues identified in #736 - many thanks to @Richard-L, @aredfan, @Rapora9 and @Topixtor for their input and work in finding these.

The config option for texture fixes will also apply to the previously implemented texture injections - so Larson's gun, the Colosseum roof, and the wrong palette colours on Cowboy, SkaterBoy and Baldy. I think this is easier than having separate options for each aspect - so it's an all-or-nothing approach to fix textures essentially, whether room or model related.

Although there are four types of room issue we are targeting, `ROOM_MESH_EDIT_TYPE` has six types because in some cases we need a combination of edits to achieve the right result. Two of these types incur a change in the room mesh size - `RMET_ADD_FACE` and `RMET_ADD_VERTEX` - hence the change in `level.c` to ensure the correct amount of space is allocated pre-injection. The framework deals with the complexity of shifting the mesh data when we are adding to it.


README.md
```diff
@@ -330,6 +330,7 @@ Not all options are turned on by default. Refer to `Tomb1Main_ConfigTool.exe` fo
 - fixed incorrect ceiling textures in Colosseum 
 - fixed inconsistencies in some enemy textures
 - fixed the animation of Lara's left arm when the shotgun is equipped
+- fixed room texture issues throughout the game, including Unfinished Business
```
rr-: Can we list the individual fixes?

rr- (approved):

Richard-L:
So excellent Lahm 👏

README.md
lahm86: Good point, I've updated it [here](https://github.com/LostArtefacts/Tomb1Main/pull/831/commits/d09bd005a3e0a04cbf57c17c39151d46c4c84e69).

rr-: I mean unless there's really a lot of them, I had in mind listing every change like so:

- Lost Valley: fix black triangle in the tunnel between the lake and the caves
- Colosseum: fix Midas textures appearing at the roof

or, (this can be automated), like so:

- Lost Valley: fix missing texture in room 58
- Colosseum: fix wrong texture in rooms 12, 57, 94 and 95

lahm86: Ah, I see, sorry. I'll put a list together, there are several so I think the second approach will be best.

lahm86: A much more detailed list [here](https://github.com/LostArtefacts/Tomb1Main/pull/831/commits/29eed02c4979696821b65e6ecb7ebb2f0e13238d). I summarised it down to two types - `incorrect` to cover actual wrong textures alongside wrong rotations, and `missing` to cover actual missing faces and poorly aligned vertices that result in gaps. Then there is just the standalone Colosseum roof fix, which I've merged in manually. Let me know if this looks OK.

rr- (approved):
This is perfect, and also leaves room for future changelog entries should we fiddle with more textures.

walkawayy (approved):
Awesome LGTM.

As a side comment, do we need to force the 2.15 installer to overwrite the user's existing files? They will need the new .bin files and gameflows to enable the new texture fix options (same for any other injection fixes, but I know we didn't in previous releases). I'm just trying to prevent any issues for people in the future where the solution will be to reinstall. Might be an @rr- question.

rr-:
> Awesome LGTM.
> 
> As a side comment, do we need to force the 2.15 installer to overwrite the user's existing files? They will need the new .bin files and gameflows to enable the new texture fix options (same for any other injection fixes, but I know we didn't in previous releases). I'm just trying to prevent any issues for people in the future where the solution will be to reinstall. Might be an @rr- question.

Yes for the new features to work. Not overwriting will not cause issues, it will just not apply any fixes and the user will have to update the .bin and the gameflow files by hand for these to work. 

Also, choosing to overwrite the existing files shouldn't reset the user settings ever since 889065e4, so the label:

>                            Overwrites any files such as level data, music files and existing user settings.

is not entirely correct.

PR #832, opened 2023-05-11
room: fix ceiling height miscalculation

### LINKED ISSUE #323: OG Bug: Lara clips inside the ceiling
In certain places Lara can clip through the visible ceiling as if the actual ceiling was higher. For this to happen, it seems the floor below must be uneven. One such place in original TR1 is in the Caves level, near the exit door of the pit with a bear (image below).

location

clipping

I don't know of any other locations in the original TR1, but this bug is also present for example in a custom level I'm working on.

lahm86:
Resolves #323.

#### Checklist

#### Description

This prevents the ceiling height calculation reading unrelated floor data when there exists a floor tilt. I have tested the following scenarios and the height corresponds to the geometry in each case.
- Tilted floor + flat ceiling
- Flat floor + flat ceiling
- Flat floor + tilted ceiling

### DIFF
--- a/src/game/room.c
+++ b/src/game/room.c
@@ -306,6 +306,7 @@ int16_t Room_GetCeiling(FLOOR_INFO *floor, int32_t x, int32_t y, int32_t z)
 {
     int16_t *data;
     int16_t type;
+    int16_t ended;
     int16_t trigger;
 
     FLOOR_INFO *f = floor;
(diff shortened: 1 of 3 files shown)


walkawayy (approved):

src/game/room.c
```diff
@@ -320,9 +320,9 @@ int16_t Room_GetCeiling(FLOOR_INFO *floor, int32_t x, int32_t y, int32_t z)
 
     if (f->index) {
         data = &g_FloorData[f->index];
-        type = *data++ & DATA_TYPE;
+        type = *data & DATA_TYPE;
 
-        if (type == FT_TILT) {
+        if (!(*data++ & END_BIT) && type == FT_TILT) {
```
rr-: Let's try to avoid pointer manipulation inside conditions and extract this to a variable.
BTW, if I understand the `END_BIT` purpose correctly, shouldn't this `END_BIT` check also apply to `FT_ROOF`?

lahm86: We don't need to worry about the roof as it doesn't look for any more entry types after that. End bit means no more functions for this sector, but there can still be value data. So it reads the slant values and does what it needs to with those. There is a loop further down which checks for objects changing the height, trapdoors and suchlike, but that starts again at the beginning of the sector's FD and the loop condition explicitly checks for the end bit.

rr- (approved):

PR #837, opened 2023-05-15
Rework game resolution management (part 1)

### LINKED ISSUE #114: Suggestion: Change Renderer Behaviour

screenshot0003

rr-:
#### Checklist

#### Description

Functional:
- removed `hi_res` option. Instead the resolution is now saved in `resolution_width/height` keys
- resolution changes are applied immediately.
- changed the desktop/user-supplied resolution to be positioned first, not last. If the user changes `resolution_width/height` to something weird, these values will override the first resolution and it will be selected.
- fixed fade to black animation for pictures ending on a non-black frame (so when Eidos logo ends fading and the main menu is about to load, we still very dark Eidos logo instead of black screen in the current release; this is fixed by this 3673ae5)

Cosmetic:
- dropped external index management in favor of direct width/height comparisons
- removed a couple of pointless functions such as `Output_InitialisePolyList` and `Output_ClearScreen`
- renamed `Output_DisplayPicture` and similar to use new term, `BackdropImage`. Changed `display` to `load` as it doesn't actually display anything.

This doesn't fix #114 but it's a start.
This is best viewed on a commit-by-commit basis.


walkawayy (approved):
Changes look nice! The instant resolution change is much better too.

I tested different resolutions and everything worked for me including screenshots.

rr-:
I think the changelog could use improvements but IDK how to include the entire scope of the changes without being too wordy about it.

lahm86 (approved):
LGTM

PR #842, opened 2023-05-18
shell: fix skipping Eidos logo

### LINKED ISSUE #541: Improve fade skipping
Very nitpicky thing, but something you do notice after prolonged testing:

Currently the fades cannot be skipped via ESC or the action button. Take for example the initial EIDOS logo: you can only skip after fade-in, or before fade-out, but not during. It would be really nice if the skip behaviour were extended to all these three states: fade-in, image showing, fade-out.

I wonder where else this has an effect. I reckon during the end credits.

rr-:
#### Checklist

#### Description

Resolves #541

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - changed screen resolution option to apply immediately (#114)
 - changed shaders to use GLSL 1.20 which should fix most issues with OpenGL 2.1 (#327, #685)
 - fixed sounds stopping instead of pausing if game sounds in inventory are disabled (#717)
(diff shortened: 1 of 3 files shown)


walkawayy (approved):
Wouldn't this allow better skipping of other pictures too? Such as credits and maybe loading screens in the future?

rr-:
Correct

walkawayy:
> Correct

So maybe the changelog should reflect that? Either way, I tested and it works great.

rr-:
Expanded the changelog to also talk about end credits. Merging

PR #843, opened 2023-05-19
docs: add braid showcase to readme

lahm86:
#### Checklist

#### Description

I'm not sure if something animated would be better here but I feel this image gives a sense of motion anyway. You can let me know what you think.


walkawayy (commented):
I'll wait for rr to approve or not (he might want an animated version like the draw distance), but I'd suggest moving up the braid to earlier in the showcase. I think more people are interested in it than most features. Maybe one of the first two rows?

rr-:
Yeah I'd say this is the most impactful change and as such it deserves even number one place. As for the sense of motion, it can stay as is, and doesn't have to be animated.

lahm86:
It's in top position now, and everything else shifted along by one:
https://github.com/lahm86/Tomb1Main/tree/braid-showcase#showcase

walkawayy (approved):

rr- (approved):

rr- (approved):

rr- (approved):

PR #846, opened 2023-05-20
filesystem: add case insensitive directory and file detection

### LINKED ISSUE #845: Case insensitive directory and file detection
`File_Open` needs to be made case insensitive. This can be made by enumerating files in the parent directory (or directories!!) if they do not exist.

Required for #716. Linux is case sensitive and [Windows can set specific folders to be case sensitive](https://www.tenforums.com/tutorials/111293-enable-disable-case-sensitive-attribute-folders-windows-10-a.html).

walkawayy:
Resolves #845.

#### Checklist

#### Description
Added case insensitive directory and file detection. I tested this on Windows with a normal folder and a folder with the case sensitive attribute enabled. This will need to be tested on Linux.

Case sensitive paths are only checked as a fallback if the normal file loading doesn't work since it's slower. Added a custom `strsep` so it can be cross platform. Memory management will need reviewed. Also, I aliased Windows calls like so: `#define strcasecmp _stricmp` and `#define strncasecmp _strnicmp`. Not sure if that's good practice, but I wanted to make this work for Windows and Linux.
...


src/filesystem.c
```diff
@@ -12,8 +12,21 @@ struct MYFILE {
...
 };
 
+void File_RemoveSubstring(char *str, const char *sub);
 static bool File_ExistsRaw(const char *path);
 
+void File_RemoveSubstring(char *str, const char *sub)
+{
+    size_t len = strlen(sub);
+    if (len == 0)
+        return;
```
rr-: Missing braces

src/filesystem.c
```diff
@@ -12,8 +12,21 @@ struct MYFILE {
...
+
+    char *p = str;
+    while ((p = strstr(p, sub)) != NULL) {
```
rr-: Unnecessary `!= NULL`

src/filesystem.c
```diff
@@ -55,6 +68,26 @@ char *File_GetFullPath(const char *path)
             size_t out_size = strlen(game_dir) + strlen(path) + 2;
             char *out = Memory_Alloc(out_size);
             sprintf(out, "%s/%s", game_dir, path);
+            if (!File_ExistsRaw(out)) {
```
rr-: Let's swap the condition and remove `else`:
```c
if (File_ExistsRaw(out)) {
    return out;
}
Memory_FreePointer(&out);
```

src/filesystem.c
```diff
@@ -55,6 +68,26 @@ char *File_GetFullPath(const char *path)
...
+    }
+
+    // Check for case sensitive path.
+    if (File_IsRelative(path)) {
+        const char *game_dir = File_GetGameDirectory();
+        if (game_dir) {
+            char *case_path = Memory_Alloc(strlen(path) + 2);
+            if (S_File_CasePath(path, case_path)) {
+                File_RemoveSubstring(case_path, "./");
+            }
```
rr-: Can't we change `S_File_CasePath` to not use `./` instead of removing it after?

src/specific/s_filesystem.c
```diff
@@ -1,19 +1,42 @@
...
 
 #include <SDL2/SDL_filesystem.h>
 #include <assert.h>
+#include <dirent.h>
 #include <string.h>
 
 #if defined(_WIN32)
     #include <direct.h>
+    #define strcasecmp _stricmp
+    #define strncasecmp _strnicmp
```
rr-: Is this really needed? Especially given our docker toolchain which is fairly posix compliant?

src/specific/s_filesystem.c
```diff
@@ -1,19 +1,42 @@
...
 const char *m_GameDir = NULL;
 
+static char *S_File_MyStrsep(char **stringp, const char *delim);
+
+static char *S_File_MyStrsep(char **stringp, const char *delim)
+{
+    char *start = *stringp;
+    char *p;
+
+    p = (start != NULL) ? strpbrk(start, delim) : NULL;
```
rr-: Can be contracted with the declaration; unnecessary `!= NULL`

src/specific/s_filesystem.c
```diff
@@ -1,19 +1,42 @@
...
+
+    if (p == NULL) {
+        *stringp = NULL;
+    } else {
+        *p = '\0';
+        *stringp = p + 1;
+    }
```
rr-:
```c
if (p) {
    *p = '\0';
    *stringp = p + 1;
} else {
    *stringp = NULL;
}
```

src/specific/s_filesystem.c
```diff
@@ -35,3 +58,71 @@ void S_File_CreateDirectory(const char *path)
...
+    if (p[0] == '/') {
+        d = opendir("/");
+        p = p + 1;
+    } else {
+        d = opendir(".");
+        r[0] = '.';
+        r[1] = 0;
+        rl = 1;
+    }
+    LOG_DEBUG("r: %s", r);
```
rr-: Unnecessary debug calls, the one in `File_Open` can stay but these should go away in the final version

src/specific/s_filesystem.c
```diff
@@ -35,3 +58,71 @@ void S_File_CreateDirectory(const char *path)
...
+    strcpy(p, path);
+    size_t rl = 0;
+
+    DIR *d;
+    if (p[0] == '/') {
+        d = opendir("/");
+        p = p + 1;
+    } else {
+        d = opendir(".");
+        r[0] = '.';
```
rr-: This is where we make the path start with `./`

rr- (commented):
I'll test it when I get back home around Tuesday, for now we can focus on cosmetic stuff.

src/filesystem.c
```diff
@@ -12,8 +12,21 @@ struct MYFILE {
     const char *path;
 };
 
+void File_RemoveSubstring(char *str, const char *sub);
```
lahm86: Should this be static?

src/specific/s_filesystem.c
walkawayy: This one I'm not 100% sure of. If I remove it, it still builds and runs for me. So, I guess the answer is yes.

src/specific/s_filesystem.c
walkawayy: My bad. Testing relics.

src/specific/s_filesystem.c
walkawayy: I'll have to do some more testing. If I remove that, it bugs the directory traversal process.

walkawayy: I updated all the names because the `c`, `d`, `p`, etc were not clear at all.

The problem seems to be removing `./` causes `case_path[case_len] = '/';` to need to be moved near `path_piece = S_File_Strsep(&path_copy, "/");`, Either way, this breaks `path_dir = opendir(case_path);`. If you have any ideas LMK.

rr-: I meant this line is only a starting point. I'll play with it back home.

rr-:
I tested it and it didn't trigger properly in case the game was run from a working directory other than the game's directory.

I addressed this and simplified the code a bit so that the loop is easier on the eyes:

```diff
diff --git a/src/filesystem.c b/src/filesystem.c
index ae1a8580..cf4001cb 100644
--- a/src/filesystem.c
+++ b/src/filesystem.c
@@ -4,6 +4,7 @@
 #include "memory.h"
 #include "specific/s_filesystem.h"
 
+#include <assert.h>
 #include <stdio.h>
 #include <string.h>
 
@@ -12,22 +13,8 @@ struct MYFILE {
     const char *path;
 };
 
-static void File_RemoveSubstring(char *str, const char *sub);
 static bool File_ExistsRaw(const char *path);
 
-static void File_RemoveSubstring(char *str, const char *sub)
-{
-    size_t len = strlen(sub);
-    if (len == 0) {
-        return;
-    }
-
-    char *p = str;
-    while ((p = strstr(p, sub))) {
-        memmove(p, p + len, strlen(p + len) + 1);
-    }
-}
-
 static bool File_ExistsRaw(const char *path)
 {
     FILE *fp = fopen(path, "rb");
@@ -63,35 +50,25 @@ bool File_Exists(const char *path)
 
 char *File_GetFullPath(const char *path)
 {
-    if (!File_ExistsRaw(path) && File_IsRelative(path)) {
+    char *full_path = NULL;
+    if (File_IsRelative(path)) {
         const char *game_dir = File_GetGameDirectory();
         if (game_dir) {
-            size_t out_size = strlen(game_dir) + strlen(path) + 2;
-            char *out = Memory_Alloc(out_size);
-            sprintf(out, "%s/%s", game_dir, path);
-            if (File_ExistsRaw(out)) {
-                return out;
-            }
-            Memory_FreePointer(&out);
+            full_path = Memory_Alloc(strlen(game_dir) + strlen(path));
+            sprintf(full_path, "%s%s", game_dir, path);
         }
     }
+    if (!full_path) {
+        full_path = Memory_DupStr(path);
+    }
 
-    // Check for case sensitive path.
-    if (File_IsRelative(path)) {
-        const char *game_dir = File_GetGameDirectory();
-        if (game_dir) {
-            char *case_path = Memory_Alloc(strlen(path) + 2);
-            if (S_File_CasePath(path, case_path)) {
-                File_RemoveSubstring(case_path, "./");
-            }
-            size_t out_size = strlen(game_dir) + strlen(case_path) + 2;
-            char *out = Memory_Alloc(out_size);
-            sprintf(out, "%s%s", game_dir, case_path);
-            Memory_FreePointer(&case_path);
-            return out;
-        }
+    char *case_path = S_File_CasePath(full_path);
+    if (case_path) {
+        Memory_FreePointer(&full_path);
+        return case_path;
     }
-    return Memory_DupStr(path);
+
+    return full_path;
 }
 
 char *File_GuessExtension(const char *path, const char **extensions)
diff --git a/src/specific/s_filesystem.c b/src/specific/s_filesystem.c
index 875c8c6f..510c0e64 100644
--- a/src/specific/s_filesystem.c
+++ b/src/specific/s_filesystem.c
@@ -16,23 +16,6 @@
 
 const char *m_GameDir = NULL;
 
-static char *S_File_Strsep(char **stringp, const char *delim);
-
-static char *S_File_Strsep(char **stringp, const char *delim)
-{
-    char *start = *stringp;
-    char *p = start ? strpbrk(start, delim) : NULL;
-
-    if (p) {
-        *p = '\0';
-        *stringp = p + 1;
-    } else {
-        *stringp = NULL;
-    }
-
-    return start;
-}
-
 const char *S_File_GetGameDirectory(void)
 {
     if (!m_GameDir) {
@@ -55,67 +38,76 @@ void S_File_CreateDirectory(const char *path)
 #endif
 }
 
-bool S_File_CasePath(char const *path, char *case_path)
+char *S_File_CasePath(char const *path)
 {
-    size_t length = strlen(path);
-    char *path_copy = Memory_Alloc(length + 1);
-    strcpy(path_copy, path);
-    size_t case_len = 0;
+    assert(path);
 
-    DIR *path_dir;
+    char *current_path = Memory_Alloc(strlen(path) + 2);
+    const char *sep = strchr(path, '/') ? "/" : "\\";
+
+    char *path_copy = Memory_DupStr(path);
+    char *path_piece = path_copy;
+
+    bool relative;
     if (path_copy[0] == '/') {
-        path_dir = opendir("/");
-        path_copy = path_copy + 1;
+        relative = false;
+        strcpy(current_path, "/");
+    } else if (strstr(path_copy, ":\\")) {
+        relative = false;
+        strcpy(current_path, path_copy);
+        strstr(current_path, ":\\")[1] = '\0';
+        path_piece += 3;
     } else {
-        path_dir = opendir(".");
-        case_path[0] = '.';
-        case_path[1] = 0;
-        case_len = 1;
+        relative = true;
+        strcpy(current_path, ".");
     }
 
-    bool last_file = false;
-    char *path_piece = S_File_Strsep(&path_copy, "/");
     while (path_piece) {
-        if (!path_dir) {
-            Memory_FreePointer(&path_copy);
-            return false;
+        char *delim = strpbrk(path_piece, "/\\");
+        char old_delim = delim ? *delim : '\0';
+        if (delim) {
+            *delim = '\0';
         }
 
-        if (last_file) {
-            closedir(path_dir);
+        DIR *path_dir = opendir(current_path);
+        if (!path_dir) {
             Memory_FreePointer(&path_copy);
-            return false;
+            Memory_FreePointer(&current_path);
+            return NULL;
         }
 
-        case_path[case_len] = '/';
-        case_len += 1;
-        case_path[case_len] = 0;
-
         struct dirent *cur_file = readdir(path_dir);
         while (cur_file) {
             if (strcasecmp(path_piece, cur_file->d_name) == 0) {
-                strcpy(case_path + case_len, cur_file->d_name);
-                case_len += strlen(cur_file->d_name);
-                closedir(path_dir);
-                path_dir = opendir(case_path);
+                strcat(current_path, sep);
+                strcat(current_path, cur_file->d_name);
                 break;
             }
             cur_file = readdir(path_dir);
         }
+        closedir(path_dir);
 
         if (!cur_file) {
-            strcpy(case_path + case_len, path_piece);
-            case_len += strlen(path_piece);
-            last_file = true;
+            strcat(current_path, sep);
+            strcat(current_path, path_piece);
         }
 
-        path_piece = S_File_Strsep(&path_copy, "/");
-    }
-
-    if (path_dir) {
-        closedir(path_dir);
+        if (delim) {
+            *delim = old_delim;
+            path_piece = delim + 1;
+        } else {
+            break;
+        }
     }
 
+finish:
     Memory_FreePointer(&path_copy);
-    return true;
+
+    char *result;
+    if (current_path[0] == '.') { /* strip leading ./ */
+        result = Memory_DupStr(current_path + 2);
+    } else {
+        result = Memory_DupStr(current_path);
+    }
+    return result;
 }
diff --git a/src/specific/s_filesystem.h b/src/specific/s_filesystem.h
index cbf04db2..114960ee 100644
--- a/src/specific/s_filesystem.h
+++ b/src/specific/s_filesystem.h
@@ -4,4 +4,4 @@
 
 const char *S_File_GetGameDirectory(void);
 void S_File_CreateDirectory(const char *path);
-bool S_File_CasePath(char const *path, char *case_path);
+char *S_File_CasePath(char const *path);
```

src/specific/s_filesystem.c
```diff
@@ -35,3 +37,78 @@ void S_File_CreateDirectory(const char *path)
...
+
+        if (delim) {
+            *delim = old_delim;
+            path_piece = delim + 1;
+        } else {
+            break;
+        }
+    }
+
+finish:
```
walkawayy: Is this needed? I don't see any `goto`s.

walkawayy:
Works well for me in the case sensitive directory.

However, in my normal directory, the game now crashes instantly with a black screen first. One time I got a popup that the game could not allocate enough memory.

EDIT: Think it's fixed now with `+1`: `full_path = Memory_Alloc(strlen(game_dir) + strlen(path) + 1);`

src/specific/s_filesystem.c
rr-: It's not needed

src/specific/s_filesystem.h
```diff
@@ -1,4 +1,7 @@
 #pragma once
 
+#include <stdbool.h>
```
rr-: This isn't needed now

rr- (commented):
Does it work now?

lahm86:
Working great for me on both sensitive and insensitive folder tests.

src/specific/s_filesystem.h
walkawayy: Oh yes. I also added `#pragma once` since it was missing for `s_filesystem.h`.

src/specific/s_filesystem.c
```diff
@@ -35,3 +37,78 @@ void S_File_CreateDirectory(const char *path)
...
+{
+    assert(path);
+
+    char *current_path = Memory_Alloc(strlen(path) + 2);
+    const char *sep = strchr(path, '/') ? "/" : "\\";
+
+    char *path_copy = Memory_DupStr(path);
+    char *path_piece = path_copy;
+
+    bool relative;
```
walkawayy: Looks like this is never actually needed either. It just gets set in the if conditions below based on the type of path, but it's never used beyond that.

rr- (approved):

PR #864, opened 2023-06-02
lara: add responsive jumping

### LINKED ISSUE #157: Feature request: TR2+ Jump Mechanics
Implement a config option that allows for TR2+ style running jumps i.e. you have freedom to jump at any point during a run once you start running.

lahm86:
Resolves #157.

#### Checklist

#### Description

The new injection file here will amend the relevant anim range low/high frames [as detailed in the issue](https://github.com/LostArtefacts/Tomb1Main/issues/157#issuecomment-988231288). It wasn't possible to combine this with the `lara_animations.bin` file as that is always injected, whereas this one needs to be tied to the config option.

Demo: https://www.youtube.com/watch?v=mqi9EDz0FYU

A bit more detail on the jump lock workings. It's effectively to allow animation 0 to behave differently the first time compared to subsequent times during continuous running jumps.

1. Lara wants to run so enters animation 6 (RUN_START)
2. The game sets a lock saying jumping is not allowed
3. Animation 6 completes and moves to animation 0 (RUN)
4. Lara wants to jump immediately
    - The jump lock is still in force, so frames 0, 1, 2, 3 play out as normal
    - Frame 4 is reached, the jump lock is lifted
    - Lara can jump from her left foot if frame < 11
    - Lara can jump from her right foot if frame >= 11
8. The jump animation completes and she (via a few other interim animations) returns to animation 0 (RUN) frame 1
9. The jump lock is lifted this time around, so she is free to jump on any frame
11. Lara stops, or grabs a ledge, or anything else
12. She wants to run again, GOTO 1


src/game/lara/lara_state.c
```diff
@@ -80,6 +81,8 @@ void Lara_State_Walk(ITEM_INFO *item, COLL_INFO *coll)
 
 void Lara_State_Run(ITEM_INFO *item, COLL_INFO *coll)
 {
+    static bool jump_permitted = true;
```
walkawayy: I think this should be declared at the top of the like `static bool m_JumpPermitted = true;`.

walkawayy (changes requested):
Just a style comment from me.

src/game/lara/lara_state.c
lahm86: Thanks, do we prefer to avoid static variables in functions? I can't think that I've seen any others now that I think about it. I don't suppose scope really matters for this.

lahm86:
I just realised that this broke the Lost Valley demo - Lara ends up not leaving the pool. I've gone for the same approach as disabling `enable_enhanced_look` temporarily while in the demo.

src/game/lara/lara_state.c
rr-: Yes we want to make the global state easily seen this way.

src/game/lara/lara_state.c
```diff
@@ -112,7 +115,21 @@ void Lara_State_Run(ITEM_INFO *item, COLL_INFO *coll)
...
+        if (anim == LA_RUN_START) {
+            jump_permitted = false;
+        } else if (anim == LA_RUN) {
+            if (item->frame_number == LF_JUMP_READY) {
+                jump_permitted = true;
+            }
+        } else {
+            jump_permitted = true;
+        }
+    }
```
rr-, line 126: I believe this boolean logic can be simplified, I'll follow up tomorrow

src/game/lara/lara_state.c
lahm86: Got it, now moved to module scope and renamed accordingly.

src/game/lara/lara_state.c
rr-: Would this work?
```c
    if (g_Config.enable_tr2_jumping) {
        int16_t anim = item->anim_number - g_Objects[item->object_number].anim_index;
        if (anim == LA_RUN_START) {
            m_JumpPermitted = false;
        } else if (anim != LA_RUN || item->frame_number == LF_JUMP_READY) {
            m_JumpPermitted = true;
        }
    }
```

lahm86: That looks much better - I'll test it out tomorrow. Thanks.

rr-: An ideal outcome would be to forgo introducing a new global variable and set it always in the local scope. But I don't think that's possible since frame number resets to 0 and cycles, so even comparing it with >= will work in an unintended way.

lahm86: Yes, it is a shame but thankfully this is a short snippet of code now so looks much tidier. I've tested it out and all is good 👍

rr- (approved):

lahm86:
Sorry, just remembered to rebase following #867 (for the changelog).

rr- (approved):

walkawayy (approved):
LGTM

PR #867, opened 2023-06-04
skatekid: check AI status before executing control

### LINKED ISSUE #866: OG bug: SkateKid can crash the game if too many enemies are active
If SkateKid is present in a level and is left alive, and the level has several other enemies and they are also left alive, eventually the game will crash. The crash does not occur in Natla's Mines as there are so few enemies, but this is currently an issue in the randomizer, and indeed could affect custom levels.

lahm86:
Resolves #866.

#### Checklist

#### Description

This ensures that SkateKid's AI status is checked before executing his control routine. Every other enemy uses this idiom, so it seems it was simply missed in this instance in OG. `CREATURE_INFO *kid = item->data;` could be `NULL`, so accessing it further down in the routine caused the crashing.


walkawayy (approved):
Nice find.

rr- (approved):

PR #872, opened 2023-06-05
missile: test damage range for overflow

### LINKED ISSUE #871: OG bug: missile range/damage overflow
If Lara is extremely far away from an exploding missile, she can still take damage from it and in some cases her health can actually increase beyond 1000. Eventually it reaches 32767, then overflows to -1 and she dies. This is reproducible most easily with Natla stuck in a certain room, where she has her standard behaviour of firing constantly, such as in this example (Natla is stuck here because of pathfinding issues in the level file, unrelated to this issue).

https://www.youtube.com/watch?v=6oZBRvoYUnA

lahm86:
Resolves #871.

#### Checklist

#### Description

Ensures Lara can only be damaged if the missile range is positive.

An aside from this, I'm wondering if a global `Lara_TakeDamage(int16_t points)` function would be a good idea, which would clamp the value in all cases - probably best under its own issue? The missile check would still be needed to test the range, but this would at least avoid overflowing her hitpoints in any case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 - fixed quick load creating an invalid save if used when no saves are present (#853)
 - fixed Lara entering body hit animations when not appropriate to do so (#857)
 - fixed SkateKid causing a game crash when too many enemies are active (#866)
+- fixed missiles damaging Lara when she is far beyond their damage range (#871)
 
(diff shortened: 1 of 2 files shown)


rr- (approved):
LGTM. I don't know if there are more n² idioms like this throughout the code, if so, it'd make sense to extract it to a function. `Lara_TakeDamage` sounds good as well.

walkawayy (commented):
LGTM.

walkawayy (approved):

PR #873, opened 2023-06-05
data: fix room mesh oversights

### LINKED ISSUE #869: OG bug: incorrect textures in Lost Valley, Atlantis and Great Pyramid
Further texture issues identified as follows:

- Lost Valley: https://github.com/LostArtefacts/Tomb1Main/discussions/736#discussioncomment-5878993
- Atlantis/Pyramid: https://github.com/LostArtefacts/Tomb1Main/discussions/736#discussioncomment-6081779

lahm86:
Resolves #869.

#### Checklist

- [ ] ~~I have added a changelog entry about what my pull request accomplishes, or it is an internal change~~

#### Description

No new changelog entry for this one as #831 covers it, but the relevant section in the README lists the additional fixes.


rr-:
I think we're overstepping our boundaries when changing the kind of things this PR addresses. WDYT?

lahm86:
It's difficult where to draw the line with these. I tend to agree with the ones here though, the room in Atlantis/Pyramid has the animated textures on the archway on the other side of the room so it does feel like it was missed. And the waterfall feels similar, although I didn't change the rotation of the green marked item, that seems fine to me as-is.

But I don't mind if you would rather draw the line with these - perhaps for the future we should only target missing faces, if any more come to light? Those do cry out "oversight" more than any of the others I feel.

rr- (approved):
Fair, but let's proceed cautiously

walkawayy (approved):
I agree that these two cases look like oversights. But we'll never know for sure since we aren't the creators.

PR #881, opened 2023-06-10
Ape vault

walkawayy:
#### Checklist

#### Description
Fixed the ape not performing the vault animation when climbing. Done with the help of ChocolateFan. `Ape_Vault` takes some minor corrections from TR3/TR4's `CreatureVault`. Since the ape does not have a vault down animation, the function now just returns a bool.

Video:
https://streamable.com/t1kpn1


Richard-L:
When you say corrections from TR3/4 what do you mean? Are these animations all in TR1's data or do they need some from the later games?

lahm86 (approved):
Look really great, nice to see this restored.

src/game/objects/creatures/ape.c
```diff
@@ -31,7 +31,7 @@
 #define APE_RADIUS (WALL_L / 3) // = 341
 #define APE_SMARTNESS 0x7FFF
 
-typedef enum {
+typedef enum APE_ANIM {
```
rr-: Breaks consistency with other enums, also I don't see it used anywhere. Let's revert it

src/game/creature.c
```diff
@@ -640,10 +644,6 @@ bool Creature_Animate(int16_t item_num, int16_t angle, int16_t tilt)
         }
 
         item->pos.x_rot = 0;
-
-        floor = Room_GetFloor(item->pos.x, item->pos.y, item->pos.z, &room_num);
-        item->floor =
-            Room_GetHeight(floor, item->pos.x, item->pos.y, item->pos.z);
```
rr-, line 646: Can you explain this change, and why the LOT->fly `if` didn't need to be amended as well?

src/game/objects/creatures/ape.c
```diff
@@ -48,77 +48,83 @@ typedef enum {
...
+            item->pos.z = (z << WALL_SHIFT) + APE_SHIFT;
         }
     }
 
+    item->floor = y;
     item->pos.y = y;
-    item->current_anim_state = APE_VAULT;
-    Item_SwitchToAnim(item, APE_VAULT_ANIM, -1);
+
+    Item_NewRoom(item_num, room_number);
```
rr-: Shouldn't we have `if (item->room_number != room_number)`?
Can we rename `room_number` to `room_num`?

rr- (commented):
TBH I'd prefer if we moved Ape_Vault to creature.c and have it be consistent with TR3, but for our current project goals this is unnecessary.

src/game/objects/creatures/ape.c
walkawayy: Oops sorry. For some reason in VS Code, this was needed to fix the intellisense. Without it, it looks like this saying it's unnamed:
image


And the types.h enums had it but I now notice other creature files don't.
```
typedef enum GAME_STATUS {
    GMS_IN_GAME = 0,
    GMS_IN_INVENTORY = 1 << 0,
    GMS_IN_PAUSE = 1 << 1,
    GMS_IN_STATS = 1 << 2,
    GMS_IN_INVENTORY_HEALTH = 1 << 3,
} GAME_STATUS;
```

src/game/creature.c
walkawayy: This was changed in TR3+'s `CreatureVault`. Two things:
1. I moved everything to a `Creature_Vault` then ended up changing back. Because, we had to make changes compared to later TR's vaulting because the ape has no vault down animation. So `Ape_Vault` no longer matches. Also, Core has the `APE_TURN_L_FLAG` and `APE_TURN_R_FLAG` stuff in `Ape_Vault`. Now this could be extracted out in the future, but for now it matches the TR1 code more. (Btw, this flags are bugged I'm pretty sure but that's for another day/PR). And ape is the only vaulting TR1 enemy.

2. Basically, the ape has one chance to perform the vault anim which is when the ape crosses an edge (square in the game world). The original code which changed the `item->floor` after the `item->pos.y` updates was one frame behind, so the ape missed the window of opportunity to vault. They corrected this in future TR games where vaulting enemies worked properly. My guess is they ran out of time for TR1 to fix this or it was just an oversight because the ape technically went up/down blocks even though the vault animation didn't work.

src/game/objects/creatures/ape.c
walkawayy: I can rename. Um, maybe? I just left it like this to make it consistent with later TRs. The vault function in TR3+ checks `if (vault)` before `Item_NewRoom`. However, the `Ape_Vault` is slightly different since we don't want have a downward vault animation. So, the function only gets to this part at the end and returns true if the ape is performing a vault up. From what ChocolateFan said, this doesn't hurt anything.

src/game/objects/creatures/ape.c
rr-: Let's revert this for now and create a follow-up task to supply this for all enums.

src/game/objects/creatures/ape.c
rr-: I think so too, but I'd like to stay consistent with the other places that use Item_NewRoom.

rr- (approved):

PR #898, opened 2023-07-23
B114 change renderer resolution

### LINKED ISSUE #114: Suggestion: Change Renderer Behaviour

oziphantom:
#### Checklist

- [ ] I have read the [coding conventions](https://github.com/rr-/Tomb1Main/blob/master/CONTRIBUTING.md#coding-conventions)

#### Description

Again, terribly sorry about this.

I have restored the branch and fixed the screen areas not clearing when you change resolutions and the screenshot now grabs the entire FBO frame.

### DIFF
--- a/bin/shaders/fbo.fsh
+++ b/bin/shaders/fbo.fsh
@@ -1,0 +1,9 @@
+#version 120
+
+varying vec2 vertTexCoords;
+
+uniform sampler2D tex0;
+
+void main(void) {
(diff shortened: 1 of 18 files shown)


rr-:
Thanks and no worries, accidents happen. This was largely a WIP, so in order to be able to continue working on it, I may close and reopen it under my account once I'm back home next week.

rr-:
Closing this in favor of #974 which continues the work of this PR.

PR #900, opened 2023-07-24
savegame: initialise gym only when present

### LINKED ISSUE #899: T1M bug: assertion error when gym level is not present
If the gym level is removed from the gameflow and the other levels are rebased to index 0, the following error shows when the game is exited.

image

lahm86:
Resolves #899.

#### Checklist

#### Description

The error message stemmed from the following line when it was being called against the legacy "Current Position" level object. During shutdown, it had `injections.length` value of 1 despite this being 0 when the gameflow was initially loaded, so it was trying to free a null injection path.

https://github.com/LostArtefacts/Tomb1Main/blob/develop/src/game/gameflow.c#L1018

In `savegame.c`, the negative index in the initialisation just so happened to mean that the address of the length value above was being accessed, so it was changing from 0 to 1 here. This probably went unnoticed in pre-injection versions as it was most likely not changing anything that incurred memory management like we have in this situation.

This is the only place where `g_GameFlow.gym_level_num` is used as an array index.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## [Unreleased](https://github.com/rr-/Tomb1Main/compare/stable...develop) - ××××-××-××
 - fixed Natla's gun moving while she is in her semi death state (#878)
+- fixed an error message from showing on exiting the game when the gym level is not present in the gameflow (#899)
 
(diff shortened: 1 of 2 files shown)


walkawayy (approved):
Ah nice find. LGTM.

rr- (approved):
Do we need to get this released as a hotfix?

PR #903, opened 2023-07-30
bear: fix the bear pat attack so it does not miss Lara

### LINKED ISSUE #450: Bug: issues with BEAR_ATTACK1 and BEAR_ATTACK2
From now on, I will refer to BEAR_ATTACK1 as "bite" and BEAR_ATTACK2 as "pat". The bear also deals charging damage whenever it's running and touches Lara, similarly to the T-Rex and Torso.

### 1st issue: the bear stops biting.
Sometimes the bear doesn't bite anymore, but only charges against Lara. I think this is unintentional, especially since the bear never stops dealing charge damage even when it's actively biting Lara. It also makes the bear an easy fight since charging barely deals any damage compared to biting. See the 3 videos below:

Bear acting as it should: biting Lara every time it gets close to her.

https://user-images.githubusercontent.com/97730054/155711694-5d03edcb-4754-4cb3-b693-9ebecfe10576.mp4

Bear only charging. **I think this should be fixed so that the bear behaves as above.**

https://user-images.githubusercontent.com/97730054/155711769-fd8a951b-d26d-42bf-a1e6-83d435a11926.mp4

Here you can see that when the bear is actively biting, it also deals charge damage whenever it touches Lara between the bites. This indicates that the charge damage is just an additional feature and not a deliberate "charge only mode" that the bear enters.

https://user-images.githubusercontent.com/97730054/155711975-ac40d601-09ae-4c75-8c98-32f285472cfa.mp4

Also, my tests indicate that if the bear is behaving correctly, it can sometimes be "broken" by going out of reach and then back. Same applies vice versa: a broken bear can sometimes be fixed by going out of its reach. Usually the bear stands up on its rear legs when this happens.

** **

### 2nd issue: bear's pat attack is not useful.
That might sound arbitrary or something that should not be touched but hear me out. It's a compilation of problems with the pat attack and I'm not sure what exactly is causing it. But I think it's a beautiful move and with some slight changes it could work so much better that it's worth fixing. The problems might stem from a similar "stuck in a no-attack mode" than issue 1, the range of the attack, the hitbox of the attack, or any compilation of these.

First problem is that the bear simply doesn't use the pat even though it looks like it could and should. Here the bear is simply walking towards Lara, not doing anything threatening. No matter whether I stay close to the bear or go further, it doesn't attack.

https://user-images.githubusercontent.com/97730054/155714094-73ac0746-e391-43c2-952d-f9b413deabaf.mp4

Second problem is that the hitbox of the pat doesn't seem very accurate. In these 2 videos I'm standing right next to the bear and you can even see Lara get "hit" by the bear but still no damage.

https://user-images.githubusercontent.com/97730054/155714700-0d8b1799-f7df-445f-9109-6c0baccf0218.mp4

https://user-images.githubusercontent.com/97730054/155714711-c47243a1-c2aa-47e9-b31e-d39c30611854.mp4

Third problem is what  you see in the beginning of the previous video (pat_no_damage2). The range of the attack is so big that the bear is usually way too far from Lara for the pat to hit her. Here is another video of this issue:

https://user-images.githubusercontent.com/97730054/155715909-e7b08b59-9e00-41f9-88d0-5b17ed882974.mp4

The pat has a range of 2 blocks. Here it thinks it should pat Lara, even though its attack will never hit her. If I take a step back, it will stop and start walking towards Lara, never attacking her again. If I get closer, the same happens: just when I'm about to enter the actual hitbox range, the bear stops attacking. And it never tries to pat again, not even when I try to regain the ~2 block distance.

[save_tr1_04.zip](https://github.com/rr-/Tomb1Main/files/8141156/save_tr1_04.zip)

Here is a save file where the bear has just stood up behind Lara, and will start its infinite patting if you don't move.

**I think this should be fixed so that the bear actually uses pat when it's close to Lara, and that the pat hitbox is good enough to hit Lara when the bear does attack.**

walkawayy:
#### Checklist

#### Description
Fixed the bear pat attack so it does not miss Lara. Fix done by ChocolateFan.
Does not solve all the things listed in the original issue because they were user preferred behavior changes and not really bug fixes. Requesting translations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## [Unreleased](https://github.com/rr-/Tomb1Main/compare/stable...develop) - ××××-××-××
 - fixed Natla's gun moving while she is in her semi death state (#878)
 - fixed an error message from showing on exiting the game when the gym level is not present in the gameflow (#899)
+- fixed the bear pat attack so it does not miss Lara (#450)
 
 ## [2.15.2](https://github.com/rr-/Tomb1Main/compare/2.15.1...2.15.2) - 2023-07-17
 - fixed Natla not leaving her semi-death state after Lara takes her down for the first time (#892, regression from 2.15.1)
--- a/README.md
+++ b/README.md
@@ -288,6 +288,7 @@ Not all options are turned on by default. Refer to `Tomb1Main_ConfigTool.exe` fo
 - fixed ceiling heights at times being miscalculated, resulting in camera issues and Lara being able to jump into the ceiling
 - fixed the ape not performing the vault animation when climbing
 - fixed Natla's gun moving while she is in her semi death state
+- fixed the bear pat attack so it does not miss Lara
 
 #### Cheats
 - added a fly cheat
--- a/src/config.c
+++ b/src/config.c
@@ -193,6 +193,7 @@ bool Config_ReadFromJSON(const char *cfg_data)
     READ_BOOL(fix_qwop_glitch, false);
     READ_BOOL(fix_alligator_ai, true);
     READ_BOOL(change_pierre_spawn, true);
+    READ_BOOL(fix_bear_ai, true);
     READ_INTEGER(fov_value, 65);
     READ_INTEGER(resolution_width, -1);
     READ_INTEGER(resolution_height, -1);
@@ -377,6 +378,7 @@ bool Config_Write(void)
     WRITE_BOOL(fix_qwop_glitch);
     WRITE_BOOL(fix_alligator_ai);
     WRITE_BOOL(change_pierre_spawn);
+    WRITE_BOOL(fix_bear_ai);
     WRITE_INTEGER(fov_value);
     WRITE_INTEGER(resolution_width);
     WRITE_INTEGER(resolution_height);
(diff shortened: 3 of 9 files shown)


walkawayy:
@ajtudela another translation request please:
`Fixes bear pat attack so it does not miss Lara.`

rr- (approved):

ajtudela:
> @ajtudela another translation request please:

Could you explain me this fix so I can make a better translation?

walkawayy:
> > @ajtudela another translation request please:
> 
> Could you explain me this fix so I can make a better translation?

It fixes when the bear is on its hind legs and "pats" Lara. The range is bugged so it does not damage. That's the name for the attack in the code. Here's the relevant video from the issue:

https://github.com/LostArtefacts/Tomb1Main/assets/81546780/00a3ccb0-9a92-4969-aabb-b6eb8befb80a

Here's a random gif of a dog "patting" another dog.

https://64.media.tumblr.com/10591d32af065d3b671836043b7a6f01/tumblr_p8dbscYrKt1vb46leo1_400.gif

rr-:
>https://64.media.tumblr.com/10591d32af065d3b671836043b7a6f01/tumblr_p8dbscYrKt1vb46leo1_400.gif

We need to get this implemented ASAP

ajtudela:
Nice gif, hahaha

The translation would be:
```
Corrige el ataque del oso para que golpee a Lara.
```

PR #905, opened 2023-08-04
savegame: add triggered music tracks to the savegame

### LINKED ISSUE #371: Prevent music triggers from playing again after loading a save file

walkawayy:
Resolves #371.

#### Checklist

#### Description
Added the triggered music tracks to the savegame so one shot tracks don't replay on load. I'm afk this weekend. But I wanted some feedback on the naming, behavior, etc of this. This saves off `g_MusicTrackFlags` for new save format saves. The original game only did this for Pierre, Cowboy, Baldy, and Larson (which for some reason uses `MX_BALDY_SPEECH` which could be a T1M bug, but I haven't looked into it yet). The implementation means the game behaves similar to how a level plays out if the player never reloads. Each `MUSIC_TRACK_ID` can only be played once in a level unless it's marked as one shot. This could be limiting to custom levels potentially, but it's also how the OG game behaves now. And custom levels do get 60 music slots to work with. Below are some examples of how it behaves and how it depends on level files.

This one shot Colosseum trigger doesn't play again on reload:
image

But this **non** one shot St Francis Folly music trigger will always play:
image

### DIFF
--- a/src/game/savegame/savegame_bson.c
+++ b/src/game/savegame/savegame_bson.c
@@ -1297,9 +1297,7 @@ bool Savegame_BSON_LoadFromFile(MYFILE *fp, GAME_INFO *game_info)
                 json_object_get_object(root_obj, "music"))) {
             goto cleanup;
         }
-    }
 
-    if (header.version >= VERSION_3) {
         if (!SaveGame_BSON_LoadMusicTrackFlags(
                 json_object_get_array(root_obj, "music_track_flags"))) {
             goto cleanup;
--- a/src/game/savegame/savegame_bson.c
+++ b/src/game/savegame/savegame_bson.c
@@ -63,6 +63,8 @@ static bool Savegame_BSON_LoadLOT(struct json_object_s *lot_obj, LOT_INFO *lot);
 static bool Savegame_BSON_LoadLara(
     struct json_object_s *lara_obj, LARA_INFO *lara);
 static bool SaveGame_BSON_LoadCurrentMusic(struct json_object_s *music_obj);
+static bool SaveGame_BSON_LoadMusicTrackFlags(
+    struct json_array_s *music_track_arr);
 static struct json_array_s *Savegame_BSON_DumpResumeInfo(
     RESUME_INFO *game_info);
 static struct json_object_s *Savegame_BSON_DumpMisc(GAME_INFO *game_info);
@@ -76,6 +78,7 @@ static struct json_object_s *Savegame_BSON_DumpAmmo(AMMO_INFO *ammo);
 static struct json_object_s *Savegame_BSON_DumpLOT(LOT_INFO *lot);
 static struct json_object_s *Savegame_BSON_DumpLara(LARA_INFO *lara);
 static struct json_object_s *SaveGame_BSON_DumpCurrentMusic(void);
+static struct json_array_s *SaveGame_BSON_DumpMusicTrackFlags(void);
 
 static void SaveGame_BSON_SaveRaw(
     MYFILE *fp, struct json_value_s *root, int32_t version)
@@ -822,6 +825,27 @@ static bool SaveGame_BSON_LoadCurrentMusic(struct json_object_s *music_obj)
                 current_track, timestamp);
         }
     }
+    return true;
+}
+
+static bool SaveGame_BSON_LoadMusicTrackFlags(
+    struct json_array_s *music_track_arr)
+{
+    if (!music_track_arr) {
+        LOG_WARNING("Malformed save: invalid or missing music track array");
+        return true;
+    }
+
+    if ((signed)music_track_arr->length != MAX_CD_TRACKS) {
+        LOG_WARNING(
+            "Malformed save: expected %d music track flags, got %d",
+            MAX_CD_TRACKS, music_track_arr->length);
+        return true;
+    }
+
+    for (int i = 0; i < (signed)music_track_arr->length; i++) {
+        g_MusicTrackFlags[i] = json_array_get_int(music_track_arr, i, 0);
+    }
 
     return true;
 }
@@ -1146,6 +1170,15 @@ static struct json_object_s *SaveGame_BSON_DumpCurrentMusic(void)
     return current_music_obj;
 }
 
+static struct json_array_s *SaveGame_BSON_DumpMusicTrackFlags(void)
+{
+    struct json_array_s *music_track_arr = json_array_new();
+    for (int i = 0; i < MAX_CD_TRACKS; i++) {
+        json_array_append_int(music_track_arr, g_MusicTrackFlags[i]);
+    }
+    return music_track_arr;
+}
+
 char *Savegame_BSON_GetSaveFileName(int32_t slot)
 {
     size_t out_size = snprintf(NULL, 0, g_GameFlow.savegame_fmt_bson, slot) + 1;
@@ -1266,6 +1299,13 @@ bool Savegame_BSON_LoadFromFile(MYFILE *fp, GAME_INFO *game_info)
         }
     }
 
+    if (header.version >= VERSION_3) {
+        if (!SaveGame_BSON_LoadMusicTrackFlags(
+                json_object_get_array(root_obj, "music_track_flags"))) {
+            goto cleanup;
+        }
+    }
+
     ret = true;
 
 cleanup:
@@ -1335,6 +1375,8 @@ void Savegame_BSON_SaveToFile(MYFILE *fp, GAME_INFO *game_info)
         root_obj, "lara", Savegame_BSON_DumpLara(&g_Lara));
     json_object_append_object(
         root_obj, "music", SaveGame_BSON_DumpCurrentMusic());
+    json_object_append_array(
+        root_obj, "music_track_flags", SaveGame_BSON_DumpMusicTrackFlags());
 
     struct json_value_s *root = json_value_from_object(root_obj);
     SaveGame_BSON_SaveRaw(fp, root, SAVEGAME_CURRENT_VERSION);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## [Unreleased](https://github.com/rr-/Tomb1Main/compare/stable...develop) - ××××-××-××
 - added the current music track and timestamp to the savegame so they now persist on load (#419)
+- added the triggered music tracks to the savegame so one shot tracks don't replay on load (#371)
 - changed the installer to always overwrite all essential files such as the gameflow and injections (#904)
 - fixed Natla's gun moving while she is in her semi death state (#878)
 - fixed an error message from showing on exiting the game when the gym level is not present in the gameflow (#899)
--- a/README.md
+++ b/README.md
@@ -368,6 +368,7 @@ Not all options are turned on by default. Refer to `Tomb1Main_ConfigTool.exe` fo
 - added an option to turn off sound effect pitching
 - added an option to use the PlayStation Uzi sound effects
 - added the current music track and timestamp to the savegame so they now persist on load
+- added the triggered music tracks to the savegame so one shot tracks don't replay on load
 - fixed the sound of collecting a secret killing the music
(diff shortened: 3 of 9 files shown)


src/game/savegame/savegame_bson.c
```diff
@@ -1267,6 +1300,11 @@ bool Savegame_BSON_LoadFromFile(MYFILE *fp, GAME_INFO *game_info)
         goto cleanup;
     }
 
+    if (!SaveGame_BSON_LoadMusicTrackFlags(
+            json_object_get_array(root_obj, "music_track_flags"))) {
+        goto cleanup;
+    }
+
```
rr-, line 1306: Since we used version bump in the previous PR we should also follow the same process here. I'd reuse version 3 and not bump the version to 4, because its introduction hasn't been released yet.

walkawayy: We ended up not bumping the savegame version.

lahm86:
Noticed a bit of an issue, I'm not sure if it's just timed triggers but it seems thar after the first time the switch here is used, the ambient track resets when you step on the tile again. _Update: this seems to be an original issue, just double checked._

https://www.youtube.com/watch?v=Al7mT2asvJA

It also somehow restored the other track when I saved and reloaded even though it had already finished.

walkawayy:
> 

Ok so there definitely are two things happening here:
1. Maybe there is a bug from the previous PR for saving / reloading the timestamp where it saves an inactive track and restores it.
2. For that Caves switch, there is definitely some weird stuff going on. The PS1 version didn't have ambient tracks, but the game plays the correct music trigger track 12 [when that switch is pulled](https://youtu.be/ECVAdDHDssk?t=760). In the PC version, there are two identical ambient tracks. `MX_CAVES_AMBIENT = 5` and `MX_ST_FRANCIS_FOLLY_AMBIENCE = 57` (looks like I might need to change those names). The gameflow gives Caves music as `"music": 57,`. This shouldn't be a big deal because the 57 looped ambience is the same as 5 (57 does have an extra 4 seconds of silence though strangely). However, in the level file, Caves actually has a trigger on level start saying to play track 5 (screenshot below). And in `Music_Play` there is a special condition where track 5 can never play. This condition was probably added because the level file's opening music trigger is screwed up because the ambience shouldn't use a music trigger. It should be a looped ambience.
``` 
if (track == MX_CAVES_AMBIENT) {
    return false;
}
```

image

So this is a bunch of background info to say that `Room_TriggerMusicTrack` is calling `Music_StopTrack(track);` when the switch timer ends from what I can tell. Which might be semi related to #783 @lahm86. Here's the relevant code from `room.c`. Is this a an OG bug? Or a side effect of the PR I linked?
```
    if ((g_MusicTrackFlags[track] & IF_CODE_BITS) == IF_CODE_BITS) {
        if (flags & IF_ONESHOT) {
            g_MusicTrackFlags[track] |= IF_ONESHOT;
        }
        Music_Play(track);
    } else {
        Music_StopTrack(track);
    }
```

lahm86:
You're right, I don't think it's related to this PR at all (apologies for confusing things). I just checked TombATI and if you step on a switch tile again as in the video above, the ambient track stops altogether, so there is definitely something lingering there. #783 could be related indeed; at least for now, restarting the ambient is better than it stopping entirely like TombATI. We should probably raise another issue to investigate that though.

For this PR, other than the potential timestamp issue, the `g_MusicTrackFlags` restore looks good imo.

rr-:
>Potential timestamp issue

Can we double-check this?

src/game/savegame/savegame_bson.c
```diff
@@ -822,6 +825,27 @@ static bool SaveGame_BSON_LoadCurrentMusic(struct json_object_s *music_obj)
...
         }
     }
+    return true;
+}
+
+static bool SaveGame_BSON_LoadMusicTrackFlags(
+    struct json_array_s *music_track_arr)
+{
+    if (!music_track_arr) {
+        LOG_WARNING("Malformed save: invalid or missing music track array");
```
rr-, line 835: Since we chose not to bump the version, this ominous warning message will show in the logs for good saves from the current stable. I think that we either should bump the version after all, or remove the warning. After giving it some more thinking, I'm leaning towards bumping the version to V3 and restoring the conditions for >= V3 for both music and music flags. @lahm86 @walkawayy

walkawayy:
I'm gonna put this on hold and probably do two new PRs before this.

The first will investigate the potential timestamp bug. I'd rather solve that separately before adding this PR to the mix if it's an unrelated bug.

The second will bump the savegame version. Then I can add the appropriate savegame checks to this PR after.

walkawayy:
> You're right, I don't think it's related to this PR at all (apologies for confusing things). I just checked TombATI and if you step on a switch tile again as in the video above, the ambient track stops altogether, so there is definitely something lingering there. #783 could be related indeed; at least for now, restarting the ambient is better than it stopping entirely like TombATI. We should probably raise another issue to investigate that though.
> 
> 
> 
> For this PR, other than the potential timestamp issue, the `g_MusicTrackFlags` restore looks good imo.

Ok I'll open an issue for it later tonight. The weird thing is your PR mentioned the triggers where the tracks would stop were caused by incorrectly flagged triggers. But this switch trigger seemed like normal flags to me? Though you know about them more than me.

lahm86:
> > You're right, I don't think it's related to this PR at all (apologies for confusing things). I just checked TombATI and if you step on a switch tile again as in the video above, the ambient track stops altogether, so there is definitely something lingering there. #783 could be related indeed; at least for now, restarting the ambient is better than it stopping entirely like TombATI. We should probably raise another issue to investigate that though.
> > For this PR, other than the potential timestamp issue, the `g_MusicTrackFlags` restore looks good imo.
> 
> Ok I'll open an issue for it later tonight. The weird thing is your PR mentioned the triggers where the tracks would stop were caused by incorrectly flagged triggers. But this switch trigger seemed like normal flags to me? Though you know about them more than me.

Yeah, it could be the way the flags are changed based on the trigger type in `Room_TriggerMusicTrack`. I think it's then calling `Music_StopTrack` and that's falling through and restarting the ambient - the initial condition there is probably passing because even though the track has stopped, `Music_CurrentTrack` still holds the last played track. Maybe there needs to be a flag to store whether or not the track has finished?

rr-:
WDYT about this?
```diff
diff --git a/src/game/music.c b/src/game/music.c
index b0000132..edd6ee4f 100644
--- a/src/game/music.c
+++ b/src/game/music.c
@@ -18,6 +18,7 @@ static int m_AudioStreamID = -1;
 static int16_t m_Track = 0;
 static int16_t m_TrackLooped = -1;
 
+static bool Music_IsBrokenTrack(int16_t track);
 static void Music_StopActiveStream(void);
 static void Music_StreamFinished(int stream_id, void *user_data);
 static char *Music_GetTrackFileName(int track);
@@ -36,6 +37,11 @@ static void Music_StopActiveStream(void)
     S_Audio_StreamSoundClose(m_AudioStreamID);
 }
 
+static bool Music_IsBrokenTrack(int16_t track)
+{
+    return track == MX_UNUSED_0 || track == MX_UNUSED_1 || track == MX_UNUSED_2;
+}
+
 static char *Music_GetTrackFileName(int track)
 {
     char file_path[64];
@@ -70,11 +76,7 @@ void Music_Shutdown(void)
 
 bool Music_Play(int16_t track)
 {
-    if (track == Music_CurrentTrack()) {
-        return false;
-    }
-
-    if (track <= MX_UNUSED_1) {
+    if (track == Music_CurrentTrack() || Music_IsBrokenTrack(track)) {
         return false;
     }
 
@@ -88,10 +90,6 @@ bool Music_Play(int16_t track)
             SFX_BALDY_SPEECH + track - MX_BALDY_SPEECH, NULL, SPM_ALWAYS);
     }
 
-    if (track == MX_CAVES_AMBIENT) {
-        return false;
-    }
-
     Music_StopActiveStream();
 
     char *file_path = Music_GetTrackFileName(track);
@@ -114,7 +112,7 @@ bool Music_Play(int16_t track)
 
 bool Music_PlayLooped(int16_t track)
 {
-    if (track == Music_CurrentTrack()) {
+    if (track == Music_CurrentTrack() || Music_IsBrokenTrack(track)) {
         return false;
     }
 
@@ -153,7 +151,7 @@ void Music_Stop(void)
 
 void Music_StopTrack(int16_t track)
 {
-    if (track != Music_CurrentTrack()) {
+    if (track != Music_CurrentTrack() || Music_IsBrokenTrack(track)) {
         return;
     }
 
diff --git a/src/global/types.h b/src/global/types.h
index e1831dde..ed363a99 100644
--- a/src/global/types.h
+++ b/src/global/types.h
@@ -482,7 +482,7 @@ typedef enum MUSIC_TRACK_ID {
     MX_TR_THEME = 2,
     MX_WHERE_THE_DEPTHS_UNFOLD_1 = 3,
     MX_TR_THEME_ALT_1 = 4,
-    MX_CAVES_AMBIENT = 5,
+    MX_UNUSED_2 = 5,
     MX_TIME_TO_RUN_1 = 6,
     MX_FRIEND_SINCE_GONE = 7,
     MX_T_REX_1 = 8,
@@ -534,7 +534,7 @@ typedef enum MUSIC_TRACK_ID {
     MX_NATLA_SPEECH = 54,
     MX_PIERRE_SPEECH = 55,
     MX_SKATEKID_SPEECH = 56,
-    MX_ST_FRANCIS_FOLLY_AMBIENCE = 57,
+    MX_CAVES_AMBIENCE = 57,
     MX_CISTERN_AMBIENCE = 58,
     MX_WINDY_AMBIENCE = 59,
     MX_ATLANTIS_AMBIENCE = 60,
```

src/game/savegame/savegame_bson.c
walkawayy: To close this loop, I added back in the savegame version bump since we decided to do it.

src/game/savegame/savegame_bson.c
walkawayy: Bumped version.

walkawayy:
Ok I did some testing and it works. The exception being triggers not marked as one shot like the Great Pyramid Torso music.

If the track finishes, the players saves, then reloads, the music will restart.
If the player saves while the track is playing, the track resumes on reload then finishes but doesn't replay because the game knows it's the current track.

This might be the best we can do? Unless we go injecting these non one shot triggers into one shot. Lmk what you guys think.

src/game/savegame/savegame_bson.c
```diff
@@ -1266,6 +1299,13 @@ bool Savegame_BSON_LoadFromFile(MYFILE *fp, GAME_INFO *game_info)
         }
     }
 
+    if (header.version >= VERSION_3) {
```
rr-: Let's join it with the condition above rather than making a new block for it.

rr- (commented):
Almost there. It'd be good to give it one final round of thorough testing, ideally in caves, with the botched lever trigger, the wolf music, the secret sfx, save/reload and check for audio clipping too.

Edit: I just noticed the comment above. Yes, that's fine. I'm not sure how to tackle the Great Pyramid Torso trigger issue, other than with injections. Which I don't think are super-necessary.

lahm86:
Converting to one shot triggers would be straight-forward, and I'd be happy to take it on. Perhaps we can make a list in #755 of exactly where these are needed, and then begin to plan implementation along with the other FD tasks. We have the `fix_floor_data_issues` option already so imo these would fall under that category.

rr-:
Sounds good to me :)

walkawayy:
Ok joined sg `VERSION_3` conditions. Should we keep the one shot trigger fixes in this PR or put those in a new one?

rr-:
Let's tackle that separately

rr- (approved):

lahm86 (approved):

PR #907, opened 2023-08-06
installer: overwrite all essential files

### LINKED ISSUE #904: Update installer to overwrite essential files
The installer should always overwrite essential files such as the gameflow and injections, as otherwise such things as sequence and injection updates can be missed.
Related to #882.

lahm86:
Resolves #904.

#### Checklist

#### Description

Changed the overwrite check approach to look at extensions so to avoid listing each file (the bin files alone would make this unmanageable I think). It will only overwrite what's in the release zip anyway, so such things as `Tomb1Main.json5` will remain unaffected. I included the shaders too - do you think this is OK, or given previous shader issues should we leave these out?

### DIFF
--- a/tools/installer/Installer/Controls/InstallSettingsStepControl.xaml
+++ b/tools/installer/Installer/Controls/InstallSettingsStepControl.xaml
@@ -127,8 +127,8 @@
                         Overwrite all existing files
                         <LineBreak />
                         <Run Style="{StaticResource small}">
-                            Overwrites any files such as level data, music files and existing user settings.
-                            If this option is off, only overwrites Tomb1Main.exe in order to update the game.
+                            Overwrites any files such as level data and music files. If this option is off, only
+                            overwrites Tomb1Main.exe and other essential files in order to update the game.
                         </Run>
                     </TextBlock>
                 </CheckBox>
--- a/tools/installer/Installer/Controls/InstallSettingsStepControl.xaml
+++ b/tools/installer/Installer/Controls/InstallSettingsStepControl.xaml
@@ -122,17 +122,6 @@
                     </TextBlock>
                 </CheckBox>
 
-                <CheckBox VerticalAlignment="Center" Margin="0,0,0,12" IsChecked="{Binding InstallSettings.OverwriteAllFiles}">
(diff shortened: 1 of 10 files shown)


tools/installer/Installer/Controls/InstallSettingsStepControl.xaml
```diff
@@ -127,8 +127,8 @@
                         Overwrite all existing files
                         <LineBreak />
                         <Run Style="{StaticResource small}">
-                            Overwrites any files such as level data, music files and existing user settings.
-                            If this option is off, only overwrites Tomb1Main.exe in order to update the game.
+                            Overwrites any files such as level data and music files. If this option is off, only
+                            overwrites Tomb1Main.exe and other essential files in order to update the game.
```
rr-: I don't think this checkbox does anything given the new changes – the release .zip contains only files from the new 
 extensions whitelist?

lahm86: There are the three PNG files too - `eidospc.png`, `titleh.png` and `titleh_ub.png`. Should we just remove the checkbox altogether?
Actually, the setting is also used here, so I suppose that ensures music and level files are retained.

https://github.com/LostArtefacts/Tomb1Main/blob/develop/tools/installer/Installer/Installers/InstallExecutor.cs#L62

rr-: IMO we should remove this box altogether, and the "install source" shouldn't overwrite the files whereas tomb1main files should. What do you think about this?

lahm86: Sounds good, I'll work on updating it.

lahm86: Updates now in place:
* Everything in the Tomb1Main embedded zip file will always be written to the target.
* Install source will only write files that aren't in the target folder already.
* Downloaded music and UB files will not be written if they already exist.

rr- (approved):

walkawayy (approved):
LGTM

PR #910, opened 2023-08-08
music: fix saving track timestamps loading tracks after they ended

### LINKED ISSUE #908: Using the save a track's timestamp option plays the track on load even if it ended
Discovered in #905. @rr- any ideas on how to resolve this? i think what's happening is `Music_GetTimestamp` is using `m_AudioStreamID` to check `m_StreamSounds[sound_id].is_playing`. But this is always returning true because even if music ends, the looped ambience starts (with the same `m_AudioStreamID`) so `is_playing` is `true`.

In `music.c` we have `m_Track`. `m_TrackLooped`, and `m_AudioStreamID` but I don't think any of these can be used safely. `m_Track` is always stored to prevent tons of replaying music issues where the level files have tons of wrongly set non one-shot triggers like the Great Pyramid torso music.

walkawayy:
Resolves #908.

#### Checklist

#### Description
Fixed saving track timestamps loading tracks after they ended. No changelog entry since this bug was never released in a major version. This also seems to fix the Caves switch issue mentioned by @lahm86 in #905 (https://www.youtube.com/watch?v=Al7mT2asvJA). I'm still going to add a separate commit to do some code cleanup with the broken track checking as suggested by @rr-.


src/game/music.c
```diff
@@ -49,9 +56,7 @@ static void Music_StreamFinished(int stream_id, void *user_data)
 
     if (stream_id == m_AudioStreamID) {
         m_AudioStreamID = -1;
-        if (m_Track == MX_SECRET) {
```
lahm86, line 52: If you disable `fix_secrets_killing_music` then collect the first secret in Caves, the secret track plays, but if you then immediately go for the second secret, it doesn't play. Is it maybe because of this particular change?
The third secret is fine as there is other music triggered in between. And equally, it's all fine with this setting enabled as standard.

lahm86 (commented):
Main issue looks good now 👍 Just another small issue introduced re secrets.

src/game/music.c
walkawayy: Ah shoot this is a repeat of #311. The problem is it skips the secret track because the game intentionally doesn't play the same track twice in a row (which doesn't affect sound effects which the fix uses). I thought `m_TrackCurrent = MX_INACTIVE;` would be fine, but I forgot we added a `Music_Play` check for `track == m_TrackLastPlayed` also.

walkawayy: I think we might need to add back in something like:
```
if (m_Track == MX_SECRET) {
    m_TrackLastPlayed = MX_INACTIVE;
}
```

src/game/music.c
```diff
@@ -103,7 +101,8 @@ bool Music_Play(int16_t track)
         return false;
     }
 
-    m_Track = track;
+    m_TrackCurrent = track;
+    m_TrackLastPlayed = track;
```
rr-: I think it'll suffice to patch this with:
```
if (track != MX_SECRET) {
    m_TrackLastPlayed = track;
}
```

src/game/music.h
```diff
@@ -35,12 +35,17 @@ void Music_Pause(void);
...
-// track, returns the "overriding" track number.
-int16_t Music_CurrentTrack(void);
+// Returns the currently playing track. Ignores looped tracks.
+MUSIC_TRACK_ID Music_CurrentTrack(void);
+
+// Returns the last played track. Ignores looped tracks.
+MUSIC_TRACK_ID Music_LastPlayedTrack(void);
+
+// Returns the looped track.
+MUSIC_TRACK_ID Music_CurrentTrackLooped(void);
```
rr-: Let's call this `Music_GetLoopedTrack` and rename this family of functions to `Music_Get…Track`. The looped track doesn't have to be currently playing.

rr- (commented):
Good call on using the enums and introducing `MX_INACTIVE`, me gusta.

src/game/music.h
walkawayy: Ya good call. `Music_GetLastPlayedTrack` and `Music_GetCurrentLoopedTrack` aren't used atm, but I'm gonna leave them in there for future use.

walkawayy:
Alright I tested a ton of music triggers, switch triggers, one shot, non one shot, and `fix_secrets_killing_music`. I think everything is behaving properly now.

EDIT: I lied. @lahm86 now that Caves switch is replaying the ominous music track when Lara steps on the tile with the switch. But only the first time after a reload. And it resets to ambience when she pulls the switch. I can't figure out if it's a problem with the trigger or what.

This switch also in Caves does not have that issue.

image

rr-:
Probably we need to reset `m_LastTrackPlayed` or something similar to `MX_INACTIVE` on game load. How about this?

```diff
diff --git a/src/game/game/game.c b/src/game/game/game.c
index a1741d67..fb992f2c 100644
--- a/src/game/game/game.c
+++ b/src/game/game/game.c
@@ -134,11 +134,6 @@ bool Game_Start(int32_t level_num, GAMEFLOW_LEVEL_TYPE level_type)
         // Item_GlobalReplace in the inventory initialization routines too early
         Savegame_InitCurrentInfo();
 
-        // prevent audio clipping that occurs when level plays a music track,
-        // and later savegame routines override it while the stream is already
-        // playing in the background thread
-        Music_Pause();
-
         if (!Level_Initialise(level_num)) {
             return false;
         }
@@ -146,7 +141,6 @@ bool Game_Start(int32_t level_num, GAMEFLOW_LEVEL_TYPE level_type)
             LOG_ERROR("Failed to load save file!");
             return false;
         }
-        Music_Unpause();
         break;
 
     case GFL_RESTART:
diff --git a/src/game/level.c b/src/game/level.c
index edf5537d..ad275bb6 100644
--- a/src/game/level.c
+++ b/src/game/level.c
@@ -803,6 +803,7 @@ bool Level_Initialise(int32_t level_num)
     Overlay_Init();
     Overlay_BarSetHealthTimer(100);
 
+    Music_Stop();
     Sound_ResetEffects();
 
     Viewport_AlterFOV(g_Config.fov_value * PHD_DEGREE);
```

Double check if this doesn't reintroduce the clipping problem.

lahm86:
I think it's to do with switches that are timed, so when the timer resets, so do the switch's flags and then the game thinks the music hasn't been played before. But then it sees that it was the last played track and so resets to the ambient.

What do you think about this in `Room_TestTriggers`?

```
case TO_CD:
    if (type != TT_SWITCH || !switch_off) {
        Room_TriggerMusicTrack(value, flags, type);
    }
    break;
```

Music should only trigger on a switch when it's in the on state. The problem is that after the timer resets and Lara stands on the tile again, `Switch_Trigger` further up in this function is passing when the switch is off and so control eventually gets down to the `TO_CD` case. The `switch_off` var seems to be used in a similar way for timed switch camera triggers.

Perhaps the issue itself is in `Switch_Trigger` but I can't see any problems there and tinkering with that is probably going to introduce more problems. At least with the `TO_CD` case above we'd be limiting it to music only?

rr-:
I don't know, while this looks OK, it also feels like it's going to bite us in the butt very shortly.

lahm86:
Yes, it's certainly not ideal. There are only 3 timed music switches in the OG levels, but this is likely used heavily in TRLEs.

LEVEL1.PHD: room 26
END.PHD: room 39
END2.PHD: room 9

The trigger data itself checks out as far as I can tell.

rr-:
Let's try not to go down that route and first see what the patch I proposed does.

walkawayy:
Ok I think rr's solution works!

lahm86 (approved):
LGTM (and sounds good to me too)👍

rr- (approved):

Issue #882, opened 2023-06-11
More streamlining for custom levels


mizunosuisei:
After noticing that tomb1main updates don't work unless you also replace the gameflow file, this leaves custom levels slightly in the lurch. Ofcourse it's easy enough to copy and paste their custom gameflow text into the updated gameflow, many wouldn't know how/bother to do this.

The only (far off) suggestion I can think of is to have all the injection data in a different cfg file that doesn't relate to per level entries, yet that feels against the purpose ofcourse - or allow the game to run regardless if injection data is/isn't there.

lahm86:
An update will he coming soon to enforce the gameflow to be updated when using the installer. Custom level builders have the tools available to them to update the script for their games, Tomb1Main cannot be responsible for that. Indeed, it is prudent for builders to check what is and what is not relevant to inject into their games. For example, `caves_textures.bin` contains spexif

Other gameflow changes such as recent remove or add new sequence types. These are again essential to how the game works, the latest being that bacon Lara's position is now customizable. This level of customization comes at the cost of

lahm86:
_Apologies for closing this before replying initially, I hit the wrong button._ 

We are looking at a change to how the gameflow is updated via the installer such that it will always be updated on new release. 

I think for level builders, careful checks should be done to see what is and what is not relevant to inject into their games as Tomb1Main cannot be responsible for that. For example, texture fix files for OG levels will not be required in custom levels, so builders should remove those entries.

Other recent changes involve gameflow sequence updates, and these are essential to how the game plays out, so it's normal imo that the game should not load when something is missing or incorrect.

mizunosuisei:
Thankyou for your response :) I'm really not knowledgeable with the indepths of it all.

Just throwing this one - it could be laughable- but wouldn't it possibly be a more novel idea to force tomb1main to load regardless if the data folder contains bin injections, and the script does not? I'm mostly thinking around this for older customs too.

rr-:
>Change the installer to have gameflow always updated

Do we have an issue for this?

>Gameflow woes

Level authors should make updates to the gameflow in a conscious and controlled manner; I don't think it needs to be said but expecting a drag-n-drop of the gameflow file to work is misplaced.  Now, I don't think our updates to the bundled gameflow are revolutionary either – drag-n-drop of a new Tomb1Main.exe to an old level should work fine. (Correct me if I'm wrong. If that's not the case, I think we should provide some reasonable amount of backwards compatibility around this.)

[...]

PR #919, opened 2023-08-12
data: allow failed injections

### LINKED ISSUE #918: Allow data injections to fail
Related to #882, the injection system should allow levels to continue to load regardless of injection success or failure.

lahm86:
Resolves #918.

#### Checklist

#### Description

Related to #882, this will allow levels to load in spite of injection failures. The overall flow for level loading is as follows:
* Initialise injections
* Read level file, using aggregated injection info
* Inject
* Complete level setup
* Cleanup injection resources

Previously, injection cleanup was only taking place if there was at least one valid injection file. Now this is pushed to the end of the flow to ensure that resources for invalid injections are released e.g. files that were opened, but where invalid data was detected at some point in the process.

I've tested the following cases.
* The gameflow points to an injection file that does not exist
* An injection file has invalid header magic
* An injection file has an unsupported version
* An injection type does not match any defined `INJECTION_TYPE`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 - added the current music track and timestamp to the savegame so they now persist on load (#419)
 - added the triggered music tracks to the savegame so one shot tracks don't replay on load (#371)
 - changed the installer to always overwrite all essential files such as the gameflow and injections (#904)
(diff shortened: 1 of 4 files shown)


rr- (approved):

walkawayy (approved):
Nice changes!

PR #921, opened 2023-08-13
gameflow: ignore legacy sequences

### LINKED ISSUE #882: More streamlining for custom levels

lahm86:
Resolves #882.

#### Checklist

#### Description

Adds handling for legacy gameflow sequence types - currently, as far as I can tell, this only relates to #790 where the Pyramid secret fix was converted to injection format. This should benefit older custom level instances where the GF setup is a copy of the original script at the time, but modified to `exit_to_title` after level 1, for example.

I did initially consider changing the "unknown sequence type" check as a whole to a `GFS_UNKNOWN` type and allowing it to continue rather than exit, but decided against it. My feeling is known legacy types should be captured, but unknown types fall under the category of poor editing so it should be more obvious to builders/players when the file is not as expected. Let me know what you think though, happy to discuss it.


rr- (approved):

walkawayy (approved):
I agree with your decision. 👍

PR #925, opened 2023-08-16
centaur: fix explosion after re-loading

### LINKED ISSUE #924: OG bug: dead centaurs explode again after reloading
This occurs in TombATI and T1M and can be recreated in Tomb of Tihocan as follows.
* Enter the temple at the end
* Cross the trigger to activate the centaur on the right
* Return outside and kill it
* Save and reload
* Cross the trigger again inside the temple
* Return outside to within the centaur's range

https://github.com/LostArtefacts/Tomb1Main/assets/33758420/1092712c-1e5f-43f9-9aea-3f243660b4d9

lahm86:
Resolves #924.

#### Checklist

#### Description

This only affects centaur statues - the control code here was always reactivating the actual centaur item on load, which was then falling through to the explosion call in `centaur.c`. Normal centaurs (and all other Atlanteans) behave normally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - fixed an error message from showing on exiting the game when the gym level is not present in the gameflow (#899)
 - fixed the bear pat attack so it does not miss Lara (#450)
 - fixed some incorrectly rotated pickups when using the 3D pickups option (#253)
+- fixed dead centaurs exploding again after saving and reloading (#924)
 - improve spanish localization and added translation for rotated pickups
 
 ## [2.15.3](https://github.com/rr-/Tomb1Main/compare/2.15.2...2.15.3) - 2023-08-15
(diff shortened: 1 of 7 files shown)


lahm86:
~~Translations to follow.~~

~~@ajtudela Could you provide a translation for the new setting? Thanks.~~

Cancel that - this will be a default setting. Thanks!

rr- (approved):

walkawayy (approved):
Good catch. Do you think this even needs a config option? I don't see a use case for the OG behavior, and it seems like an obvious bug where they forgot a kill check.

rr-:
Yes actually we can fix it without the config option.

lahm86:
Config option removed, re-requesting review for sanity checks.

rr- (approved):

walkawayy (approved):

PR #936, opened 2023-08-22
overlay: fix reset and unbind progress bars in controls menu

### LINKED ISSUE #930: T1M bug: UI bar scale setting also affect bars in the controls menu
Related to #658 (v2.12) and # #776 (2.14)

For example, when UI bar scale is set to 0.7:

Title

walkawayy:
Resolves #930.

#### Checklist

#### Description
Fixed the reset and unbind progress bars in the controls menu for non-default bar scaling. The progress bar needs to depend on text scaling instead of the bar scaling since it is positioned behind the reset and unbind text in the controls menu.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - fixed the incorrect starting animation on centaurs that spawn from statues (#926, regression from 2.15)
 - fixed jump-twist animations at times being interrupted (#932, regression from 2.15.1)
 - fixed walk-run-jump at times breaking when TR2 jumping is enabled (OG bug in TR2+) (#934)
+- fixed the reset and unbind progress bars in the controls menu for non-default bar scaling (#930)
 - improve spanish localization and added translation for rotated pickups
 
 ## [2.15.3](https://github.com/rr-/Tomb1Main/compare/2.15.2...2.15.3) - 2023-08-15
--- a/src/game/option/option_control.c
+++ b/src/game/option/option_control.c
@@ -552,6 +552,7 @@ static void Option_ControlProgressBar(TEXTSTRING *txt, int32_t timer)
     } else if (txt->flags.right) {
         x += Screen_GetResWidthDownscaledText() - width;
     }
+
     if (txt->flags.centre_v) {
         y += Screen_GetResHeightDownscaledText() / 2;
     } else if (txt->flags.bottom) {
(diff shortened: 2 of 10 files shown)


rr- (commented):
Is there a way to avoid the heavy code duplication?

rr- (commented):
This looks okay to me, but I think introducing a new enum rather than passing function pointers would be a bit cleaner; what do you think about this @lahm86 @walkawayy ?

walkawayy:
> This looks okay to me, but I think introducing a new enum rather than passing function pointers would be a bit cleaner; what do you think about this @lahm86 @walkawayy ?

I thought about that but then I'd have to add a bunch of if/else checks depending on the enum to call the correct scale function right? I can do that but thought I'd try something different. LMK what you'd prefer.

rr-:
> > This looks okay to me, but I think introducing a new enum rather than passing function pointers would be a bit cleaner; what do you think about this @lahm86 @walkawayy ?
> 
> I thought about that but then I'd have to add a bunch of if/else checks depending on the enum to call the correct scale function right? I can do that but thought I'd try something different. LMK what you'd prefer.

There could be a gateway function that takes the value to scale and an enum value how to do that, and delegates to the proper procedure. This way the users of the enum do not have to embed the if/else logic themselves.

walkawayy:
> > > This looks okay to me, but I think introducing a new enum rather than passing function pointers would be a bit cleaner; what do you think about this @lahm86 @walkawayy ?
> > 
> > 
> > I thought about that but then I'd have to add a bunch of if/else checks depending on the enum to call the correct scale function right? I can do that but thought I'd try something different. LMK what you'd prefer.
> 
> There could be a gateway function that takes the value to scale and an enum value how to do that, and delegates to the proper procedure. This way the users of the enum do not have to embed the if/else logic themselves.

I think that would work. I'll try that and PR it.

walkawayy:
Alright this simplified the function in `screen.c` a lot, but it also meant I had to touch a lot of different code to update the calls. Lmk what you think.

rr- (commented):
One last thing and we're good to go – let's make `Screen_GetRenderScaleBase` static and remove it from the .h file.

rr- (approved):
LGTM but let's wait for Lahm. When merging, please squash and adhere to the commit message standards (as we've had a few examples of commits that disregard the convention recently.)

lahm86 (approved):
Nice solution, LGTM.

Issue #942, opened 2023-08-24
OG bug: incorrect trigger types in Egypt levels


lahm86:
As discussed in #755, there are incorrect trigger types in the following levels.

**City of Khamoon**
In room 31 and 34, the trapdoor triggers are of regular type, meaning Lara need only stand below to open them. These should be dummy triggers.

**Obelisk of Khamoon**
In room 66, the trigger by the switch is of regular type, so Lara need only stand on the tile to deactivate the flipmap. This should be a switch trigger as per the unflipped room, 35.

Issue #941, opened 2023-08-24
Parse floor data into structures


lahm86:
Test the feasibility of parsing floor data into structures - for example, a linked list that `FLOOR_INFO` points to - to replace repeated `g_FloorData` looping logic in the codebase. This ticket is mainly for investigation initially and will be padded out as more details are gathered.

- [x] Rename `FLOOR_INFO` to `SECTOR_INFO` and update all relevant variables
- [x] Standardize `x_floor`/`y_floor` variables when used to calculate index into a room's 2D sector array - make `X/Z` the standard
- [x] Introduce struct in `SECTOR_INFO` for `floor` and `ceiling` to store height and tilt values
- [x] Rename `OBJECT_INFO` `floor` and `ceiling` function pointers, and the functions they point to, to be more descriptive of intent
- [x] Make a "master" floor data parser called on level load; for each sector:
  - [x] Set the `tilt` value for floor and ceiling
    - [x] Refactor usage
  - [x] Set the `room_portal` value
    - [x] Refactor usage
  - [x] Set the `is_death_sector` property
    - [x] Refactor usage
  - [x] Set the `trigger`
    - [x] Refactor usage
- [x] Delete relevant globals, such as `g_TriggerIndex`
- [x] Rewrite floor data injection to use the new structured approach (to be investigated)

walkawayy:
I think it might be a good idea to implement some sort of `on_walkable` bool or something similar into the `SECTOR_INFO` struct. That way I think we could basically implement #1429 into the `Room_GetHeight` function instead of having to traverse the sector again later. This was the way we thought originally to do it, but we didn't want to add another global.

Issue #865, opened 2023-06-03
Prevent music trigger if room is closed


Richard-L:
This occurred to me on a recent PS1 playthrough with a family member (still the only way to get a German dub without replacing audio files), and it's the same in TR1M, and made me think if it's worth adding extra code that prevents music triggers if the intended room (or moment) are still inaccessible.

Consider this case of the Neptune room in Folly (requires audio):

https://github.com/LostArtefacts/Tomb1Main/assets/4934209/a9ad52ef-311a-4776-834d-676f6e5ab4c2

I felt it's unfortunate to "waste" the music because the music trigger should maybe have been on a tile behind the door here. The music is intended to accompany the long dive down. Instead here it plays as you continue through the main room of the folly. I suppose there might be more occurrences of this throughout the game.

I assume this would either mean to inject a changed level file where the music trigger has been moved behind the door, or add code logic that only plays music tracks _if_ a specific door has been opened. Both don't seem overly desirable maybe. Let me know your thoughts.

lahm86:
An injection will be easiest imo - I think it would be too complicated to handle it automatically, especially for this kind of situation as technically you are in the Neptune room at that point.

![image](https://github.com/LostArtefacts/Tomb1Main/assets/33758420/93d2b4d5-54ea-4d45-85aa-a62f141fea63)

Can certainly tackle this as part of #755.

aredfan:
Nice find.  I agree it's a waste for the Neptune music to play prematurely.  The Thor and Damocles rooms have music triggers placed past the doors, so it would be consistent to move back the Neptune music trigger as-well.

Agree with Lahm that floordata injection is the way to tackle this issue.

lahm86:
I'm planning to bundle the fix for the 4-key music trigger into this issue as well, given that it's within scope I feel (a Folly music trigger problem basically).

https://github.com/LostArtefacts/Tomb1Main/discussions/755#discussioncomment-5275212

PR #944, opened 2023-08-25
inject: add the ability to insert floor data

### LINKED ISSUE #751: Missing trigger tile for a rat | Tomb of Tihocan
It seems I keep raising issues for @lahm86 lately, I'm so very sorry.  Also tagging @Richard-L too in case he has any thoughts to share.

This issue is about a rat which doesn't seem to trigger until Lara gets closer.

20230309_121008_Tomb_of_Tihocan

The interesting thing is the trigger tiles are there and extends all the way to the edge, see the screenshot below.

Screenshot (227)

However when flipmap is triggered the tiles are reduced.  I suspect this is a mistake.

Screenshot (228)

lahm86:
Resolves #751.

#### Checklist

#### Description

This allows us to insert new floor data values, and is used here to restore the missing rat trigger in Tomb of Tihocan room 62.


rr- (commented):
This is only capable of overwriting existing FD, are there cases where we'll want to insert or remove data and thus realign the entire thing?

lahm86:
> This is only capable of overwriting existing FD, are there cases where we'll want to insert or remove data and thus realign the entire thing?

This appends on to the end of the array - the tile in Tihocan as it stands has no floor data. A previous injection update did some forward planning here - we aggregate how much extra FD space is needed for all injections, then it's allocated in `level.c`.

https://github.com/LostArtefacts/Tomb1Main/blob/develop/src/game/inject.c#L293
https://github.com/LostArtefacts/Tomb1Main/blob/develop/src/game/level.c#L236

The only other case that's been identified will be for moving the music trigger behind the Neptune door in Folly (#865). My plan for here was to change the existing trigger in front of the door to play track `MX_UNUSED_0` (i.e. so it will effectively be skipped in `music.c`) and then create a new entry for the trigger behind the door with the correct track plus the flip command that's there already. This would leave the old values in the array, they would just be unreferenced.

The other remaining cases are for #942, but these are just type changes, so direct `int16_t` replacements.

I can imagine another large repeated FD loop to achieve removal/realignment. The injections would also need to aggregate removal size so that the FD array is set initially to the correct size in `level.c` (I don't believe `GameBuf` can realloc/free individual buffers?) and from that it could get tricky in terms of implementation order of the injections themselves. For just the Folly case, I don't feel it's worth it - what do you think?

If we manage to get somewhere with #941, this will become even tidier. So the level would load the FD array, we would inject/modify as required, and then it would be parsed into the structures. Data that's no longer pointed to would be discarded.

rr-:
Thanks for clarification.

>I don't believe GameBuf can realloc/free individual buffers?

I think so too.

> For just the Folly case, I don't feel it's worth it - what do you think?

Agreed.

> My plan for here was to change the existing trigger in front of the door to play track `MX_UNUSED_0`

What if we go even further and introduce a new trigger type that does nothing? This way we can nullify more. Perhaps we don't even need to introduce anything and just overwrite the bytes with any value that's currently unhandled (I suggest using all bits set to 1).

rr- (approved):

lahm86:
> What if we go even further and introduce a new trigger type that does nothing? 

Oh that's a great point - we could use the existing `TT_DUMMY` for that scenario. This is skipped in `Room_TestTriggers` as-is.

walkawayy (approved):

PR #946, opened 2023-08-26
inject: fix music trigger setups in Folly

### LINKED ISSUE #865: Prevent music trigger if room is closed

lahm86:
Resolves #865.

#### Checklist

#### Description

We move the Neptune room trigger here behind the door, so it's on par with the Thor room.

Additionally, the track that was intended to be played after using the 4 keys at the end of the level has been restored. The original setup has the music command on the masked triggers, but that setup is not possible because of how the music is controlled (and I don't believe it's possible in any game). The standard for custom levels is to use heavy triggers for this kind of setup, so we do the same here and make use of the boulder that opens the door.

https://www.youtube.com/watch?v=jSwCXs2Eizk


walkawayy (approved):
Nice!

Richard-L:
This is lovely Lahm, but allow two comments:
- Would it be even better to move the Neptune music trigger one further tile back into the room itself? It seems like Thor's happens only when past the door and I like it.
- I don't fully understand the need for a boulder at the end to enable the added track, but it does sound quite weird to hear one rolling somewhere in this moment. Is there no other way around this?

In any case, thank you

rr-:
Yeah, can we do something about the boulder sound?

lahm86:
Yes, it's no problem to shift the Neptune trigger further into the room.

Re the boulder, there are two options.

1. Disable the SFX on the object
2. Move the hidden room elsewhere on the map so you can't hear it

I think (2) is best, otherwise you wouldn't hear the boulder in the Atlas room either.  The room's position is arbitrary as it has no portals. This will need a little extra injection code, but I don't think it will be too much.

image

rr-:
Yeah let's go with (2).

lahm86:
Let me know what you think: https://youtu.be/ewd21yybAcU

For the additional code, we do the basic room position shift, and then move any items with the same room number to follow i.e. the boulder. If we're shifting by Y, we need to update the sectors and the vertices - although the latter isn't so important here as the room is never drawn, but it's good practice I feel (I'm not sure if we'll ever need this for anything else, but it's there if we do).

For interest, this is the room's new position.

image

rr-:
Sounds good to be on a mobile. I'll be only able to test it on a PC tomorrow.

walkawayy (approved):
LGTM!

Richard-L:
That's perfect Lahm ♥️

rr- (approved):

PR #948, opened 2023-08-28
inject: fix Atlantean Stronghold room portals

### LINKED ISSUE #227: Original TRUB levels crash on final Atlantean Stronghold jump

lahm86:
Resolves #227.

#### Checklist

#### Description

This amends the portals between rooms 74 and 12 per @SuiKaze's [comments](https://github.com/LostArtefacts/Tomb1Main/issues/227#issuecomment-1010094064). The crash is easy to reproduce with the fly cheat if you get into this area with the flip map on. It's also entirely possibly to skip over the flipoff trigger fix highlighted below if you make a mistake and fall down, and then use a setup used by speedrunners to drop into the hole.

image

Demo of the crash and the fix.
https://www.youtube.com/watch?v=ye_jnebNu3U

The modified injection function here is also used in Midas to amend the height of the portal between rooms 9 and 13 (which is linked to texture fixes in this area). The change was needed as there are 3 different portals between these Stronghold rooms, but the previous approach only allowed targeting one per room pair.


rr- (approved):

walkawayy (approved):
Nice!! This bug can finally be fixed and closed.

Issue #352, opened 2022-01-18
Strange fixed-camera positions after pulling switch


aredfan:
This happens when Lara pulls the switch to open the door leading into pyramid exterior.

https://user-images.githubusercontent.com/94776985/149934305-df1759e0-d766-4716-ad42-d587ff98b87d.mp4

Screenshot:

![vlcsnap-2022-01-18-12h03m34s754](https://user-images.githubusercontent.com/94776985/149934328-bcf70997-1d9c-4c74-af2d-d437630ad23c.png)

Richard-L:
The same bug is true in the PS1 version. Video featuring some sweet switch pulling fail :)
https://user-images.githubusercontent.com/4934209/152066392-fef65da9-ab9c-4934-8582-0e97cbb5c0eb.mp4

Could it be they intentionally hid the door opening to make it harder? Would be interesting to know if the PHD contains any info.

atombraiderfan:
The other switch in the gold maze also have a brief camera position shown for a split second before we see the door opening.

https://user-images.githubusercontent.com/94776985/152138852-e007950b-64de-45ec-b8e1-b1cf77aa6125.mp4

Screenshot:

![1](https://user-images.githubusercontent.com/94776985/152138933-d5700a7c-2d4f-43ef-b88d-3359d04648a4.png)

Rapora9:
This (the upper one) is how it looks like if camera has 1 second timer (in custom levels made with Tomb Editor). It doesn't last 1 second but just a flashing moment. Assuming this works the same way in original TR1, my guess is that the devs just put a wrong number for the camera timer.

aredfan:
I've come across three more switches with a brief camera angle, 1 is found in The Cistern and 2 in Tomb of Tihocan.

aredfan:
Given that there's renewed interest in this issue, I've put together a full list of places where this bug happens, in both TR1 and TRUB.

- LEVEL7A - # 11 - Room 15 - Key
- LEVEL7B - # 163 - Room 72 - Switch
- LEVEL7B - # 223 - Room 108 - Switch
- LEVEL10A - # 87 - Room 64 - Switch
- LEVEL10A - # 90 - Room 67 - Switch
- EGYPT - # 18 - Room 12 - Pad
- END - # 304 - Room 59 - Switch
- END - # 257 - Room 40 - Switch

lahm86:
The one frame issue when the camera shows beyond the door can be explained as follows.

[...]

PR #951, opened 2023-08-28
inject: restore City of Khamoon mummy

### LINKED ISSUE #886: Feature Request: Add the extra mummy in City of Khamoon like the PS1 version
Is there any possibility you could add the extra mummy in the City of Khamoon the same as the PS1 version? Stella's Tomb Raider site describes this room as **"POOL & BLUE ROOM WITH MANY LEDGES"**

lahm86:
Resolves #886.

#### Checklist

#### Description

This option places the mummy and its triggers as follows in City of Khamoon (triggered both in the flipped and unflipped rooms). So the mummy spawns by the switch, and is triggered on the tile below the pushblock.

image

I had to move `Stats_ObserveRoomsLoad` in `level.c` to take place post-injection for the enemy kill count to update correctly. While this wasn't a problem previously for the restored bat in Colosseum, it is here because we're changing the `index` on two sectors that don't currently have any floor data. So as a result, the kill count remains correct.

I've ticked off `fix floor data issues` in the readme for now as this is the final topic discussed in #755.


lahm86:
@ajtudela Could you provide translations for this new option? Thanks.

rr- (commented):
Can we make this option more generic, such as "restore PS1 enemies" and then only in the option description talk about the mummy in specific?

lahm86:
> Can we make this option more generic, such as "restore PS1 enemies" and then only in the option description talk about the mummy in specific?

Updated now, and I updated the vars/enums as well to be consistent. LMK what you think.

image

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 - added the current music track and timestamp to the savegame so they now persist on load (#419)
 - added the triggered music tracks to the savegame so one shot tracks don't replay on load (#371)
 - added forward/backward input detection in line with TR2+ for jump-twists (#931)
+- added an option to restore the mummy in City of Khamoon room 25, similar to the PS version (#886)
```
walkawayy: Idc personally but all other changelog entries use `PS1`. Maybe we should be consistent? Or we could retroactively change them all to `PS` or `PSX`. Whatever @rr- prefers.

tools/config/Tomb1Main_ConfigTool/Resources/Lang/en.json
```diff
@@ -239,6 +239,10 @@
       "Title": "Change Pierre spawn behavior",
       "Description": "Makes a freshly triggered (runaway) Pierre replace an already existing (runaway) Pierre."
     },
+    "restore_ps1_enemies": {
+      "Title": "Restore PS1 enemies",
+      "Description": "Adds the mummy that appears in the PlayStation version of City of Khamoon, room 25.\nNB Changing this option will require restarting the level."
```
walkawayy: What does `NB` mean?

walkawayy (changes requested):

tools/config/Tomb1Main_ConfigTool/Resources/Lang/en.json
lahm86: Nota bene - maybe too formal?

CHANGELOG.md
rr-: I'm fine with both, props for consistency

tools/config/Tomb1Main_ConfigTool/Resources/Lang/en.json
rr-: Let's just remove it, as it doesn't add much value

walkawayy:
Also going to be annoying and link #352 since I believe that is also a floor data issue.

lahm86:
> Also going to be annoying and link #352 since I believe that is also a floor data issue.

I'm not convinced this is a floor data issue. The trigger there has a timer value of 2, and changing this makes no difference to the outcome. I think there is a bug somewhere in the camera control, but haven't pinpointed it yet.

CHANGELOG.md
lahm86: Changed to PS1 to match.

tools/config/Tomb1Main_ConfigTool/Resources/Lang/en.json
lahm86: NB: NB removed :)

rr- (approved):

ajtudela:
Done #953

walkawayy (approved):

PR #958, opened 2023-08-31
build: always bump git version

rr-:
#### Checklist

#### Description

Rebuilding the exe was not updating the git version shown inside the game, which caused me some issues when I was working on the project across two machines. This new approach forces meson to always reconsider the current version, and rebuild relevant autogenerated files: `version.rc` and `init.c` prior to linking the `.exe`.

This change should require no action on the part of the Docker build users. The other systems that I know people use don't bother supplying version information so this shouldn't be relevant to them anyway.


lahm86 (approved):
LGTM

walkawayy (approved):

PR #964, opened 2023-09-01
config: move enable_game_modes to config and add gameflow override

### LINKED ISSUE #962: Move enable_game_modes option from gameflow to config tool

walkawayy:
Resolves #962.

#### Checklist

#### Description
Moved the `enable_game_modes` option from the gameflow to the config tool.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 - fixed the camera in Natla's Mines when pulling the lever in room 67 (#352)
 - fixed flame emitter saving and loading which caused rare crashing (#947)
 - fixed new game plus not working if enable_game_modes was set to false (#960, regression from 2.8)
+- moved the enable_game_modes option from the gameflow to the config tool and added a gameflow option to override (#962)
 - improve spanish localization and added translation for rotated pickups
 
 ## [2.15.3](https://github.com/rr-/Tomb1Main/compare/2.15.2...2.15.3) - 2023-08-15
--- a/bin/cfg/Tomb1Main_gameflow.json5
+++ b/bin/cfg/Tomb1Main_gameflow.json5
@@ -10,8 +10,8 @@
     "savegame_fmt_bson": "save_tr1_%02d.dat",
 
     // whether to allow user to select NG/NG+/etc. modes in the new game dialog
-    // overrides the config file option enable_game_modes
-    "override_game_modes": false,
+    // disables the config option enable_game_modes until a playthrough is completed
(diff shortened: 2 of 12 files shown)


walkawayy:
Requesting @ajtudela for a translation please. :)

Also for the future:

"Enable save crystals"
"Description": "Limits saving to the beginning of levels save crystals. Levels have limited, single use save crystals like the PS1 version. Changing this option will require restarting the level."

rr-:
This is a breaking change; level builders now can't turn this feature off.

IMO it should be present in both config and gameflow so that the level can choose if they want to support this feature and the user can choose if they can go with that.

ajtudela:
> Requesting @ajtudela for a translation please. :)
> 
> Also for the future:
> 
> "Enable save crystals" "Description": "Limits saving to the beginning of levels save crystals. Levels have limited, single use save crystals like the PS1 version. Changing this option will require restarting the level."

Done!

lahm86 (commented):
Looks good, the only thing is that if `override_game_modes` is missing, the script doesn't load. Obviously not an issue for new installs, but should we make it default to `false` for any older gameflows out there (TRLEs mainly I suppose)? Or should it be the standard going forward that it must be included? I appreciate that's a bit of a breakaway from the other settings loaded in the gameflow, just though I'd mention it.

walkawayy:
> Looks good, the only thing is that if `override_game_modes` is missing, the script doesn't load. Obviously not an issue for new installs, but should we make it default to `false` for any older gameflows out there (TRLEs mainly I suppose)? Or should it be the standard going forward that it must be included? I appreciate that's a bit of a breakaway from the other settings loaded in the gameflow, just though I'd mention it.

Ya I realized that but tried to match all other gameflow options. I think it's ok to default to `false` though since it's not a required option. I'll change that.

lahm86 (approved):

rr- (commented):
On Discord we mentioned that setting the override property to false would disable the game modes and it would be a null value or an absence of this property altogether that would have no effect on the menus. With the current implementation I think it's better to call it `force_disable_gamemodes`, or alternatively implement the above approach. WDYT?

walkawayy:
Yes good point. The name is a little clunky. I was hesitant to use disable because the new game plus menu options (in a future PR) will be available. So it's only for the first playthrough. I'll push up a change with a rename to `disable_game_modes` and the comment in the json should be enough to explain that it works only for the first playthrough.

rr- (approved):

PR #965, opened 2023-09-01
config: move enable_save_crystals to config; add override

### LINKED ISSUE #963: Move enable_save_crystals option from gameflow to config

walkawayy:
#### Checklist

#### Description
Moved the enable_save_crystals option from the gameflow to the config tool. Added a gameflow option to force a mode for custom levels. Resolves #963. In a future PR, I will rename `disable_game_modes` to follow this naming convent (`enable_...`) and do an `&` with the gameflow and config option.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 - fixed flame emitter saving and loading which caused rare crashing (#947)
 - fixed new game plus not working if enable_game_modes was set to false (#960, regression from 2.8)
 - moved the enable_game_modes option from the gameflow to the config tool and added a gameflow option to override (#962)
+- moved the enable_save_crystals option from the gameflow to the config tool (#962)
 - improve spanish localization and added translation for rotated pickups
 
 ## [2.15.3](https://github.com/rr-/Tomb1Main/compare/2.15.2...2.15.3) - 2023-08-15
--- a/bin/cfg/Tomb1Main_gameflow.json5
+++ b/bin/cfg/Tomb1Main_gameflow.json5
@@ -15,7 +15,7 @@
 
     // forces whether save crystals are enabled or not
     // overrides the config option enable_save_crystals
-    "enable_save_crystals_option": true,
+    "force_enable_save_crystals": false,
 
     // seconds to pass in the main menu before playing the demo
     "demo_delay": 16,
--- a/bin/cfg/Tomb1Main_gameflow.json5
+++ b/bin/cfg/Tomb1Main_gameflow.json5
@@ -13,8 +13,9 @@
     // disables the config option enable_game_modes until a playthrough is completed
     "disable_game_modes": false,
 
-    // disables saving except in the locations near the crystals.
-    "enable_save_crystals": false,
+    // forces whether save crystals are enabled or not
+    // overrides the config option enable_save_crystals
+    "enable_save_crystals_option": true,
 
     // seconds to pass in the main menu before playing the demo
     "demo_delay": 16,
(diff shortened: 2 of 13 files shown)


rr- (commented):
This is a breaking change; level builders now can't turn this feature off. Furthermore it makes little sense to have crystals mode in levels that do not support it.

IMO it should be present in both config and gameflow so that the level can choose if they want to support this feature and the user can choose if they can go with that.

rr- (commented):
We should avoid manipulating other objects (here: the config) inside of a function that's about loading the gameflow. It's acceptable to do it after ithe internals finish loading in the entrypoint function where it's much more visible. 

There's no way to force save crystals which is guaranteed to be requested by the builders. I think we should employ the tristate idiom here:

| User config | Gameflow settings | Outcome |
| --- | --- | --- |
| `false` | `null` or missing | Crystals off |
| `false` | `false` | Crystals off |
| `false` | `true` | Crystals on |
| `true` | `null` or missing | Crystals on |
| `true` | `false` | Crystals off |
| `true` | `true` | Crystals on |

walkawayy:
I can move the config manipulation to another function. Do you have a location in mind or should I propose a spot?

Also this gameflow setting does indeed force crystals if true. So it behaves like your table.

rr-:
> Do you have a location in mind or should I propose a spot?

For this particular case, the entrypoint function, i.e. the outermost public-facing function, is `GameFlow_LoadFromFile`. Given its involvement in global string manipulation, it appears to be a suitable choice for manipulating `g_Config`.

>Also this gameflow setting does indeed force crystals if true. So it behaves like your table.

I have some concerns regarding this – the table I proposed distinguishes between a `false` value and a `null` or missing value. Moreover, it doesn't seem to directly offer overriding specific states. So, given the current code,

```c
    g_GameFlow.enable_save_crystals_option =
￼        json_object_get_bool(obj, "enable_save_crystals_option", false);
￼    g_Config.enable_save_crystals &= g_GameFlow.enable_save_crystals_option;
```

here's how I understand the state table looks:

| User setting | Gameflow setting  | Current outcome | Desired outcome |
| ---          | ---               | ---             | ---             |
| `false`      | `null` or missing | `false`         | `false`         |
| `false`      | `false`           | `false`         | `false`         |
| `false`      | `true`            | `false`         | `true`          |
| `true`       | `null` or missing | `false`         | `true`          |
| `true`       | `false`           | `false`         | `false`         |
| `true`       | `true`            | `true`          | `true`          |

This is particularly problematic that for custom levels that do not supply this setting, it is not possible to play with save crystals enabled. Could you please confirm if that's the case?

walkawayy:
Oh I misinterpreted. I see what you would like and updated the PR to reflect that.

I moved the check to `GameFlow_LoadFromFile`. I made it so the gameflow option `force_enable_save_crystals` will force save crystals when set to true. In any other case of missing or false, it leaves it up to the user's preference. This makes the most sense to me IMO. The builder can force crystals if needed, otherwise the user chooses. How it behaves in my newly pushed PR:

| User setting | Gameflow setting  | Current outcome | Desired outcome |
| ---           | ---                      | ---        | ---        |
| `false`      | `null` or missing | `false`   | `false`   |
| `false`      | `false`                 | `false`   | `false`   |
| `false`      | `true`                  | `true`    | `true`    |
| `true`       | `null` or missing | `true`    | `true`    |
| `true`       | `false`                 | `true`    | `true`   |
| `true`       | `true`                  | `true`    | `true`    |

rr- (approved):
LGTM

lahm86 (approved):

PR #956, opened 2023-08-30
savegame: fix flame emitter item data save and load

rr-:
Resolves #947

#### Checklist

- [x] I have read the [coding conventions](https://github.com/rr-/Tomb1Main/blob/master/CONTRIBUTING.md#coding-conventions)
- [x] I have added a changelog entry about what my pull request accomplishes, or it is an internal change

#### Description

Alternative implementation of #952. Rather than trying to rely on the order of loaded items, we reuse the calculated FX IDs and adjust `item->data` as necessary.

PR #970, opened 2023-09-05
savegame: prevent loading legacy FX data

### LINKED ISSUE #968: Game freezes when jumping into water on Palace Midas level
The game freezes while only the ambient sound keeps playing. Happens every time when loading [the save](https://github.com/LostArtefacts/Tomb1Main/files/12508154/save_tr1_24.zip) and jumping into the pool below.  

I did a fresh new install (Windows release 2.15.3, GOG version, all default settings), copied the save file and the same issue occurred. 

Nothing changes in the log file before and after the freeze.

location

lahm86:
Resolves #968.

#### Checklist

#### Description

Converts the check for loading FX to only do so on version 3 or greater, so to avoid the issue of flames doubling up. Future saves will be safe as per the fix in PR #956. No changelog entry added as covered by said PR.


rr- (approved):

walkawayy (approved):
Thanks! Lgtm.

PR #972, opened 2023-09-07
natla: move facing property to item scope

### LINKED ISSUE #906: OG bug: Natla rotates uncontrollably when more than one is active
Although not an issue in OG Great Pyramid, if more than one Natla is active and one goes into her semi-death state, she can spin on the floor as shown below. In this video there is another active Natla near the beginning of the level in the flying state.

https://github.com/LostArtefacts/Tomb1Main/assets/33758420/2695930b-a247-41f8-b093-55571bb8e132

This also affects her behaviour when she wakes up again, where she doesn't enter the normal standing animation to shoot at Lara and instead spins and moves constantly regardless of Lara's position.

lahm86:
Resolves #906.

#### Checklist

#### Description

Introduced a `priv` property on `ITEM_INFO`, and we make use of this to move Natla's `facing` property from module to item scope. This prevents multiple Natlas simultaneously changing the same variable, which was leading to the spinning effect in her semi-death and second phases.

Demo of the new state: https://www.youtube.com/watch?v=VGCbKsrCBBI

### DIFF
--- a/src/game/objects/creatures/natla.c
+++ b/src/game/objects/creatures/natla.c
@@ -14,6 +14,7 @@
 #include "math/math.h"
 
 #include <stdbool.h>
+#include <stddef.h>
 
 #define NATLA_SHOT_DAMAGE 100
 #define NATLA_NEAR_DEATH 200
@@ -42,7 +43,6 @@ typedef enum {
     NATLA_DEATH = 9,
 } NATLA_ANIM;
 
-static int16_t m_Facing = 0;
 static BITE_INFO m_NatlaGun = { 5, 220, 7, 4 };
 
 void Natla_Setup(OBJECT_INFO *obj)
@@ -82,6 +82,7 @@ void Natla_Control(int16_t item_num)
     int16_t tilt = 0;
     int16_t gun = natla->head_rotation * 7 / 8;
     int16_t timer = natla->flags & NATLA_TIMER;
+    int16_t facing = (int16_t)(size_t)item->priv;
 
     if (item->hit_points <= 0 && item->hit_points > DONT_TARGET) {
         item->goal_anim_state = NATLA_DEATH;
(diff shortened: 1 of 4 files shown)


walkawayy (approved):
LGTM. Does this need to be added to the savegame? On save/load Natla recalculates her `facing` instantly, so I don't think it's needed. Just wanted to ask for future proofing.

lahm86:
> LGTM. Does this need to be added to the savegame? On save/load Natla recalculates her `facing` instantly, so I don't think it's needed. Just wanted to ask for future proofing.

Yes, that crossed my mind but it isn't currently saved so I'm thinking we leave it to be re-calculated on-the-fly as you say. But yes, if `priv` becomes used elsewhere, it may be necessary to save/load it. Perhaps we should just do that now. What do you guys think?

Edit: you wouldn't be able to save it in a generic way come to think of it, it would need to be handled similar to the flame fx number i.e. specific handling for each type. So probably best not to do anything until we find that it's actually needed on other object types.

src/game/objects/creatures/natla.c
```diff
@@ -82,6 +82,7 @@ void Natla_Control(int16_t item_num)
     int16_t tilt = 0;
     int16_t gun = natla->head_rotation * 7 / 8;
     int16_t timer = natla->flags & NATLA_TIMER;
+    int16_t facing = (int16_t)(size_t)item->priv;
```
rr-, line 85: Do we need this double cast, here and in other places?

rr- (approved):

src/game/objects/creatures/natla.c
lahm86: Without it, the following compiler warning is generated. I think if it were an `int32_t` we would be OK, but as it's an angular property I believe the range needs to be limited to `int16_t`.

```
[2/9] Compiling C object Tomb1Main.exe.p/src_game_objects_creatures_natla.c.obj
../src/game/objects/creatures/natla.c: In function ‘Natla_Control’:
../src/game/objects/creatures/natla.c:85:22: warning: cast from pointer to integer of different size [-Wpointer-to-int-cast]
   85 |     int16_t facing = (int16_t)item->priv;
```

rr-:
>Edit: you wouldn't be able to save it in a generic way come to think of it, it would need to be handled similar to the flame fx number i.e. specific handling for each type. So probably best not to do anything until we find that it's actually needed on other object types.

That's exactly what I was thinking. Let's not add it just yet.

PR #973, opened 2023-09-07
config: add new game plus unlock flag to the config

### LINKED ISSUE #966: Rework new game+ bonus flag to persist
Players may start a new game after beating the game and not want new game+ mode on. So, the new game menu should be added after the game is completed once if `enable_game_modes` is false.

A new `bonus_unlock` flag for new game+ should also be added to the savegame. That way, once the player has beaten the game at least once, the game will remember between sessions and allow the user to pick from the new game menu whether to use new game+ or not.

walkawayy:
Resolves #966.

#### Checklist

#### Description
Added a flag indicating if new game plus is unlocked to the player config which allows the player to select new game plus or not when making a new game even if the game modes option is disabled. The flag will be set if a save is loaded with the `bonus_flag` in order to set the flag for older saves. The only bug I noticed was the flag not saving to the config if the game was force exited, so I added `Config_Write();` to `Game_Stop` when the last level is completed. Let me know if you notice any other bugs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - added the triggered music tracks to the savegame so one shot tracks don't replay on load (#371)
 - added forward/backward input detection in line with TR2+ for jump-twists (#931)
 - added an option to restore the mummy in City of Khamoon room 25, similar to the PS1 version (#886)
+- added a flag indicating if new game plus is unlocked to the player config which allows the player to select new game plus or not when making a new game (#966)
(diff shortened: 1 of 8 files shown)


rr- (commented):
Come to think of it, the name `runtime` that I suggested in Discord, is the opposite of what the settings are about, eg. persisting the state across game reboots. Can we use some other name?

lahm86 (approved):

rr- (approved):

PR #974, opened 2023-09-09
gfx: add an option to render using framebuffers

### LINKED ISSUE #114: Suggestion: Change Renderer Behaviour

rr-:
#### Checklist

#### Description

Resolves #114.

Add a new rendering mode that uses framebuffers and lets the player run the game at truly lower resolutions (which then get upscaled to window size, but this is fast).
For now the framebuffer upscaling filter is tied to the game filter, however, I'd like to introduce a separate setting for this in a follow-up pull request.

Testing this should involve changing the resolutions, changing the render mode, checking how this behaves with FMVs and pictures, and finally, screenshots. Screenshots made with render mode set to framebuffer should have the image size equal to the chosen resolution.


lahm86 (commented):
Works great for me. I tried changing between each mode and resolution and didn't see any issues. FMVs and screenshots also working/generating as expected.
One thing I did notice is that a lot of errors are being logged from `sampler.c` line 27 (log attached). I appreciate this may not be directly related, but thought it worth mentioning as the functions here now check for errors.
[Tomb1Main.log.txt](https://github.com/LostArtefacts/Tomb1Main/files/12565557/Tomb1Main.log.txt)

walkawayy:
Works very well! Two minor things I noticed:

Flickering artifacts when switching resolutions in frame buffer mode (I don't think users will be changing this often, but wanted to mention) Also noticed after uploading maybe streaming the resolution change messes with OBS. But when you flip through resolutions you can notice text box artifacts and stuff:

https://github.com/LostArtefacts/Tomb1Main/assets/81546780/b0cb5dde-2f85-45ce-87a8-4f5059b4e286



Text of `Framebuffer` is overlapped with the arrow:
20230909_093542_Caves

rr-:
I removed the errors in the logs as well as improved the UI. I'm really not sure how to fix the visual flicker, but IMO it's acceptable at this point.

lahm86 (approved):
LGTM

walkawayy (approved):
Ya I think the flickering is fine, I just wanted to mention it for awareness. Better than requiring a game reboot like some games lol.

PR #976, opened 2023-09-09
gameflow: add bonus level type

### LINKED ISSUE #645: Bonus level option
Just like in TR2Gold/TR3, if player collects all secrets, the bonus level gets unlocked at the end of the game.
Would be a really nice option for level builders.

walkawayy:
Resolves #645.

#### Checklist

#### Description
Added the bonus level type for custom levels that unlocks if all main game secrets are found.


rr-:
I'd wait with this feature till 2.17.

walkawayy:
Removed the exit to bonus level from the gameflow @lahm86. This won't be in the next release so obvi no rush, but I know you mentioned wanting to play around with this. So just FYI.

I think a current limitation may be that you can't have an FMV in between multiple levels.

lahm86:
I see what you mean about the FMVs - I've attached an updated test gameflow.

[Tomb1Main_gameflow.zip](https://github.com/LostArtefacts/Tomb1Main/files/12589324/Tomb1Main_gameflow.zip)

If you don't collect all secrets, after the total stats screen, the FMVs will play whereas they shouldn't really as they're part of the bonus setup.

Would it be an idea to move the logic into the `GFS_EXIT_TO_LEVEL` case instead? That way the bonus levels are skipped over entirely.

Otherwise, it looks great 👍

lahm86:
Works great now, thanks 👍

src/game/gameflow.c
```diff
@@ -1273,6 +1293,11 @@ GameFlow_InterpretSequence(int32_t level_num, GAMEFLOW_LEVEL_TYPE level_type)
             return GF_EXIT_TO_TITLE;
 
         case GFS_EXIT_TO_LEVEL:
+            if (((int32_t)seq->data & ((1 << 6) - 1))
+                    >= g_GameFlow.first_bonus_num
+                && !g_GameInfo.bonus_level_unlock) {
+                return GF_EXIT_TO_TITLE;
+            }
             return GF_START_GAME
```
lahm86: Maybe we can do `int32_t next_level = (int32_t)(intptr_t)seq->data & ((1 << 6) - 1);` just once and use it for the condition and the later return statement.

src/game/savegame/savegame_bson.c
```diff
@@ -370,14 +371,20 @@ static bool Savegame_BSON_LoadDiscontinuedEndInfo(
...
+    struct json_object_s *misc_obj, GAME_INFO *game_info,
+    uint16_t header_version)
 {
     assert(game_info);
     if (!misc_obj) {
         LOG_ERROR("Malformed save: invalid or missing misc info");
         return false;
     }
     game_info->bonus_flag = json_object_get_int(misc_obj, "bonus_flag", 0);
+    if (header_version >= VERSION_3) {
```
lahm86: Should this be upped to V4 now as V3 is for 2.16?

src/game/stats.c
```diff
@@ -382,7 +382,7 @@ void Stats_Show(int32_t level_num)
     g_GameInfo.status &= ~GMS_IN_STATS;
 }
 
-void Stats_ShowTotal(const char *filename)
+void Stats_ShowTotal(const char *filename, int first_level, int last_level)
```
lahm86: Is it worth adding a way to show "Bonus Statistics" as the title here? So that'd be another string in the gameflow to allow builders to customize it. I know we discussed it on Discord, I'm just remembering about it now.

src/game/savegame/savegame_bson.c
```diff
@@ -370,14 +371,20 @@ static bool Savegame_BSON_LoadDiscontinuedEndInfo(
...
+        game_info->bonus_level_unlock =
+            json_object_get_bool(misc_obj, "bonus_level_unlock", 0);
+        LOG_DEBUG("Load bonus_level_unlock: %d", game_info->bonus_level_unlock);
```
lahm86: Logging perhaps no longer needed.

lahm86 (commented):
It's all looking great to me - just a few minor comments.

src/game/gameflow.c
walkawayy: Ya good point.

Also I'm sure there's some reason it's done this way, but what's the difference between `seq->data & ((1 << 6) - 1)` and `level_num` that gets passed into `GameFlow_InterpretSequence`?

walkawayy: Found a big bug while testing this. For gameflows without a bonus level, `g_GameFlow.first_bonus_num` is initialized to `-1`. So the game always returns to title after the first level. It also breaks the level type, etc. I added a `g_GameFlow.has_bonus`.

src/game/gameflow.c
```diff
@@ -1223,7 +1239,16 @@ GameFlow_InterpretSequence(int32_t level_num, GAMEFLOW_LEVEL_TYPE level_type)
         case GFS_TOTAL_STATS:
             if (g_Config.enable_total_stats && level_type != GFL_SAVED) {
                 GAMEFLOW_DISPLAY_PICTURE_DATA *data = seq->data;
-                Stats_ShowTotal(data->path);
+                LOG_DEBUG("level_type: %d", level_type);
```
lahm86: Extra debug message

lahm86 (approved):
LGTM, just one extra debug message since yesterday.

rr- (commented):
The modifications made in `Stats_ShowTotal` require the caller to take on additional responsibilities, such as selecting the start and end level numbers, as well as providing a heading. In order to simplify this, it would be better to replace the introduced parameters with just `GAMEFLOW_LEVEL_TYPE level_type`.

The changes overall appear to be inspired by the `first_level_num` and `last_level_num` idiom used in our code. However, it is important to note that these variables should only be used to retrieve the first and last level numbers, and not for direct comparison with arbitrary level numbers. This is because the order of levels should not be determined solely by their appearance in the JSON, but rather by sequences like `exit_to_level`. Simple `>=` and `<=` comparisons do not account for this. While there are instances in the code where we compare `level_num >= first_level_num` or `level_num <= last_level_num`, this is not something we should continue doing in the new code, as it reinforces a reliance on the JSON order.

Upon further examination, the checks for `first_bonus_num` and `last_bonus_num` are seemingly only meant to determine if a given `level_num` belongs to the bonus levels. It would be more efficient to check this using `g_GameFlow.levels[level_num].level_type == GFL_BONUS`. This allows us to eliminate the need for the `first_bonus_num`, `last_bonus_num` and even `has_bonus` variables entirely.

Considering these points, it would be more appropriate for `Savegame_PlayAvailableStory` to now disregard the bonus levels, rather than halting at the first encounter of a bonus level. In theory, this revision would allow the builder to place bonus levels at the beginning of the gameflow file, followed by the normal levels, without causing any issues.

Here is a proposed patch that has been implemented to address the feedback mentioned, but please note that I haven't really tested it.

```patch
diff --git a/src/game/gameflow.c b/src/game/gameflow.c
index f95e9330..8ca4762c 100644
--- a/src/game/gameflow.c
+++ b/src/game/gameflow.c
@@ -678,9 +678,6 @@ static bool GameFlow_LoadScriptLevels(struct json_object_s *obj)
     g_GameFlow.gym_level_num = -1;
     g_GameFlow.first_level_num = -1;
     g_GameFlow.last_level_num = -1;
-    g_GameFlow.has_bonus = false;
-    g_GameFlow.first_bonus_num = -1;
-    g_GameFlow.last_bonus_num = -1;
     g_GameFlow.title_level_num = -1;
     g_GameFlow.level_count = jlvl_arr->length;
 
@@ -746,11 +743,6 @@ static bool GameFlow_LoadScriptLevels(struct json_object_s *obj)
             g_GameFlow.last_level_num = level_num;
         } else if (!strcmp(tmp_s, "bonus")) {
             cur->level_type = GFL_BONUS;
-            if (g_GameFlow.first_bonus_num == -1) {
-                g_GameFlow.has_bonus = true;
-                g_GameFlow.first_bonus_num = level_num;
-            }
-            g_GameFlow.last_bonus_num = level_num;
         } else if (!strcmp(tmp_s, "cutscene")) {
             cur->level_type = GFL_CUTSCENE;
         } else if (!strcmp(tmp_s, "current")) {
@@ -1199,8 +1191,7 @@ GameFlow_InterpretSequence(int32_t level_num, GAMEFLOW_LEVEL_TYPE level_type)
                 return ret;
             }
             if (level_type == GFL_SAVED) {
-                if (g_GameFlow.has_bonus
-                    && level_num >= g_GameFlow.first_bonus_num) {
+                if (g_GameFlow.levels[level_num].level_type == GFL_BONUS) {
                     level_type = GFL_BONUS;
                 } else {
                     level_type = GFL_NORMAL;
@@ -1240,15 +1231,7 @@ GameFlow_InterpretSequence(int32_t level_num, GAMEFLOW_LEVEL_TYPE level_type)
             if (g_Config.enable_total_stats && level_type != GFL_SAVED) {
                 GAMEFLOW_DISPLAY_PICTURE_DATA *data = seq->data;
                 LOG_DEBUG("level_type: %d", level_type);
-                if (level_type == GFL_BONUS) {
-                    Stats_ShowTotal(
-                        data->path, g_GameFlow.first_bonus_num,
-                        g_GameFlow.last_bonus_num, GS_STATS_BONUS_STATISTICS);
-                } else {
-                    Stats_ShowTotal(
-                        data->path, g_GameFlow.first_level_num,
-                        g_GameFlow.last_level_num, GS_STATS_FINAL_STATISTICS);
-                }
+                Stats_ShowTotal(data->path, level_type == GFL_BONUS ? GFL_BONUS : GFL_NORMAL);
             }
             break;
 
@@ -1300,7 +1283,7 @@ GameFlow_InterpretSequence(int32_t level_num, GAMEFLOW_LEVEL_TYPE level_type)
         case GFS_EXIT_TO_LEVEL: {
             int32_t next_level =
                 ((int32_t)(intptr_t)seq->data & ((1 << 6) - 1));
-            if (g_GameFlow.has_bonus && next_level >= g_GameFlow.first_bonus_num
+            if (g_GameFlow.levels[next_level].level_type == GFL_BONUS
                 && !g_GameInfo.bonus_level_unlock) {
                 return GF_EXIT_TO_TITLE;
             }
diff --git a/src/game/gameflow.h b/src/game/gameflow.h
index 65a3d9ad..2797008b 100644
--- a/src/game/gameflow.h
+++ b/src/game/gameflow.h
@@ -60,9 +60,6 @@ typedef struct GAMEFLOW {
     int32_t gym_level_num;
     int32_t first_level_num;
     int32_t last_level_num;
-    bool has_bonus;
-    int32_t first_bonus_num;
-    int32_t last_bonus_num;
     int32_t title_level_num;
     int32_t level_count;
     char *savegame_fmt_legacy;
diff --git a/src/game/savegame/savegame.c b/src/game/savegame/savegame.c
index 912f1d70..946e6427 100644
--- a/src/game/savegame/savegame.c
+++ b/src/game/savegame/savegame.c
@@ -644,10 +644,12 @@ GAMEFLOW_OPTION Savegame_PlayAvailableStory(int32_t slot_num)
 
         gf_option = GameFlow_StorySoFar(gf_param, savegame_info->level_num);
 
-        if ((gf_param >= savegame_info->level_num
-             && gf_param <= g_GameFlow.last_level_num)
-            || (gf_param >= savegame_info->level_num
-                && gf_param <= g_GameFlow.last_bonus_num)) {
+        if (g_GameFlow.levels[gf_param].level_type == GFL_BONUS) {
+            continue;
+        }
+
+        if (gf_param >= savegame_info->level_num
+                && gf_param <= g_GameFlow.last_level_num) {
             break;
         }
 
diff --git a/src/game/shell.c b/src/game/shell.c
index db2d3514..490bbe81 100644
--- a/src/game/shell.c
+++ b/src/game/shell.c
@@ -187,8 +187,7 @@ void Shell_Main(void)
 
         case GF_START_GAME: {
             GAMEFLOW_LEVEL_TYPE level_type = GFL_NORMAL;
-            if (g_GameFlow.has_bonus
-                && gf_param >= g_GameFlow.first_bonus_num) {
+            if (g_GameFlow.levels[gf_param].level_type == GFL_BONUS) {
                 level_type = GFL_BONUS;
             }
             gf_option = GameFlow_InterpretSequence(gf_param, level_type);
diff --git a/src/game/stats.c b/src/game/stats.c
index 1d08bc37..6f3b10c6 100644
--- a/src/game/stats.c
+++ b/src/game/stats.c
@@ -383,8 +383,7 @@ void Stats_Show(int32_t level_num)
 }
 
 void Stats_ShowTotal(
-    const char *filename, int first_level, int last_level,
-    GAME_STRING_ID heading)
+    const char *filename, GAMEFLOW_LEVEL_TYPE level_type)
 {
     char buf[100];
     char time_str[100];
@@ -401,7 +400,10 @@ void Stats_ShowTotal(
 
     int16_t secret_flags = 0;
 
-    for (int i = first_level; i <= last_level; i++) {
+    for (int i = 0; i < g_GameFlow.level_count; i++) {
+        if (g_GameFlow.levels[i].level_type != level_type) {
+            continue;
+        }
         const GAME_STATS *stats = &g_GameInfo.current[i].stats;
 
         total_timer += stats->timer;
@@ -505,7 +507,7 @@ void Stats_ShowTotal(
     Text_AddOutline(txt, true, TS_BACKGROUND);
 
     // heading
-    sprintf(buf, "%s", g_GameFlow.strings[heading]);
+    sprintf(buf, "%s", g_GameFlow.strings[level_type == GFL_BONUS ? GS_STATS_BONUS_STATISTICS : GS_STATS_FINAL_STATISTICS]);
     txt = Text_Create(0, top_y + 2, buf);
     Text_CentreH(txt, 1);
     Text_CentreV(txt, 1);
diff --git a/src/game/stats.h b/src/game/stats.h
index 7183061a..651ac3c4 100644
--- a/src/game/stats.h
+++ b/src/game/stats.h
@@ -11,6 +11,4 @@ int32_t Stats_GetPickups(void);
 int32_t Stats_GetKillables(void);
 int32_t Stats_GetSecrets(void);
 void Stats_Show(int32_t level_num);
-void Stats_ShowTotal(
-    const char *filename, int first_level, int last_level,
-    GAME_STRING_ID heading);
+void Stats_ShowTotal(const char *filename, GAMEFLOW_LEVEL_TYPE level_type);
```

walkawayy:
@lahm86 would you mind posting your test gameflow if you have one off hand? I accidentally deleted mine. Everything looks good from the normal TR1 playthrough.

lahm86:
> @lahm86 would you mind posting your test gameflow if you have one off hand? I accidentally deleted mine. Everything looks good from the normal TR1 playthrough.

This one should hopefully do it.
https://github.com/LostArtefacts/Tomb1Main/files/12589324/Tomb1Main_gameflow.zip

walkawayy:
I think it looks good except for story so far with a bonus level gets stuck in an infinite loop of the Gym FMV lol.

walkawayy:
I don't think this is needed in `Savegame_PlayAvailableStory`?
```
        if (gf_direction == GF_EXIT_GAME) {
            break;
        }
```

src/game/gameflow.c
```diff
@@ -1272,9 +1281,15 @@ GameFlow_InterpretSequence(int32_t level_num, GAMEFLOW_LEVEL_TYPE level_type)
         case GFS_EXIT_TO_TITLE:
             return GF_EXIT_TO_TITLE;
 
-        case GFS_EXIT_TO_LEVEL:
-            return GF_START_GAME
-                | ((int32_t)(intptr_t)seq->data & ((1 << 6) - 1));
+        case GFS_EXIT_TO_LEVEL: {
+            int32_t next_level =
+                ((int32_t)(intptr_t)seq->data & ((1 << 6) - 1));
```
rr-: Unnecessary outer braces

src/game/stats.c
```diff
@@ -419,6 +421,10 @@ void Stats_ShowTotal(const char *filename)
         total_max_pickup_count += stats->max_pickup_count;
     }
 
+    // Check for bonus level unlock.
+    g_GameInfo.bonus_level_unlock =
+        total_secret_count >= total_max_secret_count;
```
rr-: Let's try to avoid mutating global state in functions like `Show`, `Get` etc.
To be able to reuse `total_Secret_count`, consider making a function called `Stats_Compute` that fills a struct with game statistics, and calling it twice, once in `Stats_Show`, and once to update the `g_GameInfo`. The computational cost is low, and I believe the maintenance becomes easier.

src/game/game/game.c
```diff
@@ -207,7 +216,9 @@ int32_t Game_Stop(void)
     if (g_CurrentLevel == g_GameFlow.last_level_num) {
         g_Config.profile.new_game_plus_unlock = true;
         Config_Write();
-    } else {
+    }
+
+    if (g_CurrentLevel < g_GameFlow.level_count) {
```
rr-: I understand that this got removed to carry over save information to the bonus levels, and this works because the bonus levels usually happen to be placed after the normal levels.

What happens if the game doesn't have any bonus levels though and we're on the last level? I think there will be a memory corruption and to avoid it, this should become `g_CurrentLevel + 1 < g_GameFlow.level_count`.

src/game/savegame/savegame.c
```diff
@@ -638,19 +638,14 @@ GAMEFLOW_OPTION Savegame_PlayAvailableStory(int32_t slot_num)
...
 
-        if (gf_param >= savegame_info->level_num
-            && gf_param <= g_GameFlow.last_level_num) {
-            break;
-        }
-
-        if (gf_direction == GF_EXIT_GAME) {
+        if ((g_GameFlow.levels[gf_param].level_type == GFL_NORMAL
+             || g_GameFlow.levels[gf_param].level_type == GFL_BONUS)
+            && gf_param >= savegame_info->level_num) {
```
rr-, line 648: Let's test it well to make sure this doesn't loop indefinitely especially when on the last level that has nothing after.

src/game/game/game.c
```diff
@@ -207,12 +216,12 @@ int32_t Game_Stop(void)
     if (g_CurrentLevel == g_GameFlow.last_level_num) {
         g_Config.profile.new_game_plus_unlock = true;
         Config_Write();
-    } else {
-        Savegame_CarryCurrentInfoToNextLevel(
-            g_CurrentLevel, g_CurrentLevel + 1);
-        Savegame_ApplyLogicToCurrentInfo(g_CurrentLevel + 1);
     }
 
+    Savegame_CarryCurrentInfoToNextLevel(
```
rr-: I understand that this `else` was removed to have the game carry over Lara's inventory and stats to the bonus levels as well, which usually happen to be placed after the normal levels.

What will happen, however, if the game has no bonus levels, and we've just ended the last level? I think it will create some memory corruption trying to carry the stats over to a non-existing level. Neither `Savegame_ApplyLogicToCurrentInfo` nor `Savegame_CarryCurrentInfoToNextLevel` check if the target level number is valid, so we either should make them more resilient, or convert this `else` here to `if (g_CurrentLevel + 1 < g_GameFlow.level_count)`.

src/game/gameflow.c
walkawayy: Needed for `next_level`.

```
../../src/game/gameflow.c:1285:13: error: a label can only be part of a statement and a declaration is not a statement
```

src/game/stats.c
walkawayy: Good ideas. Lmk what you think of what I pushed.

src/game/savegame/savegame.c
walkawayy: Seems to behave properly for Lahm's secret gameflow, TR1, and UB.

src/game/game/game.c
walkawayy: Pushed your second suggestion.

src/game/gameflow.c
rr-: I meant the round braces :D The ones around the value of variable declaration.

src/game/stats.h
```diff
@@ -11,4 +12,6 @@ int32_t Stats_GetPickups(void);
 int32_t Stats_GetKillables(void);
 int32_t Stats_GetSecrets(void);
 void Stats_Show(int32_t level_num);
-void Stats_ShowTotal(const char *filename);
+void Stats_ShowTotal(const char *filename, GAMEFLOW_LEVEL_TYPE level_type);
+void Stats_ComputeTotal(GAMEFLOW_LEVEL_TYPE level_type);
+bool Stats_CheckAllSecretsCollected();
```
rr-: Let's add `(void)` to make it not possible to call the function with arbitrary arguments.

src/game/stats.c
```diff
@@ -537,3 +525,38 @@ void Stats_ShowTotal(const char *filename)
...
+
+        m_TotalStats.timer += stats->timer;
+        m_TotalStats.death_count += stats->death_count;
+        m_TotalStats.total_kill_count += stats->max_kill_count;
+        m_TotalStats.total_secret_count += stats->max_secret_count;
+        m_TotalStats.total_pickup_count += stats->max_pickup_count;
+    }
+}
+
+bool Stats_CheckAllSecretsCollected()
```
rr-: Let's add `(void)` to make it not possible to call the function with arbitrary arguments.

src/game/stats.c
```diff
@@ -537,3 +525,38 @@ void Stats_ShowTotal(const char *filename)
...
+            secret_flags >>= 1;
+        }
+
+        m_TotalStats.timer += stats->timer;
+        m_TotalStats.death_count += stats->death_count;
+        m_TotalStats.total_kill_count += stats->max_kill_count;
+        m_TotalStats.total_secret_count += stats->max_secret_count;
+        m_TotalStats.total_pickup_count += stats->max_pickup_count;
+    }
+}
```
rr-: While I had something else on my mind (fill a struct passed as a function argument, or maybe even just return a struct), I really like this solution and I think it's even cleaner that my initial idea. Well done.

The only thing that crosses my mind is that the caller of `Stats_CheckAllSecretsCollected` needs to be aware to call `Stats_ComputeTotals(GFL_NORMAL)` first for it to work properly. We could alleviate this by making `Stats_ComputeTotal` private and calling `Stats_ComputeTotals` directly in `Stats_CheckAllSecretsCollected`.

src/game/gameflow.c
walkawayy: Oh lol so parenthesis. 😆 I mix them up all the time too
() parenthesis.
[] brackets.
{} braces.

src/game/stats.c
walkawayy: The reason I separated them is because `Stats_ShowTotal` needs to call `Stats_ComputeTotal`. And `Stats_ShowTotal` needs to work for both all the normal levels and all the bonus levels. And we don't necessarily want `Stats_CheckAllSecretsCollected` to be recalculated when showing total bonus level stats. I get what you are saying though, so lmk if you want it changed.

rr-: Yeah, let's change it if possible, that way the function are not dependent on the order of calls.

lahm86 (approved):
LGTM. I've been testing it out today and it all seems good; tried setting the gameflow up in different orders (from what I'd call standard to obscure) and everything carries over and flows perfectly. Story-so-far checks out, and save state seems good. Really nice feature 👍

rr- (approved):
LGTM :heart:

PR #978, opened 2023-09-10
lara: fix responsive jumping delay

### LINKED ISSUE #975: T1M bug: Lara jumps late with TR2 jumping enabled
With TR2 jumping enabled, Lara jumps later compared with normal TR1 jumping - this applies only to when entering the running animation from the run_start animation.
Reported by Sabatu on Discord: https://discord.com/channels/1071085441077481544/1075372522377322516/1149665135854616700

lahm86:
Resolves #975.

#### Checklist

#### Description

In normal TR1 jumping, Lara can jump on frame 3 of the running animation (animation 0), so this has now been restored rather than using frame 4 from TR2+. The problem was coupled with another issue - the frame number check in `Lara_State_Run` did not take into account the fact that the frame number is _about_ to be incremented in `lara_control.c`. Without the `+1` added in this fix, we were still testing the previous frame and so as a result, we were actually 2 frames out.

So the control order is:

* `Lara_HandleAboveWater`
  * `g_LaraStateRoutines[item->current_anim_state](item, coll);` ( => `Lara_State_Run`)
  * `Lara_Animate(item);`
    * `item->frame_number++;`

This shows the old responsive jumping state - Lara doesn't switch to the jumping animation until frame 4 of the running animation.
https://www.youtube.com/watch?v=WVIv_H74M4c

As well as the jump in Sabatu's report, this also affected some jumps in the original levels. Below is a before and after from Natla's Mines; a similar jump in The Great Pyramid was also affected.
https://www.youtube.com/watch?v=ZB0lPsy-zBA

And finally, here is a comparison video of the fix in detail showing that we now have like-for-like. Original TR1 jumping is on the left and responsive jumping on the right.
https://www.youtube.com/watch?v=DyowWoJr6YY

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@
 - fixed the incorrect starting animation on centaurs that spawn from statues (#926, regression from 2.15)
 - fixed jump-twist animations at times being interrupted (#932, regression from 2.15.1)
 - fixed walk-run-jump at times breaking when TR2 jumping is enabled (OG bug in TR2+) (#934)
-- fixed Lara jumping late with TR2 jumping enabled, as compared to normal TR1 jumping and when entering the run animation initially (#975)
+- fixed Lara jumping late with TR2 jumping enabled, as compared to normal TR1 jumping when entering the run animation initially (#975)
 - fixed the reset and unbind progress bars in the controls menu for non-default bar scaling (#930)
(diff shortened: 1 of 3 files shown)


rr- (approved):

walkawayy (approved):
Great videos. I appreciate them even though they were probably hell to make lol.

PR #981, opened 2023-09-10
shell: improve support for windowed mode

### LINKED ISSUE #896: Feature: windowed mode option
Hey,
Currently it supports windowed mode but I'd like some improvements:
- option in the settings (just like in original ATI version)
- or to launch in the chosen resolution (not always in 720)
- or at least when being manually stretched to maintain the AR, and also update the resolution in real time on the screen (like TR2) would be great.

For people, who stream on a single display, windowed mode is a must.
Regards!

rr-:
Resolves #896.

#### Checklist

#### Description

Fullscreen state, window size and position are now all saved across reboots. Window size and position are not updated when the game is in fullscreen state, which is intended to avoid window size resetting to desktop size upon alt-enter.

To avoid weird out-of-order config read/write sequences, I add a `loaded` flag to `g_Config` to ensure shell can only update it once it's done loading.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
 - moved the enable_game_modes option from the gameflow to the config tool and added a gameflow option to override (#962)
 - moved the enable_save_crystals option from the gameflow to the config tool (#962)
 - improved Spanish localization for the config tool
+- improved support for windowed mode (#896)
 
 ## [2.15.3](https://github.com/rr-/Tomb1Main/compare/2.15.2...2.15.3) - 2023-08-15
 - fixed Lara stuttering when performing certain animations (#901, regression from 2.14)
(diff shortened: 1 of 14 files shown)


lahm86 (commented):
Looks great 👍 

I just noticed one minor issue. If you go into windowed mode, change the window size and position, and then hit the maximize window button, when you relaunch the game, the size is restored to the maximized size, but the position is restored to the point before maximizing. This can result in the window going partially off screen. 

Could we store and retrieve the maximized state too? And possibly only store the position if the window is not maximized, so on reload it would set it to the previous floating position, and then call `SDL_MaximizeWindow`.

rr-:
@lahm86 thanks for the tests; the maximization problems should be now fixed. Coincidentally this seems to fix the other problems as well.

walkawayy (approved):
LGTM. Worth a readme mention too?

rr-:
Added a note to the README.

lahm86 (approved):

PR #983, opened 2023-09-10
filesystem: fix heap corruption

rr-:
#### Checklist

- [x] I have read the [coding conventions](https://github.com/rr-/Tomb1Main/blob/master/CONTRIBUTING.md#coding-conventions)
- [x] I have added a changelog entry about what my pull request accomplishes, or it is an internal change

#### Description

This fixes heap corruption that manifested in crashes on Linux machines.

Issue #716, opened 2023-01-22
Cross-platform compatibility


walkawayy:
- [x] Replace wgl_ext.h with cross platform variant
- [x] Remove HWND and HINSTANCE usages
- [x] Rename JSON paths to use `/` instead of `\\` separators
- [x] Case insensitive directory and file detection
- [x] Determine how to package FFmpeg and SDL2
- [x] Test on Linux
- [x] Test on MacOS

---
Looking at the README's [road map](https://github.com/rr-/Tomb1Main#road-map), it seems like we are moving towards being able to support cross-platform builds. @carlo-bramini are both of the following tasks done once #714 merges?

```
Replace wgl_ext.h with cross platform variant
Remove HWND and HINSTANCE usages
```

Do you have a list of what else needs to be done that we could track in this issue? Also it seems like you've been focusing on Linux, but is MacOS compatibility a possibility too?

carlo-bramini:
About these points:

```
Replace wgl_ext.h with cross platform variant
Remove HWND and HINSTANCE usages
```

I have compiled and executed the engine also on linux, so it is technically able to work, although the .json5 files still need to be adjusted for fixing directory separators from `'\\'` to `'/'` and some decisions should be made since the filesystem is case sensitve on other platforms.

At the moment, I compiled and tested the engine with the demo level and some hacks with these platforms:
- i686, 32-bit executable, on Windows 11, Windows 10, Windows 7 and Windows XP.
- x86_64, 64 bit executable, on Windows 11, Windows 10 and Windows 7.
- x86_64, 64 bit executable, on Linux Debian.
- ARM64, 64 bit executable, on Linux Raspberry Pi OS (needs an hack to the shaders, see #685).

I compiled my builds with CYGWIN and MinGW cross compilers. However, CYGWIN does not provide FFmpeg and the current version of SDL2 is too old for compiling Tomb1Main, but here I made public the packages made by me for CYGWIN:

https://github.com/carlo-bramini/packages-cygwin

I also compiled Tomb1Main by using the POSIX layer of CYGWIN and I run the engine with the CYGWIN/X server, but actually it was just the development tool for running it on UNIX-like systems.

So, in my opinion, those two points seems to be resolved.

About MacOS, unfortunately I cannot test if it works because I have no way to run it.

walkawayy:
Ok cool. I think the goal should be to remove the need for any hacks if we can. (Except it seems the demo support will require some, but that's a whole different issue IMO.)

[...]

PR #984, opened 2023-09-10
Linux builds

rr-:
#### Checklist

#### Description

This adds statically built (except GL, which cannot be linked statically) Linux binary to the release toolchain, partially resolving #716. Not really tested just yet. Builds on top of #983. Shouldn't be released before 2.17.


walkawayy:
Can't test on Linux, but I don't notice any issues on Windows.

walkawayy (approved):

src/game/gameflow.c
```diff
@@ -362,27 +362,27 @@ static bool GameFlow_LoadLevelSequence(
 
         if (!strcmp(type_str, "start_game")) {
             seq->type = GFS_START_GAME;
-            seq->data = (void *)level_num;
+            seq->data = (void *)(intptr_t)level_num;
```
walkawayy: Is there any purpose in using `intptr_t` instead of `uintptr_t`? I was reading about it and the one use case seemed to be for arithmetic right shift. Also do you need `(void *)` still?

rr-: It's solely to fix the warnings on Linux gcc: `cast to pointer from integer of different size [-Wint-to-pointer-cast]`.

walkawayy: Ok. I just meant can we remove `(void *)`? Seems like `intptr_t` is not in pre C99 but idk if that is an issue.

rr-: Removing the second cast to `void*` results in another error:
```
warning: assignment to ‘void *’ from ‘long int’ makes pointer from integer without a cast [-Wint-conversion]
```

lahm86 (approved):

PR #987, opened 2023-09-19
overlay: add 3d pickups animation

### LINKED ISSUE #257: Feature: Use 3D objects in overlay when 3D pickups are enabled
Currently, Tomb1Main draws pickups in the overlay as sprites even when 3D pickups are enabled. Since the 3D objects are properly upscaled by the engine, the sprites don't look as good as the 3D objects.

Optionally, the pickup objects should rotate in the overlay like they do in more recent engine versions.

In case using the 3D objects for the overlays is not feasible, the 3D objects that are used as pickups could be rendered once (for example with a slight rotation) and cached, and used for drawing the pickup sprite instead.

rr-:
#### Checklist

#### Description

Resolves #257

### DIFF
--- a/src/game/overlay.c
+++ b/src/game/overlay.c
@@ -2,6 +2,8 @@
 
 #include "config.h"
 #include "game/clock.h"
+#include "game/inventory.h"
+#include "game/objects/common.h"
 #include "game/output.h"
 #include "game/screen.h"
 #include "game/text.h"
@@ -9,6 +11,7 @@
 #include "global/const.h"
 #include "global/types.h"
 #include "global/vars.h"
+#include "src/math/matrix.h"
 #include "util.h"
 
 #include <stdbool.h>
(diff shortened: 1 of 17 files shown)


walkawayy:
I figured out why I was sometimes getting flickering pickups around the bottom of my screen that never expire. For some reason if you use the turbo speed cheat, it causes the pickups to break. So it's not really a normal game bug since I haven't seen it without using the turbo cheat, but I wanted to mention it.

walkawayy (approved):
Everything looks good now to me!

src/game/overlay.c
```diff
@@ -281,10 +307,207 @@ static void Overlay_OnFPSTextRemoval(const TEXTSTRING *textstring)
...
+{
+    float ratio = cur_frame / (float)max_frames;
+    if (ratio < 0.5f) {
+        return 2.0f * ratio * ratio;
+    }
+    float new_ratio = ratio - 1.0f;
+    return 1.0f - 2.0f * new_ratio * new_ratio;
+}
+
+static void Overlay_DrawPickup3D(DISPLAY_PICKUP_INFO *pu)
```
walkawayy, line 320: The equivalent function `Output_DrawUISprite` is in `output.c`. Should me move that function into overlay or move this into output?

lahm86 (commented):
This is really nice :)
Just one problem - if you pick something up and then quickly cross the end level trigger while the pickup animation is still playing out, the screen goes black. This also affects picking up the scion in Sanctuary, where the level ends during the Lara misc. animation. Is it worth freezing the display animation on level end, or perhaps removing it?

src/game/overlay.c
rr-: In my opinion both should actually stay where they are; Output_DrawUISprite exists as a proxy for S_Output_DrawSprite (which shouldn't be called anywhere outside output.c). The equivalent to this in the 3d space is Output_DrawPolygons which already resides in output.c.

rr-:
@lahm86 this should be now fixed.

lahm86 (approved):

PR #994, opened 2023-09-21
lara: enable rolling underwater

### LINKED ISSUE #993: Add 180° Underwater Turn
Just like in TR2+ and already added for the jumps. It would be nice to let lara roll Underwater to turn around

lahm86:
Resolves #993.

#### Checklist

#### Description

This adds the underwater rolling animation to Lara's enhanced animation injection set, and we control it with the config option through the state routines. Similar to jump-twists, this means that if you save mid-underwater-roll, switch off the option, and then reload, Lara will complete the animation gracefully but won't perform any more until the option is re-enabled.

The addition to `FX_Turn180` was needed to ensure she is facing the right way post-roll. This routine is only used by Lara across the OG levels in her on-land roll and jump-twist animations - so here `x_rot` would always be zero anyway.

In the later games, the state routines switch immediately to the roll animation rather than setting the goal, but instead we have used the standard state change behaviour i.e. it's baked in to the animation data in the injection file.


lahm86:
@ajtudela Could you provide translations for this new option? Thanks.

rr- (approved):
Hi, I don't see the injected animation? Was it injected before as part of jump twists?

lahm86:
> Hi, I don't see the injected animation? Was it injected before as part of jump twists?

It's added in this change, it's just very very small :) So the new anims are basically appended to this file.
https://github.com/LostArtefacts/Tomb1Main/pull/994/files#diff-788f0f7a19c3c96c146599354d0074f7f306d2475f38bab786a2a494f56e2d45

rr-:
Yes, I apologize – I didn't notice the .bin in the file list as GitHub rendered it in a very modest box.

walkawayy (approved):

ajtudela:
Done #996

Issue #967, opened 2023-09-03
Feature: allow customization of enemy item drops


lahm86:
Currently, pickup items that drop from enemies when they die have a number of restrictions (primarily for custom levels and the randomizer).

- Pierre, Cowboy, SkaterKid and Kold are the only enemies who can drop items 
- The dropped item types and quantities are hard-coded to each enemy type
- Dropped guns are not converted into ammo if Lara already has the weapon in her inventory
- Saving after items have been dropped but before picking them up, and then reloading can result in the items disappearing. This is not so much an issue currently because there are checks to respawn them if Lara doesn't already have the items, but it should be considered for weapon/ammo conversion and key items in the reworked approach. For example, the check for Pierre at the end of Tihocan tests if Lara has the scion from Qualopec, but this may not be applicable in a custom level. If Pierre were to drop a key item in that scenario, saving/loading before picking it up would result in a softlock.

Rework the system to allow customization using the gameflow file.

PR #1001, opened 2023-09-22
ai: make baddies drop ammo if Lara has guns

### LINKED ISSUE #1000: Bug: multiple baddies drop weapons even if Lara already carries them
This only affects rando/mods.

rr-:
#### Checklist

#### Description

Patch by padoop for the rando so that multiple skaters/cowboys/pierres drop ammo if Lara already has the given weapon. Resolves #1000.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## [Unreleased](https://github.com/rr-/Tomb1Main/compare/stable...develop) - ××××-××-××
 - added Linux builds and toolchain
 - added an option to allow Lara to roll while underwater, similar to TR2+ (#993)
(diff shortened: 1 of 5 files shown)


walkawayy (approved):
This LGTM but I know Lahm was already looking into changing how items are dropped.

I think this does cause a minor change if the player gets the magnum secret in Colosseum. Lara will get an extra magnum ammo pickup from before. It makes sense to me to make that change, but I just wanted to bring that up.

lahm86 (approved):
Looks great. Yes, I think getting the Colosseum secret and then killing Pierre later in Tihocan would give you extra ammo, but we can address that in #967.

PR #1002, opened 2023-09-22
s/clock: reduce CPU usage

### LINKED ISSUE #985: Release CPU while waiting
I was trying some profiling to investigate performance on a slower PC. With the new framebuffer rendering that is. But I actually noticed some oddity on my faster PC. CPU usage combined is around 25% all the time. One of 4 real CPU-cores being almost at maximum at all times. With or without vsync. 
GNU Profiler says it spends 70% of the time in this function "S_Clock_SyncTicks".
I am no expert, but I read that SDL_Delay() is a way to release the CPU to the scheduler. I tried to weave this into the culprit function, whilst trying to avoid it causing undesired lag. Like below. This does seem to do the trick, as total CPU usage is now around 7%. I added the +0.1 there as a hackish way to prevent the Framerate indicator dropping from 30 to the occasional 29 FPS. 

s_clock.c:
```
int32_t S_Clock_SyncTicks(int32_t target)
{
    double elapsed = 0.0;
    Uint64 last_ticks = m_Ticks;
    do {
        m_UpdateTicks();
        elapsed = (double)(m_Ticks - last_ticks) / m_Frequency;
        if ((elapsed +0.1) < (double)target) SDL_Delay(1); // NEW LINE ADDED
    } while (elapsed < (double)target);
    return elapsed;
}
```

rr-:
#### Checklist

#### Description

Fix donated by padoop, with small modifications by me. Resolves #985.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## [Unreleased](https://github.com/rr-/Tomb1Main/compare/stable...develop) - ××××-××-××
 - added Linux builds and toolchain
 - added an option to allow Lara to roll while underwater, similar to TR2+ (#993)
+- improved frame scheduling to use less CPU (#985)
 
 ## [2.16](https://github.com/rr-/Tomb1Main/compare/2.15.3...2.16) - 2023-09-20
 - added a new rendering mode called "framebuffer" that lets the game to run at lower resolutions (#114)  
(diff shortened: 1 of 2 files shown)


walkawayy:
So this is just to reduce CPU usage on fast PCs right and does not address bad performance on low end PCs?

lahm86 (approved):

walkawayy (approved):

rr-:
>So this is just to reduce CPU usage on fast PCs right and does not address bad performance on low end PCs?

I think it has the opportunity of accomplishing both (less CPU time spent in busy loops = more CPU time to draw the game), but I lack a potato computer to properly verify this. On my i7 the usage went from 95% to 60%, on my i11 from 7% to 1%.

PR #1006, opened 2023-09-24
items: use animation relative frames

### LINKED ISSUE #929: Change animation switching function to be relative from the provided animation
As discussed in #928, change `Item_SwitchToAnim` to use a frame number relative from the provided animation instead of from the object's base frame.

walkawayy:
Resolves #929.

#### Checklist

#### Description
Changed `Item_SwitchToAnim` to use frame numbers relative from the animation instead of a global animation frame number. Also, this cleans up the syntax since `-1` is no longer needed for the default `frame_base` number. Now it's just `0`.

Added `Item_TestFrame` to test frame numbers similarly using animation relative frame numbers.

Removed the frame numbers from `types.h` and moved them into the individual files since they are all only used in a single file. I kept the `LF_` (Lara Frame) prefix except for Bacon Lara's equivalent. I made it `BLF_` (Bacon Lara Frame). Even though it matches Lara, I didn't want two constants with the exact same name. Let me know if you think any of the names should be changed.


src/game/objects/general/pickup.c
```diff
@@ -99,7 +102,7 @@ void Pickup_Collision(int16_t item_num, ITEM_INFO *lara_item, COLL_INFO *coll)
         }
 
         if (lara_item->current_anim_state == LS_PICKUP) {
-            if (lara_item->frame_number != LF_PICKUP_ERASE) {
+            if (Item_TestFrame(lara_item, LF_PICKUP_ERASE)) {
```
lahm86: This should be `!Item_TestFrame...` - pickups are currently disappearing too early.

src/game/lara/lara_state.c
```diff
@@ -608,7 +610,10 @@ void Lara_State_Controlled(ITEM_INFO *item, COLL_INFO *coll)
     coll->enable_spaz = 0;
     coll->enable_baddie_push = 0;
 
-    if (item->frame_number == g_Anims[item->anim_number].frame_end - 1) {
+    if (Item_TestFrame(
```
lahm86: This feels difficult to read. Is it worth creating a function to test for the end frame? I think it's only used here but at least then it might be a bit clearer.
As an aside, I can't actually work out which animation this is - nothing in the OG levels uses this state ID (`LS_CONTROLLED`, 56).

lahm86 (changes requested):
Looks good, just a pickup issue and a suggestion for readability. I tested through each of the affected animations/scenarios.

src/game/lara/lara_state.c
walkawayy: This was ported with walk to pickups from TR4/5 for when Lara is in the controlled state. It's never actually used though in TR1, and I confirmed that during testing. Not sure if I should just delete it.

rr-: I think this can be removed. In the future this kind of function can take negative frames to test basing on end of the animation.

src/game/objects/general/pickup.c
walkawayy: Ah ya I forgot the `!` for the non-controlled case.

walkawayy:
Fixed pickup issue and reduced `Lara_State_Controlled` to match `Lara_State_Null`, `Lara_State_Gymnast`, and other similarly small collision functions.

rr- (approved):

lahm86 (approved):

PR #1012, opened 2023-09-26
ai: add configurable item drop behaviour

### LINKED ISSUE #967: Feature: allow customization of enemy item drops

lahm86:
Resolves #967.

#### Checklist

#### Description

This follows on from #1000, and allows all but a few enemies to carry and drop items for Lara to collect. The hardcoded drops have been removed and the gameflow is now responsible for the setup. The default gameflow is configured in line with OG behaviour/requirements.

`convert_dropped_guns` in the gameflow controls whether guns are converted to ammo when dropped and Lara already has that gun. This is `false` by default to avoid extra ammo pickups when killing Pierre, Cowboy and Kold in the original levels.

Enemies that are *not* capable of carrying and dropping are:
- `O_CROCODILE` `O_ALLIGATOR` `O_RAT` `O_VOLE`  - having items spawn in the water would be difficult to track with changing water levels I feel, and having floating pickups would not be in-keeping with normal experience.
- `O_WARRIOR1` `O_WARRIOR2` `O_WARRIOR3` - these explode in mid-air and don't follow the standard `Creature_Animate` routine on death. We could manually spawn the items in each object's control, and work out the floor height, but the item would appear out of nowhere if the creature were high above land when killed, so again it feels non-standard.

Pods and centaur statues can be referenced as carriers in the gameflow; the system will give the items to the enemies that actually spawn, provided those are suitable carriers (empty pods are excluded).

The statistics take into account carried items. I have added a function here for common behaviour when testing triggered killable items.

In terms of saving and loading, items will respawn accordingly if Lara hasn't picked them up yet and equally they will not if she has done so. For testing, I used legacy ATI saves, V3 of the TR1X save format (2.16) and the current V4. In each case, I tested Pierre, Cowboy, Skaterboy and Kold in six scenarios - the following with `convert_dropped_guns` set to `false` and then again with it set to `true`.
- Before killing the enemy
- After killing the enemy, but before collecting their item(s)
- After killing the enemy, and after collecting their item(s)

I also used a test level to ensure items allocated to every potential carrier type spawn as expected, and that items allocated to non-carriers are ignored.

I will follow-up this PR with a document update to create a gameflow MD file, so that builders have a go-to reference for this feature in detail, as well as other gameflow aspects.

### DIFF
--- a/src/game/carrier.c
+++ b/src/game/carrier.c
@@ -12,43 +12,51 @@
 #include <stdbool.h>
 #include <stddef.h>
 
-#define AMMO_OFFSET 4
-#define LEGACY_DROP_COUNT 4
+#define NO_OBJECT O_NUMBER_OF
 
-typedef struct LEGACY_DROP_TEST {
-    const GAME_OBJECT_ID object_id;
-    const GAME_OBJECT_ID drop_id;
-} LEGACY_DROP_TEST;
+typedef struct GAME_OBJECT_PAIR {
+    const GAME_OBJECT_ID key_id;
+    const GAME_OBJECT_ID value_id;
+} GAME_OBJECT_PAIR;
 
 static ITEM_INFO *Carrier_GetCarrier(int16_t item_num);
-static bool Carrier_IsObjectType(int16_t obj_num, const int16_t *test_arr);
+static bool Carrier_IsObjectType(
+    GAME_OBJECT_ID object_id, const GAME_OBJECT_ID *test_arr);
+static GAME_OBJECT_ID Carrier_GetCognate(
+    GAME_OBJECT_ID key_id, const GAME_OBJECT_PAIR *test_map);
 
-static const int16_t m_CarrierObjects[] = {
+static const GAME_OBJECT_ID m_CarrierObjects[] = {
     O_WOLF,   O_BEAR,   O_BAT,     O_LION,  O_LIONESS, O_PUMA,   O_APE,
     O_TREX,   O_RAPTOR, O_CENTAUR, O_MUMMY, O_LARSON,  O_PIERRE, O_SKATEKID,
-    O_COWBOY, O_BALDY,  O_NATLA,   O_TORSO, NO_ITEM
+    O_COWBOY, O_BALDY,  O_NATLA,   O_TORSO, NO_OBJECT
 };
 
-static const int16_t m_PlaceholderObjects[] = { O_STATUE, O_PODS, O_BIG_POD,
-                                                NO_ITEM };
+static const GAME_OBJECT_ID m_PlaceholderObjects[] = { O_STATUE, O_PODS,
+                                                       O_BIG_POD, NO_OBJECT };
 
-static const int16_t m_DropObjects[] = {
+static const GAME_OBJECT_ID m_DropObjects[] = {
     O_GUN_ITEM,     O_SHOTGUN_ITEM,  O_MAGNUM_ITEM,   O_UZI_ITEM,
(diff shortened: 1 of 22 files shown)


src/game/carrier.c
```diff
@@ -0,0 +1,213 @@
...
+            Carrier_TestItemDrops(item_num);
+        } else {
+            CARRIED_ITEM *item = carrier->carried_item;
+            while (item) {
+                item->status = IS_INVISIBLE;
+                item = item->next_item;
+            }
+        }
+    }
+}
```
rr-, line 241: Hi, let's add more comments in the functions what's their purpose.

src/game/carrier.c
```diff
@@ -0,0 +1,213 @@
...
+#include "game/items.h"
+#include "global/const.h"
+#include "global/types.h"
+#include "global/vars.h"
+#include "log.h"
+
+#include <stdbool.h>
+#include <stddef.h>
+
+#define AMMO_OFFSET 4
```
rr-: Can we do without this const?

src/game/carrier.c
```diff
@@ -0,0 +1,213 @@
...
+static ITEM_INFO *Carrier_GetCarrier(int16_t item_num)
+{
+    if (item_num < 0 || item_num >= g_LevelItemCount) {
+        return NULL;
+    }
+
+    ITEM_INFO *item = &g_Items[item_num];
+    if (Carrier_IsObjectType(item->object_number, m_PlaceholderObjects)) {
+        int16_t child_item_num = *(int16_t *)item->data;
+        item = &g_Items[child_item_num];
```
rr-, line 130: It's amazing that you thought of this even.

src/game/carrier.c
```diff
@@ -0,0 +1,213 @@
...
+
+static bool Carrier_IsObjectType(int16_t object_id, const int16_t *test_arr)
+{
+    for (int i = 0; test_arr[i] != NO_ITEM; i++) {
+        if (test_arr[i] == object_id) {
+            return true;
+        }
+    }
+    return false;
+}
```
rr-, line 149: Can the `int16_t` throughout this file become `GAME_OBJECT_ID`?

src/game/carrier.c
lahm86: Perhaps an explicit switch statement to convert gun to ammo would be better, or a map even.

src/game/carrier.c
lahm86: Yes, that would be much neater I think.

src/game/carrier.c
rr-: Yes that's exactly what I had on my mind.

rr- (approved):

walkawayy (approved):
Really nice!

PR #1020, opened 2023-09-30
docs: expand on gameflow usage

### LINKED ISSUE #1018: Expand on gameflow usage
Following recent additions to the gameflow such as bonus levels (#976) and enemy item drops (#1012), it would be beneficial to have a dedicated usage document for players and level builders to refer to.

lahm86:
Resolves #1018.

#### Checklist

#### Description

This is a first draft of the new gameflow documentation. I opted to move all commentary from the gameflow json file so that we don't need to maintain it in two separate places. I've expanded on each of the features as well, and opted to go for a tabular approach for all properties. I also included some json snippets for clarity of key aspects.


bin/cfg/TR1X_gameflow.json5
```diff
@@ -1,97 +1,40 @@
...
-    // enable_save_crystals.
     "force_enable_save_crystals": false,
-
-    // Seconds to pass in the main menu before playing the demo.
     "demo_delay": 16,
-
-    // Water color (R, G, B). 1.0 means pass-through, 0.0 means no value at all.
-    // [0.6, 0.7, 1.0] is original DOS version filter,
-    // [0.45, 1.0, 1.0] is default TombATI filter.
     "water_color": [0.45, 1.0, 1.0],
```
walkawayy, line 12: Isn't this water color value actually different from TombATI since TR1X better matches the PS1 TR1? The wording makes it sound like we match TombATI, but I don't think that's the case. @rr-

GAMEFLOW.md
```diff
@@ -0,0 +1,1306 @@
...
+    <td>
+      <code>normal</code>
+    </td>
+    <td>A standard level.</td>
+  </tr>
+  <tr valign="top">
+    <td>
+      <code>title</code>
+    </td>
+    <td>One - and only one - of these must be defined.</td>
```
walkawayy: Maybe add `The main menu level. One -...` as a preface? Just trying to avoid any confusion since normal levels also have a level `title` which is the name of the level.

GAMEFLOW.md
```diff
@@ -0,0 +1,1306 @@
...
+    </td>
+  </tr>
+  <tr valign="top">
+    <td>
+      <code>remove_guns</code>
+    </td>
+    <td colspan="2" align="center">N/A</td>
+    <td>
+      Removes all weapons from Lara's inventory. If present, this must appear
+      before the <code>start_game</code> sequence.
```
walkawayy: Maybe mention Lara retains the remove gun's ammo?

GAMEFLOW.md
```diff
@@ -0,0 +1,1306 @@
...
+      A list of item <em>types</em> to drop. These items will spawn dynamically
+      and do not need to be added to the level file. Duplicate IDs are permitted
+      in the same array.
+    </td>
+  </tr>
+</table>
+
+You can also toggle `convert_dropped_guns` in
+[global properties](#convert-dropped-guns) so that guns that Lara already has
+are converted to ammo when they are dropped by enemies.
```
walkawayy: `You can also toggle `convert_dropped_guns` in
[global properties](#convert-dropped-guns) so that dropped guns Lara already has
are converted to ammo when they are dropped by enemies.`

Slight reword. What do you think? The `that` sandwich was a little confusing to read.

GAMEFLOW.md
```diff
@@ -0,0 +1,1306 @@
...
+      using the 3D pickups option.
+    </td>
+  </tr>
+  <tr valign="top">
+    <td>
+      <code>*_textures.bin</code>
+    </td>
+    <td>
+      Injects various texture fixes, such as gaps in the walls or wrongly
+      colored models. Refer to the README for a full list of fixes.
```
walkawayy: Maybe add `in the original levels` here too?

GAMEFLOW.md
```diff
@@ -0,0 +1,1306 @@
...
+will not use the global injection files. If those levels have their own
+injections defined, only those will be used. And of course, if they have none
+defined, nothing will be injected.
+
+_Disclaimer: Custom levels should not use the injections mechanism and instead
+should provide data that is already correct and consistent. Reports of bugs
+about injection files not working for custom levels will not be considered. Note
+however that the injections that relate to Lara can work in custom levels
+that do not modify Lara's default mesh structure or animations. These injection
+files are based on the original Lara model._
```
walkawayy, line 1208: I agree with this disclaimer mostly but aren't injections required if builders want animations like the jump twist? This makes it seem like they should never be used. Can the animations like jump twist just be added to Lara's animations normally? Is there a required animation slot? Same for underwater roll, responsive jumping, etc.

walkawayy (changes requested):
Really nice! Just a few comments.

GAMEFLOW.md
lahm86: Yes, so I'm hoping the extra note there covers it, so builders know that provided they have used an original Lara model, the animation injections will work fine. They could create the animations manually, and all the state changes, commands etc - I'm not sure if we want to get into supporting achieving that though as it's quite intricate. And yes, you're right, it needs the new anims in specific slots.

GAMEFLOW.md
lahm86: Good catch - done.

GAMEFLOW.md
lahm86: Thanks. I've reworked this bit of the table to avoid too much repetition and expanded on it further too.

GAMEFLOW.md
lahm86: Yes, agreed, it wasn't too clear. I've made it super-explicit now - above sounded good, but the double "dropped" was making me hesitate a little 😄

GAMEFLOW.md
lahm86: Thanks, keeps it consistent 👍

bin/cfg/TR1X_gameflow.json5
lahm86: Looking at `tombati.ini` it's not clear what the blue value defaults to, but red and green are certainly as described here.

walkawayy (approved):

bin/cfg/TR1X_gameflow.json5
rr-: Blue is always 0xFF in TombATI AFAIR. The values are good and tested.

rr- (approved):

PR #1023, opened 2023-10-02
game: add an option to enable frame buffering

### LINKED ISSUE #591: Allow speedrun "buffering" in TR1M

lahm86:
Resolves #591.

#### Checklist

#### Description

Continues on from #593 to allow buffering using the F2/F3/F4 keys (or whichever keys are mapped), and via the inventory. 

This remains optional, and is disabled by default, because of the noticeable delay when pressing F2/F3/F4 for their actual purpose, plus if you use the inventory key to cancel an FMV, the inventory will open immediately in-game.

The demo below shows the ability to accurately control inputs, and to test frame numbers. F2/F3/F4 allows for one-frame buffering, and inventory for two, which ties in with TombATI. _NB the animation and frame number shown in the demo are not part of this, only added here as a showcase._

https://youtu.be/jfopcpbYdU8

### DIFF
--- a/src/game/game/game.c
+++ b/src/game/game/game.c
@@ -12,6 +12,7 @@
 #include "game/lara/lara_hair.h"
 #include "game/level.h"
 #include "game/music.h"
+#include "game/output.h"
 #include "game/overlay.h"
 #include "game/savegame.h"
 #include "game/sound.h"
@@ -20,10 +21,18 @@
 #include "global/vars.h"
 #include "log.h"
 
+#define FRAME_BUFFER(key)                                                      \
+    do {                                                                       \
+        Game_DrawScene(true);                                                  \
+        Output_DumpScreen();                                                   \
+        Input_Update();                                                        \
+    } while (g_Input.key)
+
 static const int32_t m_AnimationRate = 0x8000;
 static int32_t m_FrameCount = 0;
 
 static int32_t Game_Control(int32_t nframes, GAMEFLOW_LEVEL_TYPE level_type);
+static void Game_UpdateInput(void);
 
 static int32_t Game_Control(int32_t nframes, GAMEFLOW_LEVEL_TYPE level_type)
 {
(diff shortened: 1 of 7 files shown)


src/game/game/game.c
```diff
@@ -68,8 +77,8 @@ static int32_t Game_Control(int32_t nframes, GAMEFLOW_LEVEL_TYPE level_type)
             }
         }
 
-        if ((g_InputDB.option || g_Input.save || g_Input.load
-             || g_OverlayFlag <= 0)
+        if (((g_Config.enable_buffering ? g_Input : g_InputDB).option
+             || g_Input.save || g_Input.load || g_OverlayFlag <= 0)
```
rr-, line 81: Holy cow, what's this about?

rr- (approved):

src/game/game/game.c
lahm86: Do you think this is a better approach? It's maybe a bit more readable.
```
        if (((g_Config.enable_buffering ? g_Input.option : g_InputDB.option)
             || g_Input.save || g_Input.load || g_OverlayFlag <= 0)
```

lahm86:
@ajtudela Could you provide a translation for this new option? Thanks.

src/game/game/game.c
rr-: I mostly mean the complexity of this condition. But I think I understand what's it about.

lahm86: Yes, it's quite lengthy. Should we move it to a function?

rr-: No, what we need is a rewrite of the inventory code :D It's as good as it can get given the current form.

ajtudela:
Done #1024

walkawayy (approved):

PR #1033, opened 2023-10-06
creature: fix dead water creature placement

### LINKED ISSUE #1031: OG bug: dead crocodiles float when rooms are drained
image

lahm86:
Resolves #1031.

#### Checklist

#### Description

This moves the logic for shifting dead rats and crocodiles into shared functions in `creature.c` to resolve inconsistencies between the two and to cover all scenarios. Once a dead hybrid creature has been converted to its land type, it will remain in that state permanently.

I've also introduced an idiom in `Item_SwitchToObjAnim` that if you pass a negative frame number, this will be inferred as meaning the frame should be relative from the end of the animation.

### DIFF
--- a/src/game/creature.c
+++ b/src/game/creature.c
@@ -18,6 +18,11 @@
 
 #define MAX_CREATURE_DISTANCE (WALL_L * 30)
 
+static bool Creature_TestLandState(
+    int16_t item_num, int32_t *wh, const HYBRID_INFO *info);
+static bool Creature_TestWaterState(
+    int16_t item_num, int32_t *wh, const HYBRID_INFO *info);
+
 void Creature_Initialise(int16_t item_num)
 {
     ITEM_INFO *item = &g_Items[item_num];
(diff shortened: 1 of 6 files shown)


src/game/objects/creatures/rat.c
```diff
@@ -49,6 +49,13 @@ typedef enum {
 
 static BITE_INFO m_RatBite = { 0, -11, 108, 3 };
 
+static const HYBRID_INFO m_RatInfo = { .land.id = O_RAT,
+                                       .land.active_anim = RAT_EMPTY,
+                                       .land.death_anim = RAT_DIE_ANIM,
+                                       .land.death_state = RAT_DEATH,
+                                       .water.id = O_VOLE,
+                                       .water.active_anim = VOLE_EMPTY };
```
rr-: Will it be formatted any better if we put a trailing comma inside the struct?

src/game/creature.c
```diff
@@ -706,3 +711,80 @@ bool Creature_ShootAtLara(
...
+bool Creature_TestHybridState(
+    int16_t item_num, int32_t *wh, const HYBRID_INFO *info)
+{
+    // Test the environment for a hybrid creature. Record the water height and
+    // return whether or not a type conversion has taken place.
+    ITEM_INFO *item = &g_Items[item_num];
+    return item->object_number == info->land.id
+        ? Creature_TestLandState(item_num, wh, info)
+        : Creature_TestWaterState(item_num, wh, info);
+}
```
rr-, line 727: The name `Test` suggests that the function doesn't mutate the global state, while it changes up item animations and object numbers. Furthermore `Hybrid` is unclear until the reader already knows the concept behind it. How about `Creature_EnsureHabitat` or similar?

src/game/objects/creatures/rat.c
lahm86: It did, thanks. I've done the same in `crocodile.c` for consistency.

src/game/creature.c
lahm86: Good point, thanks. Now renamed as above, and the statics have been renamed similarly too. I've also refactored slightly to call `Room_GetWaterHeight` in the wrapper function rather than in each state function for a little less repetition.

rr- (approved):

src/game/objects/creatures/rat.c
```diff
@@ -49,6 +49,15 @@ typedef enum {
...
 static BITE_INFO m_RatBite = { 0, -11, 108, 3 };
 
+static const HYBRID_INFO m_RatInfo = {
+    .land.id = O_RAT,
+    .land.active_anim = RAT_EMPTY,
+    .land.death_anim = RAT_DIE_ANIM,
+    .land.death_state = RAT_DEATH,
+    .water.id = O_VOLE,
+    .water.active_anim = VOLE_EMPTY,
+};
```
walkawayy, line 59: Why does the water state not need a death anim/state?

walkawayy (approved):
All testing LGTM.

src/game/objects/creatures/rat.c
lahm86: We never switch to the water death state in the environment checks, so it's not needed. These are the only possible scenarios:
- Killed in water (normal control uses water death anim)
  - Room drained - switch to land death
  - Room re-flooded - remain in land death
- Killed on land
  - Room flooded - remain in land death
  - Room re-drained - remain in land death

The only difference is for crocodiles - if they're killed in water and that room is then drained, but there is still another water room below, they'll remain in the water anim death state.

PR #1036, opened 2023-10-07
build: fix failure with non-docker build

alanswanson:
#### Checklist

#### Description

Building on Linux directly with meson (as docker unnecessary) failed with;

```
/usr/lib/gcc/x86_64-pc-linux-gnu/12/../../../../x86_64-pc-linux-gnu/bin/ld: TR1X.p/src_global_vars.c.o:/srv/git/TR1X-build-meson/builddir/../src/global/vars.c:105: multiple definition of `g_TR1XVersion'; TR1X.p/meson-generated_.._init.c.o:/srv/git/TR1X-build-meson/builddir/init.c:6: first defined here
/usr/lib/gcc/x86_64-pc-linux-gnu/12/../../../../x86_64-pc-linux-gnu/bin/ld: TR1X.p/src_global_vars.c.o:/srv/git/TR1X-build-meson/builddir/../src/global/vars.c:106: multiple definition of `g_GameFlowDefaultStrings'; TR1X.p/meson-generated_.._init.c.o:/srv/git/TR1X-build-meson/builddir/init.c:8: first defined here
```

Looks like DOCKER_BUILD, as introduced in https://github.com/LostArtefacts/TR1X/commit/d87837d95169ffe96871f4bf555e1b5259dd4672, is unnecessary as meson.build always calls tools/generate_init which always generates an init.c with these two parameters. Deleting the ifndef DOCKER_BUILD section in src/global/vars.c allowed the build to complete. Left the define in entrypoint.sh files as may be useful in future.

### DIFF
--- a/docker/game-linux/entrypoint.sh
+++ b/docker/game-linux/entrypoint.sh
@@ -2,7 +2,6 @@
 set -x
 set -e
 
-export CFLAGS=-DDOCKER_BUILD
 EXE_FILE=TR1X
 
 if [ ! -f /app/build/linux/build.ninja ]; then
--- a/docker/game-win/entrypoint.sh
+++ b/docker/game-win/entrypoint.sh
@@ -2,8 +2,6 @@
 set -x
 set -e
 
-export CFLAGS=-DDOCKER_BUILD
-
 if [ ! -f /app/build/win/build.ninja ]; then
     meson --buildtype "$TARGET" /app/build/win/ --cross /app/docker/game-win/meson_linux_mingw32.txt --pkg-config-path=$PKG_CONFIG_PATH
 fi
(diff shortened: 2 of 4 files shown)


rr-:
Hi, thank you for the pull request. This preprocessor instruction is mainly for users who don't use the official toolchain, like those compiling with Visual Studio. I didn't think about the possibility of using meson outside of regular Docker builds. Instead of removing the macro completely and losing support for external toolchains, I would suggest changing the `DOCKER_BUILD` environment variable to `MESON_BUILD`, removing its declaration from the Docker scripts, and placing it in `meson.build`. What do you think?

alanswanson:
Yes, that should work. (Though had a mental block for a while thinking it would still just be the same issue!)

Coming from your Windows side the docker builds are good for your official releases but for most normal Linux users and distros they would build directly with meson and not require building and bundling already included common libraries (just ffmpeg and SDL2 now) into a single binary with upx.

rr- (approved):

lahm86 (approved):
LGTM

PR #1037, opened 2023-10-08
Gun frame tests

walkawayy:
#### Checklist

#### Description
Refactored pistols / rifle functions to use frame checking functions. Added new frame constants and renamed some existing ones to make it more clear what is going on.


src/game/items.c
```diff
@@ -791,11 +791,23 @@ void Item_TakeDamage(ITEM_INFO *item, int16_t damage, bool hit_status)
...
-    return item->frame_number >= g_Anims[item->anim_number].frame_base + start
-        && item->frame_number <= g_Anims[item->anim_number].frame_base + end;
+    return Item_TestAbsFrameRange(
+        item->frame_number, g_Anims[item->anim_number].frame_base + start,
+        g_Anims[item->anim_number].frame_base + end);
+}
+
+bool Item_TestAbsFrameRange(int16_t abs_frame, int16_t start, int16_t end)
+{
+    return abs_frame >= start && abs_frame <= end;
```
rr-: Hey, since this no longer works with ITEM_INFO, I'm leaning towards creating anim.c / anim.h pair and changing the prefix of this and other functions that do not use ITEM_INFO* to Anim_. WDYT?

walkawayy: Ok, I wasn't sure if we wanted to make new `anim.h/c` files that would only include these two functions. I can do that. They don't use `ANIM_STRUCT` either.

walkawayy:
I don't understand this linter. And my local `make imports` doesn't report any errors.

rr-:
You probably have a different version installed on your local machine. The CI output shows what needs to be done for it to pass.

lahm86:
I noticed `anims.h` and `items.h` have the same functions declared, guessing that's not intended.

walkawayy:
> I noticed `anims.h` and `items.h` have the same functions declared, guessing that's not intended.

Wow lol. Yes I do need to fix my local version because it reported no issues when I ran make imports. And there is output, so it is definitely running.

lahm86 (approved):
All looks good to me 👍

rr- (approved):

PR #1040, opened 2023-10-09
build: correct import name for python json5 module in meson build

alanswanson:
#### Checklist

#### Description

Whilst the package name is _pyjson5_, the actual module name within python is _json5_.  Caused below build failure on my Gentoo distro. Ubuntu/Debian do not make any changes to the pyjson5 package so wonder if they patched a wildcard match elsewhere in Python. If someone could please test this works on docker.

```
[24/164] Generating fake_init with a custom command
FAILED: init.c
/usr/bin/python3 /srv/git/TR1X-test-pyjson5-failure/tools/generate_init --version-file version.txt -o init.c
Traceback (most recent call last):
  File "/srv/git/TR1X-test-pyjson5-failure/tools/generate_init", line 7, in <module>
    import pyjson5
ModuleNotFoundError: No module named 'pyjson5'
```

### DIFF
--- a/tools/generate_init
+++ b/tools/generate_init
@@ -4,7 +4,7 @@ import io
 import json
 from pathlib import Path
 
-import pyjson5
+import json5
 
 REPO_DIR = Path(__file__).parent.parent
 BIN_DIR = REPO_DIR / "bin"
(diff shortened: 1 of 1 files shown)


rr-:
Nope, `ModuleNotFoundError: No module named 'json5'`
It's installed with `python3 -m pip install pyjson5`.

alanswanson:
Hmm, there are two different pyjson5 packages, https://github.com/dpranke/pyjson5/ and https://github.com/Kijewski/pyjson5, and the "wrong" one is in my distro (though it still worked)! Sorry for the noise (but this'll provide a reference to anybody else encountering the issue). Closing.

**EDIT:** Just to clarify, from https://github.com/Kijewski/pyjson5/issues/29#issuecomment-1015777249 these two modules are effectively same except one is cython compiled and faster while the other is pure python (though a few less functions which are unused here). So it's safe for users to switch between the two as needed.

PR #1046, opened 2023-10-09
Config rewrite

### LINKED ISSUE #1043: Inconsistencies between game defaults and config tool defaults
For example the braid is disabled by default in the game code, while it is enabled by default in the config tool.

rr-:
#### Checklist

#### Description

Resolves #1043.

Because we had a mismatch between the defaults provided by the config tool and the game, I found it important to have a way to be able to check it automatically. And so, I created a tool that takes TR1X.json5 and compares it with the config tool specs, reporting any mismatches. (It's meant to be run with a config file that gets created automatically upon the first game launch.) 

Turns out the braid was the only option not in sync with the config tool defaults. This got fixed in this PR as well.

Additionally, to get the config more structured, I replaced `READ_BOOL` / `WRITE_INTEGER` etc macros in the config with an array of structs, similar to what the configtool specs file has. It's both prettier and uglier at the same time, but I think I prefer the new approach. It's important to have consistent data types and their sizes due to pointer safety (so eg no writing to int8_t using `COT_INT32`, or we'll get a boo-boo). I was thinking about creating a tool to generate it from a JSON but I don't think that would be very productive, and at the same time it would render compiling outside of meson even more difficult.
LMK what you think about it – whether we want it, and if we want it formatted or not (currently it's fenced with a `clang-format off` directive).
I didn't touch the input read/write routines as they're a bit more advanced. I did move the `Input_FlashConflicts` to `Config_Init` which I believe to be a more suitable place for it, but I haven't tested if it breaks anything.


src/config_map.c
```diff
@@ -0,0 +1,155 @@
...
+static const CONFIG_OPTION_ENUM_MAP m_ScreenshotFormats[] = {
+    { "jpg", SCREENSHOT_FORMAT_JPEG },
+    { "jpeg", SCREENSHOT_FORMAT_JPEG },
+    { "png", SCREENSHOT_FORMAT_PNG },
+    { NULL, -1 },
+};
+
+const CONFIG_OPTION g_ConfigOptionMap[] = {
+    // clang-format off
+    { .name = "disable_healing_between_levels", .type = COT_BOOL,   .target = &g_Config.disable_healing_between_levels,      .default_value = &(bool){false}},
```
lahm86: There are some compiler warnings - I appreciate they're not relevant in these cases, but is it worth putting explicit `.param = NULL` here to reduce the noise?

`[9/165] Compiling C object TR1X.exe.p/src_config_map.c.obj
../../src/config_map.c:53:5: warning: missing initializer for field ‘param’ of ‘CONFIG_OPTION’ [-Wmissing-field-initializers]`

lahm86:
I think this is a great approach, and although the map itself isn't too pretty as you say, it will be easy to modify/manage this going forward. It's also nice that we only need to define it here (C# project aside) and not have to add the read/write commands as well. The tool for comparison is extremely useful as well to avoid the mismatch problem appearing again.

src/config_map.c
rr-: Sorry, I forgot to compile it for Windows. I removed the warnings by supplying `0` which tells the compiler to initialize the rest of the struct with zeros/nulls.

rr-:
Could I also ask for a sanity check of reading/writing cycle of each variable type (int, float, double, bool and enum) and that input still flash conflicts?

src/config.c
```diff
@@ -20,102 +21,20 @@
...
-const ENUM_MAP m_ScreenshotFormats[] = {
-    { "jpg", SCREENSHOT_FORMAT_JPEG },
-    { "jpeg", SCREENSHOT_FORMAT_JPEG },
-    { "png", SCREENSHOT_FORMAT_PNG },
-    { NULL, -1 },
-};
-
-static const char *Config_ProcessKey(const char *key);
 static int Config_ReadEnum(
     struct json_object_s *obj, const char *name, int8_t default_value,
```
lahm86: Should we change the `int8_t` here to `int32_t` to match what's being passed in?

src/config.c
```diff
@@ -20,102 +21,20 @@
...
-    const ENUM_MAP *enum_map);
+    const CONFIG_OPTION_ENUM_MAP *enum_map);
 static void Config_WriteEnum(
     struct json_object_s *obj, const char *name, int8_t value,
```
lahm86: And same here.

lahm86 (commented):
All looks good to me. Conflicts are still showing correctly, and the read/write looks good - just a couple of comments.

src/config.c
rr-: I changed it to an `int` because ints are commonly used for enums and we also use ints in other parts of the code. However, it's important to note that ints are not always guaranteed to be 32-bit. In order to ensure this guarantee, I'd need to change `g_Config` members to be of type `int` instead of the relevant enums. Nonetheless, I don't think this is ever going to be relevant to us.

lahm86 (approved):

walkawayy (approved):

PR #1049, opened 2023-10-10
tools: clean installer and config tool

lahm86:
#### Checklist

#### Description

This is a housekeeping update to clear all warnings and messages from the two C# projects (example [here](https://github.com/LostArtefacts/TR1X/actions/runs/6247127192)). I've left the commits separate to mark clearly what was done, but will squash after review.

The one extended change is in replacing `WebClient` for file downloads. The class is now obsolete and its replacement, `HttpClient`, doesn't have a progress tracker; I've created a wrapper for that so there is no loss of functionality.


walkawayy (approved):

rr-:
Hey, in the naming violations commit we only change EventHandler types but leave other private variables intact.  In my opinion this change introduces inconsistency. Is there somewhere I can read more about this warning?

lahm86:
> Hey, in the naming violations commit we only change EventHandler types but leave other private variables intact. In my opinion this change introduces inconsistency. Is there somewhere I can read more about this warning?

I didn't think this through properly yesterday. I've now removed the `event` keyword as that's not required in these cases, and reverted the naming. The `event` keyword does the following, which is why VS complains about the naming i.e. it's similar to property naming.

```cs
private event EventHandler Temp;
```
=>
```cs
private EventHandler _temp;
private event EventHandler Temp
{
    add
    {
        _temp += value;
    }
    remove
    {
        _temp -= value;
    }
}
```

But in the `RelayCommand` case, we are adding to/removing from the handler directly via, for example, `public event EventHandler? CanExecuteChanged`. So having the additional `event` keyword on the private variable was redundant.

rr- (approved):

PR #1050, opened 2023-10-11
movable_block: fix various height issues

### LINKED ISSUE #723: Bugs with pushable blocks that are placed on air (and triggered to fall)
Reporting some more issues with pushable blocks, this time when they fall down. I'm not sure if these bugs are caused by Tomb Editor because so far they only happen in a custom level. Then again there's only 2 places in OG levels (Thor room, TNT in Natla's Mines) where this mechanic is used, so maybe they somehow avoid it. In Thor room it might be related to how the hammer stands on the trigger. Maybe.

I will keep testing and reporting here if I find more info, and hopefully someone can take a look at these issues. Also, some of the issues might be a result of misusing the blocks? Would be nice to know what can be done and what can't.

Anyway, to the point. Here's a clip that shows some of the core issues. Under certain circumstances, if you put blocks on the air (with intent to make them fall), the data is not updated correctly and Lara can go through blocks or they can form invisible "blocks" (that can be stood/climbed on). Saving and loading makes it wrose. Below the video I will try to explain all the conditions I'm aware of. The list might seem big but I think they all stem from the same 1-3 reasons.

https://user-images.githubusercontent.com/97730054/215189076-df6da13e-caba-41cd-bb05-8b9a9c323e58.mp4

[LEVEL1.zip](https://github.com/rr-/Tomb1Main/files/10522847/LEVEL1.zip)

The blocks may act differently if they're set invisible or not, and based on where they are put: in a room where there's floor below (floor room) them or portal (portal room).

BEFORE ACTIVATING and NO SAVE/LOAD
- visible block in floor room: there's an invisible "block" below it, and the block itself has no collision
- visible block in portal room: there's flickering and camera issues below the block, and no collision for block itself
- visible block above ceiling in portal room: flickering and camera issues below the block, no collision for block itself

SAVE/LOAD and then ACTIVATING
- invisible block in floor room: falls 1 block below the floor
- visible block in floor room: lands on top of an invisible "block"
- visible block in portal room: flickering / camera issues on the same spot as above
- visible block above ceiling in portal room: flickering / camera issues on the same spot as above

SAVE/LOAD and then ACTIVATE and then SAVE/LOAD
- invisible block in floor room: there's 1 block "hole" in the floor now
- visible block in floor room: finally normal
- invisible block in portal room: has no collision now
- visible block in portal room: has fallen through the floor, flickering / camera issues on the same spot as above
- visible block above ceiling in portal room: block has no collision

ACTIVATE
- visible block in floor room: lands on top of an invisible "block"
- visible block in portal room: flickering / camera issues in the bottom of portal room (same spot as above)
- visible block above ceiling in portal room: flickering / camera issues in the bottom of portal room (same spot as above)

ACTIVATE and SAVE/LOAD
- visible block in floor room: no collision
- invisible block in floor room: no collision, still 1 invisible "block" below it
- visible block in portal room: no collision and flickering / camera issues in hte bottom of portal room (same spot as above)
- visible block above ceiling in portal room: no collision

As you might see, "invisible block above ceiling in portal room" has never any problems. This is how the block is set up in Thor room. So you could maybe argue it's the only way to use the blocks. However given how hacky the TR games are, I think this is just the first way they got it to make and never needed anything else so it stayed like that. It would be nice to have more options for level builders to use this mechanic, and also be able to have the blocks visible and still working correctly (most of the problems are for the visible blocks).

walkawayy:
#### Checklist

#### Description
Fixes various scenarios where pushables have the wrong height as described in #723. Opening as a draft for now for testing purposes. Debug logs still in there on purpose.

### DIFF
--- a/src/game/objects/traps/movable_block.c
+++ b/src/game/objects/traps/movable_block.c
@@ -79,7 +79,8 @@ void MovableBlock_Setup(OBJECT_INFO *obj)
 void MovableBlock_Initialise(int16_t item_num)
 {
     ITEM_INFO *item = &g_Items[item_num];
-    if (item->status != IS_INVISIBLE) {
+
+    if (item->status != IS_INVISIBLE && item->pos.y >= Item_GetHeight(item)) {
         Room_AlterFloorHeight(item, -WALL_L);
     }
 }
@@ -110,6 +111,11 @@ void MovableBlock_Control(int16_t item_num)
         item->status = IS_DEACTIVATED;
         FX_DinoStomp(item);
         Sound_Effect(SFX_T_REX_FOOTSTOMP, &item->pos, SPM_NORMAL);
+    } else if (
+        item->pos.y >= height && !item->gravity_status
+        && !(bool)(intptr_t)item->priv) {
+        item->status = IS_NOT_ACTIVE;
+        Item_RemoveActive(item_num);
     }
 
     if (item->room_number != room_num) {
@@ -132,6 +138,10 @@ void MovableBlock_Collision(
     int16_t item_num, ITEM_INFO *lara_item, COLL_INFO *coll)
 {
     ITEM_INFO *item = &g_Items[item_num];
+
+    if (item->current_anim_state == MBS_STILL) {
+        item->priv = (void *)false;
+    }
 
     if (!g_Input.action || item->status == IS_ACTIVE
         || lara_item->gravity_status || lara_item->pos.y != item->pos.y) {
@@ -226,6 +236,7 @@ void MovableBlock_Collision(
         item->status = IS_ACTIVE;
         Item_Animate(item);
         Lara_Animate(lara_item);
+        item->priv = (void *)true;
     }
 }
 
--- a/src/game/objects/traps/movable_block.c
+++ b/src/game/objects/traps/movable_block.c
@@ -12,6 +12,8 @@
 #include "global/vars.h"
 #include "util.h"
 
+#include <stdbool.h>
+
 #define LF_PPREADY 19
 
 typedef enum {
@@ -36,6 +38,11 @@ static int16_t m_MovingBlockBounds[12] = {
 };
 
 static bool MovableBlock_TestDoor(ITEM_INFO *lara_item, COLL_INFO *coll);
+static bool MovableBlock_TestDestination(ITEM_INFO *item, int32_t block_height);
+static bool MovableBlock_TestPush(
+    ITEM_INFO *item, int32_t block_height, DIRECTION quadrant);
+static bool MovableBlock_TestPull(
+    ITEM_INFO *item, int32_t block_height, DIRECTION quadrant);
 
 static bool MovableBlock_TestDoor(ITEM_INFO *lara_item, COLL_INFO *coll)
 {
@@ -63,6 +70,143 @@ static bool MovableBlock_TestDoor(ITEM_INFO *lara_item, COLL_INFO *coll)
         }
     }
     return false;
+}
+
+static bool MovableBlock_TestDestination(ITEM_INFO *item, int32_t block_height)
+{
+    int16_t room_num = item->room_number;
+    FLOOR_INFO *floor =
+        Room_GetFloor(item->pos.x, item->pos.y, item->pos.z, &room_num);
+    if (floor->floor == NO_HEIGHT / 256) {
+        return true;
+    }
+
+    if ((floor->floor << 8) != item->pos.y - block_height) {
+        return false;
+    }
+
+    return true;
+}
+
+static bool MovableBlock_TestPush(
+    ITEM_INFO *item, int32_t block_height, DIRECTION quadrant)
+{
+    if (!MovableBlock_TestDestination(item, block_height)) {
+        return false;
+    }
+
+    int32_t x = item->pos.x;
+    int32_t y = item->pos.y;
+    int32_t z = item->pos.z;
+    int16_t room_num = item->room_number;
+
+    switch (quadrant) {
+    case DIR_NORTH:
+        z += WALL_L;
+        break;
+    case DIR_EAST:
+        x += WALL_L;
+        break;
+    case DIR_SOUTH:
+        z -= WALL_L;
+        break;
+    case DIR_WEST:
+        x -= WALL_L;
+        break;
+    }
+
+    FLOOR_INFO *floor = Room_GetFloor(x, y, z, &room_num);
+    COLL_INFO coll;
+    coll.quadrant = quadrant;
+    coll.radius = 500;
+    if (Collide_CollideStaticObjects(&coll, x, y, z, room_num, 1000)) {
+        return false;
+    }
+
+    if (((int32_t)floor->floor << 8) != y) {
+        return false;
+    }
+
+    floor = Room_GetFloor(x, y - block_height, z, &room_num);
+    if (((int32_t)floor->ceiling << 8) > y - block_height) {
+        return false;
+    }
+
+    return true;
+}
+
+static bool MovableBlock_TestPull(
+    ITEM_INFO *item, int32_t block_height, DIRECTION quadrant)
+{
+    if (!MovableBlock_TestDestination(item, block_height)) {
+        return false;
+    }
+
+    int32_t x_add = 0;
+    int32_t z_add = 0;
+    switch (quadrant) {
+    case DIR_NORTH:
+        z_add = -WALL_L;
+        break;
+    case DIR_EAST:
+        x_add = -WALL_L;
+        break;
+    case DIR_SOUTH:
+        z_add = WALL_L;
+        break;
+    case DIR_WEST:
+        x_add = WALL_L;
+        break;
+    }
+
+    int32_t x = item->pos.x + x_add;
+    int32_t y = item->pos.y;
+    int32_t z = item->pos.z + z_add;
+
+    int16_t room_num = item->room_number;
+    FLOOR_INFO *floor = Room_GetFloor(x, y, z, &room_num);
+    COLL_INFO coll;
+    coll.quadrant = quadrant;
+    coll.radius = 500;
+    if (Collide_CollideStaticObjects(&coll, x, y, z, room_num, 1000)) {
+        return false;
+    }
+
+    if (((int32_t)floor->floor << 8) != y) {
+        return false;
+    }
+
+    floor = Room_GetFloor(x, y - block_height, z, &room_num);
+    if (((int32_t)floor->ceiling << 8) > y - block_height) {
+        return false;
+    }
+
+    x += x_add;
+    z += z_add;
+    room_num = item->room_number;
+    floor = Room_GetFloor(x, y, z, &room_num);
+
+    if (((int32_t)floor->floor << 8) != y) {
+        return false;
+    }
+
+    floor = Room_GetFloor(x, y - LARA_HEIGHT, z, &room_num);
+    if (((int32_t)floor->ceiling << 8) > y - LARA_HEIGHT) {
+        return false;
+    }
+
+    x = g_LaraItem->pos.x + x_add;
+    y = g_LaraItem->pos.y;
+    z = g_LaraItem->pos.z + z_add;
+    room_num = g_LaraItem->room_number;
+    floor = Room_GetFloor(x, y, z, &room_num);
+    coll.radius = LARA_RAD;
+    coll.quadrant = (quadrant + 2) & 3;
+    if (Collide_CollideStaticObjects(&coll, x, y, z, room_num, LARA_HEIGHT)) {
+        return false;
+    }
+
+    return true;
 }
 
 void MovableBlock_Setup(OBJECT_INFO *obj)
@@ -148,7 +292,7 @@ void MovableBlock_Collision(
         return;
     }
 
-    uint16_t quadrant = ((uint16_t)lara_item->pos.y_rot + PHD_45) / PHD_90;
+    DIRECTION quadrant = ((uint16_t)lara_item->pos.y_rot + PHD_45) / PHD_90;
     if (lara_item->current_anim_state == LS_STOP) {
         if (g_Input.forward || g_Input.back
             || g_Lara.gun_status != LGS_ARMLESS) {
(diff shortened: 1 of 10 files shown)


src/game/objects/traps/movable_block.c
```diff
@@ -65,6 +68,77 @@ static bool MovableBlock_TestDoor(ITEM_INFO *lara_item, COLL_INFO *coll)
...
+int16_t MovableBlock_GetHeight(ITEM_INFO *item)
+{
+    int16_t room_num = item->room_number;
+    FLOOR_INFO *floor =
+        Room_GetFloor(item->pos.x, item->pos.y, item->pos.z, &room_num);
+    int32_t height =
+        Room_GetHeight(floor, item->pos.x, item->pos.y, item->pos.z);
+
+    return height;
+}
```
rr-, line 89: This series is calls is used so often in the code that I believe it deserves a place in the item API.

rr-:
I asked about this in Discord, but it went unanswered. Can the new alter floor function also be used for rolling blocks? Currently, the `Room_AlterHeight` function is still present, but its usage is limited to rolling blocks and Thor's Hammer head. I believe it would be best to avoid having two functions that essentially do the same thing but have different implementations. Moreover, modifying the room's internal data from an external module doesn't seem like the ideal approach.

lahm86:
I agree that one function is desirable for all cases, and that this should belong in `room.c`. A couple of other comments:

- The current `Room_AlterHeight` function modifies the `blocked` status on box overlaps, which prevents enemies clipping on top of pushblocks (where the correct flags have been set in the level data). Currently this doesn't work - an example is in Folly, if you stand on top of the first pushblock here, the lions clip up to Lara, which doesn't happen normally.
- The `do/while` in `MovableBlock_AlterFloorHeight` seems to be a reduced version of `Room_GetFloor`, the difference being it only takes horizontal portals into account. I think that makes sense, but would you ever be able to push a block onto a vertical portal? Even if it was a floor with a sky portal, the block's `y` position should not trigger either condition for the vertical checks in `Room_GetFloor`. So perhaps the loop isn't needed and we can just use `Room_GetFloor`; otherwise, maybe a separate `Room_GetFloorIgnoreVerticalPortals` is needed (that's a terrible name, but just a placeholder).

walkawayy:
> I asked about this in Discord, but it went unanswered. Can the new alter floor function also be used for rolling blocks? Currently, the `Room_AlterHeight` function is still present, but its usage is limited to rolling blocks and Thor's Hammer head. I believe it would be best to avoid having two functions that essentially do the same thing but have different implementations. Moreover, modifying the room's internal data from an external module doesn't seem like the ideal approach.

Tbh I didn't do much testing with rolling balls in other than the one in the test level that seems to work. So I didn't want to combine room / block floor altering functions yet. I also didn't tag for review yet because I wanted some testing feedback first. There's still debug stuff all over the code, so it's not really ready for review yet. That's why I opened as a draft PR.

walkawayy:
> I agree that one function is desirable for all cases, and that this should belong in `room.c`. A couple of other comments:
> 
>     * The current `Room_AlterHeight` function modifies the `blocked` status on box overlaps, which prevents enemies clipping on top of pushblocks (where the correct flags have been set in the level data). Currently this doesn't work - an example is in Folly, if you stand on top of the first pushblock here, the lions clip up to Lara, which doesn't happen normally.
> 
>     * The `do/while` in `MovableBlock_AlterFloorHeight` seems to be a reduced version of `Room_GetFloor`, the difference being it only takes horizontal portals into account. I think that makes sense, but would you ever be able to push a block onto a vertical portal? Even if it was a floor with a sky portal, the block's `y` position should not trigger either condition for the vertical checks in `Room_GetFloor`. So perhaps the loop isn't needed and we can just use `Room_GetFloor`; otherwise, maybe a separate `Room_GetFloorIgnoreVerticalPortals` is needed (that's a terrible name, but just a placeholder).

Thanks for the feedback. I'll have to see if we can combine the two floor altering functions or not.

And I'll prob need a test level for your second point? Is there an OG level case where that occurs?

I also still need to test legacy saves.

src/game/objects/traps/movable_block.c
```diff
@@ -289,7 +301,7 @@ bool MovableBlock_TestPush(
         return false;
     }
 
-    if (((int32_t)floor->floor << 8) != y) {
+    if (Room_GetHeight(floor, x, y, z) != y) {
```
lahm86: This - and the similar change in `MovableBlock_TestPull` - fixes a separate bug in TR1, where you can push/pull a block onto a slope once. Could we make this optional? There are some secrets in the randomizer that make use of this glitch (and the other OG glitches, whose fixes have config settings).

Edit: actually, I believe this change causes the trapdoor issue as detailed above.

src/game/room.c
```diff
@@ -590,19 +590,52 @@ int16_t Room_GetWaterHeight(int32_t x, int32_t y, int32_t z, int16_t room_num)
...
+
+    int16_t data;
+    FLOOR_INFO *floor;
+    ROOM_INFO *r = &g_RoomInfo[item->room_number];
+
+    do {
+        int32_t z_floor = (item->pos.z - r->z) >> WALL_SHIFT;
+        int32_t x_floor = (item->pos.x - r->x) >> WALL_SHIFT;
+
+        if (z_floor <= 0) {
```
lahm86, line 605: Maybe some `CLAMP`s here would be a bit neater:

```c
if (z_floor <= 0) {
    z_floor = 0;
    CLAMP(x_floor, 1, r->y_size - 2);
} else if (z_floor >= r->x_size - 1) {
    z_floor = r->x_size - 1;
    CLAMP(x_floor, 1, r->y_size - 2);
} else {
    CLAMP(x_floor, 0, r->y_size - 1);
}
```

src/game/room.c
```diff
@@ -590,19 +590,52 @@ int16_t Room_GetWaterHeight(int32_t x, int32_t y, int32_t z, int16_t room_num)
...
+        }
+
+        floor = &r->floor[z_floor + x_floor * r->x_size];
+        data = Room_GetDoor(floor);
+        if (data != NO_ROOM) {
+            r = &g_RoomInfo[data];
+        }
+    } while (data != NO_ROOM);
+
+    if (floor->floor != NO_HEIGHT / 256) {
```
lahm86, line 622: Should we update to use `STEP_L` in place of `256`?

src/game/room.c
```diff
@@ -590,19 +590,52 @@ int16_t Room_GetWaterHeight(int32_t x, int32_t y, int32_t z, int16_t room_num)
...
+        floor->floor += height >> 8;
+        if (floor->floor == floor->ceiling && floor->sky_room == NO_ROOM) {
             floor->floor = NO_HEIGHT / 256;
```
lahm86, line 625: And here too for consistency.

src/game/savegame/savegame_bson.c
```diff
@@ -541,6 +541,7 @@ static bool Savegame_BSON_LoadItems(
         if (obj->save_flags) {
             item->flags = json_object_get_int(item_obj, "flags", item->flags);
             item->timer = json_object_get_int(item_obj, "timer", item->timer);
+            LOG_DEBUG("item %d -> flags: %d", i, item->flags);
```
lahm86: Leftover debug call I think.

src/game/objects/traps/movable_block.c
```diff
@@ -133,13 +139,18 @@ void MovableBlock_Collision(
 {
     ITEM_INFO *item = &g_Items[item_num];
 
+    if (Item_TestAnimEqual(lara_item, LA_STOP)) {
```
lahm86: This has a minor side effect when using the fly cheat. If you start pushing/pulling and then enter and exit fly mode, the block stops mid animation. Not really an issue for normal gameplay, but it's sometimes useful when flying quickly through a level e.g. in Natla's Mines at the drill. But it's really minor imo, not sure if it's worth handling. I don't believe it's possible otherwise for Lara's push/pull animation to be interrupted like this.

lahm86 (changes requested):
The functionality looks largely good from my testing - just a few comments though as below.

There is one new issue which seems to have been introduced - you can now push blocks onto raised trapdoors it seems, after which they lose all collision and render badly.

20231014_122125_The_Cistern

If you door cheat, you can see the floor height is being applied to the room below.
20231014_122915_The_Cistern

src/game/objects/traps/movable_block.c
walkawayy: We can do an or check for `g_Lara.water_status == LWS_CHEAT`.

This added variable is needed because Core reused item status `IS_ACTIVE` for when Lara moves a block which caused one of the bugs. `IS_ACTIVE` also is set when a triggered block falls. So we need to be able to differentiate.

src/game/room.c
walkawayy: I will make a second, separate commit for this because `NO_HEIGHT / 256` is used in other spots like `Door_Shut` and `MovableBlock_TestDestination`.

I'll also make another commit to make some `MovableBlock` functions static that are only used in the file.

src/game/objects/traps/movable_block.c
walkawayy: Yes i think this change can be reverted. It was to try and get stacking working on room portals. But since that is not supported anymore, I don't think it's needed.

src/game/room.c
lahm86: It might be good to standardize throughout the codebase as sometimes you see `h / 256` and other times `h >> 8`. So maybe that's one to tackle under a separate issue.

src/game/objects/traps/movable_block.c
lahm86: We can raise another issue for adding an option to fix being able to push onto a slope - it's fixed in TR2 onwards anyway.

src/game/room.c
walkawayy: Alright I opened #1053 for this.

src/game/objects/traps/movable_block.c
```diff
@@ -133,13 +139,19 @@ void MovableBlock_Collision(
 {
     ITEM_INFO *item = &g_Items[item_num];
 
+    if (Item_TestAnimEqual(lara_item, LA_STOP)
+        || g_Lara.water_status == LWS_CHEAT) {
```
lahm86: This still causes a problem.

https://youtu.be/fmq2PW45mxE

WDYT about just doing this in the condition?

```c
if (item->current_anim_state == MBS_STILL) {
```

This allows the block's animation to complete if Lara enters DOZY. There are only three states for the block: still (covers falling too), being pushed, being pulled. And adding to this, maybe we could remove `m_BlockPushPull` and just look at the state instead?

walkawayy: Yes I think that works. I'll push the change.

lahm86 (approved):
LGTM. Tested all the scenarios from the test level in the issue itself, and each standard case in the OG levels, everything looks good.

src/game/objects/traps/movable_block.c
```diff
@@ -65,6 +72,143 @@ static bool MovableBlock_TestDoor(ITEM_INFO *lara_item, COLL_INFO *coll)
     return false;
 }
 
+static bool MovableBlock_TestDestination(ITEM_INFO *item, int32_t blockhite)
```
rr-: Can we call it block_height?

src/game/objects/traps/movable_block.c
```diff
@@ -65,6 +72,143 @@ static bool MovableBlock_TestDoor(ITEM_INFO *lara_item, COLL_INFO *coll)
...
+
+    if ((floor->floor << 8) != item->pos.y - blockhite) {
+        return false;
+    }
+
+    return true;
+}
+
+static bool MovableBlock_TestPush(
+    ITEM_INFO *item, int32_t blockhite, uint16_t quadrant)
```
rr-: Can we change quadrant to be of type DIRECTION?

src/game/objects/traps/movable_block.c
```diff
@@ -65,6 +72,143 @@ static bool MovableBlock_TestDoor(ITEM_INFO *lara_item, COLL_INFO *coll)
...
+        break;
+    case DIR_WEST:
+        x -= WALL_L;
+        break;
+    }
+
+    FLOOR_INFO *floor = Room_GetFloor(x, y, z, &room_num);
+    COLL_INFO coll;
+    coll.quadrant = quadrant;
+    coll.radius = 500;
```
rr-, line 121: Shouldn't 500 be `WALL_L / 2`?

src/game/objects/traps/movable_block.c
```diff
@@ -65,6 +72,143 @@ static bool MovableBlock_TestDoor(ITEM_INFO *lara_item, COLL_INFO *coll)
...
+    if (Collide_CollideStaticObjects(&coll, x, y, z, room_num, 1000)) {
```
rr-, line 122: Shouldn't 1000 be `WALL_L`?

src/game/objects/traps/movable_block.c
```diff
@@ -65,6 +72,143 @@ static bool MovableBlock_TestDoor(ITEM_INFO *lara_item, COLL_INFO *coll)
...
+    floor = Room_GetFloor(x, y - blockhite, z, &room_num);
+    if (((int32_t)floor->ceiling << 8) > y - blockhite) {
+        return false;
+    }
+
+    return true;
+}
+
+static bool MovableBlock_TestPull(
+    ITEM_INFO *item, int32_t blockhite, uint16_t quadrant)
```
rr-: Can we change `quadrant` to be of type `DIRECTION`?

src/game/objects/traps/movable_block.c
walkawayy: Tbh I am not sure we should change either of these. I only moved them to a static function. I didn't change any values. Changing this to from 500 to 512, and 1000 to 1024 could have gameplay consequences.

src/game/room.c
```diff
@@ -590,19 +590,42 @@ int16_t Room_GetWaterHeight(int32_t x, int32_t y, int32_t z, int16_t room_num)
...
+
+        if (z_floor <= 0) {
+            z_floor = 0;
+            CLAMP(x_floor, 1, r->y_size - 2);
+        } else if (z_floor >= r->x_size - 1) {
+            z_floor = r->x_size - 1;
+            CLAMP(x_floor, 1, r->y_size - 2);
+        } else {
+            CLAMP(x_floor, 0, r->y_size - 1);
+        }
```
rr-, line 613: Why do we clamp the floor to 1… size-2 rather than 0…size-1?
Can we provide a high-level overview of how the function operates as a comment?

rr- (approved):
Approving, I didn't test it, but it looks good to me. I'm glad that the changes turned out to be much simpler than we had started with. Just a confirmation – were the legacy TombATI saves tested?

src/game/room.c
walkawayy: Check out [Room_GetFloor](https://github.com/LostArtefacts/TR1X/blob/2078ede62e4c88141f202fabb87915dacac4dfcb/src/game/room.c#L233). This is extremely similar to that except it excludes the Y axis traversal.

The slight difference is I used `x_floor` and `z_floor` because the floor traversal is across those axes. And now that also matches TRView. I think the OG code used x/y as in x/y plane, but it was slightly confusing because there is also the 3D world's Y axis.

rr- (approved):

PR #1061, opened 2023-11-01
fix: wrong include path

vvs-:
#### Checklist

#### Description

Trivial one-liner fixing include path. Otherwise it doesn't compile for me.

### DIFF
--- a/src/game/overlay.c
+++ b/src/game/overlay.c
@@ -3,7 +3,6 @@
 #include "config.h"
 #include "game/clock.h"
 #include "game/inventory.h"
-#include "game/objects/common.h"
 #include "game/output.h"
 #include "game/screen.h"
 #include "game/text.h"
--- a/src/game/overlay.c
+++ b/src/game/overlay.c
@@ -11,7 +11,7 @@
 #include "global/const.h"
 #include "global/types.h"
 #include "global/vars.h"
-#include "src/math/matrix.h"
+#include "math/matrix.h"
 #include "util.h"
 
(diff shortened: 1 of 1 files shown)


vvs-:
And why on Earth that linter changed unrelated line there and failed?

rr-:
Thanks. While I agree that this linter has many shortcomings, it still brings enough value to the project to keep it. As far as I know, there are currently no superior alternatives available for C. At least the suggested modification it proposes is relatively minor.

vvs-:
> While I agree that this linter has many shortcomings, it still brings enough value to the project to keep it. As far as I know, there are currently no superior alternatives available for C. At least the suggested modification it proposes is relatively minor.

I'm sorry, With all respect, I'd argue that asking random contributors to fix long standing issues using their own unrelated changes just as an excuse is not a good idea. Running linter separately from all other activity, for example, might be better.

That said, I've managed to fix that warning which I have no idea why it even happened in the first place. But other people might not be so lucky.

rr-:
If the linters are isolated and not integrated into our regular development workflow, they may be overlooked or forgotten, defeating their purpose of catching issues early on.

I perceive your behavior to be somewhat rude – I am not requesting anything from you; it is my choice to dedicate my personal time away from my family to address this pull request and enable you to compile in a way that is not officially supported. Additionally, you indicated that you have read the contributor guidelines, but you have disregarded our recommended commit message format. From now on please be more mindful of the way you frame your messages. Thanks.

PR #1067, opened 2023-11-04
data: restore missing trigger in The Cistern

### LINKED ISSUE #1066: Cistern gold key soft-lock
https://github.com/LostArtefacts/TR1X/assets/4934209/9c81b017-9094-4ff0-abbb-86c6fde7f449

Consider this scenario in Cistern I came across in a recent playthrough: If you pick up the rusty key and open the door to the right with it, you can fall down the right side to avoid dying in the spikes, and then access the room of the gold key. The intended way is to first raise the water level, to avoid the spikes.
Picking up the gold key then doesn't open the door as it does when the room is flooded. There's then no way out of the room anymore.
Perhaps would it make sense to add the door opening to this scenario too? Granted it's quite rare as the player needs to be aware how to handle spikes drops. But since this is something introduced in the Qualopec secret, I would say a fix to the door would be good to have.
@lahm86 would love your thoughts, and thanks in advance as always 🙏

lahm86:
Resolves #1066.

#### Checklist

#### Description

Added a trigger to room 56 so that it matches its flipped room (72) and therefore prevents a softlock.


walkawayy (approved):
Wow lol I'm still learning about new bugs. Whack a mole continues. 😆

rr- (approved):

Issue #1060, opened 2023-10-24
Level select feature implementation just like PS1


rahul5162Volt:
It would be better to add level select feature just like in PS1 version.

walkawayy:
If you press the right arrow key on any of your saves while in the load game passport, you can access the level select menu.

![image](https://github.com/LostArtefacts/TR1X/assets/81546780/4f7adc23-dda0-4ba0-b2e2-e3490ba9d0f7)

walkawayy:
Gonna close this since we have level select per save file, a dev console to type in whatever level you'd like, and a cheat option and input combo to skip levels.

Issue #565, opened 2022-05-19
Feature request: Play Any Level


Nickelony:
Add a global json property like:
```json
"play_any_level": true,
```
Which would allow the player / level builder to select any available level which is defined in GameFlow.
In TR2 / 3 / 4 and 5, the level selection is in the "New Game" section:
![image](https://user-images.githubusercontent.com/20436882/169311596-7c067ae5-0a28-4308-bbb9-cccee5f2023a.png)

This is a very important feature, especially for level builders, because if they want to test a level, they currently have to modify GameFlow and move the desired level to the very top.

rr-:
> if they want to test a level, they currently have to modify GameFlow and move the desired level to the very top.

…or they can enable cheats and press <kbd>L</kbd> a few times.
Or they can save the game once in the last level, and use the Select Level function to start the game from any level.

walkawayy:
I think this can be also closed due to level select in the passport, the level cheat, and now the dev console command @rr-

Nickelony:
> I think this can be also closed due to the level cheat and dev console command @rr- 

This isn't necessarily a developer only feature. Some people like to have a game with 1 level only, but multiple outfits, so when you go into the New Game tab, you get to choose the outfit you want to play with. Those outfits are the same exact level n-times, just Lara's model in the .wad file is different. I know it's a bit of a hack, but levels like "Rome wasn't built in a day" use it.

Issue #553, opened 2022-05-13
Feature request: instant kill cheat


rr-:
For TRLE builders. I suggested it after watching Sabatu test The Great Pyramid and having to go through the torso fight again and again. This would be similar to the door open cheat. Alternative implementation: have all guns deal 10000 damage when the cheat is on.

Rapora9:
This is a good cheat. I think the best way to implement the cheat is by setting health to 0 for the enemy who Lara is targeting when the corresponding cheat button is pressed. This way the player (and builders) can for example
- easily decide which enemy they want to be killed
- kill an enemy from a long range
- get a lock to an enemy by shooting it, and then killing it even when behind a wall, in a different room etc.
- kill other enemies with normal ammo and others with one-shots.

If I had more time I'd be interested in doing this but right now I can't.

rr-:
To whoever takes it, please wait until #561 gets merged because it affects the input code significantly and would conflict with this ticket.

walkawayy:
I think this can be closed since it's now a console cheat as Lahm linked in the PR @rr-

PR #1068, opened 2023-11-06
Add a developer console

rr-:
#### Checklist

#### Description

[test-2023-11-06_22.52.09.webm](https://github.com/LostArtefacts/TR1X/assets/1045476/138410cf-83c7-4dab-b91c-c05f99519240)

I've added a console to help with debugging and testing. The console is dumb:
- It only supports ASCII characters in the range of A-Z, a-z, 0-9, `-` and `.`.
- In terms of editing, it supports arrows, home/end, and backspace only.
- There's a simple caret, but since we don't have `|` or a block character, I chose an `←` arrow sprite for it. I'd like to see this improved, as well as have some kind of prompt character (`$`, `›` or anything really). I avoided including `/` as part of the user prompt, because the keybinding is a user choice – should the player bind it to <kbd>\`</kbd>, it'd be weird to see the prompts start with `/`.
- There's _no_ support for any command history whatsoever.
- There's _no_ support for user keybindings.
- There's _no_ setting to disable it – should the player loathe this feature, they'll need unbind the `/` key.
- There's _no_ `/help`. The commands are documented in `COMMANDS.md`. This is kind of deliberate, as when we scale, the list of commands may not fit on the screen and implementing a scrollbar is the last thing I'd like doing :D
- There's a simple message log that disappears after a while. Each line stays for an amount of time relative to its length rather than a fixed period. This comes with a mechanism to avoid introducing gaps in the log whenever removing expired messages (eg logging something long, something short and something long won't create a blank line in the middle – removal of the middle line will wait till everything older gets cleaned up).

Important architectural notes:
- I've removed `Text_RemoveAll` altogether. Everyone is now responsible to manage their `TEXTSTRING`s by themselves. This made it easier to avoid certain class of problems, all the while not introducing regressions, it seems (but I'd welcome more testing).
- This PR breaks the custom of not depending on SDL in `game/`. I have mixed feelings about this, but these changes structure the code in a way that I don't hate a whole lot.
    - I wonder if we could shift away from `S_Input_GetState` towards event handlers in the fashion of this PR's additions.
- I've moved parts of game logic from `Input_Update` to `Game_ProcessInput` for gameplay-related stuff, and `Shell_ProcessInput` for window/shell-related stuff. This removed the necessity for `Option_ControlIsLocked`, but I haven't extensively tested it. The most sensitive place would be rebinding the controls for bilinear filter toggle and the like and I haven't really done that.
- The current design of `console_cmd.c` command parsing is extremely simplistic – we can't, for example, iterate through all possible invocations, which would be useful in creating an automatic `/help` command. Plus it'd be nice to be able to toggle more config options via commands, and currently this requires lots of repetition.
- I've simplified management of FPS counter rendering by delegating it to `output.c`. I had problems with disabling z-buffer and had to use `glDepthMask` – for some reason `glDisable(GL_DEPTH_TEST)` was not enough (?!). Since as usual we have similar code all over the place, I'm leaning towards either removing fps counter from `overlay.c` to its own file since its lifecycle no longer matches that of `Overlay_GameDrawInfo`'s, or moving `Output_DrawBlackOverlay` and perhaps fade functions into `overlay.c`, to group overlay routines in a single place. Finally I'd move `Overlay_BarDraw` meat to output.c.

LMK your thoughts about this and possible other commands we might want to include on day one.


Richard-L:
Just to say I think it looks super lovely..!!

src/game/console_cmd.c
```diff
@@ -0,0 +1,154 @@
...
+            if (!g_Objects[O_LARA].loaded || !g_LaraItem->hit_points) {
+                return true;
+            }
+            if (room_num < 1 || room_num > g_RoomCount) {
+                Console_Log(
+                    "Invalid room: %d. Valid rooms are 1-%d", room_num,
+                    g_RoomCount);
+                return true;
+            }
+            int16_t room_idx = room_num - 1;
```
lahm86: I'm wondering if 0-based room indexing will be better - it may just be personal preference, but I would tend to use this by checking the room number in trview then entering the command. Rooms aren't always next to each other sequentially so otherwise it could be a little confusing.

lahm86 (commented):
This is really nice and will be so useful 🚀 Just a small comment on room index basis. 

Some other commands that could be useful (and could maybe close some other issues - maybe these would be better under their own PRs later).

- `/flipmap N` : turn on flipmap N
- `/flipon N` : flips on map N
- `/flipoff N` : flips off map N
- `/kill N` : kill item N (#553)
- `/level N` : jump to the start of level N (#565, #1060)

src/game/console_cmd.c
rr-: Ideally I think we should be compatible with TRViewer and level builders mindsets and tooling. I kind of assumed it's 1-indexed without really checking. Do we know what TRViewer and TombEditor use?

lahm86: Yes, both are 0 based.

image
image

rr-: Thank you. Changed to 0 both for `/tp X` and `/pos`.

rr-:
> This is really nice and will be so useful 🚀 Just a small comment on room index basis.
> 
> Some other commands that could be useful (and could maybe close some other issues - maybe these would be better under their own PRs later).
> 
> * `/flipmap N` : turn on flipmap N
> * `/flipon N` : flips on map N
> * `/flipoff N` : flips off map N

In TR1 we only have a global flipmap (dictated by `g_FlipStatus`), is it OK for me to drop `N`?

> * `/kill N` : kill item N ([Feature request: instant kill cheat #553](https://github.com/LostArtefacts/TR1X/issues/553))

What should `N` refer to? `O_ALLIGATOR` eg `/kill 11` to kill all alligators? Or maybe distance near Lara to be more useful? Should it have special behavior if the player doesn't provide an argument?
I'd probably want a `/killall` to kill all hostile creatures.

Offtop: I'd want us to implement some kind of ability to bind keyboard shortcuts to specific commands as it'd be super useful, but it's going to be a nightmare, isn't it?

> * `/level N` : jump to the start of level N ([Feature request: Play Any Level #565](https://github.com/LostArtefacts/TR1X/issues/565), [Level select feature implementation just like PS1 #1060](https://github.com/LostArtefacts/TR1X/issues/1060))

lahm86:
Yes, that sounds good for the flipmap.

I was initially thinking `N` would indicate the item number for `kill`, but that could be tricky for ones that hatch. So the object number is probably best as you say. Maybe this set of commands?

`kill 11` - kill the first active croc in the item list (or maybe nearest?)
`kill` - kill the first active/nearest enemy
`killall 11` - kill all active crocs
`killall` - kill all active enemies

We have `m_CarrierObjects` in `carrier.c` which marks all enemy objects. Perhaps this could be moved somewhere shared for convenience.

walkawayy:
Haven't been able to test yet, but what about some kind of `giveitem X N` command where X is the item number and N is the quantity?

Having to rely on knowing items numbers for some of these commands is annoying but idk if there is a better way to handle that. Also, we have to make sure the item number in questions for any commands is available in the level file.

rr-:
@lahm86 @walkawayy added the suggested commands. I didn't include `/kill 11` as I don't think it will be used very often. I also fixed a couple of bugs introduced in previous iterations (a z-index inventory ring issue, malformed ammo counter, a crash when showing total stats).
I had to repurpose `GF_SELECT_GAME` to implement `/level X`. I also had to change up `Inv_Ring_Active` to get `/give` to understand custom strings such as `/give machine cog`. I'd appreciate additional testing around it – I did my own, but you never know…

rr-:
Alright, I think I'm done with bugfixing around the new commands.

src/game/console_cmd.c
```diff
@@ -0,0 +1,495 @@
...
+{
+    if (String_Equivalent(args, "all")) {
+        int32_t num = 0;
+        for (int16_t item_num = 0; item_num < g_LevelItemCount; item_num++) {
+            struct ITEM_INFO *item = &g_Items[item_num];
+            if (Object_IsObjectType(item->object_number, g_EnemyObjects)
+                && item->hit_points > 0) {
+                Effect_ExplodingDeath(item_num, -1, 0);
+                Sound_Effect(SFX_EXPLOSION_CHEAT, &item->pos, SPM_NORMAL);
+                Item_Kill(item_num);
```
lahm86, line 354: Could we also call `Carrier_TestItemDrops(item_num);`, similar to how it's done in `mutant.c`? Would be handy for testing drops.

lahm86 (commented):
This is fantastic! Super-speedy debugging, here we come 😄 I've left a small comment for `kill` re testing item drops.

I wonder too if it should be exclusive to normal gameplay/the title screen to avoid some complications. For example, if you run `give X` while in the inventory, the display doesn't update until you exit and re-enter, or change rings.

And while commands are accepted in demo mode, hitting enter quits the demo.

Not a big deal at all imo, it's just a thought.

walkawayy (commented):
Awesome work!

Crashes:
`/give 5` without an item number

Weird but not a big deal:
* Losing focus of the TR1X window with the console open requires you to press escape before input works again. Console remains up but you can't type anything.

Thoughts:
* Should `/fly` or `/fly off` turn off fly mode if Lara is in it? I guess we have the `walk` key to turn off.
* `abortion` blowing up Lara seems like a weird name xD. Also it's not mentioned in the `COMMANDS.md`.
* I think there should be a way to unbind or turn off the console IMO. I could see players hitting it accidentally and getting confused. Or just annoying people by accidentally going into console mode and getting themselves killed.

rr-:
>abortion blowing up Lara seems like a weird name xD. Also it's not mentioned in the COMMANDS.md.

It was meant to be an easter egg and refers to "No time for abortions now!" :) I was also thinking of naming it "/natlastinks" to punish the players who dare to type this blasphemy.

rr-:
>I wonder too if it should be exclusive to normal gameplay/the title screen to avoid some complications. For example, if you run give X while in the inventory, the display doesn't update until you exit and re-enter, or change rings.

I think the ring should be made more reactive. Ending the level should close it, too. But it's a part of broader refactorings that I'd prefer to tackle in a separate PR.

>And while commands are accepted in demo mode, hitting enter quits the demo.

I think I agree about the demos. In order to test the demos, I'd like to have a `/demo` command, probably paired with `/menu` and maybe `/quit`. However this PR has grown quite big so I'd avoid putting any more features / significant enhancements here, and likewise I'd create a new ticket for it once we land this current form.

>Crashes: `/give 5` without an item number

Fixed, good catch

>Losing focus of the TR1X window with the console open requires you to press escape before input works again. Console remains up but you can't type anything.

I can't reproduce this; is appears to be Windows shenanigans?

>Should /fly or /fly off turn off fly mode if Lara is in it? I guess we have the walk key to turn off.

Yes, we can quit fly mode with the walk key. Because of this, I don't think it's necessary to add `/fly off`.

>I think there should be a way to unbind or turn off the console IMO. I could see players hitting it accidentally and getting confused. Or just annoying people by accidentally going into console mode and getting themselves killed.

I have added a setting for it.

walkawayy (approved):
LGTM.

lahm86 (approved):

PR #1080, opened 2023-11-18
movable_block: fix stacking multiple blocks on room portals

### LINKED ISSUE #1079: Movable blocks can't be stacked more than once especially on room portals
Test level from Lahm:

[LEVEL1.zip](https://github.com/LostArtefacts/TR1X/files/13401446/LEVEL1.zip)

walkawayy:
Resolves #1079.

#### Checklist

#### Description
Credit goes to @asasas9500 for the changes! Fixed bugs when trying to stack multiple movable blocks. Tested Thor room, Rapora's test level, Lahm's test level. Could possibly use a bigger test level to make sure all push/pull scenarios work, but I think everything looks good from my testing so far.

### DIFF
--- a/src/game/objects/traps/movable_block.c
+++ b/src/game/objects/traps/movable_block.c
@@ -130,7 +130,7 @@ static bool MovableBlock_TestPush(
     }
 
     floor = Room_GetFloor(x, y - block_height, z, &room_num);
-    if (Room_GetCeiling(floor, x, y - STEP_L, z) > y - block_height) {
+    if (Room_GetCeiling(floor, x, y - block_height, z) > y - block_height) {
         return false;
     }
 
@@ -179,7 +179,7 @@ static bool MovableBlock_TestPull(
     }
 
     floor = Room_GetFloor(x, y - block_height, z, &room_num);
-    if (Room_GetCeiling(floor, x, y - STEP_L, z) > y - block_height) {
+    if (Room_GetCeiling(floor, x, y - block_height, z) > y - block_height) {
         return false;
     }
 
--- a/src/game/objects/traps/movable_block.c
+++ b/src/game/objects/traps/movable_block.c
@@ -244,10 +244,10 @@ void MovableBlock_Control(int16_t item_num)
     Item_Animate(item);
 
     int16_t room_num = item->room_number;
-    FLOOR_INFO *floor =
-        Room_GetFloor(item->pos.x, item->pos.y - 128, item->pos.z, &room_num);
-    int32_t height =
-        Room_GetHeight(floor, item->pos.x, item->pos.y - 128, item->pos.z);
+    FLOOR_INFO *floor = Room_GetFloor(
+        item->pos.x, item->pos.y - STEP_L / 2, item->pos.z, &room_num);
+    int32_t height = Room_GetHeight(
+        floor, item->pos.x, item->pos.y - STEP_L / 2, item->pos.z);
 
     if (item->pos.y < height) {
         item->gravity_status = 1;
--- a/src/game/objects/traps/movable_block.c
+++ b/src/game/objects/traps/movable_block.c
@@ -77,11 +77,13 @@ static bool MovableBlock_TestDestination(ITEM_INFO *item, int32_t block_height)
     int16_t room_num = item->room_number;
     FLOOR_INFO *floor =
         Room_GetFloor(item->pos.x, item->pos.y, item->pos.z, &room_num);
-    if (floor->floor == NO_HEIGHT / 256) {
+    if (Room_GetHeight(floor, item->pos.x, item->pos.y, item->pos.z)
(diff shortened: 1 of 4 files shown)


rr- (commented):
Hi, very elegant changes. Can you say why we're subtracting 128 in one place? Also does it work with save/load considering we didn't change anything in the savegame code? Thanks

Nit: let's avoid using name `f` and name the variable something more meaningful. BTW do we have some routines in the current codebase that could do the newly added iteration for us?

src/game/objects/traps/movable_block.c
```diff
@@ -123,12 +125,12 @@ static bool MovableBlock_TestPush(
...
     }
 
-    if (((int32_t)floor->floor << 8) != y) {
+    if (Room_GetHeight(floor, x, y, z) != y) {
         return false;
     }
 
     floor = Room_GetFloor(x, y - block_height, z, &room_num);
-    if (((int32_t)floor->ceiling << 8) > y - block_height) {
+    if (Room_GetCeiling(floor, x, y - STEP_L, z) > y - block_height) {
```
walkawayy: I think this should change to `y - y - STEP_L` for consistency.

src/game/objects/traps/movable_block.c
```diff
@@ -172,12 +174,12 @@ static bool MovableBlock_TestPull(
...
```
walkawayy: `y - block_height` also.

walkawayy:
Fixed the weird `STEP_L`.

`f` was named to match `Room_GetCeiling`. I could change both in a new commit maybe?
```
    FLOOR_INFO *f = floor;
    while (f->sky_room != NO_ROOM) {
```

From what I tested save/load seems to work fine!

src/game/objects/traps/movable_block.c
```diff
@@ -243,9 +245,9 @@ void MovableBlock_Control(int16_t item_num)
 
     int16_t room_num = item->room_number;
     FLOOR_INFO *floor =
-        Room_GetFloor(item->pos.x, item->pos.y, item->pos.z, &room_num);
+        Room_GetFloor(item->pos.x, item->pos.y - 128, item->pos.z, &room_num);
```
lahm86: Am I right in thinking that `-128` is used to ensure we are getting the actual sector where the block is positioned, and so by moving up a little, we avoid `Room_GetFloor` traversing portals and hence returning sectors further down the stack? Might be an idea to change this to `-STEP_L / 2` for clarity.

src/game/objects/traps/movable_block.c
```diff
@@ -243,9 +245,9 @@ void MovableBlock_Control(int16_t item_num)
...
     int32_t height =
-        Room_GetHeight(floor, item->pos.x, item->pos.y, item->pos.z);
+        Room_GetHeight(floor, item->pos.x, item->pos.y - 128, item->pos.z);
```
lahm86: And `-128` => `-STEP_L / 2` here too.

lahm86 (commented):
Really nice fix, bring on the block puzzles :)

One thought is whether this should be optional, because it also fixes another TR1 bug with blocks where you can sometimes push them onto slopes. An example is in Tihocan room 77, where you can push the block out onto the slope in room 89. We use this in the randomizer for some glitched secrets, and with this fix that's no longer possible. But I completely understand if we want to avoid a config option for this - I can replace the secret locations in the randomizer with something else (iirc we only have two that use this).

src/game/objects/traps/movable_block.c
walkawayy: Ya good call. Fixed.

walkawayy:
> Really nice fix, bring on the block puzzles :)
> 
> One thought is whether this should be optional, because it also fixes another TR1 bug with blocks where you can sometimes push them onto slopes. An example is in Tihocan room 77, where you can push the block out onto the slope in room 89. We use this in the randomizer for some glitched secrets, and with this fix that's no longer possible. But I completely understand if we want to avoid a config option for this - I can replace the secret locations in the randomizer with something else (iirc we only have two that use this).

Um I'm not really sure how easily it would be to support both. I didn't even realize the Randomizer used this bug. Is there anywhere in the OG or UB levels this is used? From what I can think of, the blocks are in flat rooms or lowered a step like in St Francis Folly first room. I kinda want to avoid a config option unless it's needed in an OG or UB level. :/

lahm86:
It's not used in speedruns or anything like that - there's no benefit in exploiting the glitch in the OG levels. I'm fine with tweaking the rando secrets, it's unique to TR1 anyway so not a common thing that players come across. Agreed that this would become overly complicated as a config option.

rr-:
If possible let's avoid making it configurable. IMO it'd be feeding our config creep that's already out there.

rr- (approved):

lahm86 (approved):

PR #1100, opened 2023-12-05
passport: allow restart level in save crystal mode

### LINKED ISSUE #1099: Restart Level passport page not visible when save crystals are enabled
I was going to ask about implementing the Restart Level feature from the PSX version, but I see now that it is available (on death), it's just not visible when save crystals are enabled.

While I'm here, I might as well also ask about the possibility of implementing the Restart Level page without having to die first.

walkawayy:
Resolves #1099.

#### Checklist

#### Description
Added the restart level option in the passport at all times for save crystal mode. Refactored the passport code to be clearer.

### DIFF
--- a/src/game/inventory/inventory.c
+++ b/src/game/inventory/inventory.c
@@ -841,11 +841,13 @@ static int32_t Inv_ConstructAndDisplay(int inv_mode)
                      && g_SavedGamesCount) /* f6 menu */
                     || g_InvMode == INV_DEATH_MODE /* Lara died */
                     || (g_InvMode == INV_GAME_MODE /* esc menu */
-                        && g_GameInfo.passport_page
-                            != PASSPORT_PAGE_2 /* but not the save page */
+                        && g_GameInfo.passport_selection
+                            != PASSPORT_MODE_SAVE_GAME /* but not save page */
                         )
                     || g_CurrentLevel == g_GameFlow.gym_level_num /* Gym */
-                    )) {
+                    || g_GameInfo.passport_selection
+                        == PASSPORT_MODE_RESTART)) {
+                LOG_DEBUG("Fade to black!");
                 Output_FadeToBlack(false);
             } else {
                 Output_FadeToTransparent(false);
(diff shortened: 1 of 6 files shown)


rr- (approved):

lahm86 (approved):

walkawayy:
Marking this as draft since at this point it's a big refactor. I can't sayI tested every scenario yet, but a lot of the behavior seems to work as expected. This PR allows the level restart at any time if in save crystal mode. I think the code is cleaner but idk. I tried to avoid a COMPLETE rewrite and kept a lot of the subfunctions. But lmk if you think it's better or worse. Hopefully not way worse because I do not want to touch this code again lol. Also probably best viewed as a standalone file since the diff is a mess.

src/game/inventory/inventory.c
```diff
@@ -900,81 +902,27 @@ static int32_t Inv_ConstructAndDisplay(int inv_mode)
...
-                    // page 2: save game
-                    Savegame_Save(g_GameInfo.current_save_slot, &g_GameInfo);
-                    Config_Write();
-                    return GF_NOP;
-                }
-            } else {
-                // page 3: exit to title
-                return GF_EXIT_TO_TITLE;
-            }
+            return GF_EXIT_TO_TITLE;
```
rr-, line 928: Maybe we could convert it to a switch? Do you reckon we might need more precise conditions in the future?

src/game/option/option_passport.c
```diff
@@ -392,178 +565,124 @@ void Option_Passport(INVENTORY_ITEM *inv_item)
...
     if ((inv_item->goal_frame - inv_item->open_frame) % 5) {
-        page = PASSPORT_PAGE_FLIPPING;
+        m_PassportStatus.page = PAGE_FLIPPING;
     }
 
-    if (g_InvMode == INV_SAVE_MODE || g_InvMode == INV_SAVE_CRYSTAL_MODE) {
-        g_InputDB.menu_left = 0;
-        g_InputDB.menu_right = 0;
-    }
+    // TODO Out of bounds index?
```
rr-: Hmm I don't see why it should be out of bounds?

src/game/option/option_passport.c
```diff
@@ -140,118 +159,129 @@ static void Option_PassportShutdownText(void)
...
+        m_PassportStatus.modes[PAGE_3] = PASSPORT_MODE_EXIT_TITLE;
+        break;
+    default:
+        m_PassportStatus.mode = PASSPORT_MODE_BROWSE;
+        m_PassportStatus.pages[PAGE_1] = g_SavedGamesCount > 0;
+        m_PassportStatus.pages[PAGE_2] = true;
+        m_PassportStatus.pages[PAGE_3] = true;
+        m_PassportStatus.modes[PAGE_1] = PASSPORT_MODE_LOAD_GAME;
+        m_PassportStatus.modes[PAGE_2] = PASSPORT_MODE_NEW_GAME;
+        m_PassportStatus.modes[PAGE_3] = PASSPORT_MODE_EXIT_TITLE;
```
rr-: This is brilliant. Let's just add blank lines after each `break;` for easier vertical navigation.

rr- (commented):
The code looks great. I'm assuming that every `INV_MODE` and `PASSPORT_MODE` enum values have been tested.

I noticed that there is some degree of redundancy in the internal state of the passport - we have `.modes` and `.mode`; could we remove `m_PassportStatus.mode` and use only `m_PassportStatus.modes[m_PassportStatus.page]` instead? It seems to be mainly needed for `PASSPORT_MODE_BROWSE`, which is in turn never used in `.modes`; can we consolidate this together? Pinging @lahm86 for brainstorming – it might well be case that it's not worth it and the current form is better.

src/game/inventory/inventory.c
walkawayy: Yes to a switch. Idk if we will need more in the future or not, but I don't think we will?

src/game/option/option_passport.c
walkawayy: Ya this doesn't apply. I'll remove it. Originally I had an idea to make the pages array a length of 5 and always mark pages 0 and 4 as unavailable with pages 1, 2, and 3 as normal. Then we could just check `if pages[page-1]`. However, this would throw off all the frame number calculations and might be a bit confusing, so I scrapped the idea.

walkawayy (commented):
I still need to test some scenarios like a fresh install with no existing saves. But I tested normal, save crystal mode, F5 anf F6 quick save/load, passport in the gym including quick save/load, level select, and story so far. From what I tested it looked good shockingly.

Yes, as far as names, I have `mode` and `modes, and `page` and `pages`. I originally had stuff like `current_page` or `page_availability` but the names were super long especially attached to `m_PassportStatus`. So if you have any ideas on names, I am definitely open to suggestions.

And yes, I tried to just keep all the modes in one enum even though stuff like `PASSPORT_MODE_BROWSE` is only used for the current `mode`. Some kind of way to differentiate flipping through the pages (browse) vs confirming the button and bringing up the requestor is needed I think.

I did mostly keep the subfunctions like `Option_PassportShowSaves`. If you have any suggestions for those lmk.

Lastly, I want to dig in and see if g_OverlayFlag can be removed. I'm hoping Lahm's added game status can be used instead, but I haven't looked into that yet. I wanted to get some feedback on the current changes which are already a lot before nuking more code lol.

lahm86:
🚀 I really like this. Re names, I wonder if maybe making a struct like `PASSPORT_CURRENT` to store the individual `page` and `mode` fields, which are then accessed through `m_PassportStatus.current.mode = PASSPORT_MODE_BROWSE;` and suchlike. Although that may be a bit "wordier" in the long run, but it might be a little clearer for differentiating between these fields and the arrays?

walkawayy:
> 🚀 I really like this. Re names, I wonder if maybe making a struct like `PASSPORT_CURRENT` to store the individual `page` and `mode` fields, which are then accessed through `m_PassportStatus.current.mode = PASSPORT_MODE_BROWSE;` and suchlike. Although that may be a bit "wordier" in the long run, but it might be a little clearer for differentiating between these fields and the arrays?

Not a bad idea! I'll let rr chime in before doing anything.

rr-:
While this suggestion improves on semantics, it doesn't address my original concern – that `current mode` is not the same as `modes[current page]`, and it can hold different values. Is there an easy way to deal with just a single `modes[current page]` that would hold the necessary enum values? If not, I'd leave it as-is.

As for `PASSPORT_CURRENT`, we don't necessarily need it named – surprisingly C allows anonymous nested structs:
```c
typedef struct PASSPORT_INFO {
    struct {
        PASSPORT_PAGE page;
        PASSPORT_MODE mode;
    } current;
    PASSPORT_MODE modes[4];
} PASSPORT_INFO;
```
We ocassionally use this idiom throughout the code.

walkawayy:
> While this suggestion improves on semantics, it doesn't address my original concern – that `current mode` is not the same as `modes[current page]`, and it can hold different values. Is there an easy way to deal with just a single `modes[current page]` that would hold the necessary enum values? If not, I'd leave it as-is.
> 
> As for `PASSPORT_CURRENT`, we don't necessarily need it named – surprisingly C allows anonymous nested structs:
> 
> ```c
> typedef struct PASSPORT_INFO {
>     struct {
>         PASSPORT_PAGE page;
>         PASSPORT_MODE mode;
>     } current;
>     PASSPORT_MODE modes[4];
> } PASSPORT_INFO;
> ```
> 
> We ocassionally use this idiom throughout the code.

What about:
```
typedef struct PASSPORT_STATUS {
    PASSPORT_PAGE page;
    PASSPORT_MODE mode;
    bool page_available[PAGE_COUNT];
    PASSPORT_MODE page_role[PAGE_COUNT];
} PASSPORT_STATUS;
```
?
I think that makes it more clear what each variable is and we don't need to introduce `current`. And the names aren't too long. I'll have to think about a way not having `PASSPORT_MODE_BROWSE`. For the moment I think it's needed.

rr-:
Sounds good to me! This may be a subtle change in semantics but I think it ticks all the right boxes.

rr- (approved):

src/global/types.h
```diff
@@ -1181,19 +1181,24 @@ typedef enum GAME_BONUS_FLAG {
...
-    PASSPORT_PAGE_FLIPPING = -1,
-    PASSPORT_PAGE_1 = 0,
-    PASSPORT_PAGE_2 = 1,
-    PASSPORT_PAGE_3 = 2,
-    PASSPORT_PAGE_COUNT = 3,
+    PAGE_FLIPPING = -1,
+    PAGE_1 = 0,
+    PAGE_2 = 1,
+    PAGE_3 = 2,
+    PAGE_COUNT = 3,
```
rr-, line 1188: Let's keep the lengthy name, or move the type to `option_passport.c`.

lahm86 (commented):
All looks good from testing 👍

I did find one issue but it's unrelated to these changes as you can replicate it in the current release. If you have save crystals enabled and move to a new level, if you are holding left or right while the inventory opens (to offer you the chance to save at the start of the level) you can stop on, for example, the audio settings but if you then press escape the inventory closes and you have therefore lost the opportunity to save. Best tackled under its own issue do you think?

rr-:
Yes, let's tackle it separately.

lahm86:
Raised under #1108.

pas-de-2:
I built the latest master to test this out. Issues I encountered:

With `"enable_game_modes" : false`:
* can't start a new game, the game fades to black and then returns to the title screen

With `"enable_save_crystals" : true`:
* the Load Game page in the in-game passport doesn't open the menu to choose a save, instead it just loads your last save, or if you've restarted the level, it restarts the level again
* Exit to Title closes the game instead of returning to the title menu (also happens with `"enable_save_crystals" : false`)
* the ingame passport defaults to the Load Game page when opened instead of the Restart Level page (minor inconsistency with PS1 TR)

rr-:
Thanks for the report; this has to do with the recent passport refactor. I reported the bugs as #1124 #1126  and prepared pending fixes for them.

PR #1103, opened 2023-12-06
movable_block: test animation index over state

### LINKED ISSUE #1101: TR1X Bug: Lara's push animation can break
**Build :**
Release 3.0.3

**Description :**
In the latest release, you can break Lara's pushing animation by interacting with pushable blocks in a specific way, allowing Lara to freely walk around while the block moves forward on its own. This can lead to other issues as seen in the video.

https://github.com/LostArtefacts/TR1X/assets/72892666/c21d974c-858d-40ea-b7d6-4a7d05c8b472

I've only tried this in City of Vilcabamba 2, so I'm only guessing that this can happen with any pushable block.
I'm also guessing this is a TR1X Bug, since I couldn't find any info about it online.

**Steps to replicate :**
1- Get in front of a pushable block.
2- Grab and let go of it while trying to move forward. If you don't clip through the block, then repeat this step.
3- And voilà.

lahm86:
Resolves #1101.

#### Checklist

#### Description

`LS_PP_READY` refers to Lara's animation state when grabbing OR releasing a pushblock. Prior to #1006, `LF_PPREADY` was an absolute frame index and so the check for this would only pass after the grabbing animation. With relative indexing, the check can pass in either animation, and so holding action and forwards/backwards after releasing a block would allow the test to pass and the block to move independently. We now test for the specific grab animation rather than release.

I've checked other cases where this could be an issue and this is the only one. The `Item_TestFrame` calls in the PR files below highlight the animation states - `LS_PICKUP` could have potentially been an issue, but there is already a guard of sorts around this by testing water/not water, plus the frame indices are different too. `LS_PP_READY`/`LF_PPREADY` was ultimately the perfect storm.

https://github.com/LostArtefacts/TR1X/pull/1006/files
https://github.com/LostArtefacts/TR1X/blob/develop/src/game/objects/general/pickup.c#L104

I've also replaced some remaining 1024 usage in `movable_block.c` with the `WALL_L` const.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - fixed missiles damaging Lara when she is far beyond their damage range (#1090)
 - fixed Lara's meshes being swapped in the gym level when using the console to give guns (#1092)
 - fixed Midas's touch having unrestricted vertical range (#1094)
+- fixed pushblocks moving freely if Lara releases but tries to regrab during the release animation (#1101, regression from 3.0)
 
 ## [3.0.3](https://github.com/LostArtefacts/TR1X/compare/3.0.2...3.0.3) - 2023-11-27
 - fixed underwater shadow effects rendering always in the same way rather than at random (#1081)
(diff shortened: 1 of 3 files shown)


walkawayy (approved):
Works well thanks! Think we should change the pickup check in the future to an anim check just in case? That state check mirrors TR4/TR5 code, but I wouldn't be surprised if it could be prone to bugs.

rr- (approved):

PR #1107, opened 2023-12-09
flame: tie FX to Lara's room when on fire

### LINKED ISSUE #1106: OG bug: flames are not always drawn when Lara is on fire
If Lara catches fire and then enters a room other than the one where the flames originated, the flames are no longer drawn. They re-appear if she enters the original room, or when she is simply facing that room. Ultimately, the flame FX remain tied to the original room so are only drawn when that room is drawn.

https://github.com/LostArtefacts/TR1X/assets/33758420/92293928-02b2-4bb4-bb14-0cf40340c216

lahm86:
Resolves #1106.

#### Checklist

#### Description

This ensures that if Lara enters a different room while on fire, that the flame FX are tied to that room so that they will always be drawn.


rr- (approved):

walkawayy (approved):
These flames!

PR #1114, opened 2023-12-10
gameflow: enhance save crystals and game mode selections

### LINKED ISSUE #1102: Save crystal mode and Unfinished Business
Ran into a couple issues:

- Unfinished Business doesn't have save crystals, so it doesn't really make sense for the config option to prevent saving when launching the game with `-gold`.

- This led me to try disabling `enable_save_crystals` and only enabling `force_enable_save_crystals` in `TR1X_gameflow.json5`. Problem with this method is that `force_enable_save_crystals` actually overwrites your config's `enable_save_crystals` value, so if you play base TR and then later launch UB, save crystal mode will be enabled again.

I suppose another possibility would be to import the save crystal locations from b122251's PSX port of UB.

rr-:
Resolves #1102

#### Checklist

#### Description

- changed `force_enable_save_crystals` to `force_save_crystals` to let custom level authors force-disable save crystals
- changed `force_disable_game_modes` to `force_game_modes` to let custom level authors force-enable game modes
- fixed inability to save in Unfinished Business in crystals mode as UB doesn't have crystals


bin/cfg/TR1X_gameflow.json5
```diff
@@ -7,8 +7,8 @@
     "main_menu_picture": "data/titleh.png",
     "savegame_fmt_legacy": "saveati.%d",
     "savegame_fmt_bson": "save_tr1_%02d.dat",
-    "force_disable_game_modes": false,
-    "force_enable_save_crystals": false,
+    "force_game_modes": null,
+    "force_save_crystals": false,
```
lahm86: I think this should be `null` - with `false` there is no way to play with save crystals without manually changing this line (so the config option is ignored).

rr-: That's correct, my mistake with copypasting.

lahm86 (approved):

walkawayy (approved):
Looks good

walkawayy:
Only comment for changelog is, if I understand this correctly, the custom level author can force enable or also force disable crystals or modes. The changelog makes it seem like only it only allows force enable (which would mean the user could then choose in the config).

rr-:
Thanks – I'll change it

walkawayy (approved):

PR #1128, opened 2023-12-19
input: improve menu navigation for gamepad

### LINKED ISSUE #1104: Additional controller binds for menu back and menu confirm
Currently, the only way to exit a menu like the passport or sound options with a gamepad is to press the Inventory button (Share on a DS4), which is strange and unintuitive. This might be a regression from 2.12, when custom keybinding support was added, since [previously](https://github.com/walkawayy/TR1X/blob/bc2bd9cb864e4b7374923abe796952c8f1cda0aa/src/specific/s_input.c#L569C19-L569C19
) you could press Roll or Start or Touchpad to exit menus as well.

Suggestion: make both the Roll and Weapon Draw buttons trigger `menu_back`, to cover people used to modern gamepad controls (where Roll/Circle = deselect) and people used to PlayStation TR2-5 controls (where Draw/Triangle = deselect).

pas-de-2:
#### Checklist

#### Description

This is what I was getting at with https://github.com/LostArtefacts/TR1X/issues/1104. I see walkawayy wants to expand keybind customization, but these should still serve as much saner defaults for gamepad players.


src/game/inventory/inventory.c
```diff
@@ -521,7 +521,7 @@ static int32_t Inv_ConstructAndDisplay(int inv_mode)
             }
 
             if (start_level != -1 || start_demo
-                || (g_InputDB.option && g_InvMode != INV_TITLE_MODE)) {
+                || ((g_InputDB.option || g_InputDB.roll || g_InputDB.draw) && g_InvMode != INV_TITLE_MODE)) {
```
rr-: Shouldn't we replace all these checks with just a single `g_InputDB.menu_back`?

pas-de-2: Doh, yes of course

rr- (approved):
We try to start changelog entries from a few select keywords to make it easier on the players eyes, so let's reword it to start with `added (…)` and put it above `fixed` entries. For example: `added an ability to back out of menus with the roll and weapon draw buttons (#1104)`.

I'm not sure how I feel about having this behavior be present also on the keyboard, but I can't think of an elegant way to specialize it off the top of my head.

Other than that LGTM, but I don't play on a pad, so let's await an opinion from other maintainers as well.

pas-de-2:
>I'm not sure how I feel about having this behavior be present also on the keyboard

Makes sense, fixed it so it only applies to controller input

lahm86 (approved):
LGTM. I've tried it and although (as a keyboard player) it doesn't feel native to me, I can see the benefit for pad players.

One thing this has a very slight effect on is input buffering (for rolling or drawing weapons), but it just means you can buffer those inputs more easily (one less key to press); so it's not an issue. Buffering these inputs would be very rare IMO too, aside from perhaps debugging.

_Edit - just noticed the latest commit, so above is irrelevant for KB._

walkawayy (commented):
I'm good with this as a stop gap. But, the reason I removed these keybinds originally was because people can rebind all the controls now. So, say a player rebinds roll to X and draw weapons to like L1 for some reason, those buttons will now function as `menu_back` even though the intention of this PR was to make circle and triangle do menu back. Plus on PS1 and most PS2 games, triangle was back. But, since PS3/PS4. circle is now back. That puts us in a weird spot since this is a legacy game, but we have modern controls stuff.

Maybe it would be better to do this like the keyboard controls? Where they use hard coded buttons, but only for the menu controls. Users can mess up their controls this way if they rebind stuff like `menu_confirm` to triangle, but that's why the reset controls button exists. Idk what are everyone's thoughts? I think something like this might work in `S_Input_GetControllerState`. I can test.
```
        state.toggle_bilinear_filter |= S_Input_GetBindState(INPUT_ROLE_BILINEAR, cntlr_layout_num);
        state.menu_back              |= S_Input_JoyBtn(SDL_CONTROLLER_BUTTON_Y);
        state.menu_back              |= S_Input_JoyBtn(SDL_CONTROLLER_BUTTON_B);
        state.menu_back              |= S_Input_JoyBtn(SDL_CONTROLLER_BUTTON_TOUCHPAD);
```

rr- (approved):

rr-:
Thanks for looking into it @walkawayy . I refrain from commenting on the look'n'feel of gamepads since I haven't played TR on a gamepad in 25 years, so it'd be best to gather voice from other gamepad players.

walkawayy:
> I'm good with this as a stop gap. But, the reason I removed these keybinds originally was because people can rebind all the controls now. So, say a player rebinds roll to X and draw weapons to like L1 for some reason, those buttons will now function as `menu_back` even though the intention of this PR was to make circle and triangle do menu back. Plus on PS1 and most PS2 games, triangle was back. But, since PS3/PS4. circle is now back. That puts us in a weird spot since this is a legacy game, but we have modern controls stuff.
> 
> Maybe it would be better to do this like the keyboard controls? Where they use hard coded buttons, but only for the menu controls. Users can mess up their controls this way if they rebind stuff like `menu_confirm` to triangle, but that's why the reset controls button exists. Idk what are everyone's thoughts? I think something like this might work in `S_Input_GetControllerState`. I can test.
> 
> ```
>         state.toggle_bilinear_filter |= S_Input_GetBindState(INPUT_ROLE_BILINEAR, cntlr_layout_num);
>         state.menu_back              |= S_Input_JoyBtn(SDL_CONTROLLER_BUTTON_Y);
>         state.menu_back              |= S_Input_JoyBtn(SDL_CONTROLLER_BUTTON_B);
>         state.menu_back              |= S_Input_JoyBtn(SDL_CONTROLLER_BUTTON_TOUCHPAD);
> ```

Hm this method does cause issues in the controls menu. Hold Triangle to rebind doesn't work and the controls menu goes back instead. Also, you can't exit the inventory like this, you can only close the open menus like the passport. Is that expected / desired behavior?

pas-de-2:
> Maybe it would be better to do this like the keyboard controls? Where they use hard coded buttons, but only for the menu controls.

I don't mind this idea, personally, but we also need to hardcode the A(cross) button as menu confirm, otherwise you run into an issue when the action button is bound to the B or Y button, eg. I'm using TR Saturn controls on an 8bitdo Saturn pad, and action is already bound to B.

> Hm this method does cause issues in the controls menu. Hold Triangle to rebind doesn't work and the controls menu goes back instead. 

This is a conflict for sure. We can try a different button mapping to reset gamepad binds? Since A/B/Y are used, X(square)? Or Start?

> Also, you can't exit the inventory like this, you can only close the open menus like the passport. Is that expected / desired behavior?

You should be able to, and you can if you change this line in `Inv_ConstructAndDisplay` from `g_InputDB.option` to `g_InputDB.menu_back`, like so:
https://github.com/LostArtefacts/TR1X/blob/eed90afdc6bbce477d1257164e79e9e1e01baa5e/src/game/inventory/inventory.c#L523-L524

walkawayy:
> > Maybe it would be better to do this like the keyboard controls? Where they use hard coded buttons, but only for the menu controls.
> 
> I don't mind this idea, personally, but we also need to hardcode the A(cross) button as menu confirm, otherwise you run into an issue when the action button is bound to the B or Y button, eg. I'm using TR Saturn controls on an 8bitdo Saturn pad, and action is already bound to B.
> 
> > Hm this method does cause issues in the controls menu. Hold Triangle to rebind doesn't work and the controls menu goes back instead.
> 
> This is a conflict for sure. We can try a different button mapping to reset gamepad binds? Since A/B/Y are used, X(square)? Or Start?
> 
> > Also, you can't exit the inventory like this, you can only close the open menus like the passport. Is that expected / desired behavior?
> 
> You should be able to, and you can if you change this line in `Inv_ConstructAndDisplay` from `g_InputDB.option` to `g_InputDB.menu_back`, like so:
> 
> https://github.com/LostArtefacts/TR1X/blob/eed90afdc6bbce477d1257164e79e9e1e01baa5e/src/game/inventory/inventory.c#L523-L524

Let's avoid X for reset button, so we can also implement X as the hard coded menu confirm button along with the addition of the Circle/Triangle buttons as `menu_back`.

Why don't we make the reset all button something like R1 (right shoulder). In `option_control.c` we can change:
`#define RESET_ALL_BUTTON "rightshoulder"`
This works for me and on a Playstation controller shows the R1 button. That should avoid any conflicts.

And yes, let's make that `inventory.c` change as well IMO.

pas-de-2:
Pushed hardcoded version. Tests well with a DS4 and 8bitdo's M30, at least.

RE: the touchpad, I think it might make more sense to make this button the default 'Inventory' button for DS4/Dual Sense controllers, as that has been its function in games ever since the Select button was replaced with the 'Share/Create' buttons. For now, the player can just rebind to this.

walkawayy:
> Pushed hardcoded version. Tests well with a DS4 and 8bitdo's M30, at least.
> 
> RE: the touchpad, I think it might make more sense to make this button the default 'Inventory' button for DS4/Dual Sense controllers, as that has been its function in games ever since the Select button was replaced with the 'Share/Create' buttons. For now, the player can just rebind to this.

Ya looks good. My new question is if Square should function as confirm? That personally doesn't make sense for me. And that will also affect the unbinding button in the controls menu.

pas-de-2:
Oh lol, I wasn't sure about adding square either tbh, but I misread you talking about wanting to bind the X button earlier as the Xbox X button and not the PlayStation cross. I'll fix it.

walkawayy:
Ok this looks good to me. @lahm86  thoughts?

src/game/input.c
```diff
@@ -112,7 +112,6 @@ void Input_Update(void)
     g_Input.menu_down |= g_Input.back;
     g_Input.menu_left |= g_Input.left;
     g_Input.menu_right |= g_Input.right;
-    g_Input.menu_confirm |= g_Input.action;
```
lahm86, line 115: This is a breaking change for keyboard players IMO - personally, I use action all the time in menus, so I would have to re-learn using enter. Is there a way to allow KB players to retain action for menus?

lahm86 (changes requested):

src/game/input.c
walkawayy: Ah good catch. I only tested controller.

pas-de-2: Sure thing, moved the check to S_Input_GetCurrentState so action only doubles as menu_confirm for keyboard players

walkawayy: That solution makes sense to me.

walkawayy (approved):

lahm86 (approved):

PR #1132, opened 2024-01-05
inventory_func: fix items not adding if sprite is missing

### LINKED ISSUE #1130: [3.0] Lara spawning without pistols in Feder's Oasis
https://www.trle.net/sc/levelfeatures.php?lid=3481

20231225_110152_Caves

Noticed Lara spawns in this TRLE without her pistols. She can draw them at first, but if you ever equip another weapon like a shotgun, you can't reselect the pistols.

Issue starts happening in TR1X version 3.0, in 2.16 the pistols load fine.

EDIT: tracked it down to commit https://github.com/LostArtefacts/TR1X/commit/56e1ae15229645972dc52349f01e079e627c84dc and the addition of this check in `Inv_AddItem`:

https://github.com/LostArtefacts/TR1X/blob/56e1ae15229645972dc52349f01e079e627c84dc/src/game/inventory/inventory_func.c#L14-L16

walkawayy:
Resolves #1130.

#### Checklist

#### Description
Fixed items not being added to inventory if the sprite is missing from the level file. This only checks if the inventory 3D model is loaded since the sprite is only needed for pickups if 3D pickups aren't enabled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - fixed flames not being drawn when Lara is on fire but leaves the room where she caught fire (#1106)
 - fixed being able to deselect the passport in quick save, quick load, save crystal, and death modes (#1108)
 - fixed inability to save in Unfinished Business in crystals mode as UB doesn't have crystals (#1102)
+- fixed items not being added to inventory if the sprite is missing from the level file (#1130)
 
 ## [3.0.5](https://github.com/LostArtefacts/TR1X/compare/3.0.4...3.0.5) - 2023-12-13
 - fixed crash when pressing certain keys and the console is disabled (#1116, regression since 3.0)
--- a/src/game/inventory/inventory_func.c
+++ b/src/game/inventory/inventory_func.c
@@ -11,7 +11,7 @@
 bool Inv_AddItem(int32_t item_num)
 {
     int32_t item_num_option = Inv_GetItemOption(item_num);
-    if (!g_Objects[item_num].loaded || !g_Objects[item_num_option].loaded) {
+    if (!g_Objects[item_num_option].loaded) {
         return false;
     }
 


rr-:
We need to address the rendering of sprite items on the ground and in the UI when the player has 3D pickups disabled.
I'd either use some dummy sprite, or fall back to a 3D model that has to be there for the inventory anyway.

walkawayy:
> We need to address the rendering of sprite items on the ground and in the UI when the player has 3D pickups disabled. I'd either use some dummy sprite, or fall back to a 3D model that has to be there for the inventory anyway.

If I am reading this properly, I think this check is only needed for the console command that gives items? I believe pickups should be handled during the level loading process. If the code gets to this point, Lara has already picked up the item (and the sprite/3D model displayed in the overlay), or the player is using the console command. In both cases, I think the game only needs the 3D model for the inventory. The level file should probably include both pistol 3D model and sprite, but since the game technically only needs the 3D models for inventory, I don't think we need to brick starting items that don't have sprites if Lara never has to pick them up.

Maybe @lahm86 can confirm if what I am saying is accurate lol.

rr-:
I meant that the fix is good – but we also should draw _something_ in the overlay UI, when the user player up an item that has no sprite and they have 3D pickups disabled. Same for drawing the actual item on the ground. As is, the way I understand this fix is going to work, is that it'll draw invisible items and in Natla's mines for example the pistols in the cabin are gonna be transparent, leaving the player to guess their tile.

walkawayy:
> I meant that the fix is good – but we also should draw _something_ in the overlay UI, when the user player up an item that has no sprite and they have 3D pickups disabled. Same for drawing the actual item on the ground. As is, the way I understand this fix is going to work, is that it'll draw invisible items and in Natla's mines for example the pistols in the cabin are gonna be transparent, leaving the player to guess their tile.

No Natla's Mines is fine (see screenshot). This fix should only affect items already being adding to the inventory (`Inv_AddItem`). This inventory function should never require the sprite model AFAIK.

`Overlay_DrawPickups` handles all the overlay code and `Object_DrawPickupItem` / `Object_DrawSpriteItem` handle drawing the models / sprites in the game world.

image

rr- (approved):
Thanks, this makes sense. Seeing that it works either way, I don't really understand why we have a separate object id for the sprite versions then, but maybe it's better to not delve too much into it :D

lahm86 (approved):

PR #1135, opened 2024-01-13
lara: prevent look reset on target triggers

### LINKED ISSUE #1026: TR1X bug: Look at Item function is less rigid
**Update:** This bug occurs when `Look while Running` is enabled.

___________________________________

First discovered in Temple of the Cat.  Basically have Lara stand in this spot and face the pyramid, then start tapping left and stop the moment the camera locks onto item.

TR1X:
1 2 TR1X

TombATI, with the same steps repeated.  Notice the circled boulder - that is what the camera should lock on, but it's not visible in the above screenshot:
1 1 TombATI

TR1X, if I have Lara turn the other way by tapping right until the camera locks onto item:
2 2 TR1X

TombATI, with the same steps repeated as above.  Once again the circled boulder is out of view in the prior screenshot:
2 1 TombATI

After this, I investigated Caves and can confirm the bug happens there as well.  When you reach the area with the opening, have Lara face away, then start tapping left until the camera locks onto item.

TR1X:
3 2 TR1X

TombATI, same spot:
3 1 TombATI

So my conclusion is the Look at Item function is less rigid in TR1X than in TombATI.

lahm86:
Resolves #1026.

#### Checklist

#### Description

This ensures that if Lara remains on a trigger with a `TO_TARGET` entry but no fixed camera, that she continues to look at the correct position of said target. The camera's type and item is reset on each frame, but the main camera update routine occurs after Lara's control routines, hence we need to refer to the cached `last_item` of the camera.

Tested the areas described in the issue itself, plus Natla's Mines room 94. I've also checked regular camera triggers and it seems fine to me.

~~We believe this is a bug in the later games too, it's just that there isn't a dosbox/ati equivalent for those so it's gone unnoticed.~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - fixed being able to deselect the passport in quick save, quick load, save crystal, and death modes (#1108)
 - fixed inability to save in Unfinished Business in crystals mode as UB doesn't have crystals (#1102)
 - fixed items not being added to inventory if the sprite is missing from the level file (#1130)
+- fixed differences when looking at items from triggers that do not use fixed cameras when the enhanced look option is enabled (#1026)
 
 ## [3.0.5](https://github.com/LostArtefacts/TR1X/compare/3.0.4...3.0.5) - 2023-12-13
 - fixed crash when pressing certain keys and the console is disabled (#1116, regression since 3.0)
--- a/src/game/lara/lara_look.c
+++ b/src/game/lara/lara_look.c
@@ -81,7 +81,7 @@ void Lara_LookUpDownSurf(void)
 
 void Lara_ResetLook(void)
 {
-    if (g_Camera.type == CAM_LOOK) {
+    if (g_Camera.type == CAM_LOOK || g_Camera.last_item) {
         return;
     }
     if (g_Lara.head_x_rot <= -HEAD_TURN / 2


walkawayy (approved):
Great!

rr-:
We probably ought to clear last_item some time, for example when the player presses the look button, since as I understand it, the camera currently will never get reset after Lara looks at an item.

I'm not sure about the  bug being present in later TR games – Lara leans her head rather strongly in TR2 TGW, whereas with this bug her movement seems to be very restricted. While the fix seems good, I wonder if there's some other way to go around it.

walkawayy:
Maybe the only way to get a proper comparison is to remake the Caves level in TR2. Annoying but it could only be the beginning section. It wouldn't cover every case, but it could give a proof of concept.

lahm86:
Something I experimented with a while ago (just extract and rename to wall.tr2).
[CAVES.zip](https://github.com/LostArtefacts/TR1X/files/13930938/CAVES.zip)

I think you're right, the bug isn't there in TR2+.
Screenshot 2024-01-14 092356

lahm86:
In terms of the camera getting reset, `Camera_Update` is called each frame, and has this:
https://github.com/LostArtefacts/TR1X/blob/develop/src/game/camera.c#L763

So once `.item` has cleared, so will `.last_item`. For `.item` to clear:

- Lara remains on a trigger with a look at item then the player presses and releases look (per TombATI)
- Lara leaves said trigger while still locked on to the item

https://github.com/LostArtefacts/TR1X/blob/develop/src/game/camera.c#L810

These appear to be functional as-is. I'll put a test build here too - @aredfan, @Richard-L, perhaps if you have time you could perform some testing on this too?

[TR1X.zip](https://github.com/LostArtefacts/TR1X/files/13930956/TR1X.zip)

rr- (approved):

aredfan:
I tested both Caves and Temple of the Cat with Lahm's build.  The camera lock-on behaviour is working as expected in both levels.

LGTM👍

PR #1147, opened 2024-01-25
installer: add CD installation source

### LINKED ISSUE #1144: Allow mounted TR1 CD (or physical) as installer game file source
It strikes me as odd that even though you can provide the original CD inserted in your system, none of installer source options support it. I also tried tricking it by providing the CD path as a TR1X or TombATI source, and then when it didn't like that also to the respective `\DATA` on both, but it still doesn't let you proceed.

lahm86:
Resolves #1144.

#### Checklist

#### Description

I've tested it without having a CD inserted with the installer open, inserting a CD and then selecting the drive from there as well as having a CD inserted before launching the installer and both cases work fine.

I've removed a little compiler noise now present in the latest VS version, simplifying some `new` expressions and making a method static.

The only other change is increasing the initial window height slightly to show all options on the first page without having to scroll.


rr- (approved):

walkawayy (approved):
Nice! LGTM. I don't have a CD drive to test myself though.

Richard-L:
Love you Lahm.

Most happy to test some if you like, and throw a few versions at it.

Would need a binary though.

lahm86:
> Love you Lahm.
> 
> Most happy to test some if you like, and throw a few versions at it.
> 
> Would need a binary though.

Credit goes to @rr-, my input was minimal 😄 
Here is a build if you would like to try it out.

[TR1X_Installer.zip](https://github.com/LostArtefacts/TR1X/files/14053808/TR1X_Installer.zip)

Richard-L:
Hmmm, both MS Defender and Chrome report a virus. This is making this hard.

image
image

walkawayy:
Ya sometimes windows gets in that immediate delete loop for some reason. It's annoying.

lahm86:
I don't think there is much we can do about false positives in Defender. Were you able to get it running at all?

Richard-L:
Wasn't no, sorry. It won't let me access the file.

walkawayy:
You could try this? I'm pretty sure I've done this before.

https://superuser.com/a/1662420

lahm86:
We're just going to proceed and merge this, but thanks for trying to get around the AV issue. If there do happen to be any nuances with different CD versions once this is released, we can address those separately.

Richard-L:
Just tried again and this time didn't get a virus warning. Works very nicely 👍 
Potential added mini convenience could be it noticing automatically that a CD has been inserted.

PR #1148, opened 2024-01-27
gun: add target changing and target lock mode options

### LINKED ISSUE #1145: Feature request: Add TR4+ ability to switch targets by tapping the Look button
TR4 introduced the ability to change which enemy Lara aims at by pressing the Look button when there are multiple enemies in range, which is a pretty useful feature that would be nice to have in TR1 as well.

### LINKED ISSUE #1146: Feature request: Automatically switch targets after killing an enemy
It would be nice if Lara would automatically stop targeting an enemy once it dies and immediately start targeting a new enemy if one is in range, instead of continuing to shoot at an already dead enemy until Action is released.

walkawayy:
Resolves #1145 and #1146.

#### Checklist

#### Description
Decided to group these two issues since they have co-dependent features. 1 of 2 config options are translated in French with the second one pending. Requesting @ajtudela to translate the two for Spanish if possible. Thanks!

Added the option to change weapon targets by tapping the look key like in TR4+.
Added three targeting lock options:
  - full lock: always keep target lock even if the enemy moves out of sight or dies (OG TR1)
  - semi lock: keep target lock if the enemy moves out of sight but lose target lock if the enemy dies
  - no lock: lose target lock if the enemey goes out of sight or dies (TR4+)

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -51,7 +51,7 @@ typedef enum {
 typedef enum {
     TLM_FULL = 0,
     TLM_SEMI = 1,
-    TLM_NO = 2,
+    TLM_NONE = 2,
 } TARGET_LOCK_MODE;
 
 typedef struct {
--- a/src/config.h
+++ b/src/config.h
@@ -134,7 +134,7 @@ typedef struct {
     bool enable_buffering;
     bool enable_lean_jumping;
     bool enable_target_change;
-    int32_t target_lock_mode;
+    TARGET_LOCK_MODE target_mode;
 
     struct {
         int32_t layout;
--- a/src/config.h
+++ b/src/config.h
@@ -48,6 +48,12 @@ typedef enum {
     BSM_PS1 = 5,
 } BAR_SHOW_MODE;
 
+typedef enum {
+    TLM_FULL = 0,
+    TLM_SEMI = 1,
+    TLM_NO = 2,
+} TARGET_LOCK_MODE;
+
 typedef struct {
     bool loaded;
 
@@ -127,6 +133,8 @@ typedef struct {
     bool enable_uw_roll;
     bool enable_buffering;
     bool enable_lean_jumping;
+    bool enable_target_change;
+    int32_t target_lock_mode;
 
     struct {
         int32_t layout;
--- a/src/game/gun/gun_misc.c
+++ b/src/game/gun/gun_misc.c
@@ -199,6 +199,12 @@ void Gun_TargetInfo(WEAPON_INFO *winfo)
 
 void Gun_GetNewTarget(WEAPON_INFO *winfo)
 {
+    // Preserve OG targeting behavior.
+    if (g_Config.target_lock_mode == TLM_FULL && !g_Config.enable_target_change
+        && !g_Input.action) {
+        g_Lara.target = NULL;
+    }
+
     ITEM_INFO *best_target = NULL;
     int16_t best_yrot = 0x7FFF;
     int16_t num_targets = 0;
@@ -264,7 +270,11 @@ void Gun_GetNewTarget(WEAPON_INFO *winfo)
     }
 
     if (num_targets > 0) {
-        for (int slot = 0; slot < num_targets; slot++) {
+        for (int slot = 0; slot < NUM_SLOTS; slot++) {
+            if (!m_TargetList[slot]) {
+                g_Lara.target = NULL;
+            }
+
             if (m_TargetList[slot] == g_Lara.target) {
                 break;
             }
--- a/src/game/gun/gun_misc.c
+++ b/src/game/gun/gun_misc.c
@@ -3,6 +3,7 @@
 #include "config.h"
 #include "game/collide.h"
 #include "game/effects/blood.h"
+#include "game/input.h"
 #include "game/inventory.h"
 #include "game/items.h"
 #include "game/los.h"
@@ -48,6 +49,9 @@
 #define SHOTGUN_RARM_YMAX (+80 * PHD_DEGREE)
 #define SHOTGUN_RARM_XMIN (-65 * PHD_DEGREE)
 #define SHOTGUN_RARM_XMAX (+65 * PHD_DEGREE)
+
+static ITEM_INFO *m_TargetList[NUM_SLOTS];
+static ITEM_INFO *m_LastTargetList[NUM_SLOTS];
 
 WEAPON_INFO g_Weapons[NUM_WEAPONS] = {
     // null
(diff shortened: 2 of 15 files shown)


src/game/gun/gun_misc.c
```diff
@@ -241,14 +246,86 @@ void Gun_GetNewTarget(WEAPON_INFO *winfo)
...
+                g_Lara.target = NULL;
+            }
+
+            if (m_TargetList[slot] == g_Lara.target) {
+                break;
+            }
+        }
+
+        if (!g_Lara.target) {
+            g_Lara.target = bestitem;
```
rr-: Can we rename to `best_item`?

src/game/gun/gun_misc.c
```diff
@@ -241,14 +246,86 @@ void Gun_GetNewTarget(WEAPON_INFO *winfo)
...
+        Gun_TargetInfo(winfo);
+        return;
+    }
+
+    if (num_targets > 0) {
+        for (int slot = 0; slot < NUM_SLOTS; slot++) {
+            if (!m_TargetList[slot]) {
+                g_Lara.target = NULL;
+            }
+
```
rr-: Can these four lines be replaced with `for (int slot = 0; slot < num_targets; slot++) {`?

src/game/gun/gun_misc.c
```diff
@@ -241,14 +246,86 @@ void Gun_GetNewTarget(WEAPON_INFO *winfo)
...
+                found_new_target = true;
+                break;
+            }
+
+            if (m_LastTargetList[last_target] == m_TargetList[new_target]) {
+                break;
+            }
+        }
+
+        if (found_new_target == true) {
```
rr-: We don't do `== true` :D

src/game/input.c
```diff
@@ -17,9 +17,32 @@ static bool m_KeyConflict[INPUT_ROLE_NUMBER_OF] = { false };
...
+            input->look = 0;
+            m_HoldLook++;
+        }
+    } else {
+        if (m_HoldLook && m_HoldLook != 100) {
+            input->change_target = 1;
+        }
+        m_HoldLook = 0;
+    }
+}
```
rr-, line 46: Is this technique necessary, considering we use g_InputDB anyway, which is 1 only for the first frame the player holds the button? Or is it TR4+ behavior that it works with a slight delay?

walkawayy: Ya all this code is based off TR4. (I did split up the gun new target and change target functions tho to account for all 3 modes).

Basically when Lara has guns equipped, this allows tapping look to change targets and only goes into look mode if the player holds down look.

src/game/gun/gun_misc.c
walkawayy: I think that should work.

rr- (approved):
I haven't tested, but the code LGTM.

lahm86 (approved):
LGTM

ajtudela:
Done here: #1150

src/game/gun/gun_misc.c
walkawayy: Update: reverted this back to the original TR4 code since this change broke targeting.

walkawayy:
Alright I think this is ready for final review but I am going to send aredfan one last build to test first.

Changes:
Lose target of Natla during her semideath state.
Restore OG targeting behavior in Full lock mode with the target change option off. Lara will always target the "best" target instead of aiming at her first, best target.

walkawayy:
Alright aredfan confirmed that all testing looks good.

Last point of discussion is setting the target to NULL when natla enters her semi death state. I could change it to only do that in semi and no lock states to keep OG behavior in full target mode? The game sets natlas hit points to Don’t Target during her semi death state. Meaning if the player lets go of action, lara can’t target natla until she wakes up. Lmk what you think.

rr-:
I'd only set it to null for the new modes to maintain OG behavior for full mode.

src/config.h
```diff
@@ -48,6 +48,12 @@ typedef enum {
     BSM_PS1 = 5,
 } BAR_SHOW_MODE;
 
+typedef enum {
+    TLM_FULL = 0,
+    TLM_SEMI = 1,
+    TLM_NO = 2,
```
rr-: I can't explain it but to me `TLM_NONE` sounds more fitting. I think of it like "mode: [full/semi/*none*]" and this sounds better than "mode: no".

src/config.h
```diff
@@ -127,6 +133,8 @@ typedef struct {
     bool enable_uw_roll;
     bool enable_buffering;
     bool enable_lean_jumping;
+    bool enable_target_change;
+    int32_t target_lock_mode;
```
rr-: Can we change it to `TARGET_LOCK_MODE`? (I know we use int32_t for some enums, but this is actually technical debt that has been lingering since the decompilation phase.)

src/game/input.c
```diff
@@ -17,9 +17,32 @@ static bool m_KeyConflict[INPUT_ROLE_NUMBER_OF] = { false };
...
+static void Input_CheckChangeTarget(INPUT_STATE *input)
+{
+    if (g_Lara.gun_status != LGS_READY) {
+        return;
+    }
+
+    if (input->look) {
+        if (m_HoldLook >= 6)
+            m_HoldLook = 100;
+        else {
```
rr-: Missing curly braces around `if`

src/game/input.c
```diff
@@ -17,9 +17,32 @@ static bool m_KeyConflict[INPUT_ROLE_NUMBER_OF] = { false };
...
+            input->look = 0;
+            m_HoldLook++;
+        }
+    } else {
+        if (m_HoldLook && m_HoldLook != 100) {
```
rr-: Can we remove the magic constants by doing away without `100` and putting `6` in a `#define`?
Eg
```c
#define LOOK_HOLD_TIME 6
if (input->look) {
    if (m_HoldLook < LOOK_HOLD_TIME) {
        input->look = 0;
        m_HoldLook++;
    }
} else {
    if (m_HoldLook > 0 && m_HoldLook < LOOK_HOLD_TIME) {
        input->change_target = 1;
    }
    m_HoldLook = 0;
}

rr- (commented):
LGTM with minor cosmetic comments

src/config.h
walkawayy: Sure. I wasn't sure which to use, so I was trying to match the type in `config_map.c` : `.default_value = &(int32_t){TLM_FULL}`

rr-: I think that can be changed as well.

src/config.h
walkawayy: Ok I renamed it internally to `TLM_NONE`. I left it as `No` in the config tool because it's referred to as `No lock`.

rr- (approved):

walkawayy:
@rr- damnit I forgot to add in the Spanish translations to this PR.

rr-:
Can you post the diff here? I'll amend and force-push. @walkawayy

walkawayy:
> Can you post the diff here? I'll amend and force-push. @walkawayy

It's the Spanish text from #1150 instead of the English text merged here in `es.json`.

rr-:
Done, adjusted

PR #1155, opened 2024-02-16
Introduce phasers

rr-:
#### Checklist

#### Description

Currently the game has nested event loops, which makes maintaining anything related a small nightmare. For instance, if the user presses escape while in-game, the game enters an internal loop for the inventory screen that does everything by itself (input processing, rendering, …) and blocks the outer loop from continuing.

My eventual aim is to have exactly two loops, one for overaching processing of gameflow sequences, and the second one – spawned by the first – being an event loop for rendering title screen/game/inventory/pause/cutscenes and whatnot.

This pull request introduces the concept of phasers. We can have one loop called Game_Loop, and my idea is to switch between various phases each time the player does something inside of that loop. So for example when the user presses escape to open the inventory, the loop keeps spinning, and instead we just switch to different control()/draw() callback pair.

The amount of work necessary to make this shift is significant, especially around the inventory code. As a starter, I have refactored the pause screen to use this approach, to show that this architecture seems to be viable. It is able to exit to the menu and interact directly with the gameflow as intended and necessary for inventory.


lahm86 (commented):
Love the concept 👍 
Works great, however demos don't appear to play out. The levels load, but you are able to control as though it were normal gameplay.

rr-:
Thanks, good catch. Fixed. I don't like it, and ideally would remove `g_GameInfo.current_level_type` altogether if possible, but it should become easier after we convert more processes to use phasers.

src/game/phase/phase_pause.c
```diff
@@ -152,17 +105,83 @@ bool Game_Pause(void)
...
+static int32_t Phase_Pause_Control(int32_t nframes)
+{
+    Phase_Pause_UpdateText();
+
+    Input_Update();
+    Shell_ProcessInput();
+    Game_ProcessInput();
+
+    switch (m_State) {
+    case 0:
```
walkawayy: Should we use enums for the case statement's pause states?

walkawayy (changes requested):
No crashes but when I select quit game from the pause menu, the level is restarted.

src/game/phase/phase.c
```diff
@@ -0,0 +1,57 @@
...
+        m_Phaser->start();
+    }
+}
+
+int32_t Phase_Control(int32_t nframes)
+{
+    if (m_Phaser && m_Phaser->control) {
+        return m_Phaser->control(nframes);
+    }
+    return -1; // GF_NOP
```
walkawayy: Why use `-1` instead of `GF_NOP`? For a future use case reason?

rr-: I missed this. Gonna fix it in a moment

rr-: Fixed

rr-:
> No crashes but when I select quit game from the pause menu, the level is restarted.

Fixed

src/game/phase/phase_pause.c
rr-: Added

walkawayy:
Sorry there's one weird bug left (it's low priority though). If you exit to title from the in game pause screen, then immediately go to a save and choose `Story so far` from the select level menu, all the in game cutscenes are half faded / dark. The cutscenes work fine otherwise. Maybe some status flag needs cleared?

rr-:
@walkawayy fixed

walkawayy (approved):

lahm86 (approved):

PR #1156, opened 2024-02-16
Demo fix

rr-:
#### Checklist

#### Description

Fixes two regressions related to demo display recently made in develop.

### DIFF
--- a/src/global/types.h
+++ b/src/global/types.h
@@ -1226,11 +1226,17 @@ typedef struct RGBA8888 {
     uint8_t a;
 } RGBA8888;
 
-typedef struct VECTOR_3D {
+typedef struct XYZ_16 {
+    int16_t x;
+    int16_t y;
+    int16_t z;
+} XYZ_16;
+
+typedef struct XYZ_32 {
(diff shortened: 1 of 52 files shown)


src/global/types.h
```diff
@@ -1226,11 +1226,17 @@ typedef struct RGBA8888 {
     uint8_t a;
 } RGBA8888;
 
-typedef struct VECTOR_3D {
+typedef struct XYZ16 {
+    int16_t x;
+    int16_t y;
+    int16_t z;
+} XYZ16;
```
walkawayy: `XYZ_16` seems more readable IMO. Or maybe `Vector_16`? Same for 32 idk.

walkawayy (approved):
Just 1 comment. Otherwise demos and game seem to work fine.

src/global/types.h
rr-: Is it OK for me then to convert RGB888 and the family to RGB_888? @walkawayy

walkawayy: That makes sense to me. If you agree.

PR #1160, opened 2024-02-18
Convert cutscenes to phasers

rr-:
#### Checklist

#### Description

Continue phaser work with cutscenes.

I think there is an issue in the first cutscene where Larson's pistol disappears for 1 frame. But it happens so fast I'm not sure. It also rings a bell that we've had dealt with a similar bug in the psat.


src/game/phase/phase_cutscene.c
```diff
@@ -0,0 +1,76 @@
...
+}
+
+static void Phase_Cutscene_Draw(void)
+{
+    Game_DrawScene(true);
+    Output_AnimateTextures(g_Camera.number_frames);
+}
+
+PHASER g_CutscenePhaser = {
+    .start = NULL,
```
walkawayy: Why is `Game_Cutscene_Start` with `Game_Cutscene_InitialiseHair` not used for a phaser `.start` function?

walkawayy (changes requested):
Everything works as expected during normal gameplay.
 
Story So Far seems to be broken though. The FMVs play fine, but the end of the Larson cutscene seems to have sound issues, and then all future cutscenes have no sound and sometimes don't start at the beginning of the cutscene if you skip through the previous one. Seems like some variables may not be getting reinitialized in this mode? I looked at the diff, but everything looks pretty normal. So I'm not sure why it's happening.

src/game/phase/phase_cutscene.c
rr-: Thanks, good catch. Some of the cutscenes need to have their camera positions hardcoded. To achieve this, we split the cutscene loading process into the following gameflow sequences:

- **Start**, which initializes everything (the level, objects and whatnot),
- **Loop**, which runs the actual cutscene. This is the part that got converted to a phase.
- **End**, which stops the music. This got moved to the phase .end.

This way, we get to inject extra sequences like this:

- Start,
- **Set camera X/Y/Z**,
- **Set camera angle**,
- **Play synced audio**,
- **Mesh swap**,
- Loop,
- [End] (which is now a part of `Phase_End`)

At first I didn't move `Game_Cutscene_Start` into the phase because I wanted to keep `Phase_Set` right before the `Game_Loop`. However, it's why the music broke. The chronology was like this - for the first cutscene everything worked fine, but for the following one:

- Run GF `start_cine` sequence – load level, objects…
- Run GF camera, audio and mesh swap sequences… + **play synced audio**
- Run GF `loop_cine` seuqnece:
    - `Phase_Set(PHASE_CUTSCENE)`
        - End the previous phase (`Phase_Cutscene_End()` = **`Music_Stop()`**)
        - Start the new phase (do nothing as there was not `Phase_Cutscene_Start()`)
    - `Game_Loop()` (play the cutscene)

 I've fixed this by moving `Game_Cutscene_Start` to `Phase_Cutscene_Start`, and calling `Phase_Set` upon `start_cine`, while keeping `Game_Loop` in the `loop_cine`. This way the above flow gets changed to:

- Run GF `start_cine` sequence:
     - `Phase_Set(PHASE_CUTSCENE, level_num)`
          - `Phase_Cutscene_End()` = `Music_Stop()`
          - `Phase_Cutscene_Start()` = load level, objects…
- Run GF camera, audio and mesh swap sequences… + **play synced audio**
- Run GF `loop_cine` sequence:
    - `Game_Loop()` (play the cutscene)

walkawayy (changes requested):
Hm there still seems to be issues with the cutscenes.

Normal Gameplay:
I noticed the end of the Larson Qualopec cutscene still seems to have desynced audio near the end. Also, at the end of Natla's Mines, I had the cutscene audio spill over into the next Pyramid FMV.

Story So Far:
Now there is working audio for all the cutscenes, but the cutscene videos do not always start at the beginning still. Also the title menu music doesn't play upon return to the menu.

GAMEFLOW.MD:
Remove the two `stop_cine` references on lines 819 and 963.

walkawayy (approved):
Very nice! LGTM. Story so far and normal gameplay look good. Didn't notice any lag, audio desync, audio spilling, audio muting, or main menu music problems.

PR #1163, opened 2024-02-19
Music refactors

rr-:
#### Checklist

#### Description

Adds a couple of useful music related routines and also switches the timestamps from using int64_t opaque integers towards more human-readable `double seconds`.

May break saved music triggers but that's a sacrifice I'm willing to make.


walkawayy (approved):
Does in fact break existing saved music timestamps which I agree isn't a big deal.

Did not notice any issues when making new saves and loading them. The timestamps loaded properly.

PR #1164, opened 2024-02-21
Convert inventory to a phase

rr-:
#### Checklist

#### Description

This concludes refactoring out nested loops. The only usage left is `#define FRAME_BUFFER` which I haven't tested(!).


lahm86 (commented):
Save crystals seem to respawn - so if you use one, then reload, the crystal reappears and you can use it again.

Everything else looks good inventory-wise from what I can tell. 

Also re buffering, inventory buffering now advances 1 frame rather than 2, but I don't think this is an issue. It just means that any setups using it would be twice as long; F-key buffering still works fine, so that would arguably make setups quicker anyway as you needn't wait for the inventory animation. As TR1X isn't a speedrunning tool, I think overall it's fine for the "casual setup" player.

Here the first setup is what's done normally to get Lara into the right position. The second setup shows it needs double buffers. The Khamoon setup works fine using F-key buffering.
https://youtu.be/9rwbUmI7W-Y

rr-:
Thanks @lahm86 this should be now fixed.

lahm86 (approved):
Looks good!

walkawayy (approved):
I didn't notice any issues while testing!

PR #1168, opened 2024-02-26
build: switch to just

rr-:
#### Checklist

#### Description

This is to make our makefile a bit more maintainable, expansible and fix its internal inconsistencies (for example mixed use of underscores and hyphens in docker images, different processes for docker builds of tools etc).

It uses https://github.com/casey/just which I believe to much more suitable as a task runner than a Makefile.

To build for Windows on Linux or WSL: `just`
To build for Linux: `just build-linux`
To build the installer: `just build-installer`
To lint: `just lint`
To clean temporary files: `just clean`
To list all targets: `just -l`

To install just: `sudo sh -c "curl --proto '=https' --tlsv1.2 -sSf https://just.systems/install.sh | bash -s -- --to /usr/local/bin/"` (for other methods refer to their manual)


walkawayy (approved):
Worked for me in Windows 11 WSL. I only had two issues I was able to fix:
1. I had to delete all my old Docker containers out of Docker Desktop because there was some conflict happening with the `_docker_run` command.
2. For some reason `clang-format` is not in my $PATH, only `clang-format-14` and `clang-format-17` explicit versions. I just changed my local `justfile` to specify `clang-format-17` and it works.

lahm86:
Apologies for missing the review for this. FWIW, it all works fine for me.

PR #1171, opened 2024-03-01
math: define M_PI constant

izarif:
#### Checklist

#### Description

Let's define M_PI constant instead of using compiler specific features, it will make the code more portable

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -16,7 +16,6 @@ endif
 
 build_opts = [
   '-Wno-unused',
-  '-D_GNU_SOURCE',
   '-DMESON_BUILD',
 ]
 c_opts = []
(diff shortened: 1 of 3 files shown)


walkawayy:
Seems to work fine for me on Windows 11 WSL. Idk if the PI change could have impact if we reduce the number of significant figures though.

eikehein:
msvc already defines `M_PI` to the same value: https://learn.microsoft.com/en-us/cpp/c-runtime-library/math-constants

In LLVM libc it's `3.14159265358979323846264338327950288`.

That we get consistent results across platforms with this approach may be speaking for it though.

lahm86 (approved):

rr- (approved):

PR #1172, opened 2024-03-01
meson: refactor

izarif:
#### Checklist

#### Description

Minor code cleanup
Removed unnecessary dependencies
The dependency call for libbacktrace has been replaced by find_library because libbacktrace does not have a pc file

Successfully built a static binary using msys2 mingw


eikehein:
Thanks for the PR, this looks quite good.

I'm curious why you removed the explicit `warning_level` though. Did you run into issues in the  msys2 environment?

walkawayy (approved):
Looks good to me and builds fine on Windows 11 WSL. I'll leave final approval up to @rr- though.

izarif:
> I'm curious why you removed the explicit `warning_level` though. Did you run into issues in the msys2 environment?

No, but warning_level is also set in default_options

izarif:
Sorry for force push, i put zlib back in the dependency list because it is needed for non-static build

eikehein:
> No, but warning_level is also set in default_options

Right, but it's set to `1`: https://mesonbuild.com/Builtin-options.html#core-options

TR1X had it (presumably intentionally) set to a stricter `3`: https://mesonbuild.com/Builtin-options.html#details-for-warning_level

izarif:
> > No, but warning_level is also set in default_options
> 
> Right, but it's set to `1`: https://mesonbuild.com/Builtin-options.html#core-options

Looks like `warning_level` is set twice 

https://github.com/LostArtefacts/TR1X/blob/develop/meson.build#L4

and

https://github.com/LostArtefacts/TR1X/blob/develop/meson.build#L8

I removed the second assignment
The value of `warning_level` should now be 2

Or maybe I misunderstood something?

eikehein:
Ah, gotcha. No, I only saw you removing that line, and hadn't caught that it was set twice. In that case I agree this is a nice clean-up, although @rr- should probably re-visit whether he wants it to be 2 or 3.

Thumbs up from my side.

rr- (commented):
Hi, the newest version doesn't build for me because `M_PI` is undefined on either Linux or Windows.

izarif:
> Hi, the newest version doesn't build for me because `M_PI` is undefined on either Linux or Windows.

I've merged the changes from the `develop` branch and now it should build without errors

rr- (approved):

rr-:
Thanks for the cleanup.

PR #795, opened 2023-03-30
gfx: make shaders compatible with GLSL 1.20

vvs-:
#### Checklist

- [X] I have read the [coding conventions](https://github.com/rr-/Tomb1Main/blob/master/CONTRIBUTING.md#coding-conventions)
- [x] I have added a changelog entry about what my pull request accomplishes, or it is an internal change

#### Description

This fixed compatibility with OpenGL 2.1 shaders on my system.

P.S. I didn't add neither a changelog entry nor `fixes #issue` because it's not clear at this time if this will be enough. It can be added later if needed.

PR #1173, opened 2024-03-04
Add macOS support

### LINKED ISSUE #997: Feature request: native Mac builds
Currently it's possible to build the game for Linux; we should add Mac builds as well.

eikehein:
#### Checklist

#### Description

This adds support for macOS to TR1X, which involved making the following changes:

- The OpenGL 2.1 support on macOS does not support GL_ARB_explicit_attrib_location and GL_EXT_gpu_shader4. I therefore added a code path to request an OpenGL 3.3 Core Profile context and appropriate shaders, as the additions of these extensions are part of 3.3 Core. This was done using unified shader files + also unifies the vertex and fragment shaders into single files, by providing preprocessor defines to the shader files. This approach is similar to OpenLara's and was requested by @rr-.

- A few changes to the frame timing code to use higher-precision types and not underflow in the calculation, due to the very high return value of SDL_GetPerformanceFrequency() on Apple Silicon. @rr- has already introduced similar changes in the 60fps branch, so the general approach should be aligned.

- Making a few 8-byte alignment fixes to struct definitions on struct layout. This is required for 64 bit alignment rules on ARM, but macOS' llvm toolchain also generates warnings on x86-64 otherwise.

- Extending the meson build system with macOS-specific logic that enables dynamic linking (macOS applications always dynamically link at least the System framework) and produces a macOS-style .app bundle on installation. Creating the .app bundle involves copying over dynamically linked libs and rewriting their link paths, which is performed by a new utility script `tools/mac/bundle_mac_dylibs.sh` - and yes, that's really how it's commonly done on macOS.

- A version of the TR1X app icon in macOS' .icns format, used by the app bundle and the installer image.

- An expanded README.

- Adding a GitHub workflow that builds TR1X and its dependencies for both arm64 and x86-64, creates a universal binary app bundle, as well as a macOS-style .dmg installer image (using the new `tools/mac/create_mac_installer.sh`). 

  The binaries and the installer image get code-signed and notarized, which is needed to distribute software on latest macOS. The code-signing and notarization steps require an Apple developer account and certificate, which I'm happy to sponsor for the project and which @rr- has agreed to. To that end the workflow file references a few secrets (cert, account id, etc.) that need to be set in the GitHub repo settings which I will make available.

  Should I get hit by a bus or something, the project could either simply stop signing/notarizing or get another Apple developer account and swap out the credentials, so there should be no fundamental bus number concerns.

  The final installer image comes in at 17.7 MB, which is not bad for two architectures vs. the x86-only 12.1 MB Windows installer.


eikehein:
For early review by @walkawayy ad @lahm86.

You can see an example run of the GitHub workflow here:

https://github.com/eikehein/TR1X/actions/runs/8126372817

The .dmg artifact is also downloadable there. @walkawayy has already tested this a bit on their Intel Mac (and I've tested it on a M1 Mac, both native and x86-64 via the Rosetta binary emulator).

Lint run is green: https://github.com/eikehein/TR1X/actions/runs/8126370992

eikehein:
Note that to fully merge/integrate this, we'll have to decide how to integrate the macOS workflow with `.github/release.yml`, i.e. how things call each other and how the artifact handover should work, but it shouldn't be too hard.

src/gfx/gl/program.c
```diff
@@ -36,6 +37,50 @@ void GFX_GL_Program_Bind(GFX_GL_Program *program)
...
+    if (!processed_content) {
+        return NULL;
+    }
+    processed_content[0] = '\0';
+
+#ifdef __APPLE__
+    strcpy(processed_content, version_ogl33c);
+    strcat(processed_content, define_ogl33c);
+#else
+    stcpy(processed_content, version_ogl21);
```
lahm86: Typo - `strcpy`

src/specific/s_clock.c
```diff
@@ -32,8 +32,7 @@ int32_t S_Clock_SyncTicks(void)
             last_counter + TICKS_PER_FRAME * (m_Frequency / fps);
         const double elapsed_sec =
             (double)(m_Counter - last_counter) / m_Frequency;
-        const double delay_sec =
-            (double)(target_counter - m_Counter) / m_Frequency;
+        const double delay_sec = abs(target_counter - m_Counter) / m_Frequency;
```
lahm86: This generates the following warning.
```
warning: taking the absolute value of unsigned type ‘long long unsigned int’ has no effect [-Wabsolute-value]
```

lahm86 (changes requested):
A couple of code comments below. Fixing the typo yields a build, but it fails to launch for me as follows.

image

[TR1X.log](https://github.com/LostArtefacts/TR1X/files/14486794/TR1X.log)

src/gfx/gl/program.c
eikehein: Thanks, fixed!

src/specific/s_clock.c
eikehein: Thanks, fixed in the proper way.

The underlying issue here is that on macOS, one of the very early frames ends up running behind, i.e. `target_counter - m_Counter` ends up being negative. The division by `m_Frequency` then causes an underflow, so `delay_sec` ends up with a large positive value. Multiplied by `* 1000`, and rendering essentially freezes as the next frame gets scheduled far in the future.

The idea for now is to simply not introduce any delay if we're already running behind, which I think is the best way to go with the current simple frame timing code.

In the future we probably want to implement some proper frame pacing and skip frames if we're running late, but this is best left to wait on @rr-'s interpolation work.

eikehein:
@lahm86 I think I spotted the shader issue. Give it another spin? :-) If this still doesn't work I'll do a Windows build to check over here.

I've also addressed the other comments.

Edit: One more push coming up to make the code formatting linter happy.

lahm86 (commented):
Plays now on Windows 👍 
Unfortunately, I don't own a Mac so can't test that, but I'm happy to go with the testing approach already taken.

I wonder regarding the shader changes if #795 will have been affected? Perhaps we should make a test build available once the PR has gone through further reviews for others to try as well.

eikehein:
> Plays now on Windows 👍 Unfortunately, I don't own a Mac so can't test that, but I'm happy to go with the testing approach already taken.

In addition to @walkawayy and myself, there's also Frjttr on the TEN Discord who's been testing it with Sabatu’s TR1 as a Customs testcase for me :). They're also using a M1/arm64 Mac.

> I wonder regarding the shader changes if #795 will have been affected? Perhaps we should make a test build available once the PR has gone through further reviews for others to try as well.

It shouldn't :). Basically, at the moment this preserves the GLSL 1.20 shaders in all situations except on __APPLE__, where it uses OpenGL 3.3 Core w/ GLSL 1.30 shaders instead as the OGL21/GLSL120 support on macOS isn't up to our extension requirements.

lahm86 (approved):

walkawayy (approved):
I don't know anything about the shader files honestly, but the changes look good to me. Don't notice any issues on my Intel Macbook Pro 2019 or Windows 11. Awesome job! I'll tag rr for review for when he has time.

.github/workflows/build_macos.yml
```diff
@@ -0,0 +1,219 @@
...
+          /usr/bin/codesign --force --options runtime -s "${IDENTITY}" --keychain $KEYCHAIN_PATH -v TR1X-Installer.dmg
+          xcrun notarytool submit --wait --apple-id "$MACOS_APPLEID" --password "$MACOS_APP_PWD" --team-id "$MACOS_TEAMID" TR1X-Installer.dmg
+          xcrun stapler staple -v TR1X-Installer.dmg
+
+      - name: Upload signed+notarized installer image
+        uses: actions/upload-artifact@v4
+        with:
+          name: TR1X-Installer.dmg
+          path: ${{ github.workspace }}/installer/TR1X-Installer.dmg
+          compression-level: 0 # .dmg is already compressed.
```
rr-: Hi, it's best if we integrated this with the release job.
However, I'd want us to follow TR2X modernized and more modularized build process: https://github.com/LostArtefacts/TR2X/tree/develop/.github/ so it's fine to do this in a follow-up PR.

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 - added an option to the installer to install from a CD drive (#1144)
 - added stack traces to logs for better crash debugging (#1165)
 - changed the way music timestamps are internally handled – resets music position in existing saves
+- added support for arm64 and Apple Silicon
+- added support for macOS builds
```
rr-: Let's put the `added` entries above `changed`
Let's also add a changelog entry for the changed shaders structure and compatibility as I believe the proposed changes should already help with certain hardware.

meson.build
```diff
@@ -323,4 +327,14 @@ executable(
...
   gui_app: true,
+  install: true,
 )
+
+if host_machine.system() == 'darwin'
+  install_subdir('bin/cfg', install_dir : 'Contents/Resources')
+  install_subdir('bin/data', install_dir : 'Contents/Resources')
+  install_subdir('bin/shaders', install_dir : 'Contents/Resources')
+  install_data('tools/resources/icon.icns', install_dir : 'Contents/Resources')
+  install_data('tools/resources/Info.plist', install_dir : 'Contents')
```
rr-: Can we move these last two files into `tools/resources/mac/`, or `tools/mac/resources/` (whichever makes more sense to you)?

src/gfx/context.c
```diff
@@ -118,6 +118,17 @@ void GFX_Context_Attach(void *window_handle)
...
+
+// Use a OpenGL 3.3 Core Profile context on Mac OS
+// (OpenGL 2.1 on MacOS does not support newer extensions
+// (GL_ARB_explicit_attrib_location, GL_EXT_gpu_shader4))
+#ifdef __APPLE__
+    SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 3);
+    SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 3);
+    SDL_GL_SetAttribute(
+        SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_CORE);
+#endif
```
rr-: Is it possible for us to convert this `#ifdef` into an abstract function that in `src/gfx/context_mac.c` contains these lines and in `src/gfx/context_generic.c` contains a no-op implementation, and then use conditional compilation in `meson.build` to choose between the implementations?

src/gfx/gl/program.c
```diff
@@ -36,6 +37,50 @@ void GFX_GL_Program_Bind(GFX_GL_Program *program)
...
+        return NULL;
+    }
+    processed_content[0] = '\0';
+
+#ifdef __APPLE__
+    strcpy(processed_content, version_ogl33c);
+    strcat(processed_content, define_ogl33c);
+#else
+    strcpy(processed_content, version_ogl21);
+#endif
```
rr-: Ditto – I'm OK with duplicating some of the code between mac-specific and generic implementation.
Alternatively, we can create `platform.h` `platform_mac.c` `platform_linux.c` that returns an enum `PLATFORM_MAC` `PLATFORM_LINUX` respectively etc., which we then can use it for things that only need this information at runtime and not compile time.

All in all, this is to try to stay away from `#ifdefs` as much as possible.

src/global/types.h
```diff
@@ -1933,14 +1941,6 @@ typedef struct REQUEST_INFO {
...
     int16_t y;
     uint16_t flags;
-    const char *heading_text;
-    char *item_texts;
-    int16_t item_text_len;
-    TEXTSTRING *heading;
-    TEXTSTRING *background;
-    TEXTSTRING *moreup;
-    TEXTSTRING *moredown;
-    TEXTSTRING *texts[MAX_REQLINES];
```
rr-: Is there a reason why we needed to reorder the members?

src/global/types.h
```diff
@@ -1922,7 +1922,15 @@ typedef struct BOX_INFO {
     int16_t overlap_index;
 } BOX_INFO;
 
-typedef struct REQUEST_INFO {
+typedef struct __attribute__((aligned(8))) REQUEST_INFO {
```
rr-: Isn't this accomplished by `#pragma pack(push, 1)` code fence that should span this struct as well?

tools/mac/bundle_mac_dylibs.sh
```diff
@@ -0,0 +1,119 @@
...
+
+if [[ $LINKS_ONLY -eq 1 ]] || [[ $COPY_ONLY -eq 0 && $LINKS_ONLY -eq 0 ]]; then
+    for lib_path in "$FRAMEWORKS_PATH"/*; do
+        update_links "$lib_path"
+    done
+
+    update_links "$APP_BINARY_PATH"
+fi
+
+echo "Libraries for $APP_NAME copied and updated."
```
rr-: For more complex scripts, we try to stay away from Bash and Shell – can we convert this into a Python script using `argparse`?

src/specific/s_clock.c
```diff
@@ -6,11 +6,11 @@
 #include <SDL2/SDL_timer.h>
 
 static Uint64 m_Counter = 0;
-static double m_Frequency = 0.0;
+static Uint64 m_Frequency = 0.0;
```
rr-: I think it should be `0` or `0ull` or similar.

tools/mac/create_mac_installer.sh
```diff
@@ -0,0 +1,41 @@
...
+  --volname "${APP_NAME} Installer" \
+  --volicon "../tools/resources/icon.icns" \
+  --window-pos 200 120 \
+  --window-size 800 400 \
+  --icon-size 100 \
+  --icon "${APP_NAME}.app" 200 190 \
+  --hide-extension "${APP_NAME}.app" \
+  --app-drop-link 600 185 \
+  "${DMG_NAME}" \
+  "${APP_BUNDLE_PATH}"
```
rr-: Likewise, let's try to use Python

rr- (commented):
Thanks. Awesome work. I left a couple of comments to let the code be more in line of what we are using or trying to follow – otherwise looks great!

.github/workflows/build_macos.yml
eikehein: My suggestion is that after merging, you do the integration with the release workflow in a follow-up commit since you have the vision on that one :)

CHANGELOG.md
eikehein: Done

src/gfx/gl/program.c
eikehein: In the updated MR I'm using Meson's `configure_file` mechanism to generate a gfx_options.h that provides a define for the default backend, and sets it differently depending on the build platform.

The code is now ifdef-free, and the shader pre-processing function now takes a GFX_GL_BACKEND argument from an extensible enum, allowing us to do further work on e.g. GLES later on. This also means the code is now free of the platform notion again; instead, the build system opts macOS into preferring OpenGL 3.3.

src/global/types.h
eikehein: Basically, on llvm/arm64, you want pointer members aligned to multiples of the 8-byte boundary. There's 3 ways to do this:

- Adding padding members, imho icky
- Doing weird tetris like having members that add up to multiples of 8 inbetween
- Putting all the pointers first

This seems the least icky?

src/global/types.h
eikehein: `pack(push, 1)` aligns to 1-byte boundaries, that attrib aligns to 8-byte boundaries.

src/specific/s_clock.c
eikehein: Whoops, forgot this one - updating, sec

src/global/types.h
rr-: Sounds good, but what happens if they're not aligned? I think we have more structures like this BTW.

.github/workflows/build_macos.yml
rr-: Thanks, I'll do that, but I don't have a clear idea when this happens.

bin/shaders/2d.glsl
```diff
@@ -0,0 +1,36 @@
...
+#else
+#define OUTCOLOR gl_FragColor
+#define TEXTURE texture2D
+varying vec2 vertTexCoords;
+#endif
+
+void main(void) {
+    OUTCOLOR = TEXTURE(tex0, vertTexCoords);
+}
+#endif // VERTEX
```
rr-: Can we indent preprocessed fragments here and in the other shaders?
Eg

```
#ifdef VERTEX
    yadda yadda
#else 
    blabla boo boo
#endif
```

src/gfx/common.h
```diff
@@ -12,3 +14,8 @@ typedef enum GFX_RENDER_MODE {
     GFX_RM_LEGACY,
     GFX_RM_FRAMEBUFFER,
 } GFX_RENDER_MODE;
+
+typedef enum GFX_GL_BACKEND {
+    GFX_GL_21,
+    GFX_GL_33C,
+} GFX_GL_BACKEND;
```
rr-: Missing \n at EOF

src/gfx/gl/program.c
```diff
@@ -36,6 +37,50 @@ void GFX_GL_Program_Bind(GFX_GL_Program *program)
     GFX_GL_CheckError();
 }
 
+char *GFX_GL_Program_PreprocessShader(
+    const char *content, GLenum type, GFX_GL_BACKEND backend)
+{
+    const char *version_ogl21 =
+        "#version 120\n#extension GL_ARB_explicit_attrib_location: "
+        "enable\n#extension GL_EXT_gpu_shader4: enable\n";
```
rr-: Can we make it like so:
```
const char *version_ogl21 =
    "blabla\n"
    "yadda yadda yadda\n"
    "bar qux\n";
```
so that `\n` always appears at the end of the string fragment?

src/gfx/gl/program.c
```diff
@@ -50,7 +95,14 @@ void GFX_GL_Program_AttachShader(
...
 
-    glShaderSource(shader_id, 1, (const char **)&content, NULL);
+    char *processed_content =
+        GFX_GL_Program_PreprocessShader(content, type, GFX_GL_DEFAULT_BACKEND);
+    Memory_FreePointer(&content);
+    if (!processed_content) {
+        Shell_ExitSystemFmt("Failed to pre-process shader source:  %s", path);
+    }
+
+    glShaderSource(shader_id, 1, (const char **)&processed_content, NULL);
```
rr-: Is this cast necessary?

src/global/types.h
rr-: All in all this seems to be pretty easy condition to break. I wonder if removing the `#pragma push(pack, 1)` from the beginning that forces the alignment would help? It's there mostly for legacy reason from the decompilation era. We should be good to remove it altogether for the most part, but I'm almost certain there are a couple of places left that use unsafe casts and rely on the member sizes/order. IMO it's better to track these and let the OS decide on data alignment for itself rather than fiddle with it by ourselves. Please LMK your thoughts.

eikehein: Yep it's a good point, I'll investigate what this macOS toolchain does if I remove the fence and the attrib, maybe it just ends up being happy by itself

bin/shaders/2d.glsl
eikehein: Will do

src/gfx/common.h
eikehein: Will do

src/gfx/gl/program.c
eikehein: Will do

bin/shaders/2d.glsl
```diff
@@ -0,0 +1,38 @@
...
+    #define OUTCOLOR gl_FragColor
+    #define TEXTURE texture2D
+
+    varying vec2 vertTexCoords;
+#endif
+
+void main(void) {
+    OUTCOLOR = TEXTURE(tex0, vertTexCoords);
+}
+#endif // VERTEX
```
rr-: Sorry to be repetitive but we're missing an \n here as well as the other shaders too. Whole we're at it let's make sure the other files have it as well.

src/global/types.h
```diff
@@ -2036,8 +2034,6 @@ typedef struct SAMPLE_INFO {
     int16_t flags;
 } SAMPLE_INFO;
```
rr-, line 2037: I'll confirm later today if the game code uses unsafe casts anywhere that night break this.

rr-: I have checked all casts across the project and it seems the last unsafe operations were removed with the introduction of the XYZ_32 `struct`. I used this command to look for the casts:

```
grep -P '(?<!sizeof ?)\([^\(\),]*\*\)' **/*.c|grep -v -P '\((const )?(void|u?int(8|16|32)_t|char|bool|float|double) ?(\* ?)?(const ?)?\*\)'
```

rr- (approved):

eikehein:
Squashed / linearized history

PR #1183, opened 2024-03-24
build: improve toolchain structure

rr-:
#### Checklist

#### Description

- Make `tools/` only about tools and helper files strictly related to tooling
- Move `docker/` to `tools/`
- Create an umbrella directory `data/` for all resources and assets
- Rename `bin/` to `data/ship`
- Put files from `tools/resources/` to `data/`
- Put most of the code from GH actions into justfile, so that releases can be created using `just` for example
- Change GH actions to refer to the justfile and scripts where possible – the idea is to make them as lightweight as possible
- Integrate the recently donated Mac builds with the release job


justfile
```diff
@@ -29,15 +41,41 @@ _docker_run *args:
...
+    zip_name=$(tools/generate_version)-Linux.zip
+    7z a "${zip_name}" ./data/ship/* ./build/linux/TR1X
+    echo "Created ${zip_name}"
+package-win zip_name="": (build-win "release") (build-win-config)
+    #!/bin/bash
+    zip_name="{{zip_name}}"
+    if [[ -z $zip_name ]]; then
+        zip_name=$(tools/generate_version)-Windows.zip
+    fi
+    7z a "${zip_name}" ./data/ship/* ./build/win/TR1X.exe ./tools/config/out/TR1X_ConfigTool.exe
```
walkawayy: When I apt installed `7zip` in WSL, the default command name is `7zz`. Which is present in `package-win` and `package-linux`.

walkawayy (approved):
Worked really well! Only one issue I had. Idk if you have your 7zip aliased? I am running Ubuntu 22.04 in WSL.

eikehein:
The new structure looks great.

lahm86 (approved):
LGTM. I had to remove the existing docker image from my VM to force it to download again, but otherwise everything seems fine.

rr-:
@lahm86 @walkawayy @eikehein I've updated it a bit to address an issue raised by @walkawayy – requires rebuilding the images again.

Rather than guessing how the 7-zip executable is named on the target user's system, I decided to move the packaging inside the docker as a subcommand. Furthermore, since Python (which we already are tightly integrated with) provides built-in zipfile support, I decided to drop 7-zip dependency altogether and instead zip the files by hand. Lastly, as there is a lot of overlap between Linux and Windows Docker entrypoints, I've created a shared Python library `tools/shared/` that can be reused between all tools, and provides the common bits around the new entrypoints.

I'm not sure if this isn't overengineering, but I am hoping this to prove more maintainable down the line than the current setup. Please LMK your thoughts.

justfile
rr-: Addressed – please see comment below

lahm86 (approved):
LGTM. One less external dependency is always good 👍

walkawayy (approved):
LGTM! Sorry for the trouble a 7zip command caused : /

PR #1230, opened 2024-03-30
data: fix stacked cabin in Natla's Mines

### LINKED ISSUE #1052: Stacked portable buildings in Natla's Mines (Game breaking bug)
First time playing the game and ran into this game breaking bug where after putting in the third fuse, the cutscene doesn't play and the portable buildings duplicate with the one on top being blackened and because of this I was unable to progress any further. I thankfully had a save at the beginning of the level and ran though it again and this time I didn't get stuck.

I did two things differently on my second go, I got the Conveyor belt fuse second, instead last and I put in all the fuses at the same time, instead of whenever I got one.

Here's my saves with a readme file. They were created on v2.15.3, I then updated to v2.16 and got the same result.
[saves.zip](https://github.com/LostArtefacts/TR1X/files/12884162/saves.zip)

Screenshot 2023-10-12 233852

lahm86:
Resolves #1052.

#### Checklist

#### Description

I originally thought this would be tricky to resolve but on closer inspection the bug is caused by incorrect flipmap indices in room 85. When the cabin drops, it triggers [flip index 3](https://github.com/LostArtefacts/TR1X/blob/develop/src/game/objects/general/cabin.c#L36), Pulling the lever beyond Cowboy is index 0, hence the triggers here were incorrectly targeting that. So if you pulled the lever, then returned to this part of the level, it sees that flip 0 has been done so when you then return to the cabin area it is also flipped regardless of whether or not you have used the fuses.

All four triggers in this group now reference index 3, similar to room 28.

image

I played through dev-intended and otherwise and all flip areas behave as they should now.


rr- (approved):

walkawayy (approved):
Nice fix!

Issue #1034, opened 2023-10-07
OG bug: Displaced camera moments


aredfan:
It's possible for the camera to get displaced for 1 frame if it gets stuck in geometry.  Am unsure whether this can be fixed but I thought it'll be helpful to put in an issue for it.

St Francis' Folly (with guns out):

https://github.com/LostArtefacts/TR1X/assets/94776985/c5848a7c-057f-4620-8252-9d5869fefcc0

Palace Midas:

https://github.com/LostArtefacts/TR1X/assets/94776985/5098824b-43a1-48ef-bb79-2ff7698da902

Tomb of Tihocan:

https://github.com/LostArtefacts/TR1X/assets/94776985/af7dfb58-cb21-438e-a030-23ac83400b86

Richard-L:
So odd hah. And you can reproduce it in TombATI?

aredfan:
Odd indeed ;)

Yep I can reproduce all 3 instances on TombATI.  Here's the TombATI saves if you'd like to try for yourself, remember to remove the `.txt` extension:

[saveati.0.txt](https://github.com/LostArtefacts/TR1X/files/12838192/saveati.0.txt) (Run forward)
[saveati.1.txt](https://github.com/LostArtefacts/TR1X/files/12838193/saveati.1.txt) (Jump forward and grab)
[saveati.2.txt](https://github.com/LostArtefacts/TR1X/files/12838194/saveati.2.txt) (This one is tricky to pull off so may take several tries)

Richard-L:
I think I just saw it on the Foxmaster video. Is that where you've got it from?

https://youtu.be/7AOdvSzVYxk?t=18m07s

aredfan:
Nope, an amazing coincidence though.

I encountered the Midas/Tihocan camera bugs during my playthrough, and I think I seen the Folly bug before in one of Eycore's streams.

Richard-L:
Tried the Folly one on PS1 and it occurs too. Just the framebuffer entirely black and you can't see the zoomed out map.

aredfan:
It's the same on PC (all black and no zoomed out map) if the player is using the hardcoded vanilla draw distance values.  The map is visible when using the TombATI draw distance values.

BTW I found a spot to reproduce this bug in a stable way:

https://github.com/LostArtefacts/TR1X/assets/94776985/b0ba16f2-2522-443f-88a5-bdf5a4e64732

walkawayy:
[...]

PR #1253, opened 2024-04-02
interpolation: correct camera room number

### LINKED ISSUE #1240: Loading screen and underwater tunnel of Midas show briefly at level start
https://github.com/LostArtefacts/TR1X/assets/4934209/5aadff8e-87a2-4566-9edb-7ce7f8d62815

Haven't quite noticed this before. It happens in 30fps too. It seems related to the loading screens. It doesn't happen quite as noticeably in v3.1.1 (where there are no loading screens).

Here is the source clip: 

https://github.com/LostArtefacts/TR1X/assets/4934209/ecf708f5-4def-4821-b66e-6929b0033f39

### LINKED ISSUE #1241: Lost Valley Demo black screen moment
In the Lost Valley Demo, when Lara turns the corner into the open - there's a black screen moment.  This happens in 30/60 FPS.  The bug doesn't happen in 3.1.1.

Screenshot

First part of the video is 3.1.1.  Second part is the dev build.

https://github.com/LostArtefacts/TR1X/assets/94776985/9cd5d7f1-7803-4e5d-9205-ae1d92704ebf

____________________________________

https://nebula.wind.garden/serve/pub/TR1X%203.1.1-117-g3d8bdd0-Windows.zip

### LINKED ISSUE #1243: Camera issue after opening a door
Copied from #1239 

> @rr- My mistake. The displaced camera bug in https://github.com/LostArtefacts/TR1X/issues/954 happens just before a door opens.
> 
> The screenshot below shows the weird camera moment after the door opens. It's not OG, nor does it happen in 3.1.1.
> 
> It happens in both 30/60 FPS.

image

lahm86:
Resolves #1240.
Resolves #1241.
Resolves #1243.

#### Checklist

#### Description

When `Room_GetFloor` is called, it will work out the relevant room number based on the interpolated position of the camera, so we need not track the previous room. We do still need a separate `room_num` for the interpolation because any slight change in `X Y Z` may yield a different number than the one stored in `g_Camera.pos.room_number`; I opted to move it from `result`/`prev` and put it directly into `interp`, but let me know if this would be better otherwise.

A little detail on the individual issues:
- In the Lost Valley demo example, although Lara had passed from room 0 to 1, the previous camera value was still 0 even though it was technically also in room 1.
- In Midas, the camera is initially set to Lara's room so this became the remembered value although the camera is far away in room 11. Because `Room_GetFloor` can't "fix" this (no matching portals at that position), we ended up with the camera at the correct position in room 11, but with a room value of 0, hence all the other rooms were culled and you would only see the tunnel.
- In Tihocan, when the camera returns to Lara after the fixed viewpoint, its number remains at the viewpoint for one frame so you see the adjoining Pierre room as a result (which just so happens to be directly above the switch).

This should also fix the other Midas problem @Richard-L mentioned on Discord (same idea as Lost Valley).

You can pause and use , and . to move through the frames here and check. You'll notice there are other issues remaining like the camera going into the wall in Lost Valley and going beyond the door in Tihocan, but these are part of #1034 and aren't new issues.
https://youtu.be/vDfNlQdwkqU

### DIFF
--- a/src/game/game/game_draw.c
+++ b/src/game/game/game_draw.c
@@ -19,7 +19,7 @@ void Game_DrawScene(bool draw_overlay)
     Camera_Apply();
 
     if (g_Objects[O_LARA].loaded) {
-        Room_DrawAllRooms(g_Camera.interp.result.room_num);
+        Room_DrawAllRooms(g_Camera.interp.room_num);
         if (draw_overlay) {
             Overlay_DrawGameInfo();
         } else {
--- a/src/game/interpolation.c
+++ b/src/game/interpolation.c
@@ -70,11 +70,11 @@ void Interpolation_Commit(void)
         INTERPOLATE(&g_Camera, target.y, ratio, 512);
         INTERPOLATE(&g_Camera, target.z, ratio, 512);
 
-        g_Camera.interp.result.room_num = g_Camera.interp.prev.room_num;
+        g_Camera.interp.room_num = g_Camera.pos.room_number;
         Room_GetFloor(
             g_Camera.interp.result.pos.x,
             g_Camera.interp.result.pos.y + g_Camera.interp.result.shift,
-            g_Camera.interp.result.pos.z, &g_Camera.interp.result.room_num);
+            g_Camera.interp.result.pos.z, &g_Camera.interp.room_num);
     }
 
     INTERPOLATE_ROT(&g_Lara.left_arm, rot.x, ratio, PHD_45);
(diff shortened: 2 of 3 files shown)


walkawayy (approved):
Everything looks good to me.

rr- (approved):
OK, the room number is special enough to have its place directly in the `interp` struct. Thank you so much for researching this.

Richard-L:
@lahm86 Not to over-nitpick, but I needed to play your video at .25x speed to analyse the Midas camera. There's an unexpected camera adjustment happening from one frame to the next, and then the moving in onto Lara takes place.

It might be normal, the mild jump just looks a little odd.

PR #1262, opened 2024-04-03
camera: improve underwater check

### LINKED ISSUE #1251: OG bug: underwater SFX plays for one frame in Midas
While investigating #1240, I noticed there is an odd (albeit subtle) sound for one frame at the very beginning of Midas - it's the underwater SFX playing and then immediately being cut-off on the next frame. This is present in TombATI as well.

lahm86:
Resolves #1251.

#### Checklist

#### Description

This delays the camera being marked as being underwater, and hence the SFX, until the very last moment for better accuracy in room position. As a result, the `Camera_Initialise` call no longer triggers the SFX via `Camera_Update` before the level actually begins, and we don't get the Midas issue. I did try removing `Camera_Update` from `Camera_Initialise` but this crashes Great Pyramid at the start, and so I think otherwise shuffling the general order of calls will cause further problems.

I've also changed the underwater flag in `CAMERA_INFO` to a `bool` for clarity as it's only used as such.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 - fixed broken gorilla animations (#1244, regression since 2.15.3)
 - fixed saving and loading the music timestamp when the load current music option is enabled (#1237)
 - fixed the remember played music option always being enabled (#1249, regression since 2.16)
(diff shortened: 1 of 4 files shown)


rr- (approved):

walkawayy (approved):
LGTM.

PR #1267, opened 2024-04-03
vars: refactor g_InvChosen and fix Lara saying "No" on save/load

### LINKED ISSUE #1254: Key/Puzzle/Midas item item usage can be cancelled after selection due to inventory phase
https://github.com/LostArtefacts/TR1X/assets/4934209/340bc3e3-a6fe-4538-a23b-751fcef41c2f

image

### LINKED ISSUE #1255: Lara says "No" if the player saves or loads in front of a key/puzzle hole
Also happens in TombATI.

### LINKED ISSUE #1268: Lara says "no" when taking valid actions in front of a key item receptacle
Lara says "no" when using a medipack, taking out her guns, saving, loading, etc in front of a key item receptacle even though those are valid actions.

Semi-related, Lara says "No" if she tries to use medipack and her health is full already in TR4+. In TR1, the medipack animates open like it's being used.

walkawayy:
Resolves #1255.

#### Checklist

#### Description
Fixed Lara saying "no" when taking valid actions in front of a key item receptacle. I combined the saying "no" issue since they stemmed from the same code issue. No changelog for #1254 because it's unreleased.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 - fixed the remember played music option always being enabled (#1249, regression since 2.16)
 - fixed the underwater SFX playing for one frame at the start of Palace Midas (#1251)
 - fixed an incorrect frame in Lara's underwater twist animation (OG bug in TR2 onwards) (#1242)
+- fixed Lara saying "no" when taking valid actions in front of a key item receptacle (#1268)
 
 ## [3.1.1](https://github.com/LostArtefacts/TR1X/compare/3.1...3.1.1) - 2024-01-19
 - changed quick load to show empty passport instead of opening the save game menu when there are no saves (#1141)
--- a/src/game/game/game.c
+++ b/src/game/game/game.c
@@ -203,10 +203,6 @@ GAMEFLOW_OPTION Game_Stop(void)
         return GF_LEVEL_COMPLETE | g_GameInfo.select_level_num;
     }
 
-    if (!g_InvChosen) {
-        return GF_EXIT_TO_TITLE;
-    }
-
     if (g_GameInfo.passport_selection == PASSPORT_MODE_LOAD_GAME) {
         return GF_START_SAVED_GAME | g_GameInfo.current_save_slot;
     } else if (g_GameInfo.passport_selection == PASSPORT_MODE_SELECT_LEVEL) {
--- a/src/game/lara/lara.c
+++ b/src/game/lara/lara.c
@@ -334,7 +334,7 @@ void Lara_AnimateUntil(ITEM_INFO *lara_item, int32_t goal)
     } while (lara_item->current_anim_state != goal);
 }
 
-void Lara_UseItem(int16_t object_num)
+void Lara_UseItem(GAME_OBJECT_ID object_num)
 {
     LOG_INFO("%d", object_num);
     switch (object_num) {
@@ -396,6 +396,40 @@ void Lara_UseItem(int16_t object_num)
         Inv_RemoveItem(O_BIGMEDI_ITEM);
         Sound_Effect(SFX_MENU_MEDI, NULL, SPM_ALWAYS);
         break;
+
+    case O_KEY_ITEM1:
+    case O_KEY_OPTION1:
+    case O_KEY_ITEM2:
+    case O_KEY_OPTION2:
+    case O_KEY_ITEM3:
+    case O_KEY_OPTION3:
+    case O_KEY_ITEM4:
+    case O_KEY_OPTION4:
+    case O_PUZZLE_ITEM1:
+    case O_PUZZLE_OPTION1:
+    case O_PUZZLE_ITEM2:
+    case O_PUZZLE_OPTION2:
+    case O_PUZZLE_ITEM3:
+    case O_PUZZLE_OPTION3:
+    case O_PUZZLE_ITEM4:
+    case O_PUZZLE_OPTION4:
+    case O_LEADBAR_ITEM:
+    case O_LEADBAR_OPTION: {
+        int16_t receptacle_item_number = Object_FindReceptacle(object_num);
+        if (receptacle_item_number == NO_OBJECT) {
+            Sound_Effect(SFX_LARA_NO, NULL, SPM_NORMAL);
+            return;
+        }
+        ITEM_INFO *item = &g_Items[receptacle_item_number];
+        g_Lara.interact_target.item_num = receptacle_item_number;
+        g_Lara.interact_target.is_moving = true;
+        g_Lara.interact_target.move_count = 0;
+        Inv_RemoveItem(object_num);
+        break;
+    }
+
+    default:
+        break;
     }
 }
 
@@ -447,6 +481,9 @@ void Lara_Initialise(int32_t level_num)
     g_Lara.left_arm.flash_gun = 0;
     g_Lara.right_arm.lock = 0;
     g_Lara.left_arm.lock = 0;
+    g_Lara.interact_target.is_moving = false;
+    g_Lara.interact_target.item_num = NO_OBJECT;
+    g_Lara.interact_target.move_count = 0;
 
     if (g_RoomInfo[g_LaraItem->room_number].flags & RF_UNDERWATER) {
         g_Lara.water_status = LWS_UNDERWATER;
--- a/src/game/objects/common.c
+++ b/src/game/objects/common.c
@@ -51,6 +51,45 @@ const GAME_OBJECT_PAIR g_GunAmmoObjectMap[] = {
     { NO_OBJECT, NO_OBJECT },
 };
 
+const GAME_OBJECT_PAIR g_KeyItemToHoleMap[] = {
+    { O_KEY_OPTION1, O_KEY_HOLE1 },       { O_KEY_OPTION2, O_KEY_HOLE2 },
+    { O_KEY_OPTION3, O_KEY_HOLE3 },       { O_KEY_OPTION4, O_KEY_HOLE4 },
+    { O_PUZZLE_OPTION1, O_PUZZLE_HOLE1 }, { O_PUZZLE_OPTION2, O_PUZZLE_HOLE2 },
+    { O_PUZZLE_OPTION3, O_PUZZLE_HOLE3 }, { O_PUZZLE_OPTION4, O_PUZZLE_HOLE4 },
+    { O_LEADBAR_OPTION, O_MIDAS_TOUCH },  { NO_OBJECT, NO_OBJECT },
+};
+
+GAME_OBJECT_ID Object_GetCognate(
+    GAME_OBJECT_ID key_id, const GAME_OBJECT_PAIR *test_map)
+{
+    const GAME_OBJECT_PAIR *pair = &test_map[0];
+    while (pair->key_id != NO_OBJECT) {
+        if (pair->key_id == key_id) {
+            return pair->value_id;
+        }
+        pair++;
+    }
+
+    return NO_OBJECT;
+}
+
+int16_t Object_FindReceptacle(GAME_OBJECT_ID object_id)
+{
+    GAME_OBJECT_ID receptacle_to_check =
+        Object_GetCognate(object_id, g_KeyItemToHoleMap);
+    for (int item_num = 0; item_num < g_LevelItemCount; item_num++) {
+        ITEM_INFO *item = &g_Items[item_num];
+        if (item->object_number == receptacle_to_check) {
+            const OBJECT_INFO *const obj = &g_Objects[item->object_number];
+            if (Lara_TestPosition(item, obj->bounds())) {
+                return item_num;
+            }
+        }
+    }
+
+    return NO_OBJECT;
+}
+
 bool Object_IsObjectType(
     GAME_OBJECT_ID object_id, const GAME_OBJECT_ID *test_arr)
 {
--- a/src/game/objects/common.h
+++ b/src/game/objects/common.h
@@ -18,6 +18,9 @@ extern const GAME_OBJECT_ID g_PickupObjects[];
 extern const GAME_OBJECT_ID g_GunObjects[];
 extern const GAME_OBJECT_PAIR g_GunAmmoObjectMap[];
 
+GAME_OBJECT_ID Object_GetCognate(
+    GAME_OBJECT_ID key_id, const GAME_OBJECT_PAIR *test_map);
+int16_t Object_FindReceptacle(GAME_OBJECT_ID object_id);
 bool Object_IsObjectType(
     GAME_OBJECT_ID object_id, const GAME_OBJECT_ID *test_arr);
 void Object_Collision(int16_t item_num, ITEM_INFO *lara_item, COLL_INFO *coll);
--- a/src/game/objects/general/keyhole.c
+++ b/src/game/objects/general/keyhole.c
@@ -22,10 +22,6 @@ static const OBJECT_BOUNDS m_KeyHoleBounds = {
     },
 };
 
-int32_t g_PickUpX;
-int32_t g_PickUpY;
-int32_t g_PickUpZ;
-
 static const OBJECT_BOUNDS *KeyHole_Bounds(void);
 
 static const OBJECT_BOUNDS *KeyHole_Bounds(void)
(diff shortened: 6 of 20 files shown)


src/global/types.h
```diff
@@ -1856,8 +1856,8 @@ typedef struct CAMERA_INFO {
             XYZ_32 target;
             XYZ_32 pos;
             int32_t shift;
+            int16_t room_num;
         } result, prev;
-        int16_t room_num;
```
rr-: Conflicts with develop

src/game/game/game.c
```diff
@@ -203,10 +204,6 @@ GAMEFLOW_OPTION Game_Stop(void)
         return GF_LEVEL_COMPLETE | g_GameInfo.select_level_num;
     }
 
-    if (!g_InvChosen) {
-        return GF_EXIT_TO_TITLE;
-    }
```
walkawayy, line 208: @lahm86 from what I could tell, this is never triggered. I tried quitting to title from anywhere, final stats, level skipping in the inventory, etc. I think it can be removed but I'm not 100% sure. This is also related to `phase_inventory.c`'s `Inv_Construct` which is the only place where `g_InvChosen = 0;` was set. Because of the change to `GAME_OBJECT_ID`, setting it to `0` was no longer proper. But, idk if there is some special edge case where it's relevant.

src/global/types.h
walkawayy: hm idk wtf happened. I'll fix.

src/game/phase/phase_inventory.c
```diff
@@ -203,6 +203,7 @@ static GAMEFLOW_OPTION Inv_Close(void)
 
     switch (g_InvChosen) {
     case O_PASSPORT_OPTION:
+        g_InvChosen = O_INVALID;
```
walkawayy: This is the actual fix for the issue. The rest is just refactoring. I also reset it in `Game_Start` to be consistent? Lmk what you think.

rr-: She still says no if she chooses a medpack, pistols, … etc.

Also please use a separate commit for `-1`, and yet another for removing the condition + `= 0` assignment.

walkawayy: Yes on the commit separation.

But I'm confused on the first part. The issue is Lara saying "No" if you save or load in front of a key/puzzle hole. If you choose the wrong key item, she should say "No". Unless I am misunderstanding.

rr-: Yes and no – if you're standing next to a keyhole, a gorilla is trying to murder you and you take a medkit, Lara understandably will complain about the situation but that's still not the right thing to do.

walkawayy: Oh I never realized that. She also complains you select the pistol. Idk if that's technically a different issue than this PR, but I will look into it.

src/game/game/game.c
lahm86: As far as I can tell, this looks fine to be removed. I can't think of any special case where it would be needed here.

src/game/phase/phase_inventory.c
lahm86: We could maybe use a similar array construct in `common.c` to store all "key items" and perform an additional check before saying no that the chosen item is one of those. Might need some translation between the likes of `O_PUZZLE_OPTION1` <=> `O_PUZZLE_ITEM1`. Perhaps `Inv_GetItemOption` could be updated to use some sort of construct like `GAME_OBJECT_PAIR` where we map guns to ammo.

rr-: Sounds good. We also have the `Object_IsObjectType` and `g_EnemyObjects`. This should be tackled in #1268 though.

@walkawayy please rebase onto the newest develop and I'll approve. Thanks

walkawayy: Well I might rework this PR to fix her saying No for other items too since they are all related to `g_InvChosen` not getting cleared in the right scenarios. Ideally, I would like to handle all the inventory stuff in the inventory function. That way we don't have this `g_InvChosen` thing that needs passed around and cleared at different spots in the code. I just need to think of a solution for these keys / puzzles. Lahm's idea might work. Also, I might need to set `Lara.is_moving` when it's confirmed the correct item is chosen so that the player can't input cancel due to the phaser change. It seems like tomb4 has a similar implementation to both of these ideas except for the puzzles/keys.

We can move this to 4.1 if needed.

walkawayy:
Ok so this is a major refactor, so lmk if you like it or absolutely hate it lol. I did testing on key holes, puzzle holes, and Midas hand, and it does in fact work. This should solve both #1268 and #1254 if we choose to go with this method. I left debug logging in the code for now for testing purposes if people find bugs.

The reason I started the refactor was because of #1254. Due to the change to phases for inventory, game, etc, there is a delay between selecting the inventory item and placing the key/puzzle. Lara was able to cancel the interaction which was a new TR1X bug. I was looking to solve this and also poked around the `tomb4` code, and noticed they used more `is_moving` as well as different inventory approach (TR4 moved away from the ring inventory.) Even then, there was still some global inventory item chosen shenanigans. I decided to try and get rid of that variable which was causing all sorts of problems because it was constantly being reset in different places including Lara's control routine. My change keeps the chosen item isolated to the inventory code. It also adds `Lara_UseItem` conditions for the key items so they are grouped with the other inventory items.

The only slightly odd code is the new `Object_CheckForHole` (lol, if anyone has any name suggestions lmk). It does some finagling to find out if there is a nearby hole that matches the chosen inventory item. This was needed because we also have to account for the case where the player goes into the inventory to manually select the key item instead of using the action button on the hole.

The other positive of this refactor is the key/puzzle/midas functions are easier to read since there aren't tons of global chosen item checks anymore. Instead, I used a bool in the correct item's `item->priv` to set when there was a successful match.

`g_Lara.interact_target.is_moving` is used to lock Lara's movement in `Lara_State_Stop` (she has to be stopped for all 3 interactions), so the player can't cancel a successful animation. It was previously used for the walk to pickups and switches feature and is just expanded upon here.

The only potential downside I could see to this is maybe causing some headaches if we ever end up doing a TR1/2/3 codebase since those games probably all do key items similar to TR1 OG. However, I was able to easily extend this to work for puzzles and Midas after I got keys working.

Sorry for the ramble, but lmk what you think. I'll also paste a build for @LostArtefacts/qa to test if they can. If you encounter bugs or issues, upload the TR1X log because I have a bunch of debug statements.

src/game/objects/common.c
```diff
@@ -44,13 +49,94 @@ const GAME_OBJECT_ID g_GunObjects[] = {
...
+int16_t Object_CheckForHole(GAME_OBJECT_ID object_id)
+{
+    GAME_OBJECT_ID hole_to_check =
+        Object_GetCognate(object_id, g_KeyItemToHoleMap);
+    for (int item_num = 0; item_num < g_LevelItemCount; item_num++) {
+        ITEM_INFO *item = &g_Items[item_num];
+        LOG_DEBUG(
+            "Check if item_num %d aka a %d is object_id %d", item_num,
+            item->object_number, object_id);
+        if (Object_IsObjectType(item->object_number, g_HoleObjects)) {
```
rr-: This is unnecessary check as we're checking for even more specific `hole_to_check` below.

src/game/phase/phase_inventory.c
```diff
@@ -260,6 +263,21 @@ static GAMEFLOW_OPTION Inv_Close(void)
...
+    case O_KEY_OPTION2:
+    case O_KEY_OPTION3:
+    case O_KEY_OPTION4:
+    case O_PUZZLE_OPTION1:
+    case O_PUZZLE_OPTION2:
+    case O_PUZZLE_OPTION3:
+    case O_PUZZLE_OPTION4:
+    case O_LEADBAR_OPTION:
+        Lara_UseItem(inv_chosen);
+        break;
```
rr-, line 257: This can be merged with the above switch cases.

src/game/phase/phase_inventory.c
```diff
@@ -29,6 +29,8 @@
 #include <stdint.h>
 #include <string.h>
 
+#include "log.h"
```
rr-: Not needed log.h import as we don't debug anything in this file (reporting to not forget about it)

src/game/objects/common.h
```diff
@@ -12,12 +12,21 @@ typedef struct GAME_OBJECT_PAIR {
...
 extern const GAME_OBJECT_ID g_EnemyObjects[];
 extern const GAME_OBJECT_ID g_PlaceholderObjects[];
 extern const GAME_OBJECT_ID g_PickupObjects[];
 extern const GAME_OBJECT_ID g_GunObjects[];
 extern const GAME_OBJECT_PAIR g_GunAmmoObjectMap[];
-
+GAME_OBJECT_ID Object_GetCognate(
+    GAME_OBJECT_ID key_id, const GAME_OBJECT_PAIR *test_map);
+int16_t *Object_GetBoundsCognate(
+    GAME_OBJECT_ID key_id, const GAME_OBJECT_BOUNDS_PAIR *test_map);
```
rr-: Let's make this function private.

src/game/objects/common.h
```diff
@@ -12,12 +12,21 @@ typedef struct GAME_OBJECT_PAIR {
     const GAME_OBJECT_ID value_id;
 } GAME_OBJECT_PAIR;
 
+typedef struct GAME_OBJECT_BOUNDS_PAIR {
+    const GAME_OBJECT_ID key_id;
+    const int16_t *bounds;
+} GAME_OBJECT_BOUNDS_PAIR;
```
rr-: Let's make this struct private.

src/game/lara/lara.c
```diff
@@ -396,6 +396,32 @@ void Lara_UseItem(int16_t object_num)
...
+    case O_PUZZLE_OPTION4:
+    case O_LEADBAR_OPTION: {
+        int16_t hole_item_number = Object_CheckForHole(object_num);
+        if (hole_item_number == NO_ITEM) {
+            LOG_DEBUG("Inventory no");
+            Sound_Effect(SFX_LARA_NO, NULL, SPM_NORMAL);
+            return;
+        }
+        ITEM_INFO *item = &g_Items[hole_item_number];
+        item->priv = (void *)true;
```
rr-: Here we are using the `priv` member as a public API, which is not the intended use. To improve this, could we consider using `interact_target.item_num = hole_ite_number` and then checking for that in the relevant items instead? It's what we already do for switches and pickups, so not only it eliminates a design flaw, it also increases overall approach consistency.

src/game/objects/common.c
```diff
@@ -5,6 +5,9 @@
 #include "game/inventory.h"
 #include "game/items.h"
 #include "game/lara.h"
+#include "game/objects/general/keyhole.h"
+#include "game/objects/general/puzzle_hole.h"
+#include "game/objects/traps/midas_touch.h"
```
rr-: We shouldn't include these inside `common.c` – generalizations shouldn't depend on specific implementations.

To deal with `Object_CheckForHole` having to know too much about specific items, I propose another series of refactors (that should be landed in a separate PR first, I can do that):

- convert `int16_t[12]` to a proper `OBJECT_BOUNDS`  struct
- introduce `OBJECT_BOUNDS *Object_GetBounds(GAME_OBJECT_ID object_num)`
- introduce `OBJECT_BOUNDS *(*get_bounds)(void);` in `OBJECT_INFO`
- introduce specializations in each object type that needs it
- now only call `Object_GetBounds(object_num)` without having to know the exact implementation

rr- (commented):
Thank you. It still needs a bit of polishing around the edges regarding the interfacing, but it's so good to see `g_InvChosen` go away.  Good work :)

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@
 - fixed saving and loading the music timestamp when the load current music option is enabled (#1237)
 - fixed the remember played music option always being enabled (#1249, regression since 2.16)
 - fixed the underwater SFX playing for one frame at the start of Palace Midas (#1251)
+- fixed Lara saying "No" when saving or loading near a puzzle hole or keyhole (#1255)
```
rr-: We'll need changelog entries for the other issues as well.

src/game/objects/common.c
lahm86: I was thinking of a `bool (*is_interactable)(int16_t item_num)` in `OBJECT_INFO` and that way you could leave the specifics to the keyhole, puzzle slot etc, but to be honest I prefer @rr-'s solution instead as it could be used over and above the current case.

src/game/objects/common.c
```diff
@@ -44,13 +49,94 @@ const GAME_OBJECT_ID g_GunObjects[] = {
...
+            // TODO Make all Lara_TestPosition use const bounds?
+            return (int16_t *)pair->bounds;
+        }
+        pair++;
+    }
+
+    return NULL;
+}
+
+int16_t Object_CheckForHole(GAME_OBJECT_ID object_id)
```
lahm86: Re the naming, perhaps `Object_FindKeyItemSlot`? "Key" in the sense it's a crucial item as opposed to just keys. Or perhaps `Object_FindReceptacle`.

rr-: +1 for Object_FindReceptacle :)

walkawayy: Love that name. I didn't want to use key since it was confusing on if it was key vs key item.

src/game/objects/common.h
walkawayy: Dumb question, but do you mean just move the struct definition into the `common.c` file?

CHANGELOG.md
walkawayy: I combined the saying "no" issues since they stemmed from the same code problem. No changelog for #1254 because it's unreleased.

walkawayy:
@lahm86 `KeyHole_Trigger` runs for keyholes and puzzle holes when Lara uses them. But `KeyHole_Collision` continues to run while `PuzzleHole_Collision` doesn't anymore. This only results in a small minor difference where Lara says "no" if you press action in front of a used keyhole, but she doesn't say anything in front of a used puzzle hole. It's not really a big deal in the grand scheme of things, but I wanted to bring it up.

walkawayy:
Newest v2 build. (v1 was not the right build sorry)
[TR1X Key Items v2.zip](https://github.com/LostArtefacts/TR1X/files/14887064/TR1X.Key.Items.v2.zip)

lahm86:
> @lahm86 `KeyHole_Trigger` runs for keyholes and puzzle holes when Lara uses them. But `KeyHole_Collision` continues to run while `PuzzleHole_Collision` doesn't anymore. This only results in a small minor difference where Lara says "no" if you press action in front of a used keyhole, but she doesn't say anything in front of a used puzzle hole. It's not really a big deal in the grand scheme of things, but I wanted to bring it up.

Just checked in the current release and that seems to be the case already, unless I'm misunderstanding.

walkawayy:
> > @lahm86 `KeyHole_Trigger` runs for keyholes and puzzle holes when Lara uses them. But `KeyHole_Collision` continues to run while `PuzzleHole_Collision` doesn't anymore. This only results in a small minor difference where Lara says "no" if you press action in front of a used keyhole, but she doesn't say anything in front of a used puzzle hole. It's not really a big deal in the grand scheme of things, but I wanted to bring it up.
> 
> Just checked in the current release and that seems to be the case already, unless I'm misunderstanding.

Ya idk if it's really an issue, but I wanted to ask if we should change one or the other so they match. Maybe it was done that way because a used puzzle hole has a visual for a fill while you can't tell if a keyhole was used by visual only?

src/game/objects/common.h
rr-: Yes, correct

lahm86:
Agreed, that would make sense. I don't feel it's worth changing as part of this - maybe in the future if it's a request raised by players/builders.

CHANGELOG.md
rr-: We still should mention #1268 in the brackets.

src/game/objects/general/keyhole.c
```diff
@@ -25,10 +25,6 @@ int16_t g_KeyHoleBounds[12] = {
     +10 * PHD_DEGREE,
 };
 
-int32_t g_PickUpX;
-int32_t g_PickUpY;
-int32_t g_PickUpZ;
```
rr-, line 27: Oh good that you've removed this, thanks.

src/game/phase/phase_inventory.c
```diff
@@ -238,27 +239,24 @@ static GAMEFLOW_OPTION Inv_Close(void)
...
+    case O_PUZZLE_OPTION2:
+    case O_PUZZLE_OPTION3:
+    case O_PUZZLE_OPTION4:
+    case O_LEADBAR_OPTION:
+        Lara_UseItem(inv_chosen);
         break;
 
-    case O_BIGMEDI_OPTION:
-        Lara_UseItem(O_BIGMEDI_OPTION);
+    default:
```
rr-, line 259: Let's put a blank line above `default:`

aredfan:
I tested every scenario I can think of while having Lara stand in front of a receptacle.  Lara doesn't say "No" when saving/loading, or when selecting guns or medi packs (full and reduced health). I tried many different inputs and none of these disrupted Lara's interaction.  Lara still says "No" for every plot item, if selected incorrectly, except for the Scion.

**Lara when incorrectly using the Scion:**

| .           | TombATI | 3.1.1   | v2      |
|-------------|---------|---------|---------|
| Keyhole     | "No"    | "No"    | Silence |
| Puzzle hole | "No"    | "No"    | Silence |

**Edit:** I removed the Midas part of the table as it's not really relevant.

rr- (approved):
Please let's remember to use squash when merging. Thanks

lahm86 (approved):
LGTM

PR #1272, opened 2024-04-04
data: fix underwater twist animation

### LINKED ISSUE #1242: Wrong Lara rotation during UW twist
During the underwater roll/twist animation, Lara appears to be rotated at a random angle for a single frame in the second half of the animation.

@lahm86 can you please check? This is more evident with 60 fps and turbo=-2 in the latest develop.

lahm86:
Resolves #1242.

#### Checklist

#### Description

This fixes issues in the underwater twist animation from TR2.
- There were incorrect rotations on a frame
- It was triggering the 180 degree flip effect on the wrong animation and frame
- The bubbles flip effect was set on the wrong frame - it never came into effect as the frame index was above the animation length
- The end animation was linking to the wrong frame index of the idle animation, resulting in a jolt in Y pos.

I've put details in the issue so we can refer back when fixing the same in TR2X.

https://youtu.be/RsED6MSmiPI


walkawayy (approved):
Looks amazing.

rr-:
I'm sorry to be that guy, but the resulting animation is longer and still suffers from jerking at the end:

[out.webm](https://github.com/LostArtefacts/TR1X/assets/1045476/6b73ab4e-fca1-401a-96df-83d2a74d236b)

lahm86:
The jerking isn't clear to me - can you try the updated one...better or worse?

As for the increase in length, I can't explain that. No new frames have been added, but perhaps because we're now linking things up properly this is how it was intended. Either that or WadTool rectified another original issue I'm not aware of.

rr-:
The jerking is gone (please observe her right foot near the end of the animation in the clip above to see the difference):

[out.webm](https://github.com/LostArtefacts/TR1X/assets/1045476/d8df8feb-a192-4150-a8bb-4b929aafa329)

Any chance we can reduce the animation length somehow to keep it with OG? Remove some keyframe maybe?

lahm86:
Oh yes, I see now 👀 Thanks.
I'll see if I can figure out where the extra length is coming from 👍

lahm86:
Figured it out - it's another mistake in the original import from TR2. We were skipping a frame, so the current TR1X release is too short compared to actual OG TR2. Here is a video comparison of this PR with TR2, frame-by-frame. We are still one frame longer - you can see Lara needs one more frame to finish lining up at the very end - but I believe it's because we have fixed the linking and crucially the horrible mess in the middle that you'll see in TR2.

https://youtu.be/vmYCBiMGRHU

I can't see a way to edit it to remove this extra frame without rebuilding the animations completely (beyond my skill). I tried a few things like flipping slightly earlier but that means we have to do it in the first animation again, which re-introduces the odd rotation frame (and pancake Lara). If we flip slightly later and skip a frame, the camera doesn't have enough time to catch up and it looks odd. I think this is as good as we can get, unless we want to perhaps ask the custom level community if they can improve it? It maybe needs a bit more expertise in WadTool.

rr-:
Thank you for looking into this. IMO the best course of action is to merge this PR in its current state, so that we have some immediate improvement, and in the meantime throw the idea for further enhancement on Discord, not necessarily LostArtefacts – I think one of the TRLE servers will be best suited for this. If the community members manage to improve on it, we can then bundle it in 4.1 or later versions.

lahm86:
Sounds like a good plan.

lahm86:
I've raised #1275 to track this, so we can close the original issue against the fixes here.

PR #1277, opened 2024-04-05
Object bounds refactor

rr-:
#### Checklist

- [X] I have read the [coding conventions](https://github.com/LostArtefacts/TR1X/blob/master/CONTRIBUTING.md#coding-conventions)
- [X] I have added a changelog entry about what my pull request accomplishes, or it is an internal change

#### Description

For @walkawayy's #1267

PR #1279, opened 2024-04-06
Improve frame structuring

rr-:
#### Checklist

#### Description

**Meant to be reviewed after #1277** as it builds on top of it.
Also meant to be squashed as the history is a mess, and there seems to be no sensible way to make this more incremental.

Internal change that replaces `int16_t*` with `FRAME_INFO`. This makes the game logic in general much more readable. However, this enhancement comes at a cost - `FRAME_INFO` has a dynamic size based on the number of mesh rotations, where the array size is specified by nmeshes, and as such I need to deal with this in two fundamental places: level load routines, and savegame routines, which sadly still deal with raw data offsets rather than frame indices.


lahm86 (commented):
This all looks great bar one issue. I tested everything I could think of, so all animations, enemy behaviour, saving and loading and suchlike. 

The one issue is if you save with guns drawn, then reload, Lara's arms don't restore properly (and equally saving/loading while unarmed, you get bad arm animations when you next draw). I've done some testing and we can greatly simplify this part while solving it. I don't see a need to save the `frame_base` on the arm, as it's always set to the corresponding animation's frame base.

The diff below resolves the issue. Creating the extra function may be more than necessary, but it is needed in  `Savegame_Legacy_WriteArm` (although as far as I can tell we never write out legacy files).

I have tested this against saving and loading in every gun scenario - unarmed, armed, shooting and re-holstering - both with (old and new) TR1X saves and TombATI saves.

```diff
diff --git a/src/game/gun.h b/src/game/gun.h
index c76cce66..4a554dd0 100644
--- a/src/game/gun.h
+++ b/src/game/gun.h
@@ -13,3 +13,4 @@ int32_t Gun_FireWeapon(
     int32_t weapon_type, ITEM_INFO *target, ITEM_INFO *src, PHD_ANGLE *angles);
 void Gun_HitTarget(ITEM_INFO *item, GAME_VECTOR *hitpos, int16_t damage);
 void Gun_DrawFlash(LARA_GUN_TYPE weapon_type, int32_t clip);
+GAME_OBJECT_ID Gun_GetLaraAnimation(LARA_GUN_TYPE gun_type);
diff --git a/src/game/gun/gun.c b/src/game/gun/gun.c
index 15acf369..7415bc07 100644
--- a/src/game/gun/gun.c
+++ b/src/game/gun/gun.c
@@ -180,29 +180,29 @@ void Gun_InitialiseNewWeapon(void)
     g_Lara.right_arm.frame_number = LF_G_AIM_START;
     g_Lara.target = NULL;
 
-    switch (g_Lara.gun_type) {
-    case LGT_PISTOLS:
-    case LGT_MAGNUMS:
-    case LGT_UZIS:
-        g_Lara.right_arm.frame_base = g_Objects[O_PISTOLS].frame_base;
-        g_Lara.left_arm.frame_base = g_Objects[O_PISTOLS].frame_base;
-        if (g_Lara.gun_status != LGS_ARMLESS) {
-            Gun_Pistols_DrawMeshes(g_Lara.gun_type);
-        }
-        break;
+    const GAME_OBJECT_ID anim_type = Gun_GetLaraAnimation(g_Lara.gun_type);
+    g_Lara.right_arm.frame_base = g_Objects[anim_type].frame_base;
+    g_Lara.left_arm.frame_base = g_Objects[anim_type].frame_base;
 
-    case LGT_SHOTGUN:
-        g_Lara.right_arm.frame_base = g_Objects[O_SHOTGUN].frame_base;
-        g_Lara.left_arm.frame_base = g_Objects[O_SHOTGUN].frame_base;
-        if (g_Lara.gun_status != LGS_ARMLESS) {
+    if (g_Lara.gun_status != LGS_ARMLESS) {
+        if (anim_type == O_SHOTGUN) {
             Gun_Rifle_DrawMeshes();
+        } else {
+            Gun_Pistols_DrawMeshes(g_Lara.gun_type);
         }
-        break;
+    }
+}
 
+GAME_OBJECT_ID Gun_GetLaraAnimation(LARA_GUN_TYPE gun_type) {
+    switch (gun_type) {
+    case LGT_PISTOLS:
+    case LGT_MAGNUMS:
+    case LGT_UZIS:
+        return O_PISTOLS;
+    case LGT_SHOTGUN:
+        return O_SHOTGUN;
     default:
-        g_Lara.right_arm.frame_base = g_Objects[O_LARA].frame_base;
-        g_Lara.left_arm.frame_base = g_Objects[O_LARA].frame_base;
-        break;
+        return O_LARA;
     }
 }
 
diff --git a/src/game/savegame/savegame_bson.c b/src/game/savegame/savegame_bson.c
index 5f385a7a..683b9d50 100644
--- a/src/game/savegame/savegame_bson.c
+++ b/src/game/savegame/savegame_bson.c
@@ -7,7 +7,6 @@
 #include "game/inventory.h"
 #include "game/items.h"
 #include "game/lara.h"
-#include "game/level.h"
 #include "game/lot.h"
 #include "game/music.h"
 #include "game/room.h"
@@ -693,19 +692,6 @@ static bool Savegame_BSON_LoadArm(struct json_object_s *arm_obj, LARA_ARM *arm)
         return false;
     }
 
-    const int32_t frame_base = json_object_get_int(arm_obj, "frame_base", -1);
-    if (frame_base != -1) {
-        bool found = false;
-        const LEVEL_INFO *const level_info = Level_GetInfo();
-        for (int i = 0; i < level_info->anim_frame_count; i++) {
-            if (level_info->anim_frame_offsets[i] == frame_base) {
-                arm->frame_base = &g_AnimFrames[i];
-                found = true;
-            }
-        }
-        assert(found);
-    }
-
     arm->frame_number =
         json_object_get_int(arm_obj, "frame_num", arm->frame_number);
     arm->lock = json_object_get_int(arm_obj, "lock", arm->lock);
@@ -1163,11 +1149,7 @@ static struct json_array_s *SaveGame_BSON_DumpFx(void)
 static struct json_object_s *Savegame_BSON_DumpArm(LARA_ARM *arm)
 {
     assert(arm);
-    const LEVEL_INFO *const level_info = Level_GetInfo();
-    const int32_t frame_base =
-        level_info->anim_frame_offsets[arm->frame_base - g_AnimFrames];
     struct json_object_s *arm_obj = json_object_new();
-    json_object_append_int(arm_obj, "frame_base", frame_base);
     json_object_append_int(arm_obj, "frame_num", arm->frame_number);
     json_object_append_int(arm_obj, "lock", arm->lock);
     json_object_append_int(arm_obj, "x_rot", arm->rot.x);
diff --git a/src/game/savegame/savegame_legacy.c b/src/game/savegame/savegame_legacy.c
index a63bf798..b3e04a69 100644
--- a/src/game/savegame/savegame_legacy.c
+++ b/src/game/savegame/savegame_legacy.c
@@ -375,18 +375,8 @@ static void Savegame_Legacy_ReadLara(LARA_INFO *lara)
 
 static void Savegame_Legacy_ReadArm(LARA_ARM *arm)
 {
-    int32_t frame_base;
-    Savegame_Legacy_Read(&frame_base, sizeof(int32_t));
-
-    bool found = false;
-    const LEVEL_INFO *const level_info = Level_GetInfo();
-    for (int i = 0; i < level_info->anim_frame_count; i++) {
-        if (level_info->anim_frame_offsets[i] == frame_base) {
-            arm->frame_base = &g_AnimFrames[i];
-            found = true;
-        }
-    }
-    assert(found);
+    // frame_base is superfluous
+    Savegame_Legacy_Skip(sizeof(int32_t));
 
     Savegame_Legacy_Read(&arm->frame_number, sizeof(int16_t));
     Savegame_Legacy_Read(&arm->lock, sizeof(int16_t));

```

rr-:
Thanks @lahm86 works perfectly. Indeed we don't ever write the legacy save files. Do you think it's OK to remove this forever? (If so, I'd do that in a follow-up PR.)

lahm86:
> Thanks @lahm86 works perfectly. Indeed we don't ever write the legacy save files. Do you think it's OK to remove this forever? (If so, I'd do that in a follow-up PR.)

That sounds good, I can't think we'd ever want to have TR1X generate TombATI files? It would have problems anyway e.g. saving mid jump-twist and loading that into TombATI would probably not go down well.

lahm86 (approved):

PR #1282, opened 2024-04-07
output: animate room sprites

### LINKED ISSUE #449: Feature request: Animate sprite sequence 193 (Plant 3)
In Cistern and Tomb of Tihocan, there is a sprite sequence 193 called _Plant 3._ These plants are found in many water areas like here (Cistern, central pool) and here (Tomb of Tihocan, after the long slide).

Näyttökuva (348)
Näyttökuva (349)

As you see in this video, the sprite has actually 4 frames that play out a waving animation.

https://user-images.githubusercontent.com/97730054/155404864-e6d256c1-ced8-4b9d-a283-536ce6639de4.mp4

Yet, none of these plants are moving. And it also seems – as supported by the screenshots above – that all of the plants use the first frame. I have found no variation. This is the case on Ps1 too (based on some YouTube videos I watched).

Why are the extra frames not used, neither for animation or for having different looking plants? In the lack of better knowledge, I guess it's an error from Core. It would be nice to utilise them and that's why I'm requesting to animate the plants and bring them to life.

If you have more information or explanation what are the extra frames for, be sure to share it :)

A word of warning though. At least in custom level building when playing with sprites, they tend to break stuff in unexpected ways. So, if something is done to this, be careful? []:

lahm86:
Resolves #449.

#### Checklist

#### Description

This allows room sprites that are part of a sequence with more than one texture to be animated. This occurs at the same rate as regular animated textures. This implements behaviour in The Cistern and Tomb of Tihocan where the level files always contained these sequences, but there was no logic to cycle through them.

The `loaded` flag is needed now in `STATIC_INFO` to ensure we don't continue trying to manipulate stale objects after changing levels. The changes in `inject.c` are to reflect `level.c`, both ultimately have been tidied up. Crucially, in reading the level, we were previously skipping 2 bytes that contain the (negative) number of "meshes" so we now store that and use it to cycle through. Regular statics have a positive count here. 

The behaviour for The Cistern and Tomb of Tihocan can be switched off in the config tool (this is an independent option). The injection file here will set the number of textures to `-1` on that particular sprite sequence (193) when the option is disabled; when enabled, we just let the OG level data through. Custom level builders need only create sprite sequences with more than one texture, and should not ship the bin file.


lahm86:
I meant to add, I've tested on a level generated by TombEditor as well - attached for reference.
[level1.zip](https://github.com/LostArtefacts/TR1X/files/14898948/level1.zip)

src/game/inject.c
```diff
@@ -487,24 +491,24 @@ static void Inject_TextureData(
...
-            level_info->sprite_info_count++;
+            OBJECT_INFO *object = &g_Objects[object_num];
+            object->nmeshes = num_meshes;
+            object->mesh_index = mesh_index + level_info->sprite_info_count;
+            object->loaded = 1;
+        } else if (object_num - O_NUMBER_OF < STATIC_NUMBER_OF) {
+            STATIC_INFO *object = &g_StaticObjects[object_num - O_NUMBER_OF];
+            object->nmeshes = num_meshes;
+            object->mesh_number = mesh_index + level_info->sprite_info_count;
+            object->loaded = true;
```
rr-, line 509: This field is part of a bit field, so it still has to be `1` I think. Not sure how C will handle `true` in this case.

src/game/level.c
```diff
@@ -484,6 +484,7 @@ static bool Level_LoadObjects(MYFILE *fp)
         File_Read(&object->c.min.z, sizeof(int16_t), 1, fp);
         File_Read(&object->c.max.z, sizeof(int16_t), 1, fp);
         File_Read(&object->flags, sizeof(int16_t), 1, fp);
+        object->loaded = true;
```
rr-, line 487: Ditto

src/game/level.c
```diff
@@ -528,18 +529,22 @@ static bool Level_LoadSprites(MYFILE *fp)
...
-                fp);
+            OBJECT_INFO *object = &g_Objects[object_num];
+            object->nmeshes = num_meshes;
+            object->mesh_index = mesh_index;
+            object->loaded = 1;
+        } else if (object_num - O_NUMBER_OF < STATIC_NUMBER_OF) {
+            STATIC_INFO *object = &g_StaticObjects[object_num - O_NUMBER_OF];
+            object->nmeshes = num_meshes;
+            object->mesh_number = mesh_index;
+            object->loaded = true;
```
rr-, line 547: Ditto

src/game/output.c
```diff
@@ -943,6 +943,21 @@ void Output_AnimateTextures(void)
...
             ptr++;
         }
+
+        for (i = 0; i < STATIC_NUMBER_OF; i++) {
+            STATIC_INFO *static_info = &g_StaticObjects[i];
+            if (!static_info->loaded || static_info->nmeshes >= -1) {
+                continue;
+            }
+
+            int32_t num_meshes = -static_info->nmeshes;
```
rr-: Let's try to apply `const` where possible, also for pointers: `const type *const var = …` to mark that the pointer itself doesn't change

src/game/inject.c
lahm86: This is the new flag I added to `STATIC_INFO` as a bool. I wanted to change `OBJECT_INFO` to match then realised it was a bit field there.

src/game/output.c
lahm86: Will do 👍

src/game/inject.c
rr-: I think we should remove bit fields altogether, as they're about optimizing for memory consumption and IMO we need code readability first these days.

lahm86: Yes, agreed, it would be much more readable. I can follow that up separately throughout.

rr-: Yes – definitely a separate PR, not in scope of this.

src/game/output.c
lahm86: Updated. I'd like to rework regular animated textures i.e. potentially parse into structures to make the original part of this function more readable - it's effectively doing the same as the sprite rotations. Another follow-up task I think.

rr- (approved):

lahm86:
Fixed an issue where `Output_AnimateTextures` would exit early if a level had no regular animated textures.

walkawayy (approved):
Looks great.

rr- (approved):

PR #1285, opened 2024-04-08
Add OSD messages + rework game strings

### LINKED ISSUE #1220: Console logs for cheats and special actions
Idea floated by @rr- on discord.  The cheats/special actions are listed below.

- Fly Cheat
- Item Cheat
- Level Skip
- Turbo Speed

"Texture Filter ON/OFF" could be added too if it's within the scope of this issue.  It'll be good to let players know what they just enabled/disabled with an input.

____________________________

I also think it's fine to have console logs even if dev console is disabled because cheats are optional.

rr-:
#### Checklist

#### Description

Resolves #1220


src/game/shell.c
```diff
@@ -300,16 +301,37 @@ void Shell_ProcessInput(void)
...
+            g_Config.rendering.enable_perspective_filter
+                ? GS(OSD_PERSPECTIVE_FILTER_ON)
+                : GS(OSD_PERSPECTIVE_FILTER_OFF));
         Config_Write();
     }
 
     if (g_InputDB.toggle_fps_counter) {
         g_Config.rendering.enable_fps_counter ^= true;
+        Console_Log(
+            g_Config.rendering.enable_perspective_filter
```
walkawayy: This is the only bug I noticed. The FPS message is incorrectly linked to the `enable_perspective_filter` option.

walkawayy (changes requested):
Only noticed one issue. Everything else seems to work great! Only one question besides the bug I commented, is the console supposed to advance its timer in the inventory? If you use a `give guns` or some cheat in the inventory, the console log never disappears. This may be intended, but I wanted to bring it up just in case.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 - added the ability to move the look camera while targeting an enemy in combat (#1187)
 - added the ability to skip fade-out in stats screens
 - added support for animated room sprites in custom levels and an option to animate plant sprites in The Cistern and Tomb of Tihocan (#449)
+- added OSD messages for certain actions (#1220)
```
walkawayy: Might be worth writing out `on screen display` or `console`. Tbh, I didn't even know what OSD was at first til I googled it lol.

rr-:
>Only noticed one issue. Everything else seems to work great! Only one question besides the bug I commented, is the console supposed to advance its timer in the inventory? If you use a give guns or some cheat in the inventory, the console log never disappears. This may be intended, but I wanted to bring it up just in case.

The text takes long time to disappear but eventually it does. The time is the same both inside and outside the inventory.

I noticed it's possible to use the item cheat in the title menu which does not make sense. I'll fix that.

lahm86 (commented):
Looks great.

One other cheat I thought of was opening doors while in fly mode - `Door_OpenNearest` in `door.c`. Might be awkward though to log all the activated doors - maybe just a single general message if it succeeds or not?

Also, is it worth perhaps adding a table in `GAMEFLOW.md` with the available string definitions? Otherwise level builders may need to refer to old gameflow files if they are looking to override any.

rr-:
>Also, is it worth perhaps adding a table in GAMEFLOW.md with the available string definitions? Otherwise level builders may need to refer to old gameflow files if they are looking to override any.

I thought about this, and I've changed the `package` Docker command to rewrite the gameflow JSON file and include the strings from the `.def` file. This means the pre-releases and stable releases will contain the `strings` table populated as before. Unfortunately, Mac users will not benefit from this change due to the unique setup we use for its builds. I didn't see it as a priority to port this solution to this platform.

IMO this isn't the cleanest approach. I can easily see it breaking when we expect it the least, so if you can think of a way to land this table in `GAMEFLOW.md` instead without having to declare the same strings twice, I'm all for it. Maybe a pre-commit hook?

Added messages for the door cheat – good catch :)

lahm86 (approved):

walkawayy (approved):
LGTM. Tested only on Mac with no Docker though.

PR #1309, opened 2024-04-18
camera: remove LOS check on fixed cameras

### LINKED ISSUE #954: OG bug: Various doors have a one-frame displacement
I've been asked by @lahm86 to raise this issue so the second OG bug can be separated from #352 .

List of triggers that have the one frame "through the door" problem:

- LEVEL7A - # 11 - Room 15 - Key
- LEVEL7B - # 163 - Room 72 - Switch
- LEVEL7B - # 223 - Room 108 - Switch
- LEVEL10A - # 87 - Room 64 - Switch
- EGYPT - # 18 - Room 12 - Pad
- END - # 304 - Room 59 - Switch
- END - # 257 - Room 40 - Switch

lahm86:
Resolves #954.

#### Checklist

#### Description

LOS checks are removed in TR3 onwards for fixed cameras, so implemented here to fix the camera being thrown through doors that are yet to open. This also improves other areas of the game, for example if Lara is jumping down a slope with a low-ceiling and a fixed camera like in Tihocan or The Great Pyramid, or if she begins a level on a fixed camera but is just out of sight like in [Natla's Mines](https://github.com/LostArtefacts/TR1X/issues/1034#issuecomment-2028661547).

The one thing this could affect is custom levels with fixed cameras where Lara can go behind geometry, but even so it still feels like an improvement rather than having a fixed camera moving all over the place. I checked all the fixed cameras in OG and TRUB and they behave well with this change. Let me know your thoughts though if this should be optional, my feeling is it shouldn't.

The following demo shows the result. I implemented a test command to toggle between old and new, so keep an eye on the console to see the current state.

https://youtu.be/9OTPPHdjbCc

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## [Unreleased](https://github.com/LostArtefacts/TR1X/compare/stable...develop) - ××××-××-××
 - changed injection files to be placed in its own directory (#1306)
 - fixed camera vibrations when using the teleport command in 60 FPS (#1274)
+- fixed the camera being thrown through doors for one frame when looked at from fixed camera positions (#954)
 
 ## [4.0.3](https://github.com/LostArtefacts/TR1X/compare/4.0.2...4.0.3) - 2024-04-14
 - fixed flickering sprite pickups (#1298)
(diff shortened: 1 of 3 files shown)


rr- (approved):
LGTM. I agree there's a limit of what should be optional.

walkawayy (approved):
Wow it looks and plays great! That opening camera of Natla's Mines is finally fixed!

PR #1326, opened 2024-04-20
output: fix initial flicker

### LINKED ISSUE #1322: Title screen flashes briefly at a wrong viewport size during game launch
The game is rendered at what seems to be the initial window size, rather than desktop size.

For the FBO render mode, this lasts 1 frame.
For the window size render mode, this lasts 2 frames.

rr-:
#### Checklist

#### Description

Addresses #1322 – decreases the flicker down to what feels like a single frame. This is the best I can do at the moment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - fixed passport inventory item not being animated in 60 FPS (#1314)
 - fixed compass needle being too fast in 60 FPS (#1316, regression from 4.0)
 - fixed black screen flickers that can occur in 60 FPS (#1295)
+- decreased initial flicker upon game launch (#1322)
 
 ## [4.0.3](https://github.com/LostArtefacts/TR1X/compare/4.0.2...4.0.3) - 2024-04-14
 - fixed flickering sprite pickups (#1298)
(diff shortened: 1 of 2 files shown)


lahm86 (commented):
Looks good, but this happens if you launch maximized or in full screen mode:

image

walkawayy (approved):
Seems much better to me. I don't get flickering on W11 as far as I can tell.

rr-:
> Looks good, but this happens if you launch maximized or in full screen mode:
> 
> image

I can't reproduce on Windows (through Wine + virtual desktop) or Linux, on 640x480 or desktop resolution.

lahm86:
> > Looks good, but this happens if you launch maximized or in full screen mode:
> > image
> 
> I can't reproduce on Windows (through Wine + virtual desktop) or Linux, on 640x480 or desktop resolution.

The only way I can solve it is by undoing the changes, but I feel I'm missing something.

Here is a demo of it. So full-screen and maximized window mode result in the small display. Even if you reset the window, when you launch a level it goes back to a small output.

https://youtu.be/m8MEXSSNzFE

I've also noticed it doesn't remember my window size - so resizing a window, closing and reopening, the original size is restored.

@walkawayy assuming it was all good on your system?

rr-:
I've reworked the fix – can you please retest?

lahm86 (approved):
Works perfectly, thanks 😄

Issue #1295, opened 2024-04-10
Partial black screen moments - 60 FPS


aredfan:
There are a few places where Lara can encounter partial black screen moments, while playing in 60 FPS mode.

The issue in the first 2 videos can be reproduced consistently, the 3rd video less so.

- [x] Natla's Mines, room 72.

https://github.com/LostArtefacts/TR1X/assets/94776985/93bbec4e-55cc-4bdd-9b7a-8113d5eca775

- [x] Natla's Mines, room 68.

https://github.com/LostArtefacts/TR1X/assets/94776985/8224cc1f-f638-41b2-8616-496f5e64c1e1

- [x] Palace Midas, room 3.

https://github.com/LostArtefacts/TR1X/assets/94776985/cecec94f-d2a6-49e5-88af-93d7a03f4a37

____________________________

**New videos:**

- [x] St. Francis' Folly, room 42.

https://github.com/LostArtefacts/TR1X/assets/94776985/3ea67860-9e7c-4447-a326-be02632f75ac

- [x] Obelisk of Khamoon, room 35.

https://github.com/LostArtefacts/TR1X/assets/94776985/92d0852e-252e-4fa1-8348-5a96a971c5f3

aredfan:
I think this spot in Return to Egypt have the same issue as the one in Palace Midas, so it's another place to test when a fix is found.

- [x] Return to Egypt, room 15.

https://github.com/LostArtefacts/TR1X/assets/94776985/36594ef2-ebe2-4344-9d85-165273ba53b7

**New video:**

- [x] Lara's Home, room 12.

https://github.com/LostArtefacts/TR1X/assets/94776985/84496c4a-2d7d-4ef8-ba75-6c9962b9e48d

aredfan:
I've rechecked each of the scenarios shown above and the bug is fixed. Closing this issue as complete.

PR #1327, opened 2024-04-20
room_draw: include camera target room

lahm86:
Part of #1295.

#### Checklist

#### Description

This is a workaround for flickers in 60 FPS to ensure that the camera's target room is included in drawing checks. It does not solve _every_ issue, but the majority i.e. those where the visibility portal checks deem the target room to not be visible. What it doesn't solve is this:

image

This is all the same room, but the clipping is for some reason deciding to cull most of the left side.

To be clear too, this does not solve the OG issues detailed in #1034.

The camera going out of bounds is not a new phenomenon in TR1. This PR hopefully puts us in a situation that's no(t much) worse than OG/30 FPS until we get a better solution for the camera collision control in place.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - fixed console not retaining changed user settings across game relaunches (#1318)
 - fixed passport inventory item not being animated in 60 FPS (#1314)
 - fixed compass needle being too fast in 60 FPS (#1316, regression from 4.0)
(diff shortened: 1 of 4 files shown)


walkawayy (approved):
Seems really nice no. I didn't notice the excess flickering that 60 FPS brought in. Great job.

rr- (approved):

PR #1383, opened 2024-06-13
inject: increase Scion collision radius

### LINKED ISSUE #1381: OG bug: shooting the scion with the shotgun is difficult
If shooting the scion with the shotgun, most shots will miss. Players will generally have the pistols to fall back on, but if it's used in a custom level where pistols are removed, this can potentially lead to softlock (assuming the scion has a trigger for progression per OG).

See https://github.com/LostArtefacts/TR-Rando/pull/697 for reference.

lahm86:
Resolves #1381.

#### Checklist

#### Description

Adds the ability to edit the centre point and collision radius of meshes. I've applied this to the Scion object in the Great Pyramid to allow it to be more easily shot with the shotgun. We only increase the radius for the Scion (from 63 to 256) but it made sense to add the ability to edit the centre point too in case that's ever required. Existing mesh edits will default to 0 here for these new properties.

Custom levels can edit this directly in WadTool, so this is only relevant to OG. I'm also unsure about making this optional, WDYT?


rr- (approved):
LGTM. I agree that this feature can be permanent.

walkawayy (approved):
Nice! I agree it should be permanent.

PR #1386, opened 2024-06-17
Rendering cleanup

rr-:
#### Checklist

#### Description

- Removes some insane buffer swap logic around background surfaces, instead assumes 2D renderer gets called first, then the 3D renderer (sprite statics are 3D, 2D is only used for background images and FMVs)
- Removes `Output_DrawBackdropImage` as like stated above, this is now called always before 3D logic
    **Note** this may be a questionable decision – in the future if we want to draw an overlay 2D image such as for TR3 loading screens fades, we'll need to also be able to draw an overlay image not just underlay, but I'd postpone supporting this until we actually need that
- Removes `RenderToggle` crap, replaces with `BeginScene` `EndScene` which is easier to follow
- Removes backbuffer shenanigans from 2D_Surface, callers now need to manage such things by themselves
- Removes backbuffers and their management from `s_output.c` and `s_fmv.c`, as they're not really needed – we can write to primary surfaces directly, as double-buffering is provided out-of-the-box by `SDL_GL_SwapWindow` that we use
- Removes conditional statements around rendering mode, introduces strategy pattern instead (`renderers/fbo_renderer.c`, `renderers/legacy_renderer.c`)
- Fixes FBO treating background images significantly different
- Fixes wireframe mode to always show background images and FMVs
- Makes wireframe width a configurable setting
- Changes background color to white for wireframe mode
    **Note** this looks weird with FMVs on screens that have mismatching aspect ratio. LMK if I should keep black side bars in that scenario.

The wireframe mode is not yet released, and as such, changes to it were not added to the changelog. The rest is internal stuff.


src/game/game/game.c
```diff
@@ -21,9 +21,10 @@
 
 #define FRAME_BUFFER(key)                                                      \
     do {                                                                       \
+        Output_BeginScene();                                                   \
         Game_DrawScene(true);                                                  \
         Input_Update();                                                        \
-        Output_DumpScreen();                                                   \
+        Output_EndScene();                                                     \
```
walkawayy, line 27: Looks like extra trailing whitespace here.

walkawayy (approved):
The changes seem nice! Stats screens, final stats, FMVs, cutscenes, inventory fades, loading screens, and wireframe all seem to work without issue.

src/game/game/game.c
rr-: Thanks. It's easy to miss, but we use `\` as a continuation mark for multiline macros, and our formatter aligns those `\` in a single column to the right. The extra spaces serve to add up to that extra alignment.

lahm86 (approved):
Everything LGTM 👍

PR #1389, opened 2024-06-24
Memory fixes, high number of savegame slot fixes, and requestor fixes and rework

walkawayy:
#### Checklist

#### Description
This PR fixes memory issues that occurred with savegames as well as a 64 savegame slot limit due to `int32_t gf_param = gf_option & ((1 << 6) - 1);` which was due to the bitflow operations in the gameflow. There is a second small commit to fix two small memory leaks in the gameflow as well.

The memory leak occurred if the number of savegame slots was over 16. Requestor's `item_flags` max value was 16 which was less than the default savegame amount in TR1X. Reworked much of the requestor code to account for this. Also reworked the `item_texts` to be safer.

The 64 slot limit was fixed by reworking the `GAMEFLOW_OPTION`s to no longer use bit shifting. This PR introduces a new gameflow info struct that contains that information. It's a new global struct and changed the phasers, so they don't return. But it seemed necessary since some of the issue was only being able to return one value.

This will need heavily tested, but lmk if you think the code changes look good before I ask for testing help. I'm surprised we didn't have more issues reported because the memory was severely corrupted. Maybe it only really showed up badly when I started testing with a really high number of savegames.

Also, there is a memory leak in `packer.c`, but I was unable to fix it. The `TEX_INFO *info = Memory_Alloc(sizeof(TEX_INFO));` are never freed, and there is some pointer to malloc'd pointer stuff going on in there.


src/global/types.h
```diff
@@ -965,17 +965,17 @@ typedef enum GAMEFLOW_OPTION {
...
+    GF_START_FMV = 2,
+    GF_START_DEMO = 3,
+    GF_EXIT_TO_TITLE = 4,
+    GF_LEVEL_COMPLETE = 5,
+    GF_EXIT_GAME = 6,
+    GF_START_SAVED_GAME = 7,
+    GF_RESTART_GAME = 8,
+    GF_SELECT_GAME = 9,
+    GF_START_GYM = 10,
+    GF_STORY_SO_FAR = 11,
```
rr-, line 978: We can remove the concrete int values altogether.

src/game/phase/phase_pause.h
```diff
@@ -3,3 +3,6 @@
 #include "game/phase/phase.h"
 
 extern PHASER g_PausePhaser;
+
+void Phase_Pause_Init(void);
+void Phase_Pause_Shutdown(void);
```
rr-, line 8: This makes the outermost game shell code too reliant on very specific parts of the code, i.e. the pause phaser. As such I'd try to avoid introducing these new functions and instead move the requestor initiation and destruction to `Phase_Pause_Start` and `Phase_Pause_End`.

rr- (changes requested):
Let's break this into smaller PRs – one for the requestor changes; one for gameflow changes; one for the memory leaks that can be merged immediately.

PR #1393, opened 2024-06-27
requester: fix memory corruption and refactor

### LINKED ISSUE #1374: Game can crash with more than 16 savegame slots
Like in title, game either outright crashes or after playing startup FMVs. Also, when anything above 218 is set, game will crash only after trying to start a new game or load a save.

walkawayy:
Fixes issues with incorrectly sized items_flags, standardizes requestor functions, and switches to snprintf during savegame scanning.

#### Checklist

#### Description
Part 2 of separating #1389 into more manageable pull requests. Fixes issues with incorrectly sized `items_flags`, standardizes requestor functions, and switches to snprintf during savegame scanning. Part of the fix for #1374 as this fixes incorrectly sized `items_flags` which could cause memory corruption. The next and final PR will be the final fix for gameflow bit shifting limiting savegame slots to 64.


src/game/requester.c
```diff
@@ -174,34 +189,28 @@ void Requester_SetHeading(REQUEST_INFO *req, const char *string)
     Text_Remove(req->heading);
     req->heading = NULL;
 
-    if (string) {
-        req->heading_text = string; // unsafe
-    } else {
-        req->heading_text = NULL;
-    }
+    size_t out_size = snprintf(NULL, 0, "%s", string) + 1;
```
rr-: `const`

src/game/requester.c
```diff
@@ -6,27 +6,41 @@
...
+
 #include <stdbool.h>
+#include <stdio.h>
 #include <string.h>
 
 #define BOX_BORDER 2
 #define BOX_PADDING 10
 
-void Requester_Init(REQUEST_INFO *req)
+void Requester_Init(REQUEST_INFO *req, uint16_t max_items)
```
rr-: `const uint16_t max_items`

src/game/phase/phase_pause.c
```diff
@@ -78,23 +84,27 @@ static int32_t Phase_Pause_DisplayRequester(
...
         Requester_SetHeading(&m_PauseRequester, header);
         Requester_AddItem(&m_PauseRequester, option1, 0);
         Requester_AddItem(&m_PauseRequester, option2, 0);
+        m_IsTextReady = true;
+        g_InputDB = (INPUT_STATE) { 0 };
+        g_Input = (INPUT_STATE) { 0 };
+    }
 
-        is_pause_text_ready = true;
+    // Don't allow back because it breaks the requestor items.
```
rr-: I don't quite understand this scenario, can you elaborate?

src/game/savegame/savegame.c
```diff
@@ -509,7 +515,9 @@ void Savegame_ScanSavedGames(void)
 
                 char *full_path =
                     Memory_Alloc(strlen(SAVES_DIR) + strlen(filename) + 2);
-                sprintf(full_path, "%s/%s", SAVES_DIR, filename);
+                size_t out_size =
+                    snprintf(NULL, 0, "%s/%s", SAVES_DIR, filename) + 1;
+                snprintf(full_path, out_size, "%s/%s", SAVES_DIR, filename);
```
rr-: The `snprintf(NULL, 0, …)` storage size calculation is not necessary here, as the buffer is guaranteed to have sufficient capacity in the allocation call above.

src/game/savegame/savegame.c
```diff
@@ -432,8 +433,13 @@ bool Savegame_Save(int32_t slot_num, GAME_INFO *game_info)
...
         req->item_flags[slot_num] &= ~RIF_BLOCKED;
-        sprintf(
-            &req->item_texts[req->item_text_len * slot_num], "%s %d",
+        size_t out_size =
+            snprintf(
+                NULL, 0, "%s %d", g_GameFlow.levels[g_CurrentLevel].level_title,
+                g_SaveCounter)
+            + 1;
+        snprintf(
+            req->item_texts[slot_num], out_size, "%s %d",
```
rr-: It isn't necessary to use `snprintf` in this instance. Initially, we determine the string's size and then print the calculated number of bytes into the buffer, which is effectively the same as just using `sprintf`. A better approach might be to allocate the buffer based on the calculated string size. However, in this case, we have a preallocated buffer of `req->item_text_len` size. Therefore, we should do `snprintf(req->item_texts[slot_num], req->item_text_len, …)`. Or better yet, postpone the string allocation from `Requester_Init` until we actually need to `snprintf` the strings at call sites, and instead of ceding this responsibility to the callers, wrap the entire operation (calculate item text size → allocate the item text buffer → printf the desired string to the item text buffer) in a new function.
(Actually it looks like we already have such function – `Requester_ChangeItem`).

src/game/savegame/savegame.c
```diff
@@ -544,29 +552,33 @@ void Savegame_ScanSavedGames(void)
...
             }
+
+            flags = req->item_flags[req->items_used] & ~RIF_BLOCKED;
+            size_t out_size = snprintf(
+                                  NULL, 0, "%s %d", savegame_info->level_title,
+                                  savegame_info->counter)
+                + 1;
+            snprintf(
+                savegame_text, out_size, "%s %d", savegame_info->level_title,
+                savegame_info->counter);
```
rr-, line 558: Here we use an intermediate buffer `savegame_text`, but the same principles apply, only with `MAX_LEVEL_NAME_LENGTH`.

src/game/requester.c
```diff
@@ -174,34 +189,28 @@ void Requester_SetHeading(REQUEST_INFO *req, const char *string)
...
-    if (string) {
-        strcpy(&req->item_texts[req->items * req->item_text_len], string);
-        req->item_flags[req->items] = flag;
-    } else {
-        req->item_flags[req->items] = 0;
+    if (req->items_used < req->max_items && string) {
+        size_t out_size = snprintf(NULL, 0, "%s", string) + 1;
+        snprintf(req->item_texts[req->items_used], out_size, "%s", string);
+        req->item_flags[req->items_used] = flag;
+        req->items_used++;
```
rr-: Don't we need to zero the item flags and the text buffer in case the string is NULL?

src/game/requester.c
```diff
@@ -174,34 +189,28 @@ void Requester_SetHeading(REQUEST_INFO *req, const char *string)
...
 void Requester_AddItem(REQUEST_INFO *req, const char *string, uint16_t flag)
 {
-    if (string) {
-        strcpy(&req->item_texts[req->items * req->item_text_len], string);
-        req->item_flags[req->items] = flag;
-    } else {
-        req->item_flags[req->items] = 0;
+    if (req->items_used < req->max_items && string) {
+        size_t out_size = snprintf(NULL, 0, "%s", string) + 1;
+        snprintf(req->item_texts[req->items_used], out_size, "%s", string);
```
rr-: This is the place where I'd do the string allocation. To make this function even more useful, I'd make it accept a `format` string and a variable number of arguments using the `va_start()` family of functions (see how `Log_Message` is implemented). This would allow the callers to do `Requester_AddItem(req, flag, "%s", level_title)` without having to care about memory allocation at all. Naturally this idiom would also need to be applied to `Requester_ChangeItem` that would have to reallocate the item text buffer.

walkawayy: Nice I didn't know about this function in C. ~~`Requester_ChangeItem` I noticed is never used. Do we want to keep it?~~

EDIT: nvm , I see your other comment.

src/game/requester.c
walkawayy: `if (req->items_used < req->max_items && string) {`

`Requester_AddItem` just doesn't run at all `string` is `NULL`. Is that ok or do you want to change it?

src/game/savegame/savegame.c
walkawayy: Ok. I was just gonna standardize to `snprintf` to ensure we don't run over the buffer size and ensure null termination. But I will change it back.

src/game/savegame/savegame.c
walkawayy: Yes I did it this way so we could use `Requester_AddItem` instead of editing the requestor itself directly. The if/else use different string formats. But if we make the change to your format idea in `Requester_AddItem`, that would eliminate the need to do this. So I'll make that change.

src/game/phase/phase_pause.c
walkawayy: Oops forgot to answer this sorry for spams.

So `Requester_Display` clears the requester if `menu_back` is pressed.
```
    } else if (g_InputDB.menu_back) {
        Requester_ClearTextstrings(req);
        return -1;
    }
```

If you test the pause function in `develop`, you'll notice pressing `menu_back` will cause the requester to flash. That's because the text is cleared and has to be reinitialized. Since the player can't exit the pause requestor with `menu_back`, they have to select `Continue` or `No` on the second page, I decided to just zeroize the input if they press it. That fixes the requester flashing. Also, because we `malloc` now, I didn't want the requester malloc and freeing in this scenario.

rr-: Thanks, this makes sense. Can we provide this context in the comment?

src/game/requester.c
rr-: Sorry - this makes sense. I don't think we'll ever try to add a NULL item anyway.

src/game/option/option_passport.c
```diff
@@ -122,16 +125,16 @@ static void Option_PassportFlipLeft(INVENTORY_ITEM *inv_item);
 
 void Option_PassportInit(void)
 {
-    g_SavegameRequester.item_texts = Memory_Alloc(
-        g_Config.maximum_save_slots * g_SavegameRequester.item_text_len);
-    m_SelectLevelRequester.item_texts = Memory_Alloc(
-        (g_GameFlow.level_count + 1) * m_SelectLevelRequester.item_text_len);
+    Requester_Init(&g_SavegameRequester, g_Config.maximum_save_slots);
+    Requester_Init(&m_SelectLevelRequester, g_GameFlow.level_count + 1);
+    Requester_Init(&m_NewGameRequester, MAX_GAME_MODES);
```
walkawayy, line 130: Do I need to traverse each `item_blocked` and `item_texts` and initialize them to `NULL`? Should I also be doing that in `Requester_ClearTextstrings`?

src/game/requester.c
```diff
@@ -174,34 +185,61 @@ void Requester_SetHeading(REQUEST_INFO *req, const char *string)
     Text_Remove(req->heading);
     req->heading = NULL;
 
-    if (string) {
-        req->heading_text = string; // unsafe
-    } else {
-        req->heading_text = NULL;
-    }
+    const size_t out_size = snprintf(NULL, 0, "%s", string) + 1;
+    snprintf(req->heading_text, out_size, "%s", string);
```
walkawayy, line 214: Should I also use malloc here in `Requester_SetHeading`? It's the only init malloc left using ` eq->item_text_len`.

src/game/requester.c
```diff
@@ -6,27 +6,38 @@
...
-    req->moredown = NULL;
-    for (int i = 0; i < MAX_REQLINES; i++) {
-        req->texts[i] = NULL;
-        req->item_flags[i] = 0;
-    }
+    req->max_items = max_items;
+    req->heading_text = Memory_Alloc(sizeof(char) * req->item_text_len);
+    req->item_blocked = Memory_Alloc(sizeof(bool) * max_items);
+    req->item_texts = Memory_Alloc(sizeof(char *) * max_items);
+    Requester_ClearTextstrings(req);
```
walkawayy, line 44: Should I traverse each `item_blocked` and `item_texts` and initialize them to `NULL`? Should I also be doing that in `Requester_ClearTextstrings`?

src/game/requester.c
rr-: I would say so, yes.

src/game/requester.c
rr-: No need to initialize to NULL, since `Memory_Alloc` zeroes the memory for us.
I don't think `ClearTextstrings` is required to deallocate the memory, it's only important that we don't leave it dangling and call the `Shutdown` route.

src/game/option/option_passport.c
```diff
@@ -48,55 +48,58 @@ static PASSPORT_STATUS m_PassportStatus = {
...
-    .items = MAX_GAME_MODES,
+    .items_used = 0,
+    .max_items = MAX_GAME_MODES,
     .requested = 0,
     .vis_lines = MAX_GAME_MODES,
     .line_offset = 0,
     .line_old_offset = 0,
     .pix_width = 162,
     .line_height = TEXT_HEIGHT + 7,
+    .blockable = false,
```
rr-: Let's call it `is_blockable` to indicate it's a bool

src/game/option/option_passport.c
```diff
@@ -331,12 +334,12 @@ static void Option_PassportInitSelectLevelRequester(void)
...
+    Requester_ClearTextstrings(req);
     Requester_SetHeading(req, GS(PASSPORT_SELECT_MODE));
-    Requester_AddItem(req, GS(PASSPORT_MODE_NEW_GAME), 0);
-    Requester_AddItem(req, GS(PASSPORT_MODE_NEW_GAME_PLUS), 0);
-    Requester_AddItem(req, GS(PASSPORT_MODE_NEW_GAME_JP), 0);
-    Requester_AddItem(req, GS(PASSPORT_MODE_NEW_GAME_JP_PLUS), 0);
+    Requester_AddItem(req, 0, "%s", GS(PASSPORT_MODE_NEW_GAME));
+    Requester_AddItem(req, 0, "%s", GS(PASSPORT_MODE_NEW_GAME_PLUS));
+    Requester_AddItem(req, 0, "%s", GS(PASSPORT_MODE_NEW_GAME_JP));
+    Requester_AddItem(req, 0, "%s", GS(PASSPORT_MODE_NEW_GAME_JP_PLUS));
```
rr-: `0` → `false`, since the new param is a bool

src/game/phase/phase_pause.c
```diff
@@ -78,23 +84,28 @@ static int32_t Phase_Pause_DisplayRequester(
...
-        Requester_Init(&m_PauseRequester);
+    if (!m_IsTextReady) {
+        Requester_ClearTextstrings(&m_PauseRequester);
         Requester_SetSize(&m_PauseRequester, 2, -48);
         m_PauseRequester.requested = requested;
         Requester_SetHeading(&m_PauseRequester, header);
-        Requester_AddItem(&m_PauseRequester, option1, 0);
-        Requester_AddItem(&m_PauseRequester, option2, 0);
+        Requester_AddItem(&m_PauseRequester, 0, "%s", option1);
+        Requester_AddItem(&m_PauseRequester, 0, "%s", option2);
```
rr-: `0` → `false`

src/game/requester.c
```diff
@@ -174,34 +185,61 @@ void Requester_SetHeading(REQUEST_INFO *req, const char *string)
...
-    } else {
-        req->heading_text = NULL;
-    }
+    const size_t out_size = snprintf(NULL, 0, "%s", string) + 1;
+    snprintf(req->heading_text, out_size, "%s", string);
 }
 
 void Requester_ChangeItem(
-    REQUEST_INFO *req, int32_t idx, const char *string, uint16_t flag)
+    REQUEST_INFO *req, int32_t idx, bool blocked, const char *fmt, ...)
```
rr-: `const bool is_blocked`, `const char *const fmt` (please remember **not** to add these consts to the header file, as the usefulness of `const` in function definitions is very different from just their declarations)

src/game/requester.c
```diff
@@ -174,34 +185,61 @@ void Requester_SetHeading(REQUEST_INFO *req, const char *string)
...
+    const size_t out_size = vsnprintf(NULL, 0, fmt, va) + 1;
+    req->item_texts[idx] = Memory_Alloc(sizeof(char) * out_size);
+    va_end(va);
+    Memory_FreePointer(&req->item_texts[idx]);
+
+    if (req->item_texts[idx] != NULL) {
+        va_start(va, fmt);
+        vsnprintf(req->item_texts[idx], out_size, fmt, va);
+        req->item_blocked[idx] = blocked;
+        va_end(va);
```
rr-: I think we can simplify this function a bit:
1. `Memory_Alloc` is guaranteed to crash if the game fails to allocate the memory, so the NULL check can be removed.
2. After applying 1., since we no longer care about special scenarios, we can forgo the middle `va_end` / `va_start` calls and do `va_start()`, `vsnprintf(NULL, 0, …)`, `vsnprintf(buffer, out_size, …)`, `va_end()` in succession. Please double check my claim with the documentation, though – I might be wrong.

src/game/requester.c
```diff
@@ -174,34 +185,61 @@ void Requester_SetHeading(REQUEST_INFO *req, const char *string)
...
+
+    if (req->item_texts[req->items_used] != NULL) {
+        va_start(va, fmt);
+        vsnprintf(req->item_texts[req->items_used], out_size, fmt, va);
+        req->item_blocked[req->items_used] = blocked;
+        req->items_used++;
+        va_end(va);
+    } else {
+        Memory_FreePointer(&req->item_texts[req->items_used]);
+    }
```
rr-: Ditto

src/game/requester.h
```diff
@@ -4,11 +4,12 @@
 
 #include <stdint.h>
 
-void Requester_Init(REQUEST_INFO *req);
-void Requester_Remove(REQUEST_INFO *req);
+void Requester_Init(REQUEST_INFO *req, const uint16_t num_items);
```
rr-: In this context, `const` is unnecessary – it indicates that the value of `num_items` will not be modified within the function body, however, since this is only a function declaration, there is no function body to consider. Because of this, it's a common practice to avoid declaring primitive types as `const` in header files. (Please note that pointers are a different story – declaring them as `const` in declarations promises that the function body won't alter the content they point to, and that is a useful promise.)

rr- (commented):
I'd like to propose one last improvement - currently we have three array of pointers:

```c
    bool *item_blocked;
    char **item_texts;
    TEXTSTRING *texts[MAX_REQLINES];
```

I think it'd be elegant to wrap these together in a new structure, `REQUESTER_ITEM`, and then allocate it once like so:

```c
    req->items = Memory_Alloc(sizeof(REQUESTER_ITEM) * max_items);
```

walkawayy:
Ok I think this is looking a lot better. I removed even more unused vars, malloc the text header, and renamed the blocks for clarity. Tested in valgrind and this code has no memory leaks.

rr- (approved):

PR #1400, opened 2024-07-02
output: read vertex buffer size from level

### LINKED ISSUE #1398: TR1X bug: game crashes with large rooms
If a room has more than 1500 vertices, the game crashes on load. In TombATI, levels can load where rooms exceed this limit, but this seems to be down to luck with the original memory layout, so it's in effect an OG bug/limitation.

lahm86:
Resolves #1398.

#### Checklist

#### Description

This replaces the fixed vertex buffer size of `1500` with a maximum value read from the level - so the "biggest" room, object or static mesh will determine the array size. In most cases it will be the biggest room, but this covers all bases. An alternative to doing the object/static check would be setting an initial value of `1500` and just iterating over the room meshes, so let me know if that's preferable.

### DIFF
--- a/src/game/level.c
+++ b/src/game/level.c
@@ -891,7 +891,7 @@ static void Level_CompleteSetup(int32_t level_num)
     // Configure enemies who carry and drop items
     Carrier_InitialiseLevel(level_num);
 
-    size_t max_vertices = Level_CalculateMaxVertices();
+    const size_t max_vertices = Level_CalculateMaxVertices();
     LOG_INFO("Maximum vertices: %d", max_vertices);
     Output_ReserveVertexBuffer(max_vertices);
 
--- a/src/game/level.c
+++ b/src/game/level.c
@@ -54,7 +54,7 @@ static bool Level_LoadTexturePages(MYFILE *fp);
 static bool Level_LoadFromFile(
     const char *filename, int32_t level_num, bool is_demo);
 static void Level_CompleteSetup(int32_t level_num);
-static void Level_CalculateMaxVertices(void);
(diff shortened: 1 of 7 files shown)


src/game/level.c
```diff
@@ -925,6 +928,39 @@ static void Level_CompleteSetup(int32_t level_num)
...
+            MAX(max_vertices, *(g_Meshes[static_info->mesh_number] + 5));
+    }
+
+    for (int32_t i = 0; i < g_RoomCount; i++) {
+        max_vertices = MAX(max_vertices, *g_RoomInfo[i].data);
+    }
+
+    LOG_INFO("Maximum vertices: %d", max_vertices);
+    Output_ReserveVertexBuffer(max_vertices);
+}
```
rr-, line 963: Let's have this function return an `int32_t` and call the `Output_…` function inside higher up in the hierarchy.

src/game/output.h
```diff
@@ -15,6 +15,7 @@ void Output_DownloadTextures(int page_count);
 RGBA_8888 Output_RGB2RGBA(const RGB_888 color);
 void Output_SetPalette(RGB_888 palette[256]);
 RGB_888 Output_GetPaletteColor(uint8_t idx);
+void Output_ReserveVertexBuffer(const size_t size);
```
rr-: Using `const` for primitive types in a function forward declaration has no effect and should be removed.

lahm86: Of course, silly mistake

walkawayy (approved):
LGTM. Might be worth a README mention when talking about raised limits?

rr- (approved):

PR #1411, opened 2024-07-12
installer: use existing installation path

### LINKED ISSUE #1350: Installation path when upgrading TR1X should be the same as the initial installation
When you first install TR1X, the program suggest you Documents\TR1X as a destination folder but you are free to change that to a different folder.

If you do that and you run the installer of a newer version to upgrade it, you can select "TR1X" as an installation source since you already have all the files there. 
After you select that tho, the program STILL defaults Documents\TR1X for the _upgrade_ and so you are making a second identical installation if you advance that, and so you have to select your personal installation folder to apply the upgrade on a single installation.

The installer should just respect the personal installation folder once you select "TR1X" instead of forcing you to re-select your installation folder just to upgrade it.
It should be easy to implement since the program is already able to detect your TR1X installation in the first place, this change would make sense and make the upgrading system more streamlined.

Thank you :)

lahm86:
Resolves #1350.

#### Checklist

#### Description

The existing installation path - if it exists - will be suggested when performing an upgrade by selecting the TR1X install source.

Should we just leave the reg key as `Tomb1Main`? It's out of sight as opposed to a folder name, so adding extra code for backwards compatibility seems like unnecessary overhead imo.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - added unobtainable secrets stat support in the gameflow (#1379)
 - added a wireframe mode
 - changed console caret blinking rate (#1377)
+- changed the TR1X install source in the installer to suggest using the existing installation directory (#1350)
 - fixed config tool and installer missing icons (#1358, regression from 4.0)
 - fixed looking forward too far causing an upside down camera frame (#1338)
 - fixed the enemy bear behavior in demo mode (#1370, regression since 2.16)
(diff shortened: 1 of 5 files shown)


rr-:
>Should we just leave the reg key as Tomb1Main? It's out of sight as opposed to a folder name, so adding extra code for backwards compatibility seems like unnecessary overhead imo.

I'd say no, in the future event we might want to use a registry-based config. In the past it was highlighted that JSON files do not translate well between TRLEs, where you're supposed to reconfigure your key mappings between every game, as opposed to just having one configuration shared between all games, which is a valid concern.

If anything, I'd rename the registry key to `TR1X`. But in such a case I'd leave backwards compatibility with a comment which version we changed this in, so that we eventually drop this code after a sufficient time has passed.

lahm86:
I've added in a check to rename the legacy key (well, copy the values over to the new one) and have moved all the reg handling in general into `InstallUtils`.

rr- (approved):

walkawayy (approved):

PR #1412, opened 2024-07-16
lava: add deadly water feature

### LINKED ISSUE #1404: Add deadly water feature
Currently death tiles in water rooms are ignored, but in TR2 onwards they will kill Lara (e.g. Xian lava rooms or Diving Area acid pools). We could in theory just add the feature without making it configurable as there is nowhere like this in the original game, but there may well be existing custom levels where it could cause problems. So possibly a simple gameflow setting to control it would be best.

lahm86:
Resolves #1404.

#### Checklist

#### Description

This allows death tiles in water rooms to be acknowledged, similar to TR2+. There is nowhere in the original levels where this will have an impact, and I did consider adding a gameflow option for custom levels but decided against that. The only scenario I could think of would be an unflipped room with water where the builder would want Lara to be safe, but then the flipped/drained room would have death tiles; altogether, in that case the death tiles need only be defined in the flipped room.

One difference with TR2+ is that we exit early from `Lava_Burn` if Lara is underwater, which avoids flames appearing for a couple of frames as Lara enters the water, which IMO looks odd.

Test level attached for reference.

[LEVEL1.zip](https://github.com/user-attachments/files/16247096/LEVEL1.zip)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ## [Unreleased](https://github.com/LostArtefacts/TR1X/compare/stable...develop) - ××××-××-××
+- added deadly water feature from TR2+ for custom levels (#1404)
 
 ## [4.2](https://github.com/LostArtefacts/TR1X/compare/4.1.2...4.2) - 2024-07-14
 - added creating minidump files on crashes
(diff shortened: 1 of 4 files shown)


rr- (approved):
Looks great IMO.

PR #1414, opened 2024-07-17
output: add skybox support

### LINKED ISSUE #94: Feature request: skybox/horizon support

lahm86:
Resolves #94.

#### Checklist

#### Description

This adds support for skybox rendering similar to TR2. A default injection is provided for Lost Valley and Colossuem; custom level builders can use object slot 184, and in that case the config option is effectively overwritten so the builder has control over the level design.

The refactoring in `output.c` and `s_output.c` are courtesy of @rr- (thanks again) and fix a persistent clipping issue at the top of the screen we found in initial testing. An additional change in `Output_CalcRoomVertices` avoids room geometry disappearing beyond fog end, which can be quite jarring when fog drawing changes as Lara moves - this way the geometry is always visible in front of the sky, but draw distance still honoured.

The additional `.bin` change for Lost Valley fixes some missing textures that we had previously overlooked - these were at the top of the valley, and only really noticeable with the sky in place.


src/game/output.c
```diff
@@ -320,7 +346,6 @@ static const int16_t *Output_CalcRoomVertices(const int16_t *obj_ptr)
             int32_t depth = zv_int >> W2V_SHIFT;
             if (depth > Output_GetDrawDistMax()) {
                 m_VBuf[i].g = 0x1FFF;
-                clip_flags |= 16;
```
rr-, line 323: 1. Can we restore the original code and fence it with a condition `if (!g_Config.enable_skybox)`? This should add clarity to the code, especially when cross-referencing it with TR2+ outside flag.
2. Do we know how this is going to affect the rendering performance – do we now draw the entire level geometry, or do we still clip unseen rooms?

src/game/room_draw.c
```diff
@@ -248,6 +250,24 @@ static void Room_PrepareToDraw(int16_t room_num)
...
+static void Room_DrawSkybox(void)
+{
+    const OBJECT_INFO skybox = g_Objects[O_SKYBOX];
+    if (!skybox.loaded) {
+        return;
+    }
+
+    Output_SetupAboveWater(g_Camera.underwater);
+    Matrix_Push();
+    g_MatrixPtr->_03 = g_MatrixPtr->_13 = g_MatrixPtr->_23 = 0;
```
rr-: Let's avoid chaining assignments in a single statement.

src/game/output.c
lahm86: 1. Done - I've referenced the object loaded state though in case a custom level has the object, but the player disables the setting, as that might come across as a level issue. LMK if this looks OK. My idea was the config option only controls the injection.
2. Performance seems good to me. Yes, `Room_SetBounds` will cull rooms that are too far away anyway, so the above only applies to rooms in the "immediate" vicinity.
https://github.com/lahm86/TR1X/blob/develop/src/game/room_draw.c#L101

rr- (commented):
I'm leaning towards extending the skybox feature to apply to custom levels as well rather than injections alone. This will make the option more versatile and enable players who choose so, to experience true TR1 vibe in custom levels as well. Since the game settings are bound to each game directory, rather than being shared across all games, it shouldn't cause unnecessary surprises either.

Let's hear what others think.
@walkawayy @Richard-L @aredfan

Richard-L:
I never play TRLEs so don't have the most useful opinion I'm afraid. I would only extend it to them if the cost in code hygiene and maintainability is well manageable for you guys. I do like the general sentiment though.

Bigger question for me, perhaps worthy of a discord thread: what skyboxes would be used?

rr-:
IMO the best choices are skyboxes that have a 1996 aesthetic, and preferably coming from the original games. Of all the classic skyboxes available, I believe Lahm has  made the perfect choice with his selection as it looks perfect in both levels. Originating from TR2 Gold, it is also particularly appealing as it can be freely redistributed.

For the reference, here are WADs with all the classic skyboxes available: https://www.trsearch.org/item/5529
The only other option I'd consider is the Meteorite Cavern skybox, with a starry night sky and the Moon.

Richard-L:
Sounds good. I appreciate your desire to stick to the OG classics look.

I think some of them may _just_ benefit from certain Photoshop love perhaps. Not as much as what we did with the loading screens, but in that vein.

lahm86:
Latest commits:
- Three separate skyboxes for the relevant levels
- The gaps in Obelisk into City will now have perma-fog in effect (bundled with texture fixes as it works out quite a nice improvement imo)
- I've made the config option all-powerful, so if players don't want it in custom levels, they can control it. I think it does make sense, as you say, if they want the OG experience

rr- (approved):

walkawayy (approved):
Can't test atm but code looks good.

rr- (approved):

PR #1421, opened 2024-07-24
stats: fix ingame timer not working correctly

### LINKED ISSUE #1420: TR1X bug: In-game clock is twice as fast - 30/60 FPS
Issue reported by Danath on discord.

The clock desync only happens in-game. If the inventory is open then the clock is moving at normal speed instead.

rr-:
#### Checklist

#### Description

The timer would become inaccurate when opening the inventory due to the internal stats timer not being reset on launch. As a result, the previous value was meaningless, leading to incorrect elapsed time calculations on the opening frame.

Also, refactored for simpler usage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,10 @@
 ## [Unreleased](https://github.com/LostArtefacts/TR1X/compare/stable...develop) - ××××-××-××
 - added deadly water feature from TR2+ for custom levels (#1404)
 - added skybox support, with a default option provided for Lost Valley, Colosseum and Obelisk of Khamoon (#94)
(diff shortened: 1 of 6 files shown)


lahm86 (approved):
LGTM. Played Caves and time was spot on.

walkawayy (approved):
LGTM. Just FYI, turning on turbo cheat doesn't double speed the timer. Idk if you want to support that or not since it's a cheat anyway.

rr-:
>Just FYI, turning on turbo cheat doesn't double speed the timer.

Thanks, I haven't considered that. Let's have it not affect the gameplay time – this perspective is just as valid as the alternative, and supporting the other way would make the code messier again. I'll mention it in the changelog.

PR #1422, opened 2024-07-24
door: fix checking for Lara position

### LINKED ISSUE #1419: TR1X bug: Lara can reach trigger through closed UW gate
A regression since [1.1.5](https://github.com/LostArtefacts/TR1X/releases/tag/1.1.5), possibly [1.1.4](https://github.com/LostArtefacts/TR1X/releases/tag/1.1.4) too but I'm unable to check the latter version because it crashes.

The issue is not present in [1.1.3](https://github.com/LostArtefacts/TR1X/releases/tag/1.1.3).

____________________________

This bug manifests in 2 ways:

- The camera doesn't get pushed out after the gate closes.
- Lara can swim against the gate to reach the trigger on the other side which opens the gate.

**TombATI**

https://github.com/user-attachments/assets/8329ab5c-ce02-433a-ba69-8c63a90acdc8

**TR1X**

https://github.com/user-attachments/assets/5771fd33-4306-4689-ac10-6c621ed15425

rr-:
#### Checklist

#### Description

Resolves #1419.

Removes the proximity test and instead checks if Lara is on the same tile where the ghost block will spawn.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - fixed adjacent Midas Touch objects potentially allowing gold bar duplication in custom levels (#1415)
 - fixed the excessive pitch and playback speed correction for music files with sampling rate other than 44100 Hz (#1417, regression from 2.0)
 - fixed the ingame timer being skewed upon inventory open (#1420, regression from 4.1)
+- fixed Lara able to reach triggers through closed doors (#1419, regression since 1.1.4)
 
 ## [4.2](https://github.com/LostArtefacts/TR1X/compare/4.1.2...4.2) - 2024-07-14
 - added creating minidump files on crashes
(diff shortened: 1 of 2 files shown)


walkawayy (approved):
This LGTM from what I tested, but I'm certainly no expert tester.

lahm86 (commented):
This looks good to me. There is another issue though (and it's present in the current release too). If you save and load behind a door (in this non-portal scenario), Lara will void. The problem I believe is that `Door_Initialise` is called on level load and so Lara's position at that point is at the start of the level. The savegame logic will move her to the door's block, but then there are no further checks for door status. Should we tackle this separately?

rr-:
I have resolved the savegame issue mentioned above.
However, I've also observed that the camera still shakes when swimming up against the current in the Midas pool when the gate isn't fully closed. It's because we don't place the invisible block during the door animation that prevents Lara from reaching the trigger. It works as expected after the gate fully shuts, though, which is what this PR corrects (along with the save/load issue). I think this is consistent with the OG; can someone verify this?

lahm86:
Yes, just tried TombATI and you get the same behaviour in that scenario.

lahm86 (approved):

PR #1426, opened 2024-07-30
collide: fix collision issues with bridges, drawbridges, and trapdoors

### LINKED ISSUE #606: Fix glitches related with trapdoors and bridges
**1st issue:**
When trapdoor is placed a little bit above the slope and it has dummy trigger on it Lara slides.
Same goes for the bridge objects.
Would be cool if there was an optional option to fix this issue.

https://user-images.githubusercontent.com/111443302/185347808-ed262f21-fb55-41f3-b7fe-b588a07b5042.mp4

**2nd issue**
When Bridge (tilt = 1) and Bridge (tilt = 2) objects are placed 1 sector or so higher above the slope, they are glitchy and Lara is unable to go upwards along it.
She cannot walk with shift as well.
She can go normally downwards still though.

https://user-images.githubusercontent.com/111443302/185348103-6b2089af-39ae-4d94-a77f-b1aab9c89e82.mp4

**3rd issue:**
When there are two bridge objects that are on the same sector but on different height, if you trigger the lower bridge, the game crashes as soon as you enter to the level.
The higher bridge can still be triggered and it doesn't crash though.

walkawayy:
#### Checklist

#### Description
Fixed collision issues with drawbridges, trapdoors, and bridges when stacked and over slopes.

Attached @lahm86's test level that has a huge number of scenarios to test with.
[Bridges Test Level.zip](https://github.com/user-attachments/files/16420855/Bridges.Test.Level.zip)
EDIT: More extensive test level: [LEVEL1.zip](https://github.com/user-attachments/files/16426768/LEVEL1.zip)

### DIFF
--- a/src/game/collide.c
+++ b/src/game/collide.c
@@ -1,5 +1,6 @@
 #include "game/collide.h"
 
+#include "config.h"
 #include "game/items.h"
 #include "game/room.h"
 #include "global/const.h"
@@ -27,6 +28,7 @@ void Collide_GetCollisionInfo(
 
     FLOOR_INFO *floor = Room_GetFloor(x, ytop, z, &room_num);
     int32_t height = Room_GetHeight(floor, x, ytop, z);
+    int32_t room_height = height;
     if (height != NO_HEIGHT) {
         height -= ypos;
     }
@@ -41,9 +43,15 @@ void Collide_GetCollisionInfo(
     coll->mid_type = g_HeightType;
     coll->trigger = g_TriggerIndex;
 
-    int16_t tilt = Room_GetTiltType(floor, x, g_LaraItem->pos.y, z);
-    coll->tilt_z = tilt >> 8;
-    coll->tilt_x = (int8_t)tilt;
+    if (!g_Config.fix_bridge_collision
+        || !Room_IsOnWalkable(floor, x, ytop, z, room_height)) {
+        int16_t tilt = Room_GetTiltType(floor, x, g_LaraItem->pos.y, z);
+        coll->tilt_z = tilt >> 8;
+        coll->tilt_x = (int8_t)tilt;
+    } else {
+        coll->tilt_z = 0;
+        coll->tilt_x = 0;
+    }
 
     int32_t xleft;
     int32_t zleft;
@@ -98,10 +106,12 @@ void Collide_GetCollisionInfo(
         break;
     }
 
+    // Front.
     x = xpos + xfront;
     z = zpos + zfront;
     floor = Room_GetFloor(x, ytop, z, &room_num);
     height = Room_GetHeight(floor, x, ytop, z);
+    room_height = height;
     if (height != NO_HEIGHT) {
         height -= ypos;
     }
@@ -114,23 +124,29 @@ void Collide_GetCollisionInfo(
     coll->front_floor = height;
     coll->front_ceiling = ceiling;
     coll->front_type = g_HeightType;
-    if (coll->slopes_are_walls && coll->front_type == HT_BIG_SLOPE
-        && coll->front_floor < 0) {
-        coll->front_floor = -32767;
-    } else if (
-        coll->slopes_are_pits && coll->front_type == HT_BIG_SLOPE
-        && coll->front_floor > 0) {
-        coll->front_floor = 512;
-    } else if (
-        coll->lava_is_pit && coll->front_floor > 0 && g_TriggerIndex
-        && (g_TriggerIndex[0] & DATA_TYPE) == FT_LAVA) {
-        coll->front_floor = 512;
+
+    if (!g_Config.fix_bridge_collision
+        || !Room_IsOnWalkable(floor, x, ytop, z, room_height)) {
+        if (coll->slopes_are_walls && coll->front_type == HT_BIG_SLOPE
+            && coll->front_floor < 0) {
+            coll->front_floor = -32767;
+        } else if (
+            coll->slopes_are_pits && coll->front_type == HT_BIG_SLOPE
+            && coll->front_floor > 0) {
+            coll->front_floor = 512;
+        } else if (
+            coll->lava_is_pit && coll->front_floor > 0 && g_TriggerIndex
+            && (g_TriggerIndex[0] & DATA_TYPE) == FT_LAVA) {
+            coll->front_floor = 512;
+        }
     }
 
+    // Left.
     x = xpos + xleft;
     z = zpos + zleft;
     floor = Room_GetFloor(x, ytop, z, &room_num);
     height = Room_GetHeight(floor, x, ytop, z);
+    room_height = height;
     if (height != NO_HEIGHT) {
         height -= ypos;
     }
@@ -143,23 +159,29 @@ void Collide_GetCollisionInfo(
     coll->left_floor = height;
     coll->left_ceiling = ceiling;
     coll->left_type = g_HeightType;
-    if (coll->slopes_are_walls && coll->left_type == HT_BIG_SLOPE
-        && coll->left_floor < 0) {
-        coll->left_floor = -32767;
-    } else if (
-        coll->slopes_are_pits && coll->left_type == HT_BIG_SLOPE
-        && coll->left_floor > 0) {
-        coll->left_floor = 512;
-    } else if (
-        coll->lava_is_pit && coll->left_floor > 0 && g_TriggerIndex
-        && (g_TriggerIndex[0] & DATA_TYPE) == FT_LAVA) {
-        coll->left_floor = 512;
+
+    if (!g_Config.fix_bridge_collision
+        || !Room_IsOnWalkable(floor, x, ytop, z, room_height)) {
+        if (coll->slopes_are_walls && coll->left_type == HT_BIG_SLOPE
+            && coll->left_floor < 0) {
+            coll->left_floor = -32767;
+        } else if (
+            coll->slopes_are_pits && coll->left_type == HT_BIG_SLOPE
+            && coll->left_floor > 0) {
+            coll->left_floor = 512;
+        } else if (
+            coll->lava_is_pit && coll->left_floor > 0 && g_TriggerIndex
+            && (g_TriggerIndex[0] & DATA_TYPE) == FT_LAVA) {
+            coll->left_floor = 512;
+        }
     }
 
+    // Right.
     x = xpos + xright;
     z = zpos + zright;
     floor = Room_GetFloor(x, ytop, z, &room_num);
     height = Room_GetHeight(floor, x, ytop, z);
+    room_height = height;
     if (height != NO_HEIGHT) {
         height -= ypos;
     }
@@ -172,17 +194,21 @@ void Collide_GetCollisionInfo(
     coll->right_floor = height;
     coll->right_ceiling = ceiling;
     coll->right_type = g_HeightType;
-    if (coll->slopes_are_walls && coll->right_type == HT_BIG_SLOPE
-        && coll->right_floor < 0) {
-        coll->right_floor = -32767;
-    } else if (
-        coll->slopes_are_pits && coll->right_type == HT_BIG_SLOPE
-        && coll->right_floor > 0) {
-        coll->right_floor = 512;
-    } else if (
-        coll->lava_is_pit && coll->right_floor > 0 && g_TriggerIndex
-        && (g_TriggerIndex[0] & DATA_TYPE) == FT_LAVA) {
-        coll->right_floor = 512;
+
+    if (!g_Config.fix_bridge_collision
+        || !Room_IsOnWalkable(floor, x, ytop, z, room_height)) {
+        if (coll->slopes_are_walls && coll->right_type == HT_BIG_SLOPE
+            && coll->right_floor < 0) {
+            coll->right_floor = -32767;
+        } else if (
+            coll->slopes_are_pits && coll->right_type == HT_BIG_SLOPE
+            && coll->right_floor > 0) {
+            coll->right_floor = 512;
+        } else if (
+            coll->lava_is_pit && coll->right_floor > 0 && g_TriggerIndex
+            && (g_TriggerIndex[0] & DATA_TYPE) == FT_LAVA) {
+            coll->right_floor = 512;
+        }
     }
 
     if (Collide_CollideStaticObjects(
--- a/src/game/objects/general/bridge.c
+++ b/src/game/objects/general/bridge.c
@@ -11,15 +11,13 @@
 
 #include <stdbool.h>
 
-static bool Bridge_IsSameSector(
-    int32_t x, int32_t y, int32_t z, const ITEM_INFO *item);
-static bool Bridge_OnDrawBridge(ITEM_INFO *item, int32_t x, int32_t y);
+static bool Bridge_IsSameSector(int32_t x, int32_t z, const ITEM_INFO *item);
+static bool Bridge_OnDrawBridge(ITEM_INFO *item, int32_t x, int32_t z);
 static int32_t Bridge_GetOffset(
     ITEM_INFO *item, int32_t x, int32_t y, int32_t z);
 static void Bridge_FixEmbeddedPosition(int16_t item_num);
 
-static bool Bridge_IsSameSector(
-    int32_t x, int32_t y, int32_t z, const ITEM_INFO *item)
+static bool Bridge_IsSameSector(int32_t x, int32_t z, const ITEM_INFO *item)
 {
     int32_t sector_x = x / WALL_L;
     int32_t sector_z = z / WALL_L;
@@ -29,24 +27,23 @@ static bool Bridge_IsSameSector(
     return sector_x == item_sector_x && sector_z == item_sector_z;
 }
 
-static bool Bridge_OnDrawBridge(ITEM_INFO *item, int32_t x, int32_t y)
+static bool Bridge_OnDrawBridge(ITEM_INFO *item, int32_t x, int32_t z)
 {
-    int32_t ix = item->pos.z >> WALL_SHIFT;
-    int32_t iy = item->pos.x >> WALL_SHIFT;
-
+    int32_t ix = item->pos.x >> WALL_SHIFT;
+    int32_t iz = item->pos.z >> WALL_SHIFT;
     x >>= WALL_SHIFT;
-    y >>= WALL_SHIFT;
+    z >>= WALL_SHIFT;
 
-    if (item->rot.y == 0 && y == iy && (x == ix - 1 || x == ix - 2)) {
+    if (item->rot.y == 0 && x == ix && (z == iz - 1 || z == iz - 2)) {
         return true;
-    }
-    if (item->rot.y == -PHD_180 && y == iy && (x == ix + 1 || x == ix + 2)) {
+    } else if (
+        item->rot.y == -PHD_180 && x == ix && (z == iz + 1 || z == iz + 2)) {
         return true;
-    }
-    if (item->rot.y == PHD_90 && x == ix && (y == iy - 1 || y == iy - 2)) {
+    } else if (
+        item->rot.y == PHD_90 && z == iz && (x == ix - 1 || x == ix - 2)) {
         return true;
-    }
-    if (item->rot.y == -PHD_90 && x == ix && (y == iy + 1 || y == iy + 2)) {
+    } else if (
+        item->rot.y == -PHD_90 && z == iz && (x == ix + 1 || x == ix + 2)) {
         return true;
     }
 
@@ -148,7 +145,8 @@ void Bridge_DrawBridgeFloor(
     if (item->current_anim_state != DOOR_OPEN) {
         return;
     }
-    if (!Bridge_OnDrawBridge(item, z, x)) {
+
+    if (!Bridge_OnDrawBridge(item, x, z)) {
         return;
     }
 
@@ -163,7 +161,8 @@ void Bridge_DrawBridgeCeiling(
     if (item->current_anim_state != DOOR_OPEN) {
         return;
     }
-    if (!Bridge_OnDrawBridge(item, z, x)) {
+
+    if (!Bridge_OnDrawBridge(item, x, z)) {
         return;
     }
 
--- a/src/game/objects/general/bridge.c
+++ b/src/game/objects/general/bridge.c
@@ -14,7 +14,7 @@
 static bool Bridge_IsSameSector(int32_t x, int32_t z, const ITEM_INFO *item);
 static bool Bridge_OnDrawBridge(ITEM_INFO *item, int32_t x, int32_t z);
 static int32_t Bridge_GetOffset(
-    ITEM_INFO *item, int32_t x, int32_t y, int32_t z);
+    const ITEM_INFO *item, int32_t x, int32_t y, int32_t z);
 static void Bridge_FixEmbeddedPosition(int16_t item_num);
 
 static bool Bridge_IsSameSector(int32_t x, int32_t z, const ITEM_INFO *item)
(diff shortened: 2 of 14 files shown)


walkawayy:
I left some TODO comments in the PR for some questions I had. I also will clean up some of the files where functions changed from static, but I didn't want to blow up the diff for the PR. I also am expecting some discussion on the code changes, so I figure the usage might change depending on the code review anyway.

rr- (commented):
Any chance we can make these improvements a part of the `fix_bridge_collision` setting?

src/game/collide.c
```diff
@@ -40,10 +43,27 @@ void Collide_GetCollisionInfo(
...
     coll->trigger = g_TriggerIndex;
-
-    int16_t tilt = Room_GetTiltType(floor, x, g_LaraItem->pos.y, z);
-    coll->tilt_z = tilt >> 8;
-    coll->tilt_x = (int8_t)tilt;
+    bool skip_slope_check = false;
+
+    // TODO Do we need to check all walkable heights or is knowing Lara is above
+    // one bridge enough? Can a slope ever be below one bridge but above
+    // another?
```
rr-: I don't think this is possible unless a level builder embeds an object in the room geometry, which serves no purpose. Or they embed the bridge in a way so that it protrudes from a slope and is to use mixed room/object collision, but it's unreasonable to expect that to work IMO.
So all in all I think it's good to remove this TODO.

src/game/collide.c
```diff
@@ -40,10 +43,27 @@ void Collide_GetCollisionInfo(
...
+    VECTOR *walkable_heights = Room_GetWalkableHeights(floor, x, ypos, z);
+    for (int32_t i = 0; i < walkable_heights->count; i++) {
+        int32_t item_height = (int32_t)walkable_heights->items[i];
+        // TODO Need check within WALL_L range? Can walkables ever be below a
+        // slope?
```
rr-: It can't unless of course we talk about sloped ceilings.

src/game/collide.c
```diff
@@ -40,10 +43,27 @@ void Collide_GetCollisionInfo(
...
+        if (ABS(item_height - ypos) < WALL_L
+            && ABS(item_height - ypos) <= STEPUP_HEIGHT) {
+            skip_slope_check = true;
```
rr-: We can `break;` here as the only thing we do in this loop is set this bool.

src/game/collide.c
```diff
@@ -40,10 +43,27 @@ void Collide_GetCollisionInfo(
...
+        }
+    }
+    if (!skip_slope_check) {
+        int16_t tilt = Room_GetTiltType(floor, x, g_LaraItem->pos.y, z);
+        coll->tilt_z = tilt >> 8;
+        coll->tilt_x = (int8_t)tilt;
+    }
+    Vector_Clear(walkable_heights);
```
rr-: This should've been `Vector_Free`, since we'll later override the content of this variable with a separately allocated vector.

src/game/collide.c
```diff
@@ -114,19 +135,32 @@ void Collide_GetCollisionInfo(
...
-        && coll->front_floor > 0) {
-        coll->front_floor = 512;
-    } else if (
-        coll->lava_is_pit && coll->front_floor > 0 && g_TriggerIndex
-        && (g_TriggerIndex[0] & DATA_TYPE) == FT_LAVA) {
-        coll->front_floor = 512;
+    skip_slope_check = false;
+    walkable_heights = Room_GetWalkableHeights(floor, x, ypos, z);
+    for (int32_t i = 0; i < walkable_heights->count; i++) {
+        int32_t item_height = (int32_t)walkable_heights->items[i];
```
rr-: `const int32_t` (but only in the variable type, not in the cast, as making primitive casts const serves no purpose)

src/game/collide.c
```diff
@@ -114,19 +135,32 @@ void Collide_GetCollisionInfo(
...
+        if (ABS(item_height - ypos) <= STEPUP_HEIGHT) {
+            skip_slope_check = true;
```
rr-: `break;`

src/game/collide.c
```diff
@@ -114,19 +135,32 @@ void Collide_GetCollisionInfo(
...
+            coll->slopes_are_pits && coll->front_type == HT_BIG_SLOPE
+            && coll->front_floor > 0) {
+            coll->front_floor = 512;
+        } else if (
+            coll->lava_is_pit && coll->front_floor > 0 && g_TriggerIndex
+            && (g_TriggerIndex[0] & DATA_TYPE) == FT_LAVA) {
+            coll->front_floor = 512;
+        }
     }
+    Vector_Clear(walkable_heights);
```
rr-: `Vector_Free`

src/game/collide.c
```diff
@@ -143,19 +177,32 @@ void Collide_GetCollisionInfo(
...
-        && coll->left_floor > 0) {
-        coll->left_floor = 512;
-    } else if (
-        coll->lava_is_pit && coll->left_floor > 0 && g_TriggerIndex
-        && (g_TriggerIndex[0] & DATA_TYPE) == FT_LAVA) {
-        coll->left_floor = 512;
+    skip_slope_check = false;
+    walkable_heights = Room_GetWalkableHeights(floor, x, ypos, z);
+    for (int32_t i = 0; i < walkable_heights->count; i++) {
+        int32_t item_height = (int32_t)walkable_heights->items[i];
```
rr-: `const int32_t`

src/game/collide.c
```diff
@@ -143,19 +177,32 @@ void Collide_GetCollisionInfo(
...
+            coll->slopes_are_pits && coll->left_type == HT_BIG_SLOPE
+            && coll->left_floor > 0) {
+            coll->left_floor = 512;
+        } else if (
+            coll->lava_is_pit && coll->left_floor > 0 && g_TriggerIndex
+            && (g_TriggerIndex[0] & DATA_TYPE) == FT_LAVA) {
+            coll->left_floor = 512;
+        }
     }
+    Vector_Clear(walkable_heights);
```
rr-: `Vector_Free`

src/game/collide.c
```diff
@@ -172,18 +219,30 @@ void Collide_GetCollisionInfo(
...
-        && coll->right_floor > 0) {
-        coll->right_floor = 512;
-    } else if (
-        coll->lava_is_pit && coll->right_floor > 0 && g_TriggerIndex
-        && (g_TriggerIndex[0] & DATA_TYPE) == FT_LAVA) {
-        coll->right_floor = 512;
+    skip_slope_check = false;
+    walkable_heights = Room_GetWalkableHeights(floor, x, ypos, z);
+    for (int32_t i = 0; i < walkable_heights->count; i++) {
+        int32_t item_height = (int32_t)walkable_heights->items[i];
```
rr-: `const int32_t`

src/game/collide.c
```diff
@@ -172,18 +219,30 @@ void Collide_GetCollisionInfo(
...
+        if (ABS(item_height - ypos) <= STEPUP_HEIGHT) {
+            skip_slope_check = true;
```
rr-: `break;`

src/game/objects/general/bridge.c
```diff
@@ -163,13 +177,16 @@ void Bridge_DrawBridgeCeiling(
...
         return;
     }
 
-    if (y > item->pos.y) {
-        *height = item->pos.y + STEP_L;
+    if (y <= item->pos.y || item->pos.y <= *height) {
+        return;
     }
+
+    *height = (int16_t)item->pos.y + STEP_L;
```
rr-: I don't think this cast is necessary as the variable type is already `int16_t`

src/game/objects/general/bridge.c
```diff
@@ -197,67 +215,73 @@ void Bridge_FlatFloor(
...
```
rr-: Unnecessary cast

src/game/objects/general/bridge.c
```diff
@@ -197,67 +215,73 @@ void Bridge_FlatFloor(
...
     ITEM_INFO *item, int32_t x, int32_t y, int32_t z, int16_t *height)
 {
-    if (g_Config.fix_bridge_collision && !Bridge_IsSameSector(x, y, z, item)) {
+    if (g_Config.fix_bridge_collision && !Bridge_IsSameSector(x, z, item)) {
         return;
     }
 
-    int32_t level = item->pos.y + (Bridge_GetOffset(item, x, y, z) >> 2);
-    if (y > level) {
+    int32_t offset_height = item->pos.y + (Bridge_GetOffset(item, x, y, z));
```
rr-: `const int32_t`

src/game/objects/general/bridge.c
```diff
@@ -197,67 +215,73 @@ void Bridge_FlatFloor(
...
 {
-    if (g_Config.fix_bridge_collision && !Bridge_IsSameSector(x, y, z, item)) {
+    if (g_Config.fix_bridge_collision && !Bridge_IsSameSector(x, z, item)) {
         return;
     }
 
-    int32_t level = item->pos.y + (Bridge_GetOffset(item, x, y, z) >> 2);
-    if (y > level) {
-        *height = level + STEP_L;
+    int32_t offset_height = item->pos.y + (Bridge_GetOffset(item, x, y, z));
```
rr-: `const int32_t`

src/game/objects/general/bridge.c
```diff
@@ -197,67 +215,73 @@ void Bridge_FlatFloor(
...
+    if (y <= offset_height || item->pos.y <= *height) {
+        return;
     }
+
+    *height = (int16_t)offset_height + STEP_L;
```
rr-: Unnecessary cast

src/game/objects/general/bridge.c
```diff
@@ -197,67 +215,73 @@ void Bridge_FlatFloor(
...
     ITEM_INFO *item, int32_t x, int32_t y, int32_t z, int16_t *height)
 {
-    if (g_Config.fix_bridge_collision && !Bridge_IsSameSector(x, y, z, item)) {
+    if (g_Config.fix_bridge_collision && !Bridge_IsSameSector(x, z, item)) {
         return;
     }
 
-    int32_t level = item->pos.y + (Bridge_GetOffset(item, x, y, z) >> 1);
-    if (y > level) {
+    int32_t offset_height = item->pos.y + (Bridge_GetOffset(item, x, y, z));
```
rr-: `const int32_t`

src/game/objects/general/bridge.c
```diff
@@ -197,67 +215,73 @@ void Bridge_FlatFloor(
...
 {
-    if (g_Config.fix_bridge_collision && !Bridge_IsSameSector(x, y, z, item)) {
+    if (g_Config.fix_bridge_collision && !Bridge_IsSameSector(x, z, item)) {
         return;
     }
 
-    int32_t level = item->pos.y + (Bridge_GetOffset(item, x, y, z) >> 1);
-    if (y > level) {
-        *height = level + STEP_L;
+    int32_t offset_height = item->pos.y + (Bridge_GetOffset(item, x, y, z));
```
rr-: `const int32_t`

src/game/objects/general/bridge.c
```diff
@@ -197,67 +215,73 @@ void Bridge_FlatFloor(
...
+    if (y <= offset_height || item->pos.y <= *height) {
+        return;
     }
+
+    *height = (int16_t)offset_height + STEP_L;
```
rr-: Unnecessary cast

src/game/objects/general/bridge.h
```diff
@@ -4,6 +4,8 @@
 
 #include <stdint.h>
 
+int32_t Bridge_GetOffset(ITEM_INFO *item, int32_t x, int32_t y, int32_t z);
+bool Bridge_OnDrawBridge(ITEM_INFO *item, int32_t x, int32_t z);
```
rr-: I don't think we ever use `Bridge_OnDrawBridge` publicly, so this can be made `static` back.

src/game/objects/general/trapdoor.h
```diff
@@ -4,6 +4,7 @@
 
 #include <stdint.h>
 
+bool TrapDoor_StandingOn(ITEM_INFO *item, int32_t x, int32_t z);
```
rr-: I don't think we ever use `TrapDoor_StandingOn` publicly, so it can be made `static` back.

src/game/objects/general/trapdoor.c
```diff
@@ -65,6 +67,8 @@ void TrapDoor_Ceiling(
     if (!TrapDoor_StandingOn(item, x, z)) {
         return;
     }
+
+    // TODO Invert condition?
```
rr-: Yes, to be consistent with the bridge code change.

src/game/objects/general/trapdoor.c
```diff
@@ -53,9 +53,11 @@ void TrapDoor_Floor(
...
```
rr-: Yes, to be consistent with the bridge code change.

src/game/objects/general/trapdoor.c
```diff
@@ -53,9 +53,11 @@ void TrapDoor_Floor(
...
     if (y <= item->pos.y && item->current_anim_state == DOOR_CLOSED
         && item->pos.y < *height) {
-        *height = item->pos.y;
+        *height = (int16_t)item->pos.y;
```
rr-: Unnecessary cast

src/game/room.c
```diff
@@ -977,3 +981,66 @@ void Room_TestTriggers(int16_t *data, bool heavy)
         }
     }
 }
+
+VECTOR *Room_GetWalkableHeights(
+    FLOOR_INFO *floor, int32_t x, int32_t y, int32_t z)
+{
+    ROOM_INFO *r;
```
rr-: Double declaration of `r`

src/game/room.c
```diff
@@ -977,3 +981,66 @@ void Room_TestTriggers(int16_t *data, bool heavy)
...
+    while (floor->pit_room != NO_ROOM) {
+        ROOM_INFO *r = &g_RoomInfo[floor->pit_room];
```
rr-: `const ROOM_INFO *const r`

src/game/room.c
```diff
@@ -977,3 +981,66 @@ void Room_TestTriggers(int16_t *data, bool heavy)
...
+        int32_t x_floor = (z - r->z) >> WALL_SHIFT;
+        int32_t z_floor = (x - r->x) >> WALL_SHIFT;
```
rr-: `const int32_t` for both values

src/game/room.c
```diff
@@ -977,3 +981,66 @@ void Room_TestTriggers(int16_t *data, bool heavy)
...
+            trig_flags = *floor_data;
+            floor_data++;
+            trig_type = (type >> 8) & 0x3F;
+            do {
+                trigger = *floor_data++;
+                if (TRIG_BITS(trigger) != TO_OBJECT) {
+                    if (TRIG_BITS(trigger) == TO_CAMERA) {
+                        trigger = *floor_data++;
+                    }
+                } else {
```
rr-: This can be rearranged to

````c
if (TRIG_BITS(trigger == TO_OBJECT) {
    int16_t item_num = ...
    ...
} else if (TRIG_BITS(trigger) == TO_CAMERA) {
    trigger = *floor_data++;
}
```

src/game/room.c
```diff
@@ -977,3 +981,66 @@ void Room_TestTriggers(int16_t *data, bool heavy)
...
+                    int16_t item_num = trigger & VALUE_BITS;
+                    ITEM_INFO *item = &g_Items[item_num];
+                    if (Object_IsObjectType(
+                            item->object_number, g_TrapdoorObjects)) {
+                        int32_t item_height = item->pos.y;
```
rr-: `const int32_t`

src/game/room.c
```diff
@@ -977,3 +981,66 @@ void Room_TestTriggers(int16_t *data, bool heavy)
...
+            trig_type = (type >> 8) & 0x3F;
+            do {
+                trigger = *floor_data++;
+                if (TRIG_BITS(trigger) != TO_OBJECT) {
+                    if (TRIG_BITS(trigger) == TO_CAMERA) {
+                        trigger = *floor_data++;
+                    }
+                } else {
+                    int16_t item_num = trigger & VALUE_BITS;
+                    ITEM_INFO *item = &g_Items[item_num];
```
rr-, line 1024: `const ITEM_INFO *const`

src/game/room.c
```diff
@@ -977,3 +981,66 @@ void Room_TestTriggers(int16_t *data, bool heavy)
...
+            floor_data++;
+            trig_type = (type >> 8) & 0x3F;
+            do {
+                trigger = *floor_data++;
+                if (TRIG_BITS(trigger) != TO_OBJECT) {
+                    if (TRIG_BITS(trigger) == TO_CAMERA) {
+                        trigger = *floor_data++;
+                    }
+                } else {
+                    int16_t item_num = trigger & VALUE_BITS;
```
rr-: `const int16_t`

src/game/room.c
```diff
@@ -977,3 +981,66 @@ void Room_TestTriggers(int16_t *data, bool heavy)
...
+                    ITEM_INFO *item = &g_Items[item_num];
+                    if (Object_IsObjectType(
+                            item->object_number, g_TrapdoorObjects)) {
+                        int32_t item_height = item->pos.y;
+                        Vector_Add(heights, (void *)item->pos.y);
+                    } else if (Object_IsObjectType(
+                                   item->object_number, g_BridgeObjects)) {
+                        int32_t item_height =
```
rr-: `const int32_t item_height`

src/game/room.c
```diff
@@ -977,3 +981,66 @@ void Room_TestTriggers(int16_t *data, bool heavy)
         }
     }
 }
+
+VECTOR *Room_GetWalkableHeights(
+    FLOOR_INFO *floor, int32_t x, int32_t y, int32_t z)
```
rr-: `const FLOOR_INFO *const floor, const int32_t x, const int32_t y, const int32_t z`

src/game/room.h
```diff
@@ -30,3 +32,5 @@ void Room_AlterFloorHeight(ITEM_INFO *item, int32_t height);
 
 void Room_TestTriggers(int16_t *data, bool heavy);
 void Room_FlipMap(void);
+VECTOR *Room_GetWalkableHeights(
+    FLOOR_INFO *floor, int32_t x, int32_t y, int32_t z);
```
rr-:
```c
VECTOR *Room_GetWalkableHeights(
    const FLOOR_INFO *floor, int32_t x, int32_t y, int32_t z);
```

src/game/room.c
```diff
@@ -977,3 +981,66 @@ void Room_TestTriggers(int16_t *data, bool heavy)
...
+                        Vector_Add(heights, (void *)item_height);
+                    }
+                }
+            } while (!(trigger & END_BIT));
+            break;
+        }
+    } while (!(type & END_BIT));
+
+    return heights;
+}
```
rr-, line 1042: Rather than consulting bridge code directly from room which is a big no-no, we should introduce a new function pointer in OBJECT_INFO, for example, `get_walkable_height`, that we'd set to `Bridge_GetOffset` in relevant `…_Setup` functions for the `g_BridgeObjects` members, and call it here via `object->get_walkable_height`. This way we'd probably no longed need `g_BridgeObjects`.

rr- (commented):
Any chance we can make these improvements a part of the `fix_bridge_collision` setting?

---

When reviewing `Room_GetWalkableHeights` I noticed that we seem to have already something in the code that sounds like it serves the same purpose – namely, `OBJECT_INFO`'s `void (*floor)(ITEM_INFO *item, int32_t x, int32_t y, int32_t z, int16_t *height);` used by `Room_GetHeight`. The slope checks also call `Room_GetHeight`. So, is there any chance we could bake the introduced checks into `Room_GetHeight`?

src/game/collide.c
```diff
@@ -40,10 +43,27 @@ void Collide_GetCollisionInfo(
...
+        if (ABS(item_height - ypos) < WALL_L
+            && ABS(item_height - ypos) <= STEPUP_HEIGHT) {
+            skip_slope_check = true;
+        }
+    }
+    if (!skip_slope_check) {
+        int16_t tilt = Room_GetTiltType(floor, x, g_LaraItem->pos.y, z);
+        coll->tilt_z = tilt >> 8;
+        coll->tilt_x = (int8_t)tilt;
+    }
```
rr-, line 54: I think we should reset the tilt values back to zero in case we deem Lara is not a slope. Without that, I think it might be the case that when Lara slides into a bridge, she'll continue sliding.

rr- (commented):
One last question: what will happen if we only use the first `skip_floor_check` and let the others (front/left/right) stay as they were? I think these other checks are mostly about preventing Lara from traversing up or down the slopes depending on her animations.

Also upon reading the code it'd be good to have another test in @lahm86 room:
- A slope and a slightly tilted floor that directly continues onto a bridge/trap door
- Same as above but with a slight positive or negative height difference
- Various geometries for checking if Lara can or cannot fit underneath a bridge or a trapdoor

src/game/collide.c
lahm86: Test updates attached.
[LEVEL1.zip](https://github.com/user-attachments/files/16426768/LEVEL1.zip)

walkawayy:
I will address the smaller comments as I incorporate changes and test the code. I am going to introduce a separate commit for the trapdoor and drawbridge cleanups / condition inversions. I was just in that code a lot as I was investigating this issue, so that's why I even bothered with them.

I agree linking these fixes to `fix_bridge_collision` will be nice. I just need to think of a clean way to do this since there's a bunch of difference code affected.

As far as incorporating `Room_GetWalkableHeights`into `Room_GetHeight`, that's possibly a good idea. I just was hesitant to make any big changes in fear of changing the game collision behavior. Especially as we will need to account for future TRs which have slight changes as well. I guess if we can make it configurable with the option, maybe that can be more cleanly done. TBD on if it will work.

src/game/objects/general/bridge.c
walkawayy: `item->pos.x/y/z` are `int32_t` I believe. Is the cast unneeded? There are some mismatched types around the codebase similar to this it seems. I just changed this since I was already in the relevant code.

src/game/room.c
walkawayy: Ok. I reused code from the GetFloor and GetHeight functions here which do it like this.

I had a wish to have a common floor traversal function since they are all slightly different and repeated, but each time it happens there are different things being done while traversing. Stats sets killables, secrets, and pickups, injection does its own thing, etc.

src/game/room.c
walkawayy: I will double check, but the ceiling and floor fixes for bridges I made last. I wonder if those are fixed, if we can simplify this part of the code somehow.

src/game/objects/general/bridge.c
walkawayy: `height` is `int16_t` while `offset_height` is `int32_t`. Should this have a cast?

src/game/collide.c
walkawayy: This actually fixed a small bug in Lahm's newest test level where Lara would slide for a few frames when walking over tilted bridges when slightly above a slope. Nice catch.

walkawayy:
`what will happen if we only use the first skip_floor_check and let the others (front/left/right) stay as they were? I think these other checks are mostly about preventing Lara from traversing up or down the slopes depending on her animation`

This change completely breaks Lara's movement. Lara will be pushed away or invisible walled from any bridge over a tilt as her collision views the movement as blocked.

walkawayy:
Ok so now that the bridges' floor functions are fixed, I believe I can simplify the skip slope check by caching the room height like this:
```
    int32_t height = Room_GetHeight(floor, x, ytop, z);
    const int32_t room_height = height;
```

And then checking:
```
        if (room_height == ypos) {
            skip_slope_check = true;
            break;
        }
```

However, I still need to add an `||` to that condition to also check if the floor is equal to a walkable height. I need to know if the positions are equal because Lara is on a walkable item or just equal because Lara is standing on a normal floor. So, I still think I need to traverse the walkable items on the floor tile and get their heights. The only other solution I see is if we set a global variable in `Room_GetHeight` whenever a `TO_OBJECT` is reached and its floor function runs:
```
if (object->floor) {
    object->floor(item, x, y, z, &height);
}
```

I'm not sure if we want to do that, but I will propose it. Maybe there's a cleaner way to use that information somehow. Also, I did just notice `g_TriggerIndex` is a variable in there that could at least simplify `Room_GetWalkableHeights` since we could just check the trigger data instead of having to traverse the floor again. It's a small optimization. I believe that's what it contains anyway.

src/game/objects/general/bridge.c
rr-: Most often yes, type promotion is implicit.

rr-:
I'm seeing most of the comments are resolved, but there are no corresponding changes – can I ask to push what we have so far for easier bookkeeping?

walkawayy:
The diff is best viewed commit by commit. The first few are just minor cleanups. I tested Lahm's level with the bridge fix config option enabled and disabled. With it disabled, OG behavior seemed preserved (and broken). I think this solves just about all remaining issues we saw. There are only a few small outliers where Lara can land specific jumps into a tight space and sometimes bug on top of a bridge similar to the corner bug if a bridge is placed extremely low over a high slope. But I don't think that is a realistic level design and Lara can jump normally on top of that bridge anyway. I think it's also just the original engine collision. I didn't want to mess with the game fundamentals too much with this. LMK what you think.

rr- (approved):

src/game/room.c
```diff
@@ -13,7 +15,9 @@
 #include "global/const.h"
 #include "global/vars.h"
 
+#include <libtrx/log.h>
 #include <libtrx/utils.h>
+#include <libtrx/vector.h>
```
lahm86, line 20: I think this is redundant now?

lahm86: I think this is redundant now? And the log include too.

src/game/room.h
```diff
@@ -3,6 +3,8 @@
 #include "global/const.h"
 #include "global/types.h"
 
+#include <libtrx/vector.h>
```
lahm86: And here.

tools/config/TR1X_ConfigTool/Resources/Lang/en.json
```diff
@@ -155,7 +155,7 @@
     },
     "fix_bridge_collision": {
       "Title": "Fix bridge collision",
-      "Description": "Fixes Lara not being able to grab parts of some bridges and invisible walls at the edge."
+      "Description": "Fixes Lara not being able to grab parts of some bridges and invisible walls at the edge. Also fixes issues collision issues with drawbridges, trapdoors, and bridges when stacked over each other, over slopes, and near the ground."
```
lahm86: `...fixes issues collision issues...`

lahm86 (commented):
Just some minor comments, but otherwise everything looks 👍

lahm86 (approved):

rr- (approved):
The end result has become very elegant IMO.

rr- (changes requested):
Hold on – please remove the subprojects/libtrx change.

walkawayy:
@rr- I think I fixed the submodule issue. If it looks good, feel free to merge.

rr- (approved):

PR #1428, opened 2024-08-04
room: rename floor info

lahm86:
Part of #941.

#### Checklist

#### Description

This renames `FLOOR_INFO` to `SECTOR_INFO` to help distinguish it in cases where floor is used in reference to height. The naming is more in line with Tomb Editor, the TRosettaStone documentation and so forth.

Related variables and functions have been renamed too, and I've tried to add in `const` where applicable; `inject.c` is the exclusion to that rule as I want to tidy this up further as we continue refactoring floor data interpretation. Hopefully we can squash down the FD injection functions later.


src/game/lara/lara_misc.c
```diff
@@ -547,17 +547,17 @@ bool Lara_LandedBad(ITEM_INFO *item, COLL_INFO *coll)
 {
     int16_t room_num = item->room_number;
 
-    FLOOR_INFO *floor =
-        Room_GetFloor(item->pos.x, item->pos.y, item->pos.z, &room_num);
+    const SECTOR_INFO *const sector =
+        Room_GetSector(item->pos.x, item->pos.y, item->pos.z, &room_num);
 
-    int oy = item->pos.y;
```
rr-, line 553: oi!

src/game/objects/creatures/bat.c
```diff
@@ -44,9 +44,9 @@ static void Bat_FixEmbeddedPosition(int16_t item_num)
     const int32_t y = item->pos.y;
     const int32_t z = item->pos.z;
     int16_t room_number = item->room_number;
-    FLOOR_INFO *floor = Room_GetFloor(x, y, z, &room_number);
-    Room_GetHeight(floor, x, y, z);
-    const int16_t ceiling = Room_GetCeiling(floor, x, y, z);
+    const SECTOR_INFO *const sector = Room_GetSector(x, y, z, &room_number);
+    Room_GetHeight(sector, x, y, z);
```
rr-: No need to call this as it has no side effects... Right? or is this the function that sets the g_TriggerIndex?

src/game/objects/traps/thors_hammer.c
```diff
@@ -114,8 +114,9 @@ void ThorsHandle_Control(int16_t item_num)
         int32_t old_z = z;
 
         int16_t room_num = item->room_number;
-        FLOOR_INFO *floor = Room_GetFloor(x, item->pos.y, z, &room_num);
-        Room_GetHeight(floor, x, item->pos.y, z);
+        const SECTOR_INFO *const sector =
+            Room_GetSector(x, item->pos.y, z, &room_num);
+        Room_GetHeight(sector, x, item->pos.y, z);
         Room_TestTriggers(g_TriggerIndex, true);
```
rr-, line 120: Ditto

rr- (commented):
I think we should also rename `x_floor`/`y_floor` sprinkled across the codebase to `x_sector` and `z_sector`, or `sx` and `sz` to make it less wordy (though this may collide with `sx` used for sin(x) sometimes).

src/game/objects/creatures/bat.c
lahm86: Yes, it sets `g_TriggerIndex`. It will get nuked in a later PR 😌

lahm86:
> I think we should also rename `x_floor`/`y_floor` sprinkled across the codebase to `x_sector` and `z_sector`, or `sx` and `sz` to make it less wordy (though this may collide with `sx` used for sin(x) sometimes).

Agreed. Shall I tackle it in its own PR to keep things incremental with these changes, or OK just to combine here?

src/game/objects/creatures/bat.c
lahm86: Actually, there is no trigger call here so I think you're right. I'll double check this as I believe it was added to fix bat positions.

src/game/objects/traps/thors_hammer.c
lahm86: This one's definitely needed for the subsequent trigger call.

walkawayy:
Are you planning on keeping `OBJECT_INFO` `floor`? Such as `obj->floor = TrapDoor_Floor;`. Would functions like this and the bridge functions end up being `_Sector`?

There's also still `item->floor`. I know some of these names might be addressed in the future, but I just wanted to get an idea of the rename plans.

rr-:
> > I think we should also rename `x_floor`/`y_floor` sprinkled across the codebase to `x_sector` and `z_sector`, or `sx` and `sz` to make it less wordy (though this may collide with `sx` used for sin(x) sometimes).
> 
> Agreed. Shall I tackle it in its own PR to keep things incremental with these changes, or OK just to combine here?

Let's amend this into the same commit, as it's just a rename of a couple of variables, which we already do in the same commit as `FLOOR_INFO` rename, as opposed to heavier refactors.

I agree with @walkawayy 's comments.

lahm86:
Pushed commits to remove the unnecessary height call in bat setup, and to standardize `x_floor`/`z_floor` naming. This also renames the size variables in `ROOM_INFO` to be absolutely clear (Z is first in the data, then X - see `level.c`). 

For `floor` and `ceiling` in `OBJECT_INFO` I would like to tackle this in the next PR as these are height specific. I'm going to make `floor` and `ceiling` a struct in `SECTOR_INFO`, which will initially have a `height` value (will later also have `tilt`). I think we should then rename the `OBJECT_INFO` function pointers to something like `alterFloorHeight` and `alterCeilingHeight` because that's what they're doing specifically to the sector (or rather the height value passed to it). So the likes of `TrapDoor_Floor`/`TrapDoor_Ceiling` would become `TrapDoor_AlterFloorHeight` and `TrapDoor_AlterCeilingHeight`. I know this is a bit wordier but it's more descriptive IMO.

`floor` in `ITEM_INFO` is similar, so that's a height value rather than a sector, so I think we should rename it to `floor_height`.

walkawayy (approved):
Looks great! Didn't realize how many cases existed across the codebase. Looked like a ton of work but it's much cleaner now.

rr- (approved):

PR #1430, opened 2024-08-06
types: refactor sector floor and ceiling properties

lahm86:
Part of #941.

#### Checklist

#### Description

This introduces a struct in `SECTOR_INFO` which currently stores the click height for the floor and ceiling values. The next phase will overhaul floor data parsing, and with that ,`tilt` will be added to this struct so that it's readily available. I've updated the issue itself with a task list for everything. I think perhaps just two more PRs after this one, all going well.

We have also replaced `height << 8` throughout with regard to sector clicks and instead store the height in world units in the sector.

Object floor and ceiling functions have been renamed to make clear what they do to the passed in height value.

### DIFF
--- a/src/game/inject.c
+++ b/src/game/inject.c
@@ -1370,15 +1370,15 @@ static void Inject_RoomShift(INJECTION *injection, int16_t room_num)
 
     // Update the sector floor and ceiling clicks to match.
     const int8_t click_shift = y_shift / STEP_L;
-    const int8_t wall_height = NO_HEIGHT / STEP_L;
     for (int i = 0; i < room->z_size * room->x_size; i++) {
         SECTOR_INFO *sector = &room->sectors[i];
-        if (sector->floor == wall_height || sector->ceiling == wall_height) {
+        if (sector->floor.height == WALL_CLICKS
+            || sector->ceiling.height == WALL_CLICKS) {
             continue;
         }
 
-        sector->floor += click_shift;
-        sector->ceiling += click_shift;
+        sector->floor.height += click_shift;
+        sector->ceiling.height += click_shift;
     }
 
     // Update vertex Y values to match; x and z are room-relative.
--- a/src/game/inject.c
+++ b/src/game/inject.c
@@ -1346,6 +1346,10 @@ static void Inject_RoomShift(INJECTION *injection, int16_t room_num)
     File_Read(&z_shift, sizeof(uint32_t), 1, fp);
     File_Read(&y_shift, sizeof(int32_t), 1, fp);
 
+    x_shift = ROUND_TO_SECTOR(x_shift);
+    y_shift = ROUND_TO_CLICK(y_shift);
+    z_shift = ROUND_TO_SECTOR(z_shift);
+
     ROOM_INFO *room = &g_RoomInfo[room_num];
     room->x += x_shift;
     room->z += z_shift;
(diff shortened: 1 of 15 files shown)


src/game/inject.c
```diff
@@ -1368,17 +1368,16 @@ static void Inject_RoomShift(INJECTION *injection, int16_t room_num)
...
-        if (sector->floor == wall_height || sector->ceiling == wall_height) {
+        if (sector->floor.height == NO_HEIGHT
+            || sector->ceiling.height == NO_HEIGHT) {
             continue;
         }
 
-        sector->floor += click_shift;
-        sector->ceiling += click_shift;
+        sector->floor.height += y_shift;
+        sector->ceiling.height += y_shift;
```
rr-, line 1384: Do we need to `& ~(STEP_L - 1)` here so that the height is always a multiple of `STEP_L`?

src/global/types.h
```diff
@@ -1643,9 +1644,9 @@ typedef struct OBJECT_INFO {
     FRAME_INFO *frame_base;
     void (*initialise)(int16_t item_num);
     void (*control)(int16_t item_num);
-    void (*floor)(
+    void (*floor_height_routine)(
         ITEM_INFO *item, int32_t x, int32_t y, int32_t z, int16_t *height);
-    void (*ceiling)(
+    void (*ceiling_height_routine)(
```
rr-: Can we name this `_func`? (`draw_routine` would become `draw_func` soon after)

src/game/room.c
```diff
@@ -661,13 +661,13 @@ void Room_AlterFloorHeight(ITEM_INFO *item, int32_t height)
...
-        if (sector->floor == sky_sector->ceiling) {
-            sector->floor = NO_HEIGHT / 256;
+    if (sector->floor.height != NO_HEIGHT) {
+        sector->floor.height += height;
+        if (sector->floor.height == sky_sector->ceiling.height) {
+            sector->floor.height = NO_HEIGHT;
         }
     } else {
-        sector->floor = sky_sector->ceiling + (height >> 8);
+        sector->floor.height = sky_sector->ceiling.height + height;
```
rr-: Do we need to `(height & ~(STEP_L - 1))`?

src/game/room.c
```diff
@@ -661,13 +661,13 @@ void Room_AlterFloorHeight(ITEM_INFO *item, int32_t height)
         sector = &r->sectors[z_sector + x_sector * r->z_size];
     }
 
-    if (sector->floor != NO_HEIGHT / 256) {
-        sector->floor += height >> 8;
-        if (sector->floor == sky_sector->ceiling) {
-            sector->floor = NO_HEIGHT / 256;
+    if (sector->floor.height != NO_HEIGHT) {
+        sector->floor.height += height;
```
rr-: Do we need to `(height & ~(STEP_L - 1))`?

rr- (approved):
The last bit I'd consider changing is turning `void (*ceiling_height_func)(const ITEM_INFO *item, int32_t x, int32_t y, int32_t z, int16_t *height)` into `int16_t (*ceiling_height_func)(const ITEM_INFO *item, int32_t x, int32_t y, int32_t z, int16_t height)` eg make the new `height` to be returned rather than an out parameter. But it's optional.

(As usual please squash when merging.)

src/game/room.c
```diff
@@ -661,13 +661,14 @@ void Room_AlterFloorHeight(ITEM_INFO *item, int32_t height)
...
-            sector->floor = NO_HEIGHT / 256;
+    if (sector->floor.height != NO_HEIGHT) {
+        sector->floor.height += height & ~(STEP_L - 1);
+        if (sector->floor.height == sky_sector->ceiling.height) {
+            sector->floor.height = NO_HEIGHT;
         }
     } else {
-        sector->floor = sky_sector->ceiling + (height >> 8);
+        sector->floor.height =
+            sky_sector->ceiling.height + (height & ~(STEP_L - 1));
```
walkawayy: is there any way we can make this less ugly?

walkawayy (approved):
LGTM and I agree with rr.

src/game/room.c
rr-: We could introduce a macro for this, such as round to click or something.

rr- (approved):

PR #1433, opened 2024-08-08
room: refactor tilt and portal interpretation

lahm86:
Part of #941

#### Checklist

#### Description

This moves floor and ceiling tilts directly into sectors to avoid repeated floor data loops to retrieve their values. It also introduces some additional functions to make traversing up/down to get a relevant sector easier, and so removes more repeated logic there.

For portals, sectors now have a struct with `pit`, `sky` and `wall` values. `pit` and `sky` are taken from the original data layout, while `wall` is now retrieved from the floor data once (and is initialised to `NO_ROOM` by default). This allows us to eliminate `Room_GetDoor` for horizontal portals.

There are some `TODO`s left related to the parsing: the next PR will move lava and trigger FD into sectors too, and we can then eliminate the FD globals.

I had one gotcha with the signs in `Room_GetCeilingTiltHeight` in relation to the new `NEG_TILT`/`POS_TILT` macros, where the tilts were inverted in-game. Perhaps these could be better named (and of course open to renaming anything else here). For testing ceilings I used the fly cheat and pushed Lara to the roof then swam and kept holding down to make sure she followed the geometry.

### DIFF
--- a/src/game/objects/general/door.c
+++ b/src/game/objects/general/door.c
@@ -55,8 +55,8 @@ static void Door_Shut(DOORPOS_DATA *const d)
     sector->box = NO_BOX;
     sector->floor.height = NO_HEIGHT;
     sector->ceiling.height = NO_HEIGHT;
-    sector->sky_room = NO_ROOM;
-    sector->pit_room = NO_ROOM;
+    sector->portal_room.sky = NO_ROOM;
+    sector->portal_room.pit = NO_ROOM;
 
     const int16_t box_num = d->block;
     if (box_num != NO_BOX) {
@@ -119,7 +119,7 @@ void Door_Initialise(int16_t item_num)
     z_sector = ((item->pos.z - r->z) >> WALL_SHIFT) + dx;
     x_sector = ((item->pos.x - r->x) >> WALL_SHIFT) + dy;
     door->d1.sector = &r->sectors[z_sector + x_sector * r->z_size];
-    room_num = Room_GetDoor(door->d1.sector);
+    room_num = door->d1.sector->portal_room.wall;
     if (room_num == NO_ROOM) {
         box_num = door->d1.sector->box;
     } else {
@@ -139,7 +139,7 @@ void Door_Initialise(int16_t item_num)
         z_sector = ((item->pos.z - r->z) >> WALL_SHIFT) + dx;
         x_sector = ((item->pos.x - r->x) >> WALL_SHIFT) + dy;
         door->d1flip.sector = &r->sectors[z_sector + x_sector * r->z_size];
-        room_num = Room_GetDoor(door->d1flip.sector);
+        room_num = door->d1flip.sector->portal_room.wall;
         if (room_num == NO_ROOM) {
             box_num = door->d1flip.sector->box;
         } else {
@@ -157,7 +157,7 @@ void Door_Initialise(int16_t item_num)
         door->d1flip.sector = NULL;
     }
 
-    room_num = Room_GetDoor(door->d1.sector);
+    room_num = door->d1.sector->portal_room.wall;
     Door_Shut(&door->d1);
     Door_Shut(&door->d1flip);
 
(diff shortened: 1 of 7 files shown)


src/game/room.c
```diff
@@ -536,11 +453,75 @@ int16_t Room_GetHeight(
...
+static int16_t Room_GetFloorTiltHeight(
+    const SECTOR_INFO *sector, const int32_t x, const int32_t z)
+{
+    int16_t height = sector->floor.height;
+    if (sector->floor.tilt == 0) {
+        return height;
+    }
+
+    const int32_t zoff = sector->floor.tilt >> 8;
+    const int32_t xoff = (int8_t)sector->floor.tilt;
```
rr-: Let's name these `z_off` and `x_off`, here and in other functions

src/game/objects/general/door.c
```diff
@@ -119,7 +119,7 @@ void Door_Initialise(int16_t item_num)
     z_sector = ((item->pos.z - r->z) >> WALL_SHIFT) + dx;
     x_sector = ((item->pos.x - r->x) >> WALL_SHIFT) + dy;
     door->d1.sector = &r->sectors[z_sector + x_sector * r->z_size];
-    room_num = Room_GetDoor(door->d1.sector);
+    room_num = door->d1.sector->portal_room.wall;
```
walkawayy, line 122: Below it looks like there is a lot of repeated code like:
```
        b = &g_RoomInfo[room_num];
        z_sector = ((item->pos.z - b->z) >> WALL_SHIFT) + dx;
        x_sector = ((item->pos.x - b->x) >> WALL_SHIFT) + dy;
        box_num = b->sectors[z_sector + x_sector * b->z_size].box;
```

Is it worth another function in `room.c` similar to `Room_GetPitSector`. Something like `Room_GetBoxNumber`?


EDIT: Nvm probably not. `box_num` is set differently throughout.

src/game/room.c
```diff
@@ -536,11 +453,75 @@ int16_t Room_GetHeight(
...
 
+static int16_t Room_GetFloorTiltHeight(
+    const SECTOR_INFO *sector, const int32_t x, const int32_t z)
+{
+    int16_t height = sector->floor.height;
+    if (sector->floor.tilt == 0) {
+        return height;
+    }
+
+    const int32_t z_off = sector->floor.tilt >> 8;
```
walkawayy, line 464: We mentioned getting rid of shifts, but it looks like it's still used here and in other spots with `WALL_SHIFT`? I might be misunderstanding what we discussed.

`why do we shift left by 8? should this be * STEP_L?`

`Yeah, it's moving it into world units. Thinking about it, should we just do this once on level load too? It seems (almost) every reference to it does the shift, so we could just store it as an int16_t and do away with all the shifting`

src/game/room.c
```diff
@@ -162,27 +177,13 @@ static void Room_RemoveFlipItems(ROOM_INFO *r)
...
 
-    while (sector->pit_room != NO_ROOM) {
-        r = &g_RoomInfo[sector->pit_room];
-        sector = &r->sectors
-                      [((z - r->z) >> WALL_SHIFT)
-                       + ((x - r->x) >> WALL_SHIFT) * r->z_size];
-    }
-
-    if ((y + 512) < sector->floor.height) {
+    if ((y + STEP_L * 2) < sector->floor.height) {
```
walkawayy, line 182: Maybe after the refactors we can combine the `walkable` function check here. We should have better floordata structure that maybe will allow us to easily add another if check on whether to return no tilt.

rr- (approved):

walkawayy (approved):

src/game/room.c
lahm86: We could store X/Z tilts as `int8_t` in the sector. `Room_GetTiltType` expects the `int16_t` though but I'll see what can be done there.

src/game/objects/general/door.c
lahm86: Yeah, there are some cases like this left. What would be ideal in `C` is lambda functions so we could iterate commonly and "do the thing" per-case.

PR #1434, opened 2024-08-09
room: refactor lava and trigger interpretation

### LINKED ISSUE #941: Parse floor data into structures

lahm86:
Resolves #941.

#### Checklist

#### Description

This completes the floor data migration into properties in `SECTOR_INFO`. The raw `int16_t` array from the level files is now only used temporarily while the level is loaded; once all FD is parsed, it is discarded. This has allowed `g_FloorData` to be removed.

We do some forward lookup when reading `TRIGGER` data to know how many `TRIGGER_CMD`s are needed to be stored. The benefit is that in retrieving data later in the codebase, we just need simple `for` loops rather than re-interpreting FD values repeatedly.

`Room_TestTriggers` now expects the "triggerer" (`ITEM_INFO`) to be passed to it and it works out the sector, "heavy" state and suchlike from there. This greatly simplifies calls to it, as we don't need to have guaranteed that `g_TriggerIndex` has been set via `Room_GetHeight`. That global - and the related pointer in `COLL_INFO` - have also been removed as they are no longer required.

I did wonder if we could add function pointers to `TRIGGER_CMD` to tidy the switch statement in `Room_TestTriggers` e.g. when parsing, we allocate the function pointer based on the command type. But this probably needs some further refactoring, such as making a new module, plus we need to track flip status, flip effect and suchlike throughout the switch statement, so it maybe needs some further thought.

One other niggle is that `index` in `SECTOR_INFO` is only used in the initial parsing. I can't think of a good way to store these temporarily and then discard, but this is maybe OK for now.

For floor data injection, no `.bin` changes are needed, despite the version being updated. This is only so that subsequent updates don't need to have a global FD injection size. We no longer append to `g_FloorData` during injection; instead, because it now takes place after sector/FD parsing, we just update the existing properties or have `room.c` parse the provided injection data where new entries are needed.

### DIFF
--- a/src/game/objects/general/door.c
+++ b/src/game/objects/general/door.c
@@ -51,12 +51,12 @@ static void Door_Shut(DOORPOS_DATA *const d)
         return;
     }
 
-    sector->index = 0;
     sector->box = NO_BOX;
     sector->floor.height = NO_HEIGHT;
     sector->ceiling.height = NO_HEIGHT;
     sector->portal_room.sky = NO_ROOM;
     sector->portal_room.pit = NO_ROOM;
+    sector->portal_room.wall = NO_ROOM;
 
     const int16_t box_num = d->block;
     if (box_num != NO_BOX) {
--- a/src/game/room.c
+++ b/src/game/room.c
@@ -1,6 +1,7 @@
 #include "game/room.h"
 
 #include "game/camera.h"
+#include "game/gamebuf.h"
 #include "game/items.h"
 #include "game/lot.h"
 #include "game/music.h"
@@ -17,6 +18,7 @@
 
 #include <libtrx/utils.h>
 
+#include <assert.h>
 #include <stddef.h>
 
 #define NEG_TILT(T, H) ((T * (H & (WALL_L - 1))) >> 2)
@@ -28,7 +30,7 @@ int32_t g_FlipEffect = -1;
 int32_t g_FlipStatus = 0;
 int32_t g_FlipMapTable[MAX_FLIP_MAPS] = { 0 };
 
-static void Room_TriggerMusicTrack(int16_t track, int16_t flags, int16_t type);
+static void Room_TriggerMusicTrack(int16_t track, const TRIGGER *const trigger);
 static void Room_AddFlipItems(ROOM_INFO *r);
 static void Room_RemoveFlipItems(ROOM_INFO *r);
 
@@ -36,17 +38,15 @@ static int16_t Room_GetFloorTiltHeight(
     const SECTOR_INFO *sector, const int32_t x, const int32_t z);
 static int16_t Room_GetCeilingTiltHeight(
     const SECTOR_INFO *sector, const int32_t x, const int32_t z);
-static SECTOR_INFO *Room_GetPitSector(
-    const SECTOR_INFO *sector, int32_t x, int32_t z);
 static SECTOR_INFO *Room_GetSkySector(
     const SECTOR_INFO *sector, int32_t x, int32_t z);
 
 static void Room_PopulateSectorData(
     SECTOR_INFO *sector, const int16_t *floor_data);
 
-static void Room_TriggerMusicTrack(int16_t track, int16_t flags, int16_t type)
+static void Room_TriggerMusicTrack(int16_t track, const TRIGGER *const trigger)
 {
-    if (track == MX_UNUSED_0 && type == TT_ANTIPAD) {
+    if (track == MX_UNUSED_0 && trigger->type == TT_ANTIPAD) {
         Music_Stop();
         return;
     }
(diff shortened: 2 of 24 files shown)


src/game/room.c
```diff
@@ -752,68 +730,65 @@ static void Room_PopulateSectorData(
...
-            switch_off = g_Items[value].current_anim_state == LS_RUN;
+            switch_off =
+                g_Items[trigger->item_index].current_anim_state == LS_RUN;
             break;
         }
 
         case TT_PAD:
         case TT_ANTIPAD:
-            if (g_LaraItem->pos.y != g_LaraItem->floor) {
+            if (item->pos.y != item->floor) {
```
walkawayy, line 782: Could this have an implication where an item other than Lara could trigger an antipad?

walkawayy (approved):
This looks amazing!!

After all the refactors, we should make sure testers like @aredfan and us do some proper play testing to make sure there are no regressions.

src/game/room.c
lahm86: That'd make for some interesting skips for sure 😄 
Other items will enter the `is_heavy` condition above (line 758) so it's only ever Lara in this `else` part 😌 I just thought a couple of references fewer to a global was nice where possible.

walkawayy: Oh nice. I recently saw a playthrough of a custom level and the person tried to trigger a floor pad with a dead animal. Kinda genius but it didn't work. :laughing: 

So could we change `Lava_Burn(g_LaraItem);` to `Lava_Burn(item);` just to get rid of the global in that whole function?

lahm86: Oh yes. Although I think we'll have to cast to non const.

rr- (commented):
Call me overzealous, but what do you think about making TRIGGER.parameter a `void*`, and store the `camera` member in a linear GameBuffer to which we point then through `parameter`?

src/game/objects/general/door.c
```diff
@@ -51,12 +51,12 @@ static void Door_Shut(DOORPOS_DATA *const d)
         return;
     }
 
-    sector->index = 0;
```
rr-, line 54: Can you explain a bit more regarding why this assignment is no longer necessary?

src/game/objects/general/door.c
```diff
@@ -51,12 +51,12 @@ static void Door_Shut(DOORPOS_DATA *const d)
...
     sector->box = NO_BOX;
     sector->floor.height = NO_HEIGHT;
     sector->ceiling.height = NO_HEIGHT;
     sector->portal_room.sky = NO_ROOM;
     sector->portal_room.pit = NO_ROOM;
+    sector->portal_room.wall = NO_ROOM;
```
rr-, line 59: Is it safe to introduce this assignment?

lahm86: Ah, apologies, I forgot to discuss this above. This was the last reference to the sector's index, along with in `Room_GetSector`, and I overlooked it in the last PR. When we had `Room_GetDoor`, setting the sector's index to 0 meant that that returned `NO_ROOM` when the door is shut. So this is just doing the same but more explicitly. When the door is opened, the original sector gets restored.
I think this "FD nulling" is in place otherwise Lara could glitch through closed doors and then `Room_GetSector` would push her through the portal.

lahm86:
> Call me overzealous, but what do you think about making TRIGGER.parameter a `void*`, and store the `camera` member in a linear GameBuffer to which we point then through `parameter`?

Nice idea. I will try it out tomorrow.

src/game/objects/general/door.c
lahm86: This goes hand in hand with the line below.

lahm86:
I've changed `parameter` to `void *` in the command struct, and although there is a little extra casting needed here and there, I think overall it's much better. We're not allocating redundant space for camera data in every command crucially.

rr- (approved):

PR #1445, opened 2024-08-15
inventory: set any weapon to default on pickup if unarmed

### LINKED ISSUE #1443: Set default weapon on pickup
If Lara picks up a gun that's not the pistols, it should (perhaps optionally) become the default gun.
See https://github.com/LostArtefacts/TR1X/pull/1442#pullrequestreview-2237460540

I'm not sure that this is a bug, rather a missing feature from OG.

walkawayy:
Resolves #1443.

#### Checklist

#### Description
Set Lara's default weapon to any weapon she picks up when unarmed, not just pistols.

If you want to move this to 4.4 that's fine, but it seems like it's easy to implement and related to the other gun fixes. I thought we could maybe squeeze it in if there's no bugs.  I attached a build for @aredfan to test.

[default_weapon_v1 TR1X-4.2-47-g49b582f-Windows.zip](https://github.com/user-attachments/files/16621456/default_weapon_v1.TR1X-4.2-47-g49b582f-Windows.zip)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - added weapons to Lara's empty holsters on pickup (#1291)
 - added options to quiet or mute music while underwater (#528)
 - changed the turbo cheat to no longer affect the gameplay time (#1420)
+- changed weapon pickup behavior when unarmed to set any weapon as the default weapon, not just pistols (#1443)
 - fixed adjacent Midas Touch objects potentially allowing gold bar duplication in custom levels (#1415)
 - fixed the excessive pitch and playback speed correction for music files with sampling rate other than 44100 Hz (#1417, regression from 2.0)
 - fixed the ingame timer being skewed upon inventory open (#1420, regression from 4.1)
@@ -16,7 +17,6 @@
 - fixed carrying over unexpected guns in holsters to the next level under rare scenarios (#1437, regression from 2.4)
 - fixed item cheats not updating Lara holster and backpack meshes (#1437)
 - improved initial level load time by lazy-loading audio samples (LostArtefacts/TR2X#114)
-- set Lara's default weapon to any weapon she picks up when unarmed, not just pistols (#1443)
 
 ## [4.2](https://github.com/LostArtefacts/TR1X/compare/4.1.2...4.2) - 2024-07-14
 - added creating minidump files on crashes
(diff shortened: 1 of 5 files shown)


aredfan (approved):
Very nice [@walkawayy](https://github.com/walkawayy)👍 

I tested every combination of 4 weapons picked up, then again with 3 weapons w/o pistols, Lara will always end up drawing the first weapon. This PR also addresses Lahm's anomaly where Lara has magnums/uzis visually holstered and instead draws pistols when starting Atlantis (with pistols as primary weapon disabled).

LGTM.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 - fixed carrying over unexpected guns in holsters to the next level under rare scenarios (#1437, regression from 2.4)
 - fixed item cheats not updating Lara holster and backpack meshes (#1437)
 - improved initial level load time by lazy-loading audio samples (LostArtefacts/TR2X#114)
+- set Lara's default weapon to any weapon she picks up when unarmed, not just pistols (#1443)
```
rr-: - `added`/`changed`/`fixed`/`removed`/`improved`

I'd say in this change we should go with `changed`

src/game/inventory/inventory_func.c
```diff
@@ -14,6 +14,11 @@ bool Inv_AddItem(const GAME_OBJECT_ID object_id)
 {
     if (Object_IsObjectType(object_id, g_GunObjects)) {
         Gun_UpdateLaraMeshes(object_id);
+        if (g_Lara.gun_type == LGT_UNARMED) {
+            g_Lara.gun_type = Gun_GetType(object_id);
+            g_Lara.gun_status = LGS_ARMLESS;
+            Gun_InitialiseNewWeapon();
```
rr-, line 20: Is this call necessary? It resets the meshes and some Lara's arm information, and I don't think it's relevant in this case.

rr- (commented):
Very nice, thank you for this. We'll squeeze it into 4.3.

lahm86 (approved):

src/game/inventory/inventory_func.c
walkawayy: Yes otherwise Lara's arms are screwed up when drawing her weapon especially the shotgun. Setting `g_Lara.gun_status = LGS_ARMLESS` before the call prevents the relevant `DrawMeshes` from being called in that function that would add the weapon to her hand mesh. This call is used in the `Lara_InitialiseInventory`, `Gun_Control` when drawing a weapon, and `Lara_RevertToPistolsIfNeeded`. I don't think it's an issue?

walkawayy: Maybe I can call `Gun_Pistols_Ready` or `Gun_Rifle_Ready` instead? Those are only called during `Gun_X_Draw` functions though, so I didn't want the code to look confusing.

rr-: No it's fine as-is, thanks for explaining.

rr- (approved):

Issue #1456, opened 2024-08-27
Make level loader consistent with TR2X


rr-:
- Split the `Level_Load...` functions into even smaller functions (use TR2X for reference)
- Introduce benchmarks to every function
- Avoid file seeks by using intermediate arrays (use TR2X for reference)
- Finally, convert the loader routines to use the new VFile module

PR #1458, opened 2024-08-28
level: refactor level loader and introduce benchmarking

walkawayy:
#### Checklist

#### Description
Internal change. Breaks up the level loader into smaller functions to match TR2X. Adds benchmarking to functions. Also visually reordered a few functions to match the order the level loaded is loaded in. This is part 1 of the refactor described in #1456. I wanted to gather some benchmarking data before the change to a virtual file in a part 2 PR.

### DIFF
--- a/src/game/level.c
+++ b/src/game/level.c
@@ -25,6 +25,7 @@
 #include "global/types.h"
 #include "global/vars.h"
 
+#include <libtrx/benchmark.h>
 #include <libtrx/filesystem.h>
 #include <libtrx/log.h>
 #include <libtrx/memory.h>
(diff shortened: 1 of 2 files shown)


src/game/level.c
```diff
@@ -89,10 +99,50 @@ static bool Level_LoadFromFile(
...
         return false;
     }
 
+    if (!Level_LoadStaticObjects(fp)) {
+        return false;
+    }
+
+    if (!Level_LoadTextures(fp)) {
+        return false;
+    }
```
rr-: Actually we don't need to make these return a bool, as any error during level load is fatal and should cause the game to exit immediately.

rr- (commented):
Thanks. Can we also adjust this pull request to avoid seeks in `Level_LoadAnimFrames` in preparation for VFiles, as a separate commit? (Possibly remove all seeks altogether.)
This will require `Level_LoadAnims` to output `frame_pointers` information like in TR2X.

rr- (approved):

lahm86 (approved):
LGTM

PR #1459, opened 2024-08-29
console: use the new string functions

rr-:
#### Checklist

#### Description

A loose proposal how to refactor parsing the arguments of the console commands using https://github.com/LostArtefacts/libtrx/pull/19. I refactored a few commands to showcase how this works.

I'm not sure if I like this, honestly, and I'm eager to hear your opinion. It works great for boolean commands, but might make string handling more cumbersome. For instance, the `/play` command accepts a level name, but the new argument parsing requires spaces to be enclosed in quotes (e.g., `/play "great pyramid"`). The problem is, the console does not support all characters, and it's actually impossible to type this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - fixed cameras with glide values sometimes moving in the wrong direction (#1451, regression from 4.3)
 - fixed `/give` console command giving duplicate items under some circumstances (#1463, regression from 3.0)
 - fixed `/give` console command confusing logging around mismatched items (#1463, regression from 3.0)
+- fixed `/flip` console command misreporting an already enabled flipmap as off (regression from 4.0)
 - fixed console commands causing improper ring shutdown with selected inventory item (#1460, regression from 3.0)
 - improved logs module names readability
 - improved crash debug information on Windows
--- a/src/game/console.c
+++ b/src/game/console.c
@@ -15,6 +15,7 @@
 #include <libtrx/log.h>
 
 #include <SDL2/SDL_keycode.h>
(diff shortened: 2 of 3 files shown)


rr-:
I've simplified the proposal to use only argument parser primitives, restoring the previous behavior, and it looks acceptable. However, this still adds a lot of code to libtrx. I'm hesitant to introduce it since the code, while functional, is currently unused and likely will remain unused due to the intended flexibility of the command arguments.

Maybe the introduced functions should just become a couple of `bool String_ParseBool(const char *input, bool *target);` without introducing this new module at all.

 LMK what you think.

lahm86:
It looks really good, but as you say maybe for now it's best to go with the simpler parsing functions for ease. No harm to leave it on a branch on libtrx for potential future integration?

rr-:
Gotcha – I'll open a new PR with `String_Parse(…)`, use that to simplify existing commands, and rebase this on top of that one.

rr-:
Done

walkawayy:
It works really nice! Only odd thing I noticed was doing something like `give 10 medi`. I think it's fine to give both small and large medi with this command since it's ambiguous, but it looks like it happens twice for small and large. This console log screenshot is from one `give` command.

image

rr-:
That's a separate bug, which happens because `Object_IdsFromName` doesn't ensure that the matched objects are unique.

Logged as https://github.com/LostArtefacts/TR1X/issues/1463.

walkawayy (approved):
lgtm

lahm86 (approved):

Issue #1346, opened 2024-04-27
Black screen and crash at UB start with custom EGYPT.PHD


bursoft-git:
Hello!
Thanks for your cool work guys!
I translate Tomb1Main-TR1X into russian, and use custom *.PHD files for this
Early versions of TR1X work fine with my custom *.PHD, but version 4.1 crash when i start UB
[EGYPT.ZIP](https://github.com/LostArtefacts/TR1X/files/15138149/EGYPT.ZIP)

TR1X.log
`../../src/specific/s_shell.c 160 S_Shell_ShowFatalError GameBuf_Alloc(): OUT OF MEMORY Cinematic frames -82272`

Other files, if needed, you can grab on my translation page
https://bursoft-portable.blogspot.com/p/tomb-raider-i-gold.html

lahm86:
This does appear to be a regression since 4.1 - many thanks for reporting the problem.

PR #1467, opened 2024-08-30
level: use virtual files during loading process

### LINKED ISSUE #1456: Make level loader consistent with TR2X

walkawayy:
Resolves #1456.

#### Checklist

#### Description
Internal change. This is part 2 of the refactor described in #1456. This introduces virtual files to the level loading process. I didn't notice any speedup on my machine so testing is appreciated. Maybe we need to do some benchmarking with everything on our PC closed. I definitely have a lot of windows and tabs open.

### DIFF
--- a/src/game/level.c
+++ b/src/game/level.c
@@ -719,13 +719,15 @@ static void Level_LoadAnimatedTextures(VFILE *file)
             file, range->textures, sizeof(int16_t) * range->num_textures);
     }
 
-cleanup:
+cleanup: {
     // Ensure to read everything intended by the level compiler, even if it
     // does not wholly contain accurate texture data.
-    if (end_position - VFile_GetPos(file) > 0) {
-        VFile_Skip(file, MAX(end_position, VFile_GetPos(file)));
+    const int32_t skip_length = end_position - VFile_GetPos(file);
+    if (skip_length > 0) {
+        VFile_Skip(file, skip_length);
     }
     Benchmark_End(benchmark, NULL);
+}
 }
 
 static void Level_LoadItems(VFILE *file)
--- a/src/game/level.c
+++ b/src/game/level.c
@@ -26,10 +26,10 @@
 #include "global/vars.h"
 
 #include <libtrx/benchmark.h>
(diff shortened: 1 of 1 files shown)


src/game/level.c
```diff
@@ -710,27 +710,29 @@ static void Level_LoadAnimatedTextures(MYFILE *fp)
...
-            fp, range->textures, sizeof(int16_t), range->num_textures);
+        VFile_Read(
+            file, range->textures, sizeof(int16_t) * range->num_textures);
     }
 
 cleanup:
     // Ensure to read everything intended by the level compiler, even if it
     // does not wholly contain accurate texture data.
-    File_Seek(fp, MAX(end_position, File_Pos(fp)), SEEK_SET);
+    if (end_position - VFile_GetPos(file) > 0) {
```
lahm86: This should be something like the following I think. The EGYPT.PHD file in #1346 crashes.
```c
const int32_t skip_length = end_position - VFile_GetPos(file);
if (skip_length > 0) {
    VFile_Skip(file, skip_length);
}
```

Or we implement `VFile_Seek` and use the previous `MAX` approach.

aredfan (approved):
I tested this PR build with level skip to make sure every level and cutscene loads ok. I also tried saving and loading a few times in Caves, Cistern and Atlantis. Both demos are loading fine as well. 

LGTM.

> I didn't notice any speedup on my machine so testing is appreciated.

I did notice something of a welcome improvement. Before this PR, I was having  frame drops of 1-2 when saving, or when opening the load game page. These frame drops are mostly gone.

src/game/level.c
walkawayy: Ok I added in your suggestion thanks!

lahm86 (approved):

rr- (approved):

PR #1469, opened 2024-08-30
data: switch from WEBP to JPEG

rr-:
#### Checklist

#### Description

The files appear identical visually, but the title menu loads much faster with this format. On my machine it goes down from 236 ms to around 150 ms. However, this bloats the asset size by 34% (9.4 MiB compared to 7.0 MiB).


walkawayy (commented):
Looks good except you missed the title, credits, and Eidos images in the UB gameflow file.

walkawayy (approved):

aredfan (approved):
LGTM.

lahm86 (approved):

rr-:
With the recent improvements on develop and #1470 the savings of this pull request went down to 15 milliseconds (130 ms vs 115 ms). I'll hold on for a bit longer to see if we can do better on other fronts prior to making this switch.

rr-:
All in all I don't think it's worth it anymore as the savings aren't so dramatic with the recent improvements on develop – especially considering the file size increase. Closing for now.

PR #1496, opened 2024-09-06
output: improve vertex movement

### LINKED ISSUE #1493: Improve vertex movement
When looking through water portals, all vertices in the opposite water status rooms move, which results in gaps where the rooms join. In TR2, room vertices have attributes for controlling this, but TR1 room vertices only have position and shade. Add an initial improvement to this, to target locking vertices that are on room edges.

image

Note that it won't feasibly be possible to fix every situation, as indeed you can see in TR2. This is a level design issue where vertices don't always match up between faces, and so the movement becomes out of sync. Fixing this would mean manually adjusting faces, vertices and tex infos, which is a very big task.

image

lahm86:
Resolves #1493.

#### Checklist

#### Description

This allows locking specific vertices such that they don't move when looked at through a water portal. We make use of an unused bit in vertex shading, and ensure shading is clamped to the max value.

This only covers vertices on water edges, as mentioned in #1493 there are other situations which would require a big undertaking of work to inject face/vertex/texture changes, and I don't think that it's feasible. This brings things into line with TR2 though, so it's a definite improvement.

I've tied this to the texture fix option in the config, but LMK if you think this deserves its own setting.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 - improved level load times
 - improved logs module names readability
 - improved crash debug information on Windows
+- improved vertex movement when looking through water portals (#1493)
 
 ## [4.3](https://github.com/LostArtefacts/TR1X/compare/4.2...4.3) - 2024-08-15
 - added deadly water feature from TR2+ for custom levels (#1404)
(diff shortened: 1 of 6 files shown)


aredfan (approved):
A subtle but definitive improvement for TR1. 

I've checked some of the water rooms in Cistern, Obelisk and Return to Egypt, all LGTM.

rr- (approved):

rr-:
I think we can leave it linked with textures.

walkawayy (approved):

PR #1497, opened 2024-09-07
room: test sector range for Torso boss

### LINKED ISSUE #1236: OG bug: Torso death can lead to a softlock
I was reminded of this recently where killing Torso on a specific tile can lead to a softlock. If he dies on the tile where Lara spawns, the door does not open.

image

I see two possible fixes:
1. Change the logic [here](https://github.com/LostArtefacts/TR1X/blob/develop/src/game/objects/creatures/torso.c#L257) to test triggers on all tiles surrounding the one on which he dies.
2. Replace the trigger under Lara with another heavy trigger for the door, and implement some extra logic in the data to trigger the camera, music and egg on starting the level.

The first is easiest, but it may have a negative impact on custom levels (bit of a guess).
The second would need an extra item (a boulder in a hidden room) so it could affect existing saves.

I'm inclined to opt for the second as it is ultimately a data issue, and maybe put a warning in the release regarding old saves.

lahm86:
Resolves #1236.

#### Checklist

#### Description

This provides an option to test a range of sectors around the one on which a large and heavy item dies. It ultimately resolves the potential softlock when killing Torso boss in Great Pyramid. For custom levels, it should be switched off in the gameflow as builders have full control of FD design (although it may have some uses, the freedom is there of course).

The only other solution to the softlock that I could see would involve item injection (a hidden boulder somewhere to trigger the camera/egg/music at the start) but this would break existing and TombATI saves, and I couldn't think of a way around that problem.

Other heavy trigger activators aren't affected here as they're all less than the 1.5 tile limit introduced here.

Just for testing - getting Adam to actually end up on the problematic tile is time-consuming, so attached is an exe that forces him here on death.
[TR1X.zip](https://github.com/user-attachments/files/16917520/TR1X.zip)

### DIFF
--- a/src/game/room.c
+++ b/src/game/room.c
@@ -40,6 +40,8 @@ static int16_t Room_GetCeilingTiltHeight(
     const SECTOR_INFO *sector, const int32_t x, const int32_t z);
 static SECTOR_INFO *Room_GetSkySector(
     const SECTOR_INFO *sector, int32_t x, int32_t z);
+static void Room_TestSectorTrigger(
+    const ITEM_INFO *item, const SECTOR_INFO *sector);
 
 static void Room_TriggerMusicTrack(int16_t track, const TRIGGER *const trigger)
 {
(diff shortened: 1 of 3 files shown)


src/game/room.c
```diff
@@ -736,11 +739,41 @@ void Room_PopulateSectorData(
...
 void Room_TestTriggers(const ITEM_INFO *const item)
 {
-    const bool is_heavy = item->object_id != O_LARA;
     int16_t room_num = item->room_num;
-    const SECTOR_INFO *const sector =
+    const SECTOR_INFO *sector =
         Room_GetSector(item->pos.x, MAX_HEIGHT, item->pos.z, &room_num);
 
+    Room_TestSectorTrigger(item, sector);
+    if (item->object_id == O_LARA || !g_GameFlow.use_extended_triggers) {
```
rr-: Thanks. Is the torso boss the only enemy for which this criterion is true? If so, we could just hardcode it to always trigger this behavior for it only, and skip the gameflow changes.

lahm86: Yes, good point. That'll be cleaner, I'll sort that out.

rr- (approved):

aredfan:
I was able to lure the torso boss onto the starting trigger. This fix checks out. 👍

PR #1504, opened 2024-09-12
Objects cleanup

rr-:
#### Checklist

#### Description

- Moves `FX_INFO*`-based objects from `game/objects/*/` to `game/objects/effects/`.
- Moves `ITEM_INFO*`-based objects out of `game/objects/effects/` to other folders.
- Tries to keep only one object type per file - splits bridges, scions etc into multiple files. We still have the following objects being defined in a shared file:
    - Lion/Lioness/Puma
    - Mutants
    - Rat/Vole
    - Alligator/Crocodile
    - Switch / Underwater lever
- Renames the object `Lava` → `Ember`.
- Renames the object `LavaEmitter` → `EmberEmitter`.
- Renames the function `Lava_TestFloor` → `Room_TestLava`.
- Renames the function `Lava_Burn` → `Lara_CatchFire`.
- Removes the function `Pod_SetupBig`, since it's the same as `Pod_Setup`.
- Removes the function `LeadBar_Setup` since it's just `Pickup_Setup`.
- Merges `CutscenePlayer` setup routines into one.
- Fixes `Cog_Control` being used by drawbridges just because it happened to use the same code. There's one more weird usage of `Cog_Control` – `O_MOVING_BAR`, but I am not sure what this object is about.
- Moves `setup.c` to the `objects/` namespace.
- Changes the door cheat to also include drawbridges.


aredfan (approved):

rr-:
@aredfan thank you for testing. Just to be sure, can you please confirm if there are no regressions regarding:

- cutscenes
- Scions and Scion pedestals
- bridges
- drawbridges
- cog machine
- flames and embers
- darts
- lava wedge (the moving lava block during escape from The Great Pyramid)
- Thor's hammer
- Natla's fight

? The top elements are the riskiest, bottom ones are the least risky. Thanks again.

aredfan:
@rr- no problem at all, it's my pleasure to help out.

I did miss some of the things on your list so I checked them and double checked everything else as well.  I also made sure to test the barricades at the end of Folly and the drill in Natla's Mines, for good measure. All LGTM.

src/game/level.c
```diff
@@ -905,7 +905,7 @@ static void Level_CompleteSetup(int32_t level_num)
     Stats_ObserveRoomsLoad();
 
     // Must be called after all g_Anims, g_Meshes etc initialised.
-    Setup_AllObjects();
+    Object_SetupAllObjects();
```
walkawayy, line 908: Is the second "object" redundant? I guess "Object_SetupAll" might be a little unclear.

src/game/objects/common.c
```diff
@@ -128,6 +128,7 @@ const GAME_OBJECT_ID g_TrapdoorObjects[] = {
     O_TRAPDOOR,
     O_TRAPDOOR2,
     O_BIGTRAPDOOR,
+    O_DRAW_BRIDGE,
```
walkawayy: I think we can do one word "drawbridge". https://en.wikipedia.org/wiki/Drawbridge

It's usually spelled as one word (or with a hyphen according to Wikipedia). Also it's one word in its own .c file.

src/game/objects/general/scion1.c
```diff
@@ -0,0 +1,85 @@
...
+};
+
+static const OBJECT_BOUNDS *Scion1_Bounds(void);
+
+static const OBJECT_BOUNDS *Scion1_Bounds(void)
+{
+    return &m_Scion1_Bounds;
+}
+
+void Scion1_Setup(OBJECT_INFO *obj)
```
walkawayy, line 34: What do you think about having some sort of comment in each Scion file describing which one it is? Example - Qualopec pedestal Scion, Pierre dropped Scion, etc? Just the number has been confusing in the past. @lahm86 what do you think also?

walkawayy (approved):

src/game/objects/general/scion1.c
rr-: I'm all for it, I just don't know what each of them does :D

src/game/objects/common.c
rr-: Gotcha, will adjust libtrx as well

src/game/level.c
rr-: I don't have a strong opinion on either, maybe @lahm86 can chime in?

lahm86: Either is fine IMO. `Object_SetupAllObjects` is very clear as to what it does.

src/game/objects/general/scion1.c
lahm86: That would be great 😌 

- `O_SCION_ITEM` - ToQ and Sanctuary, triggers `O_LARA_EXTRA` pedestal pickup anim
- `O_SCION_ITEM2` - normal pickup
- `O_SCION_ITEM3` - the Great Pyramid shootable variety
- `O_SCION_ITEM4` - Atlantis, triggers `O_LARA_EXTRA` reach anim

rr-: Done

src/game/objects/common.c
rr-: Done

src/game/level.c
rr-: Let's leave as-is.

rr-:
I've updated with the remarks and also simplified `stats.c` a bit to use `Object_IsObjectType`. @aredfan can you please check that the total pickup and kill counts are as expected for each level?

walkawayy (approved):

src/game/stats.c
```diff
@@ -104,39 +84,34 @@ static void Stats_CheckTriggers(
 
             const ITEM_INFO *const item = &g_Items[item_num];
 
-            // Add Pierre pickup and kills if oneshot
-            if (item->object_id == O_PIERRE && sector->trigger->one_shot) {
-                Stats_IncludeKillableItem(item_num);
+            if (item->object_id == O_PIERRE) {
+                // Add Pierre pickup and kills if oneshot
```
walkawayy, line 88: Tiny thing but I think the comment can just say `Add Pierre kill if oneshot`. @lahm86 since the change to de-hardcode item drops, Pierre's pickup in the OG game is tallied from the gameflow? I think it used to be added here which is why the comment references it.

lahm86: Yes, it's handled in `Stats_IncludeKillableItem`.

lahm86 (approved):

aredfan:
@rr- I've found 2 irregularities with total pickups:

- Tomb of Qualopec total pickups is 7 in this PR, and 8 in dev snapshot.
- Sanctuary of the Scion total pickups is 28 in this PR, and 29 in dev snapshot.

rr-:
@aredfan fixed. Coincidentally we also accidentally fixed a bug where the `/give` command would not work with Scion in these two levels.

aredfan:
Total pickups is now correct in those 2 levels and the `/give scion` command is working too. LGTM. 👍

Issue #902, opened 2023-07-28
Weird black lines between distant blocks


Reyks21:
Basically you can see some horizontal lines in between the distant blocks, not sure if this is a T1Main or OG Bug.

Its very visible in the far snow in the first level, and some other places.

![image](https://github.com/LostArtefacts/Tomb1Main/assets/10340735/99409f1e-3f41-4ea3-b986-8911ab5ef302)
![image](https://github.com/LostArtefacts/Tomb1Main/assets/10340735/fb11b3d4-933e-44b1-99bb-d99017e5ab7c)

Richard-L:
@Rapora9 had found out these lines occurred to him when increasing Anisotropic Filtering too much. If you set it back to 1 it should work:

![](https://media.discordapp.net/attachments/948590366255374416/1074856965974532116/Nayttokuva_235.png)

Reyks21:
@Richard-L You're right, that fixed it! Thanks!

aredfan:
Not a T1M bug btw, TombATI have it too.

It's not the first time I seen this issue raised and reducing AF to 1.0 is a workaround.  While this issue is still open I would like to ask whether there's a way to fix this bug because the game looks great with AF set to 16.0, or would it be beyond the scope of this project?

Richard-L:
Would be worth limiting what can be entered in the config tool to guide users.

pas-de-2:
I ran into this when playing at a sub native resolution like 850x480, setting AF to 1x fixed it, thanks!

pas-de-2:
The sky in Temple of the Cat is another example of the conflict between >1 AF and Pretty Pixels. Reducing AF to 1 or turning off Pretty Pixels fixes it.

![Image](https://github.com/user-attachments/assets/79a29750-5e6b-46ff-a01f-1c4442daa945)

rr-:
Hey folks, is this still happening on TR1X 4.10?

pas-de-2:
Seems to be resolved from my brief testing of Caves and Temple of the Cat.

pas-de-2:
I noticed these lines return when using AF >1 and an upscaling factor >1

aredfan:
This bug can be fixed by setting `Supersampling` to `2` or more.  The option is only available in development snapshots for now.

PR #1512, opened 2024-09-14
output: improve anisotropic filter rendering

### LINKED ISSUE #1507: The Pretty pixels option adds gaps between textures
Hello everyone!
I use the build TR1X-4.3-68-g1cb8ae9-Windows
I prefer to play with the filtering turned off. In this mode, gaps between textures are sometimes visible, especially on moving textures.
Pretty pixels OFF
Pretty pixels OFF+
But with the Pretty Pixels option enabled, the gaps become even more noticeable! 
Pretty pixels ON+

Apparently, these half pixels just serve to hide these gaps.
Is there a way to mask these gaps?
Thanks!

rr-:
#### Checklist

#### Description

Resolves #1507. Improves #902.

For a definite solution, we'd need to artifically add a border around each texture face inside texture atlases, hence I keep #902 as open for now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 - improved logs module names readability
 - improved crash debug information on Windows
 - improved vertex movement when looking through water portals (#1493)
+- improved anisotropic filter rendering (#902, #1507)
 
 ## [4.3](https://github.com/LostArtefacts/TR1X/compare/4.2...4.3) - 2024-08-15
 - added deadly water feature from TR2+ for custom levels (#1404)
--- a/src/game/output.c
+++ b/src/game/output.c
@@ -432,16 +432,30 @@ static const int16_t *Output_CalcRoomVertices(const int16_t *obj_ptr)
     int32_t vertex_count = *obj_ptr++;
 
     for (int i = 0; i < vertex_count; i++) {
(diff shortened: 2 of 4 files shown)


rr-:
Actually, hold on – this makes the issue worse in the Gym near the staircase to the piano room. The problem is that the walls are constructed like this:

```
##
####
######
###### | ######
###### | ######
###### | ######
###### +-------
###### | ######
###### | ######
```

Since there's 1 quad for the left side and 2 quads for the right side, this causes the calculations to misalign the faces along the `|` seam.

lahm86:
> Actually, hold on – this makes the issue worse in the Gym near the staircase to the piano room. The problem is that the walls are constructed like this:
> 
> ```
> ##
> ####
> ######
> ###### | ######
> ###### | ######
> ###### | ######
> ###### +-------
> ###### | ######
> ###### | ######
> ```
> 
> Since there's 1 quad for the left side and 2 quads for the right side, this causes the calculations to misalign the faces along the `|` seam.

I was just testing a TRLE and it's a similar problem there too, and with a skybox present the gaps are much more obvious. I'd imagine this kind of texturing is quite common.

rr-:
Restored high precision room vertex computing – I don't think this has downgraded the original fix.

lahm86 (approved):

PR #1513, opened 2024-09-14
inject: add PS1 crystal data

### LINKED ISSUE #1506: Restore PS1 purple colouring on save crystals
Now that we have the reflections so truthfully added to TR1X, the only thing missing about the crystals is their characteristic purple colour.

https://github.com/user-attachments/assets/e090a307-d9dd-447c-9af2-4bb312b689d2

@walkawayy I believe you'd identified the shade pretty precisely when doing the PS1 style menu UI, no? Else we can help.

lahm86:
Resolves #1506.

#### Checklist

#### Description

~~This allows injections to replace an unused palette slot with a specific color, and so allows an alternative to matching to the existing level palette. Data provided for each case of the savegame crystal, and the relevant config options updated/provided. The colour used is `#4040FF`.~~

~~I initially had in-game code to find the unused palette slots, but it was considerably expensive on loading times, so instead opted for the individual injection files with the information pre-calculated.~~

~~I also adjusted the teleport command to check crystal status to avoid tp'ing to ones that are consumed, mainly to allow checking how the crystal reacts to the lighting in different areas of the level.~~

This adds an option to replace the default savegame crystal model with one that matches PS1, using palette colour `#4040FF` on the mesh.


aredfan (approved):

src/game/console_cmd.c
```diff
@@ -182,7 +182,8 @@ static COMMAND_RESULT Console_Cmd_Teleport(const char *const args)
              item_num++) {
             const ITEM_INFO *const item = &g_Items[item_num];
             const bool is_pickup =
-                Object_IsObjectType(item->object_id, g_PickupObjects);
+                Object_IsObjectType(item->object_id, g_PickupObjects)
+                || item->object_id == O_SAVEGAME_ITEM;
```
rr-: This should be tracked in a separate issue + changelog entry. I suggest amending the condition below to spell `(is_pickup || item->object_id == O_SAVEGAME_ITEM)`, or rename this variable to something that reflects its purpose better (`is_consumable` for example).

rr- (commented):
Let's hold on with this approach for now – I believe that rather than forcing the injected texture colors into the level palette, we really should start supporting true color textures.

lahm86:
I've squashed out the console change and opened #1518 for that. Shall I just ditch this PR, or leave it open for reference for the time being?

rr-:
@lahm86 let's leave it open, I should be able to deliver the true color texturing today.

rr-:
Should be easier to implement with #1520 :relieved:  Our options include adding a new function to append a palette color, performing a texture swap, or ideally just injecting the entire object.

lahm86:
Updated to make use of the new palette logic - so much easier!

aredfan (approved):
I've checked one crystal per level, and all are showing the right colour as expected whether the purple tint option is enabled or disabled.

rr- (approved):

PR #1517, opened 2024-09-14
creature: perform safer conversion of hybrids

### LINKED ISSUE #1444: Missing rat and crocodile models can cause a crash
This only affects non-OG levels.

The game assumes that when `O_RAT` is loaded, `O_VOLE` is also loaded, and vice-versa. The same applies to `O_CROCODILE` and `O_ALLIGATOR`. If the counterpart model is missing, the game continues to switch the types depending on the room's water status, and so crashes can happen. A check should be added to ensure the model is present and if not, log a warning and remove the active item.

A similar issue exists with skaterboy if the skateboard model isn't present. The game assumes that it is, so it can crash when absent. Perhaps the check here should just warn, and then simply skip drawing the skateboard.

Ref: https://github.com/LostArtefacts/TR-Rando/pull/747

lahm86:
Resolves #1444.

#### Checklist

#### Description

A check will be performed before switching rats and crocodiles to ensure the target object is loaded. A similar check is done for skate kid's skateboard before trying to draw it.

Tidied up some `const`s too.

Attached is a couple of test levels, the first has the water models present and the land models not, the second vice-versa. There is a flipmap pad to test. Also, in the first, the skateboard is missing. 

[hybrids.zip](https://github.com/user-attachments/files/17002445/hybrids.zip)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 - fixed a potential softlock when killing the Torso boss in Great Pyramid (#1236)
 - fixed Bacon Lara re-spawning after saving and loading (#1500, regression from 0.7)
 - fixed config JSON not sanitizing some numeric values (#1515)
+- fixed potential crashes in custom levels if hybrid creature objects are not present in the level (#1444)
 - changed `/tp` console command to look for the closest place to teleport to when targeting items (#1484)
 - changed the door cheat to also target drawbridges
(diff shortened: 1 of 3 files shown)


aredfan (approved):
I've tried out the test levels and experienced no crashes. LGTM.

walkawayy (approved):

walkawayy (approved):

rr- (approved):

PR #1544, opened 2024-09-25
music: fix the main music playing at the wrong volume with certain sound options

### LINKED ISSUE #1540: TR1X bug: main menu music is affected when exiting to title while UW
**Regression since:** [4.3](https://github.com/LostArtefacts/TR1X/releases/tag/4.3)

___________________

If there's no music playing (not counting ambient) while Lara is underwater then exiting to title will cause the main menu music to be affected. This issue happens if any of the following 4 UW music options are used.

- Quiet
- Full but no ambient
- Quiet but no ambient
- None

**Quick way to reproduce:**

1. `/play 7`
2. Wait until the camera is submerged then use `/title`

walkawayy:
Fixed quiet or mute main menu music if a level was exited while underwater and the quiet, full but no ambient, quiet but no ambient, or none underwater music behavior option was set.

Resolves #1540.

#### Checklist

#### Description
Fixed the main menu music being quiet or mute if a level was exited while underwater and the quiet, full but no ambient, quiet but no ambient, or none underwater music behavior option was set (regression from 4.4). I tried to apply the fix to the volume in a logical place which is where the other config option to mute main menu music is. `M_EnsureEnvironment` couldn't be called because `Game_DrawScene` isn't called when in the title menu inventory so`Camera_Apply` is never called. Might need to be hotfixed though it's not a game breaking issue, and it should be a rare bug.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - fixed double "Fly mode enabled" message when using `/fly` console command (regression from 4.0)
 - fixed crash in the `/set` console command (regression from 4.4)
 - fixed toggling fullscreen not always saving (regression from 4.4)
+- fixed main menu music volume when exiting while underwater with certain music settings (#1540, regression from 4.4)
 
 ## [4.4](https://github.com/LostArtefacts/TR1X/compare/4.3-102-g458cd96...4.4) - 2024-09-20
 - added `/exit` command (#1462)
--- a/src/game/level.c
+++ b/src/game/level.c
@@ -1149,6 +1149,7 @@ bool Level_Initialise(int32_t level_num)
     Overlay_BarSetHealthTimer(100);
 
     Music_Stop();
+    Music_SetVolume(g_Config.music_volume);
     Sound_ResetEffects();
 
     Viewport_SetFOV(Viewport_GetUserFOV());
--- a/src/game/level.c
+++ b/src/game/level.c
@@ -1152,11 +1152,13 @@ bool Level_Initialise(int32_t level_num)
     Music_SetVolume(g_Config.music_volume);
     Sound_ResetEffects();
 
-    Viewport_SetFOV(Viewport_GetUserFOV());
-
-    if (g_GameFlow.levels[level_num].music) {
+    const bool disable_music = level_num == g_GameFlow.title_level_num
+        && !g_Config.enable_music_in_menu;
+    if (g_GameFlow.levels[level_num].music && !disable_music) {
         Music_PlayLooped(g_GameFlow.levels[level_num].music);
     }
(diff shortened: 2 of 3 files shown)


aredfan (approved):
LGTM. 👍

rr-:
Thanks. I think that a more fitting place for both this volume fix and the title level music check would be in `Level_Initialise`, like so:

```diff
diff --git a/src/game/level.c b/src/game/level.c
index 42c50c0f..c60f2143 100644
--- a/src/game/level.c
+++ b/src/game/level.c
@@ -1149,11 +1149,14 @@ bool Level_Initialise(int32_t level_num)
     Overlay_BarSetHealthTimer(100);
 
     Music_Stop();
+    Music_SetVolume(g_Config.music_volume);
     Sound_ResetEffects();
 
     Viewport_SetFOV(Viewport_GetUserFOV());
 
-    if (g_GameFlow.levels[level_num].music) {
+    const bool disable_music = level_num == g_GameFlow.title_level_num
+        && !g_Config.enable_music_in_menu;
+    if (g_GameFlow.levels[level_num].music && !disable_music) {
         Music_PlayLooped(g_GameFlow.levels[level_num].music);
     }
 
diff --git a/src/game/music.c b/src/game/music.c
index c24810a5..382bf82f 100644
--- a/src/game/music.c
+++ b/src/game/music.c
@@ -131,11 +131,6 @@ bool Music_PlayLooped(MUSIC_TRACK_ID track)
         return false;
     }
 
-    if (g_CurrentLevel == g_GameFlow.title_level_num
-        && !g_Config.enable_music_in_menu) {
-        return false;
-    }
-
     M_StopActiveStream();
 
     char *file_path = M_GetTrackFileName(track);
```

walkawayy:
Ok done. Amended the issue commit and added a second commit which moves the `enable_music_in_menu` option check to `Level_Initialise`. This looks much better. I just didn't want to move it without permission lol.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 - fixed ability to crash the game with extreme FOV values (regression from 0.9)
 - fixed double "Fly mode enabled" message when using `/fly` console command (regression from 4.0)
 - fixed crash in the `/set` console command (regression from 4.4)
+- fixed the main menu music being quiet or mute if a level was exited while underwater and the quiet, full but no ambient, quiet but no ambient, or none underwater music behavior option was set (regression from 4.4)
```
rr-: One more comment – Can we shorten this a bit? For example `fixed main menu music volume when exiting underwater with certain music settings`, and link to the issue #1540.

lahm86 (approved):

PR #1661, opened 2024-10-04
Fix wibble effect seams in level geometry

rr-:
#### Checklist

#### Description

#1493

Uses double rendering of underwater rooms with the wibble effect disabled (and Z-buffer writes off, to avoid depth races) to provide sensible pixels in the seams.

This has a chance of affecting performance in pools.


lahm86 (approved):

aredfan (approved):
I've mostly checked the water room in the 3 levels with skyboxes because that's where the seams are most evident. The seams issue is fixed in nearly all cases.

There's a few stray seams remaining in Obelisk of Khamoon, room 42.  These seams are also present in OG too so nothing new there. No worries if there's no fix to be found for this case.

Screenshot

Regardless, this is a massive improvement compared to how the wibble effect worked in OG, thank you. 👍

PR #1677, opened 2024-10-05
tr1/camera: fix potential division by zero

### LINKED ISSUE #1671: OG bug: camera clipping can cause a crash
This is very specific to certain level design and isn't an issue in the original levels AFAIK. In the attached, if you pull the block and hold look, the game crashes to desktop.

[LEVEL1.zip](https://github.com/user-attachments/files/17266252/LEVEL1.zip)

The pushblock triggers a flipmap, which has a wall where the block sits, but the pushblock is also deactivated at the same time.

lahm86:
Resolves #1671.

#### Checklist

#### Description

This avoids the crash in the custom level in the issue itself. The camera still shows the void briefly before returning to Lara but the check in place is clearly safer for this kind of situation.

### DIFF
--- a/src/tr1/game/camera.c
+++ b/src/tr1/game/camera.c
@@ -782,14 +782,21 @@ static void M_Clip(
     const int32_t target_y, const int32_t left, const int32_t top,
     const int32_t right, const int32_t bottom)
 {
+    const int32_t x_diff = *x - target_x;
+    const int32_t y_diff = *y - target_y;
+
     if ((right > left) != (target_x < left)) {
-        *y = target_y + (*y - target_y) * (left - target_x) / (*x - target_x);
+        if (x_diff) {
+            *y = target_y + (left - target_x) * y_diff / x_diff;
+        }
         *x = left;
     }
 
     if ((bottom > top && target_y > top && *y < top)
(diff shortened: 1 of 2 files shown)


src/tr1/game/camera.c
```diff
@@ -782,14 +782,21 @@ static void M_Clip(
     const int32_t target_y, const int32_t left, const int32_t top,
     const int32_t right, const int32_t bottom)
 {
+    const int32_t x_diff = *x - target_x;
+    const int32_t y_diff = *y - target_y;
+
     if ((right > left) != (target_x < left)) {
-        *y = target_y + (*y - target_y) * (left - target_x) / (*x - target_x);
+        if (x_diff) {
+            *y = target_y + (*y - target_y) * (left - target_x) / x_diff;
```
rr-: Can it be simplified to `*y = target_y + (left - target_x) * y_diff / x_diff`?

lahm86: Yes 😄

walkawayy (approved):

aredfan (approved):

rr- (approved):

PR #1683, opened 2024-10-07
misc: reconfigure UI toggling mechanism

lahm86:
_No issue raised as this is all unreleased so far. As this renames the previous general UI option, your current setting will be reset effectively._

#### Checklist

#### Description

This separates the general game UI state and photo mode UI state settings, and provides a mechanism for any future cases where specific UIs need to be togglable. There is perhaps a better place than the common UI module for the function, LMK what you think.


rr- (approved):
This is really elegant and clever. Thanks for coming up with it – I'd never have thought of this solution myself.

aredfan (approved):

walkawayy (approved):

PR #1690, opened 2024-10-08
Split input code into separate modules for each backend (keyboard and controller)

rr-:
#### Checklist

#### Description

This pull request aims to simplify the input API for easier integration into TR2 in the future.

It now separates the controller and keyboard-related code into their own modules for improved clarity. The new API no longer exposes backend details, like keyboard scancodes or controller buttons and axes. Instead, users can utilize opaque APIs like `Input_ReadAndAssignRole` or `Input_IsPressed` that only require the layout number and input roles. Notably, the control dialog's "unbind key" and "reset all keybindings" functions now use special `INPUT_ROLE`s designed for these actions. These roles cannot be rebound and do not trigger conflict checks, allowing the player to assign keys like backspace to other roles as well.

Importantly, the order of `Init()` calls for various modules has been rearranged. `Config_Read` now occurs right before the window is created and the output system is initialized. This ensures the config can reliably interact with most modules after their initialization, reducing most of the uncertainty.


aredfan (approved):

lahm86 (approved):
LGTM

PR #1699, opened 2024-10-09
Fix /give not working with weapons

### LINKED ISSUE #1698: TR2X bug – /give doesn't work with weapons
In TR1X this works because every level includes gun pickup sprites. In TR2, they are only loaded if the appear in the level, and the game otherwise deals with inventory items instead.

rr-:
#### Checklist

#### Description

Resolves #1698. Also explain why we missed 0.4.

### DIFF
--- a/docs/tr2/CHANGELOG.md
+++ b/docs/tr2/CHANGELOG.md
@@ -1,4 +1,5 @@
 ## [Unreleased](https://github.com/LostArtefacts/TRX/compare/tr2-0.5...develop) - ××××-××-××
+- fixed `/give` not working with weapons (regression from 0.5)
 
 ## [0.5](https://github.com/LostArtefacts/TRX/compare/afaf12a...tr2-0.5) - 2024-10-08
 - added `/sfx` command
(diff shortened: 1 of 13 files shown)


lahm86:
It doesn't seem to work in Floating Islands and Dragon's Lair, e.g.

```
play 17
give m16
```

Seems good elsewhere (HSH excluded, as it only has the shotgun anyway). Floater and DL don't have the sprites, so I think it might still be dependent on this.

rr-:
> It doesn't seem to work in Floating Islands and Dragon's Lair, e.g.
> 
> ```
> play 17
> give m16
> ```
> 
> Seems good elsewhere (HSH excluded, as it only has the shotgun anyway). Floater and DL don't have the sprites, so I think it might still be dependent on this.

Thank you for testing this. We previously had an extra `->loaded` sanity check in the matches evaluation, which became redundant with the introduction of `M_CanTargetObjectPickup`. Removing it resolved the issue.

lahm86 (approved):

PR #1719, opened 2024-10-15
tr1/carrier: add support for TR2-style item drops

### LINKED ISSUE #1713: Improve item drop support in line with TR2
The current approach for defining item drops in the gameflow is well-suited for the original levels, but for custom levels, managing this can be cumbersome. Item numbers change while levels are being developed and so with each change, the gameflow needs to be manually checked and revised.
Continue to offer the gameflow approach for OG, but add an option to enable TR2-style item drops, so that builders need only place the items in the level itself under enemies who should drop them.

lahm86:
Resolves #1713.

#### Checklist

#### Description

This allows item drops to be defined in the level data in the same fashion as TR2, so any pickup item that is in the same position as an enemy will be carried by that enemy. The gameflow approach is retained for OG levels, and the provided gameflow setting will default to this.

For OG, Pierre in Tihocan and the enemies in Natla's Mines should continue to drop as normal. If the gameflow setting is enabled in OG, then they will not drop anything (and indeed the ape in room 82 of Tihocan will drop shotgun shells).

Attached is a custom level with every enemy type and some drops allocated to each. Similarly, if the gameflow setting is disabled here, then the items will appear normally (albeit some will float as they're positioned with the enemy).
[LEVEL1.zip](https://github.com/user-attachments/files/17383340/LEVEL1.zip)

### DIFF
--- a/data/tr1/ship/cfg/TR1X_gameflow.json5
+++ b/data/tr1/ship/cfg/TR1X_gameflow.json5
@@ -22,6 +22,7 @@
         "data/injections/uzi_sfx.bin",
         "data/injections/explosion.bin",
     ],
+    "enable_tr2_item_drops": false,
     "convert_dropped_guns": false,
 
     "levels": [
--- a/data/tr1/ship/cfg/TR1X_gameflow_demo_pc.json5
+++ b/data/tr1/ship/cfg/TR1X_gameflow_demo_pc.json5
@@ -23,6 +23,7 @@
         "data/injections/uzi_sfx.bin",
         "data/injections/explosion.bin",
     ],
+    "enable_tr2_item_drops": false,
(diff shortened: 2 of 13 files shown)


rr- (approved):

aredfan (approved):
I found an issue when tr2 item drops are enabled. The Qualopec mummy doesn't drop the pickup when Lara bumps into it.

Apart from that, LGTM. 👍

lahm86:
> I found an issue when tr2 item drops are enabled. The Qualopec mummy doesn't drop the pickup when Lara bumps into it.
> 
> Apart from that, LGTM. 👍

Thank you, good catch. That should be fixed now.

aredfan (approved):

PR #1722, opened 2024-10-16
tr2/lara: revert camera after extra anim unless heavy

### LINKED ISSUE #1580: OG bug: cinematic camera issue when Lara strikes the gong | Ice Palace
Level 14, room 10.

_________________

To see the intended cine camera behaviour, execute `/fly` while Lara is striking the gong.

https://github.com/user-attachments/assets/9cc2af3b-be22-48cc-bdfd-8181f39f03c7

In OG, the cine camera only appears for a frame or so.

https://github.com/user-attachments/assets/1d2e3f99-a1a2-49d2-a821-890d4781e57d

__________________

Here's what Lahm wrote on discord:

> I think there must be something to do with checking the cam_type somewhere, but the call to test the triggers happens before the cine camera has finished so it doesn't trigger. Entering fly mode cancels the cinematic camera, so the heavy trigger works as was intended. Probably we can raise an issue and investigate properly, would be nice to fix

lahm86:
Resolves #1580.

#### Checklist

#### Description

The `Lara_State_Extra_Breath` function serves as a bridging animation between an extra animation and Lara returning to normal. This change will respect any heavy camera triggers set during the extra animation rather than blindly resetting to chase type. So as a result it fixes the gong camera in Ice Palace.

For testing, we can check WadTool where this could apply elsewhere. Any `Lara misc animation` that has the `next anim` set to `0` is relevant:

image

So this gives the following. The camera should return to normal after these animations.
- Bartoli detonator
- Offshore start
- Offshore wheel door
- Pulling the dagger (although slightly different, as Lara herself activates the exit door camera here)
- HSH start (same again, Lara triggers a camera after the start anim)

### DIFF
--- a/docs/tr2/CHANGELOG.md
+++ b/docs/tr2/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## [Unreleased](https://github.com/LostArtefacts/TRX/compare/tr2-0.5...develop) - ××××-××-××
 - fixed `/give` not working with weapons (regression from 0.5)
+- fixed the camera being cut off after using the gong hammer in Ice Palace (#1580)
 - improved switch object names
     - Switch Type 1 renamed to "Airlock Switch"
     - Switch Type 2 renamed to "Small Switch"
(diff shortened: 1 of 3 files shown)


aredfan (approved):
I've checked all the animations listed in the OP, all LGTM.

Thank you @lahm86. 👍

rr- (approved):

PR #1728, opened 2024-10-19
tr1/lara: add wading support

### LINKED ISSUE #1537: Feature request: Allow Lara to go from deep water to one-click water like in TR2+
Since one-click water support was added in #1005, it would be nice if the ability to move from deep water to one-click water was available for custom levels. From what I can tell, two animations would have to be injected: `UNDERWATER_TO_STAND` (192) and `ONWATER_TO_WADE_LOW` (193).

lahm86:
Resolves #1537.

#### Checklist

#### Description

This adds support for wading, taken from TR2, with the required animations and SFX provided in the updated `lara_animations.bin` injection.

Attached are a couple of test levels - TR1 and TR2 with identical geometry for easy comparison. Note there are no changes in this PR to the way TR2 Lara behaves.
[WadingTests.zip](https://github.com/user-attachments/files/17445698/WadingTests.zip)

This was quite involved and required some minor changes in such things as normal swim collision, so it'd be good to check OG swim behaviour as well. And there are a couple of locations at the end of Tihocan that Lara can now reach. I haven't made this optional as I'm not sure that we could efficiently separate out the logic, but LMK if you think it's needed and I'll take a deeper look.

One thing that came to light was the TR2 code has logic to make Lara bonk against a wall if she wades into it, but the frame checks there are wrong. I've corrected that for TR1, and can follow up with a fix for TR2 later.

This also highlights how we can fix another TR1 annoyance where Lara can get stuck against a wall (on land) if you bump into it but keep holding forward and a direction. In TR2 she deflects properly in the direction you're holding, the code for which is in `M_CollideStop` in `lara/col.c`. I'll tackle this separately too to provide the fix for TR1 and to make it optional there.


aredfan (approved):

src/tr1/game/lara/common.c
```diff
@@ -124,7 +165,15 @@ void Lara_Control(void)
             g_Lara.torso_rot.y = 0;
             Item_UpdateRoom(item, -LARA_HEIGHT / 2);
             Sound_Effect(SFX_LARA_BREATH, &item->pos, SPM_ALWAYS);
-        } else {
+        }
+    } break;
```
rr-: `break` should be placed inside curly braces.

src/tr1/game/lara/common.c
```diff
@@ -139,22 +188,69 @@ void Lara_Control(void)
...
+            item->gravity = 0;
+            item->fall_speed = 0;
+            g_Lara.dive_timer = 0;
+            g_Lara.torso_rot.y = 0;
+            g_Lara.torso_rot.x = 0;
+            g_Lara.head_rot.y = 0;
+            g_Lara.head_rot.x = 0;
+            Item_UpdateRoom(item, -LARA_HEIGHT / 2);
+        }
+    } break;
```
rr-: Ditto

rr-:
Are the now-reachable places in Tihocan important gameplay-wise?

src/tr1/game/lara/common.c
lahm86: Thanks, fixed.

lahm86:
> Are the now-reachable places in Tihocan important gameplay-wise?

I would say no, if anything it gives it a bit more polish as it avoids Lara getting embedded in the floor (although you can recover from that by swimming away). It maybe gives a slight advantage for killing the crocodile, but I think that's negligible as the ledge where you're supposed to exit the water is pretty close anyway.

Here is a comparison: https://youtu.be/gScRYMtgWno

lahm86:
One side effect from the swim col updates is that the fly cheat suffered from the same issue we originally had in TR2 (https://github.com/LostArtefacts/TR2X/issues/166). where Lara faces upwards. I've pushed a fix for that in line with https://github.com/LostArtefacts/TR2X/pull/167.

aredfan (approved):
> One side effect from the swim col updates is that the fly cheat suffered from the same issue we originally had in TR2 ([LostArtefacts/TR2X#166](https://github.com/LostArtefacts/TR2X/issues/166)). where Lara faces upwards. I've pushed a fix for that in line with [LostArtefacts/TR2X#167](https://github.com/LostArtefacts/TR2X/pull/167).

Good catch.

rr-:
> > Are the now-reachable places in Tihocan important gameplay-wise?
> 
> I would say no, if anything it gives it a bit more polish as it avoids Lara getting embedded in the floor (although you can recover from that by swimming away). It maybe gives a slight advantage for killing the crocodile, but I think that's negligible as the ledge where you're supposed to exit the water is pretty close anyway.
> 
> Here is a comparison: https://youtu.be/gScRYMtgWno

Thank you for the video. The bonk at 00:57 looked a bit awkward – I think Lara should just stop. The only similar place in TR2 I was able to find is Catacombs of Talion, room 21, tile 6/4/59 but it collides with a static rather than a ceiling slope. Can we confirm if this is in line with TR2 behavior?

lahm86:
> > > Are the now-reachable places in Tihocan important gameplay-wise?
> > 
> > 
> > I would say no, if anything it gives it a bit more polish as it avoids Lara getting embedded in the floor (although you can recover from that by swimming away). It maybe gives a slight advantage for killing the crocodile, but I think that's negligible as the ledge where you're supposed to exit the water is pretty close anyway.
> > Here is a comparison: https://youtu.be/gScRYMtgWno
> 
> Thank you for the video. The bonk at 00:57 looked a bit awkward – I think Lara should just stop. The only similar place in TR2 I was able to find is Catacombs of Talion, room 21, tile 6/4/59 but it collides with a static rather than a ceiling slope. Can we confirm if this is in line with TR2 behavior?

You know, thank bonk didn't even register with me 😄 

I found the issue, the collision check was wrong for this, so it's now fixed to check for `HT_WALL` in line with the same in `Lara_Col_Run`. Bonking while wading isn't a thing in TR2 because the anim frame checks that are there are wrong (and it's an OG issue, not anything from decompiling). My plan was to fix TR2 after this.

rr- (approved):

Richard-L:
I'm super confused by this. What is the point?

Lahm's video of Tihocan is incredibly awkward. She either sticks in the ground (this was OG behaviour??), or slides off ledges back into water which is also totally weird.

I'm also confused why this wouldn't be made optional. So far the focus has always been that an "absolutely vanilla" mechanic and physic of TR1 could still be maintained. This would now be a departure from that stance which I'd regret a lot. I understand that mixing the games has potential to reduce workload, but please let's not lose the truthfulness of every individual game along the way.

rr-:
> I'm super confused by this. What is the point?
> 
> Lahm's video of Tihocan is incredibly awkward. She either sticks in the ground (this was OG behaviour??), or slides off ledges back into water which is also totally weird.
> 
> I'm also confused why this wouldn't be made optional. So far the focus has always been that an "absolutely vanilla" mechanic and physic of TR1 could still be maintained. This would now be a departure from that stance which I'd regret a lot. I understand that mixing the games has potential to reduce workload, but please let's not lose the truthfulness of every individual game along the way.

I understand your concerns and appreciate your passion for maintaining the authenticity of TR1. Let me clarify a few points: The specific issue you're mentioning mainly affects Tomb of Tihocan, particularly the first three ledges, as there are no other places in the original game levels that would let Lara wade. In the original game, Lara's behavior was obviously glitched, which is highlighted in the first half of Lahm's video. The second half shows the improvements with new code.

The purpose of this update isn't just about fixing glitches but also giving custom level builders the flexibility to use more diverse geometries. Our goal is to support TR1 and TR2 levels with different engine features using a single exe. For example, in a future update it can be expected that we'll include flares code in TR1, though it won't be noticeable – simply because the original levels don't have flares. Your feedback is valuable, and I hope this helps explain our intentions.

EDIT: Upon discussing it internally a bit more, we'll look into making this feature optional for TR1 to maintain the OG visual experience and be in line with the optional nature of other OG glitch fixes.

lahm86:
This is now optional, so with the config option disabled, Lara will behave per OG.

src/tr1/game/lara/common.c
```diff
@@ -139,22 +192,71 @@ void Lara_Control(void)
...
             g_Lara.torso_rot.x = 0;
             g_Lara.torso_rot.y = 0;
+        } else {
+            g_Lara.water_status = LWS_WADE;
+            Item_SwitchToAnim(item, LA_STAND_IDLE, 0);
+            item->current_anim_state = LS_STOP;
+            item->goal_anim_state = LS_WADE;
+            Item_Animate(item);
+            item->fall_speed = 0;
         }
```
rr-, line 202: Can we flip this condition so that we test for the positive wading scenario first?

lahm86: Done 👍

aredfan:
- [x] If wading is disabled - Lara gets softlocked in Tomb of Tihocan 13/4/64 if you have her swim up the slope as far as possible. I'm not able to to reproduce this in the latest dev snapshot. [save_tr1_01.zip](https://github.com/user-attachments/files/17459646/save_tr1_01.zip)

lahm86:
> * [ ]  If wading is disabled - Lara gets softlocked in Tomb of Tihocan 13/4/64 if you have her swim up the slope as far as possible. I'm not able to to reproduce this in the latest dev snapshot. [save_tr1_01.zip](https://github.com/user-attachments/files/17459646/save_tr1_01.zip)

Thank you. This should be fixed now, but note that you won't be able to recover from that save - it shouldn't have allowed you to go so far up the slope and so Lara is effectively clamped. Latest push re-implements the original check for this scenario, meaning you shouldn't be able to get into that position in the first place.

Richard-L:
Thanks a lot for clarifying, and sorry if my message was a bit dramatic.

I tested this on PS1 and I see the troublesome OG behaviour. A fix for the mesh embed softlock itself is of course super welcome 🙏 

I'm glad this is now optional. I'm afraid it's very easy for us to miss a would-become-wade spot in TR1 and UB.

One unrelated question I had: when it says "your review was requested" on this, does that mean a comment alone is sufficient?

aredfan (approved):

rr-:
> One unrelated question I had: when it says "your review was requested" on this, does that mean a comment alone is sufficient?

Yes – that is correct. Thanks

PR #1733, opened 2024-10-22
Decompile skidoo

rr-:
#### Checklist

#### Description

#1691


src/tr2/decomp/skidoo.c
```diff
@@ -0,0 +1,1039 @@
...
+        fx->pos.x = skidoo->pos.x;
+        fx->pos.y = skidoo->pos.y;
+        fx->pos.z = skidoo->pos.z;
+        fx->speed = 0;
+        fx->frame_num = 0;
+        fx->counter = 0;
+        fx->object_id = O_EXPLOSION;
+    }
+
+    Effect_ExplodingDeath(g_Lara.skidoo, ~SKIDOO_GUN_MESH, 0);
```
lahm86: If either the red or black skidoo explodes, you see the main part of the red mesh appear. In OG, you only see "side bits" of the model fly off. Possibly the `mesh_bits` here aren't quite right for this call?

image

rr-: Thanks – the value indeed was wrong, rather than ~4 it should've been -4 or ~(4 - 1). I have pushed a correction, but I'm unable to test it at the moment and would appreciate info if it was enough to fix the issue.

rr-: BTW the name `SKIDOO_GUN_MESH` I think is wrong and should be changed – but I don't fully understand its usage. For sure it has to do with the track mesh, but I'm not certain as to why `track_mesh & 4` yields a truthy value for armed snowmobiles and a falsy value for normal snowmobiles.

lahm86: Hmm, I don't follow it either TBH. Will need some studying.

lahm86 (approved):
LGTM

aredfan:
There's a desync in demo 3 - the skidoo misses a small jump 3 seconds into the video.

**Latest dev snapshot:**

https://github.com/user-attachments/assets/d4105b98-2bae-48c2-8658-2a2a81992363

**PR:**

https://github.com/user-attachments/assets/e7feccf5-d3b2-401d-bd7f-b0c5f5cb3bf2

rr-:
@aredfan fixed

aredfan (approved):
> @aredfan fixed

Confirmed 👍

PR #1746, opened 2024-10-24
phase_pause: add the ability to pause during cutscenes

### LINKED ISSUE #1673: Feature request: allow pausing during cutscenes

walkawayy:
Resolves #1673.

#### Checklist

#### Description
Added the ability to pause during cutscenes.

This works as is, but I marked as a draft for now since it's kind of a proof of concept. It reuses the photo mode structures since that's already in place to resume cutscenes. But, I don't know if we want to rename the `PHASE_PHOTO_MODE_ARGS` struct to something more generic or possibly make a separate one for pausing. Lmk your thoughts.

### DIFF
--- a/docs/tr1/README.md
+++ b/docs/tr1/README.md
@@ -391,6 +391,7 @@ Not all options are turned on by default. Refer to `TR1X_ConfigTool.exe` for det
 - added ability to wade, similar to TR2+
 - added Lara's exit-water-to-medium-height animation from TR2+
 - added a pause screen
+- added the ability to pause during cutscenes
 - added a choice whether to play NG or NG+ without having to play the entire game
 - added Japanese mode (guns deal twice the damage, inspired by JP release of TR3); available for both NG and NG+
 - added ability to restart level on death
--- a/docs/tr1/README.md
+++ b/docs/tr1/README.md
@@ -391,7 +391,6 @@ Not all options are turned on by default. Refer to `TR1X_ConfigTool.exe` for det
 - added ability to wade, similar to TR2+
 - added Lara's exit-water-to-medium-height animation from TR2+
 - added a pause screen
-- added the ability to pause during cutscenes
 - added a choice whether to play NG or NG+ without having to play the entire game
 - added Japanese mode (guns deal twice the damage, inspired by JP release of TR3); available for both NG and NG+
 - added ability to restart level on death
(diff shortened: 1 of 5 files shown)


rr-:
> Resolves #1673.
> 
> #### Checklist
> * [x]  I have read the [coding conventions](https://github.com/LostArtefacts/TR1X/blob/master/CONTRIBUTING.md#coding-conventions)
> * [x]  I have added a changelog entry about what my pull request accomplishes, or it is an internal change
> 
> #### Description
> Added the ability to pause during cutscenes.
> 
> This works as is, but I marked as a draft for now since it's kind of a proof of concept. It reuses the photo mode structures since that's already in place to resume cutscenes. But, I don't know if we want to rename the `PHASE_PHOTO_MODE_ARGS` struct to something more generic or possibly make a separate one for pausing. Lmk your thoughts.

Thanks. I'd copy the struct and avoid reusing it. I think it's small enough to allow a bit of duplication at the moment.

Richard-L:
Sorry to just ask in here without testing fully myself but two things I thought about regarding this:
- Does it behave like a normal pause, ie fade to transparent black, show "paused" at the bottom.
- Does the audio fade out/in too on pause toggle (I believe this is the case for when we use pause in-game, right?)

rr-:
> Sorry to just ask in here without testing fully myself but two things I thought about regarding this:
> 
> * Does it behave like a normal pause, ie fade to transparent black, show "paused" at the bottom.
> * Does the audio fade out/in too on pause toggle (I believe this is the case for when we use pause in-game, right?)

1. Yep, it behaves like a normal pause with the fade and "paused" message.
2. Nope, the audio doesn't fade in and out. We don't have audio fading anywhere in the game.

docs/tr1/README.md
```diff
@@ -391,6 +391,7 @@ Not all options are turned on by default. Refer to `TR1X_ConfigTool.exe` for det
 - added ability to wade, similar to TR2+
 - added Lara's exit-water-to-medium-height animation from TR2+
 - added a pause screen
+- added the ability to pause during cutscenes
```
rr-: I think it kind of duplicates the "added a pause screen" feature – no need to mention a specialized case as a separate feature IMO

rr- (approved):

docs/tr1/README.md
walkawayy: True but we don't have pausing in FMVs yet. So it's not a universal
Pause at the moment.

rr-: We usually use the changelog to capture version-to-version updates, whereas this document focuses more on the major differences from the original game. Since the original game didn't have a pause mode altogether, my thinking is that we don't need to include this update here.

lahm86 (approved):

aredfan (approved):

PR #1759, opened 2024-10-26
dragon: fix dragon reviving after the dagger is pulled

### LINKED ISSUE #1572: OG Bug: Dragon can live even after pulling the dagger
Maybe if you pull the dagger right when the fall down state ends? Video:

https://clips.twitch.tv/OptimisticLuckyWaspNinjaGrumpy-ihBNzN8AxFoD3v_E

walkawayy:
Resolves #1572.

#### Checklist

#### Description
Fixed the dragon reviving itself after Lara removes the dagger in a 1 frame window.

Explanation:
Dragon_Collision:
`item->frame_num - g_Anims[item->anim_num].frame_base <= DRAGON_ALMOST_LIVE // 100`

Dragon_Control:
```
creature->flags++;
if (creature->flags == DRAGON_LIVE_TIME) { // 330
    dragon_front_item->goal_anim_state = DRAGON_ANIM_STOP;
}

if (creature->flags >= DRAGON_LIVE_TIME + DRAGON_ALMOST_LIVE) { // 330 + 100
    dragon_front_item->hit_points =
        g_Objects[O_DRAGON_FRONT].hit_points / 2;
}
```

So what was happening simultaneously was:
1. `Dragon_Control` runs and `creature->flags` increments to `430` which is `>= DRAGON_LIVE_TIME + DRAGON_ALMOST_LIVE` (330+100). 
2. `Dragon_Collision` runs the same frame and `item->frame_num - g_Anims[item->anim_num].frame_base` is `100` which is <=`DRAGON_ALMOST_LIVE` (100) so `M_PullDagger` runs .

So, `M_PullDagger` runs, but that really only prepares the dragon for death by setting `creature->flags = -1;`. However, in `Dragon_Control` already, the hit points were reset to `300/2` because the `creature->flags` hit `430`. So, the `else... then if (dragon_front_item->hit_points <= 0) {` block is no longer run which checks for` if (creature->flags > -20) {`  when the dragon has no HP and kills the dragon.

So, the fix is to change in `Dragon_Control`:
`if (creature->flags >= DRAGON_LIVE_TIME + DRAGON_ALMOST_LIVE) { // 330 + 100`
to
`if (creature->flags > DRAGON_LIVE_TIME + DRAGON_ALMOST_LIVE) { // 330 + 100`
that way we don't lose 1 frame in the window to pull the dagger and kill the dragon. I cannot screenshot or capture video of the old DirectX SW rendered window for some reason.

### DIFF
--- a/src/tr2/game/objects/creatures/dragon.c
+++ b/src/tr2/game/objects/creatures/dragon.c
@@ -33,9 +33,6 @@
 #define DRAGON_LIVE_TIME 330
 #define DRAGON_HITPOINTS 300
 #define DRAGON_RADIUS (WALL_L / 3) // = 341
-#define DRAGON_ANIM_DIE 21
-#define DRAGON_ANIM_DEAD 22
-#define DRAGON_ANIM_RESURRECT 23
 
 typedef enum {
     // clang-format off
@@ -51,6 +48,9 @@ typedef enum {
     DRAGON_ANIM_SWIPE_LEFT  = 9,
     DRAGON_ANIM_SWIPE_RIGHT = 10,
     DRAGON_ANIM_DEATH       = 11,
+    DRAGON_ANIM_DIE         = 21,
+    DRAGON_ANIM_DEAD        = 22,
+    DRAGON_ANIM_RESURRECT   = 23,
     // clang-format on
 } DRAGON_ANIM;
 
--- a/src/tr2/game/objects/creatures/dragon.c
+++ b/src/tr2/game/objects/creatures/dragon.c
@@ -33,6 +33,9 @@
 #define DRAGON_LIVE_TIME 330
 #define DRAGON_HITPOINTS 300
 #define DRAGON_RADIUS (WALL_L / 3) // = 341
+#define DRAGON_ANIM_DIE 21
+#define DRAGON_ANIM_DEAD 22
+#define DRAGON_ANIM_RESURRECT 23
 
 typedef enum {
     // clang-format off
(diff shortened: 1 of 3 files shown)


src/tr2/game/objects/creatures/dragon.c
```diff
@@ -48,9 +51,6 @@ typedef enum {
     DRAGON_ANIM_SWIPE_LEFT  = 9,
     DRAGON_ANIM_SWIPE_RIGHT = 10,
     DRAGON_ANIM_DEATH       = 11,
-    DRAGON_ANIM_DIE         = 21,
-    DRAGON_ANIM_DEAD        = 22,
-    DRAGON_ANIM_RESURRECT   = 23,
     // clang-format on
 } DRAGON_ANIM;
```
lahm86: I suppose this enum should be renamed `DRAGON_STATE` as that's what it's matching up to rather than anim index (I was wondering why you had split out the anim numbers, but it makes sense). Maybe a separate task though with other objects if they're similar.

lahm86 (approved):
LGTM. Recording below catching the dagger on frame 100 and missing it on frame 101.

https://youtu.be/q0Bt4J_n-9Y

DLL with frame display attached too.
[TR2X.zip](https://github.com/user-attachments/files/17531709/TR2X.zip)

src/tr2/game/objects/creatures/dragon.c
walkawayy: Yes I just moved these out because these are animation numbers. It seems like even in TR1, the enums are all called `X_ANIM` even though they are technically animation states. And the `#defines` include the few animation numbers that might be hard coded. We could probably use a large rename in the future to your suggestion or `X_ANIM_STATE` or something.

aredfan (approved):

src/tr2/game/objects/creatures/dragon.c
rr-: Let's either refrain from this change or make a DRAGON_STATE and DRAGON_ANIM enums.

walkawayy: Ok I'll just kick the can down the line since this is going to require a rename of every object.

PR #1768, opened 2024-10-28
Switch DirectInput to SDL

### LINKED ISSUE #1695: Switch DirectInput to SDL

rr-:
#### Checklist

#### Description

Resolves #1695. Can cause regressions in TR1.

I still need to add an UI for the controller settings, but other than that, the code for the entire thing is out there.


walkawayy (changes requested):
Everything LGTM in TR1.

In TR2, everything seems to work except there seems to be a bug with the console. Try doing:
1. New game or `play 1`
2. Open inventory and go the controls menu. You don't even have to change anything.
3. Exit the controls menu and go back to the game.
4. Open the console.

You should now be stuck in a state where no inputs register except oddly up and down to scroll past console commands. You can't type, enter doesn't work, etc. You have to force quit the game.

`I still need to add an UI for the controller settings, but other than that, the code for the entire thing is out there.`
Also, my PS5 controller did not work in TR2. Idk if it should be or not? I reading this as the UI isn't done but the functionality should be working. LMK if I read it wrong.

walkawayy:
```
[258/269] Compiling C object TR2X.dll.p/game_ui_controllers_controls.c.obj
../../../src/tr2/game/ui/controllers/controls.c: In function ‘M_NavigateInputsDebounce’:
../../../src/tr2/game/ui/controllers/controls.c:118:5: warning: implicit declaration of function ‘Shell_ProcessEvents’ [-Wimplicit-function-declaration]
  118 |     Shell_ProcessEvents();
```

Getting this build issue. I think `controls.c` needs `shell.h` included.

rr-:
> ```
> [258/269] Compiling C object TR2X.dll.p/game_ui_controllers_controls.c.obj
> ../../../src/tr2/game/ui/controllers/controls.c: In function ‘M_NavigateInputsDebounce’:
> ../../../src/tr2/game/ui/controllers/controls.c:118:5: warning: implicit declaration of function ‘Shell_ProcessEvents’ [-Wimplicit-function-declaration]
>   118 |     Shell_ProcessEvents();
> ```
> 
> Getting this build issue. I think `controls.c` needs `shell.h` included.

Thanks. Rebased + fixed.

walkawayy:
Nice the console soft lock is fixed.

Two things:
1. Backspace in the console is deleting two characters instead of 1.
2. This may be a separate PR, but we talked about possibly changing the flare key from period `.` to comma `,` to match TR3/4/5. TR2 used the slash `/` key which we now use for console.

walkawayy:
Controller is working well on W11.

Stella's TR2 site:
https://tombraiders.net/stella/walks/controls/TR2controls.html

Right now, the controller's L2/R2 buttons are working like TR1X. L2 is sidestep left and R2 is sidestep right. Stella says for PS1 TR2, R2 + D-pad left/right is sidestep, and L2 is flare.

rr-:
@walkawayy updated the UI to offer editing controller layouts. It suffers from bad text overlaps due to the two-column layout and lengthy controller key names. I'd like to tackle this before merging as it feels unfinished as it is. At the same time, I'd like to keep the current two-column UI layout to avoid too many changes as this PR is already too big. Can you help me shorten the names in `libtrx/game/input/backends/controller.c`? CC @lahm86 
I believe we can use at most 5 characters. Also, with the injection framework, we could inject new sprites, even non-existing ones.

Richard-L:
There are 4 control schemes available in PS1 TR2, of which one is the one you quoted above. If you use the scheme where sidesteps are still available from both shoulder buttons, you are forced to use flares from the inventory. 

TR3 remedied this by introducing a shortcut of WALK+SELECT to use a flare directly, while maintaining shoulder button sidesteps. I think it'd be a worthwhile addition to TRX.

rr-:
> There are 4 control schemes available in PS1 TR2, of which one is the one you quoted above. If you use the scheme where sidesteps are still available from both shoulder buttons, you are forced to use flares from the inventory.
> 
> TR3 remedied this by introducing a shortcut of WALK+SELECT to use a flare directly, while maintaining shoulder button sidesteps. I think it'd be a worthwhile addition to TRX.

Correct – my plan is to add TR3 sidesteps in a follow-up PR as this one is already massive.

rr-:
> Nice the console soft lock is fixed.
> 
> Two things:
> 
> 1. Backspace in the console is deleting two characters instead of 1.
> 2. This may be a separate PR, but we talked about possibly changing the flare key from period `.` to comma `,` to match TR3/4/5. TR2 used the slash `/` key which we now use for console.

1 – Thanks – fixed.
2 – This PR already switches flares to `,` as per our internal conversation on Discord.

rr-:
Alright – I've shortened the names. While they're not the prettiest, we're eventually going to inject icons for the other buttons anyway.

aredfan (approved):

walkawayy (approved):

src/libtrx/include/libtrx/game/input/state.h
```diff
@@ -0,0 +1,82 @@
...
+        uint64_t equip_uzis : 1;
+#if TR_VERSION == 2
+        uint64_t equip_harpoon : 1;
+        uint64_t equip_m16 : 1;
+        uint64_t equip_grenade_launcher : 1;
+#endif
+        uint64_t use_small_medi : 1;
+        uint64_t use_big_medi : 1;
+#if TR_VERSION == 2
+        uint64_t use_flare : 1;
```
lahm86, line 53: The Venice demo doesn't play out as expected: Lara doesn't draw a flare at the beginning, and once she goes through the window, her movement is wrong (she hits the wall and dies because the goon has free rein).
I am assuming this will need to remain in the original bit position (`1 << 19` I believe) otherwise the demo data in the level would need to be updated. Whether this affects her movement too, I'm not sure - maybe because it's not registering, other inputs are running sooner?

lahm86: The Lost Valley demo also has an issue - Lara doesn't exit the water.
Vilcabamba, Maria Doria, and Tibet demos look OK as far as I can tell.

rr-: @lahm86 fixed

lahm86: TR2 looks good. The Lost Valley issue remains, but I've just checked develop and the problem is there too. I'll raise that as a separate issue and work out when it was introduced.

lahm86 (approved):

PR #1815, opened 2024-11-02
tr1/savegame: save and load ambient music timestamp

### LINKED ISSUE #1769: TR1X bug: environment BGM is not saved
I'm not sure if it's a bug or a feature, but the fix is simple:

```diff
diff --git a/src/tr1/game/savegame/savegame_bson.c b/src/tr1/game/savegame/savegame_bson.c
index bc06d1f3..36572229 100644
--- a/src/tr1/game/savegame/savegame_bson.c
+++ b/src/tr1/game/savegame/savegame_bson.c
@@ -1248,7 +1248,7 @@ static JSON_OBJECT *M_DumpCurrentMusic(void)
 {
     JSON_OBJECT *current_music_obj = JSON_ObjectNew();
     JSON_ObjectAppendInt(
-        current_music_obj, "current_track", Music_GetCurrentTrack());
+        current_music_obj, "current_track", Music_GetCurrentPlayingTrack());
     JSON_ObjectAppendDouble(
         current_music_obj, "timestamp", Music_GetTimestamp());
 
```

lahm86:
Resolves #1769.

#### Checklist

#### Description

This will save the current music track and a flag to indicate whether or not it's looped, so that on load, the correct start-up for it is applied. If this isn't stored, the ambient will start correctly at the timestamp on load, but it won't loop.

NB: the diff in the issue slipped into [this commit](https://github.com/LostArtefacts/TRX/commit/f8efbc90057d380949d5f7f99bd9bdabd3739dcd#diff-1f45c5bd7b4a0e0855f15181cdc479302417ed40b8d3cd04eb9f246cac065d91), but as this feature/fix needed a little extra, and said commit is currently unreleased, I figured it would be OK to just proceed and update it here. But LMK if it's preferred to adjust the history.


rr- (commented):
I'm sorry I didn't provide enough details initially. I find that resuming ambient tracks can be more distracting than one-shot tracks. Could we make loading them optional by turning the `load_current_music` option into an enum like this?

```c
typedef enum {
    // equivalent of .load_current_music = false in the current implementation
    MUSIC_LOAD_NEVER,

    // equivalent of .load_current_music = true in the current implementation
    MUSIC_LOAD_NOT_AMBIENT,

    // behavior introduced in this PR
    MUSIC_LOAD_ALWAYS,
} MUSIC_LOAD_CONDITION;
```

I think it's acceptable not to have backwards compatibility with the previous user configs for this option. But ideally, I'd name the enum-based option something new, and look for the old values in `M_LoadLegacyOptions` in `src/tr1/config.c` (reader alone).

Thank you!

lahm86:
> I'm sorry I didn't provide enough details initially. I find that resuming ambient tracks can be more distracting than one-shot tracks. Could we make loading them optional by turning the `load_current_music` option into an enum like this?
> 
> ```c
> typedef enum {
>     // equivalent of .load_current_music = false in the current implementation
>     MUSIC_LOAD_NEVER,
> 
>     // equivalent of .load_current_music = true in the current implementation
>     MUSIC_LOAD_NOT_AMBIENT,
> 
>     // behavior introduced in this PR
>     MUSIC_LOAD_ALWAYS,
> } MUSIC_LOAD_CONDITION;
> ```
> 
> I think it's acceptable not to have backwards compatibility with the previous user configs for this option. But ideally, I'd name the enum-based option something new, and look for the old values in `M_LoadLegacyOptions` in `src/tr1/config.c` (reader alone).
> 
> Thank you!

No problem, I'll update this today 👍

lahm86:
Should be good for checking/testing now.

aredfan (approved):

rr- (approved):
LGTM. One cosmetic remark – is not ambient correct, or would non-ambient sound more grammatically correct in this case? Thanks

lahm86:
Yes, non-ambient sounds better. I've changed this throughout so everything matches.

rr- (approved):

PR #1820, opened 2024-11-03
tr1/objects/scion3: save scion hitpoints

### LINKED ISSUE #1819: OG bug: possible to increment kills when destroying the scion
When Lara shoots the scion - save while the explosion sprites are present.  Load the savegame and Lara remains locked on the (now invisible) scion and has another opportunity to destroy it again.

Lara staying locked on the scion is an OG bug present in TombATI. I've not confirmed if this affects the kill stats but it's highly likely.

https://github.com/user-attachments/assets/589929d2-f64e-4b23-8e29-41ea7835403a

lahm86:
Resolves #1819.

#### Checklist

#### Description

This ensures the shootable scion's hitpoints value is saved so if it's still active on reload, Lara can't continue to target it.
Existing saves would continue to have this bug, so it's a going-forward fix.

### DIFF
--- a/src/tr1/game/objects/general/scion3.c
+++ b/src/tr1/game/objects/general/scion3.c
@@ -12,6 +12,7 @@ void Scion3_Setup(OBJECT *obj)
     obj->control = Scion3_Control;
     obj->hit_points = 5;
     obj->save_flags = 1;
+    obj->save_hitpoints = 1;
 }
 
 void Scion3_Control(int16_t item_num)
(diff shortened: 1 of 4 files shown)


aredfan (approved):
This bugfix is working perfectly, thank you. 👍

src/tr1/game/objects/general/scion3.c
```diff
@@ -12,6 +12,7 @@ void Scion3_Setup(OBJECT *obj)
     obj->control = Scion3_Control;
     obj->hit_points = 5;
     obj->save_flags = 1;
+    obj->save_hitpoints = 1;
```
rr-, line 15: Oh no – we can't do it like this, or else it'll break legacy saves. We need to force disable saving the hitpoints in the legacy reader.

Should be easy to test – just grab a TombATI save from that level, try to load it with TRX containing the fix from this PR, and it shouldn't crash.

lahm86: Good catch, thanks!
Pushed a fix, and test saves attached. This loads OK (and can confirm it certainly did crash before fixing this).

[saveati.zip](https://github.com/user-attachments/files/17611740/saveati.zip)

aredfan (approved):

rr- (approved):

PR #1826, opened 2024-11-04
tr1/lara: add responsive swim cancel option

### LINKED ISSUE #1004: Feature request: Add option for TR2 cancel swimming animation
Requested by RetroRaider279 on discord.

lahm86:
Resolves #1004.

#### Checklist

#### Description

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/7

This injects additional state changes to the swimming animation to allow Lara to enter the glide state on more frame ranges, similar to TR2+.

image

I didn't add a README entry as it's pretty much already covered with the mention of TR2 swimming. I did keep it as a separate option though, so players can mix and match with OG and TR2 swimming, but LMK if we think this is config-bloat.


rr- (commented):
What are the chances we'll run into a conflict with frame numbers between two responsive states? My intuition is to create a separate state for both anims, but I'm not confident about it since I don't fully understand the ins and outs of this solution.

lahm86:
> What are the chances we'll run into a conflict with frame numbers between two responsive states? My intuition is to create a separate state for both anims, but I'm not confident about it since I don't fully understand the ins and outs of this solution.

We should be good because we have separated the states. So for OG, only the first two state changes apply as 18 will be Lara's goal anim state. When enabled, the goal state is 60 so the ranges for 18 are ignored (but we've duplicated those anyway into 60, so they'll still be captured that way).

The three extra new anims that state 60 links to are interim anims that precede the underwater idle anim (108). These have state 18, so that becomes Lara's actual state when she reaches these anims, then everything else plays out normally.

image

This approach overall means we can retain a single source of truth for Lara's extended anims (similar to #1824). Otherwise, we'd have to create a separate injection file to overwrite the state changes in OG, something I want to steer away from in general.

rr- (approved):

aredfan (approved):

PR #1827, opened 2024-11-04
tr2/overlay: support 3d pickup animations

### LINKED ISSUE #1633: Add 3D PickUp animations
Cf TR1X

rr-:
#### Checklist

#### Description

Resolves #1633. The dragon statuettes are kept as sprites for now.

### DIFF
--- a/docs/tr1/README.md
+++ b/docs/tr1/README.md
@@ -468,6 +468,7 @@ Not all options are turned on by default. Refer to `TR1X_ConfigTool.exe` for det
 - added a cheat to explode Lara like in TR2 and TR3
 
 #### Input
+- added ability to hold forward/back to move through menus more quickly
 - added ability to move camera around with W,A,S,D
 - added additional custom control schemes
 - added the ability to unbind unessential keys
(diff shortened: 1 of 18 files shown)


rr-:
I'm not clearing the Z-Buffer as currently there's no API for this. As a consequence, there may be some clipping issues with certain geometries / positions, but I haven't been able to stumble upon this.

lahm86 (commented):
Looks great!
There is some clipping as you mention, such as the shotgun and the key cards in Offshore Rig.
The only major issue I've noticed is the prayer wheels (puzzle1) in Barkhang, they don't appear on screen at all. But looking into the data, it seems they have no animations/frames, so I suppose behaviour for that is pretty much undefined. I had no idea this was an issue in the game. We will be able to fix this with an injection eventually, but I wonder if there is a workaround for that pickup in the meantime?

rr-:
@lahm86 Fixed by moving away from viewport manipulation, opting for a different rendering approach (basically, we're just moving the viewport center point now). Additionally, I've added z-buffer clearing. While the results are acceptable, there are some sizing issues left that we'll need to address with injections. For example, the prayer wheels remain too large, and the dragon seal seems disproportionately small, among others.

On a related note, the bounding box for the 3D model of the prayer wheel appears to be broken (X and Y planes are reported to be 0). Since the bounding boxes are essential for rendering of 3D items, I've implemented a manual calculation for it. The function perhaps isn't ideally placed, but it's there if we need it later. Plus, it can help us establish the expected bounding box values to be injected once we get there.

walkawayy:
I just tried to test this on the latest push (f10873a) and my game crashes when the game tries to load any level. Savegame, new game, or console.

walkawayy (approved):
LGTM! I tried to test every key item in every level. I didn't see any issues.

aredfan (approved):

lahm86 (approved):

PR #1854, opened 2024-11-09
libtrx/config: support enforced config settings

### LINKED ISSUE #1846: Separate Build for TRX and Custom Levels?
As a level designer, I have a clear vision of how my level should be played and do not want some options of the global ConfigTool to be applied to my level.

Would it be possible to provide two different binaries? One for TR1X - TR1 Community edition and one specifically for custom levels where all the config is read from gameflow.json?

### LINKED ISSUE #1847: Removing Animation injection and enabling underwater roll results in sliding when pressing roll
When disabling animation injections but keep the option for underwater roll enabled, you can hold ROLL underwater to make lara slide at a steady rate

https://github.com/user-attachments/assets/80d721dc-83f8-4ba7-8bb4-ebf92f0d367a

lahm86:
Resolves #1846.
Resolves #1847.

#### Checklist

#### Description

This allows an enforced object to be defined in the config file, within which any regular config setting can be defined, and the values from here will be enforced in the game. This applies to TR1 and TR2.

Example snippet for `TR1X.json5`:
```json
{
  "enable_uw_roll" : true,
  "enforced" : {
    "enable_uw_roll" : false
  }
}
```
Here the initial value is replaced by the one in the `enforced` object when the game loads. When the game writes back the config, both entries will be preserved, meaning if players want to remove these enforcements they can do so and their original settings will not need to be re-edited.

The config tool will grey out any enforced options, and display a message so the user knows why they cannot be edited.

I think we can remove `force_game_modes` and `force_save_crystals` from the gameflow and allow builders to implement it in the same way as above. But probably best handled in its own PR and after review/testing of this one.

Re #1847, this same approach can be used by builders to remove features available only with Lara's animation injection. The alternative to this is testing animation goals, which IMO would result in too many additional code checks.

### DIFF
--- a/docs/tr1/CHANGELOG.md
+++ b/docs/tr1/CHANGELOG.md
@@ -3,6 +3,7 @@
 - added the ability to pause during cutscenes (#1673)
 - added an option to enable responsive swim cancellation, similar to TR2+ (#1004)
 - added a special target, "pickup", to item-based console commands
+- added support for custom levels to enforce values for any config setting (#1846)
 - changed OpenGL backend to use version 3.3, with fallback to 2.1 if initialization fails (#1738)
 - changed text backend to accept named sequences. Currently supported sequences (limited by the sprites available in OG):
     - `\{umlaut}`
--- a/docs/tr1/GAMEFLOW.md
+++ b/docs/tr1/GAMEFLOW.md
@@ -10,6 +10,7 @@ Jump to:
 - [Bonus levels](#bonus-levels)
 - [Item drops](#item-drops)
 - [Injections](#injections)
+- [User configuration](#user-configuration)
 
 ## Global properties
 The following properties are in the root of the gameflow document and control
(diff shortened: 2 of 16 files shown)


aredfan (approved):

walkawayy (changes requested):
I think everything looks good and works well!

My one suggestion is to change how the config tool shows the enforced option. Right now, the tool shows the user's version of the option greyed out. However, I think the tool should show the enforced option greyed out. For example, using your example setting from the PR force disables the underwater roll:
```
  "enforced" : {
    "enable_uw_roll" : false
  }
```

That enforced option is properly enforced in game so Lara can't roll. But, the config tool's greyed out option makes it seem like the author force enabled the underwater roll as seen here:

image

This is confusing because the player would assume the underwater roll is enforced on. I think the opposite can also occur if the user has an option disabled, but the author force enabled it. The player would think it's off but it's actually always on.

lahm86:
Pushed an update @walkawayy. Unfortunately, because of the abstraction in how the properties are managed, I couldn't think of a good way to store the enforced value but retain writing the actual value back when needed. But I agree it's clearer to show what's actually in place, so let me know if you think the update is OK. It just shows the raw enforced value along with a message. It's clear there is no UI input as such.

walkawayy:
Hm it still seems to show my user config option instead of the forced option:

image

I have:

```
{
  "enable_uw_roll" : true,
  "enforced" : {
    "enable_uw_roll" : false
  }
}
```

lahm86:
Should show like this - maybe you're running the wrong download?
image

walkawayy:
Oh oops I didn't notice I had to do a git pull with a rebase. My normal pull failed and I wasn't paying attention. Yes it works perfectly now.

walkawayy (approved):

rr- (commented):
Looks good but we need to describe this mechanism in the gameflow docs for it to be adopted.

lahm86:
Updated `GAMEFLOW.md` with some details on this. I've used save crystals as an example, although I'd still like to remove the other dedicated options for enforcing that and game modes in a separate PR. I'll update the gameflow doc again there as we mention those forced variables at the top of that document.

rr- (approved):

PR #1858, opened 2024-11-10
libtrx/config: refactor enforced config approach

### LINKED ISSUE #1857: Remove gameflow forced variables
Following #1854, `force_game_modes` and `force_save_crystals` are no longer required in the gameflow as these can be enforced in the config JSON instead by builders.

lahm86:
Resolves #1857.

#### Checklist

#### Description

~~This removes the dedicated enforced game modes and save crystals settings from the gameflow files, as these can now be controlled by builders in the main config file. The only exception is the UB gameflow, as we need to continue enforcing save crystals to be disabled there, and this can't be done through the config file. As a result, backwards compatibility is in place as standard with this, while the documentation points to the correct procedure for enforcing settings.~~

~~LMK if you think we can remove backwards compatibility for `force_game_modes`, I'm inclined to perhaps leave it for a while.~~

This refactors the enforced config control from #1854 into the gameflow and removes the legacy settings for enforcing game modes and save crystals.

### DIFF
--- a/src/libtrx/config/file.c
+++ b/src/libtrx/config/file.c
@@ -5,16 +5,20 @@
 #include "log.h"
 #include "memory.h"
 
+#include <assert.h>
 #include <string.h>
 
-#define ENFORCED_KEY "enforced"
+#define EMPTY_ROOT "{}"
+#define ENFORCED_KEY "enforced_config"
 
 static bool M_ReadFromJSON(
-    const char *json, void (*load)(JSON_OBJECT *root_obj));
+    const char *def_json, const char *enf_json,
+    void (*load)(JSON_OBJECT *root_obj));
 static void M_PreserveEnforcedState(
-    JSON_OBJECT *root_obj, JSON_VALUE *old_root);
+    JSON_OBJECT *root_obj, JSON_VALUE *old_root, JSON_VALUE *enf_root);
 static char *M_WriteToJSON(
-    void (*dump)(JSON_OBJECT *root_obj), const char *old_data);
+    void (*dump)(JSON_OBJECT *root_obj), const char *old_data,
+    const char *enf_data);
 static const char *M_ResolveOptionName(const char *option_name);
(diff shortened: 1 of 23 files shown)


rr-:
What if we try to set the enforced config options in the gameflow file instead? Probably a tough cookie, as we read that file much later. Maybe a separate file such as TR1X-lock.json (or less clumsy naming)? That way the files are free to be shipped and override existing installations, which removes the burden of maintaining backwards compatibility with existing user setups.

lahm86:
Ok, I think it's doable and I think it will be nicer in the long-run. If it wasn't for the UB snag we'd be home and dry pretty much 😄 

No need to test just yet @walkawayy, @aredfan - I'll let you know when it's updated and ready.

lahm86:
Ok, we should be good for testing this now. The gameflow now accepts the `enforced_config` object as opposed to the main config file. We do this by reading the gameflow at the same time as the config and merging the settings together. The gameflow is then read again later in the load process, and the enforced object is ignored at that point.

The config tool will continue to show where settings are enforced, as it also reads both files now.

src/libtrx/config/file.c
```diff
@@ -165,6 +193,9 @@ bool ConfigFile_Write(const char *path, void (*dump)(JSON_OBJECT *root_obj))
     if (old_data != NULL) {
         Memory_FreePointer(&old_data);
     }
+    if (enforced_data != NULL) {
+        Memory_FreePointer(&enforced_data);
+    }
```
rr-: Memory_Free family already checks for nulls so no need for this double safety.

rr- (approved):

src/libtrx/config/file.c
lahm86: Thanks, updated.

rr- (approved):

aredfan (approved):

PR #1864, opened 2024-11-11
libtrx/console: fix play_level command action

### LINKED ISSUE #1861: TR1X bug: resume info is not set when using console to jump to a level
For example, from the title screen, run `/play 2` - Lara is missing her pistols.

Another issue is:
- `/set disable-healing on` to disable healing between levels
- `/play 2`
 
Lara dies immediately when the level loads.

Regression from 4.5.

lahm86:
Resolves #1861.

#### Checklist

#### Description

This sets the action for `/play` commands to `GF_SELECT_GAME` rather than `GF_START_GAME`, so that resume info can be properly calculated. While TR2 wasn't affected, it should also be tested here as it's impacted by the change.

### DIFF
--- a/docs/tr1/CHANGELOG.md
+++ b/docs/tr1/CHANGELOG.md
@@ -53,6 +53,7 @@
 - fixed being able to shoot the scion multiple times if save/load is used while it blows up (#1819)
 - fixed certain erroneous `/play` invocations resulting in duplicated error messages
 - fixed Lara getting stuck in her hit animation if she is hit while mounting the boat or skidoo (#1606)
+- fixed the `/play` console command resulting in Lara starting the target level without pistols (#1861, regression from 4.5)
 - improved enemy item drops by supporting the TR2+ approach of having drops defined in level data (#1713)
 - improved Italian localization for the Config Tool
 - improved the injection approach for Lara's responsive jumping (#1823)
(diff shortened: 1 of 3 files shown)


rr- (approved):

rr-:
>While TR2 wasn't affected, it should also be tested here as it's impacted by the change.

Actually I think TR2 suffers from the same problem. In the GW, the player starts off with pistols, shotgun, flares and one of each medpacks, while `/play 2` only gives pistols.

lahm86:
> > While TR2 wasn't affected, it should also be tested here as it's impacted by the change.
> 
> Actually I think TR2 suffers from the same problem. In the GW, the player starts off with pistols, shotgun, flares and one of each medpacks, while `/play 2` only gives pistols.

I think somewhere TR2 has it hard-coded to add the shotgun, flares and medis on the first level (not decompiled yet, I think). The original gameflow does not have the usual "give item" commands in Great Wall. I know builders work around this by adding a "remove guns" command followed by "give pistols". So I think in TR2 when we do a `/play` and the level is beyond 1, it's only checking that the gameflow doesn't have a "remove guns" command. Might need some more investigation.

rr-:
> > > While TR2 wasn't affected, it should also be tested here as it's impacted by the change.
> > 
> > 
> > Actually I think TR2 suffers from the same problem. In the GW, the player starts off with pistols, shotgun, flares and one of each medpacks, while `/play 2` only gives pistols.
> 
> I think somewhere TR2 has it hard-coded to add the shotgun, flares and medis on the first level (not decompiled yet, I think). The original gameflow does not have the usual "give item" commands in Great Wall. I know builders work around this by adding a "remove guns" command followed by "give pistols". So I think in TR2 when we do a `/play` and the level is beyond 1, it's only checking that the gameflow doesn't have a "remove guns" command. Might need some more investigation.

Thanks. Let's log a separate ticket for it and let's put it in the backlog.

rr- (approved):

walkawayy (approved):
Great fix!

aredfan (approved):

PR #1872, opened 2024-11-11
tr1/gameflow: add support for item descriptions

### LINKED ISSUE #1821: Feature request: ability to examine pickup items
This suggestion came from Leoc1995. Add the ability to define additional/multiline text for pickup/puzzle/key items so players can examine them in the inventory.

lahm86:
Resolves #1821.

#### Checklist

#### Description

This adds the ability to define descriptions for any top-ring inventory item (keys, puzzles, pickups, the lead bar and the scion) in the gameflow. Players can examine items that have a description in the inventory. Indicators are shown to make this clear, the idea being builders may want to use this feature to offer hints. There is a config option in case players don't want it.

Example level and GF attached:
[Examine.zip](https://github.com/user-attachments/files/17707852/Examine.zip)

Description text is automatically wrapped to fit the screen, and very long text will be paginated. New line and feed escape sequences are supported in the text for manual layout.

This also removes the necessity of having an empty `strings` object in levels that don't have any key items. We use the same function for loading regular and examine strings, so it avoids logging errors when they are missing, plus regular items now default if they are missing anyway.

I did consider merging the strings, so something like the following, but put that on hold as it's quite a big deviation for levels that are out there already. LMK what you think though.

```json
"strings": {
  "key1": {
    "name": "Silver Key",
    "description": "Lorem ipsum",
  }
}
```

The regex change in the `update_gameflow` tool allows the use of curly braces inside strings, and indeed nested objects if we were to go for the `strings` change above.


rr-:
> I did consider merging the strings, so something like the following, but put that on hold as it's quite a big deviation for levels that are out there already. LMK what you think though.
> 
> ```json
> "strings": {
>   "key1": {
>     "name": "Silver Key",
>     "description": "Lorem ipsum",
>   }
> }
> ```

It looks okay but let's hold off with it for now – I was also thinking of augmenting it at some point by introducing object aliases like so: `"key1": ["Silver Key", "Key", "Pickup"]` so that it can get targeted with `/give key` and `/tp pickup` without having to hardcode `"pickup"` like we do now.

src/libtrx/game/objects/names.c
```diff
@@ -21,6 +22,9 @@ static void M_ClearNames(void)
     for (GAME_OBJECT_ID object_id = 0; object_id < O_NUMBER_OF; object_id++) {
         M_NAME_ENTRY *const entry = &m_NamesTable[object_id];
         Memory_FreePointer(&entry->name);
+        if (entry->description != NULL) {
+            Memory_FreePointer(&entry->description);
+        }
```
rr-: No need for the NULL check.

src/libtrx/game/objects/names.c
```diff
@@ -32,12 +36,29 @@ void Object_SetName(const GAME_OBJECT_ID object_id, const char *const name)
     entry->name = Memory_DupStr(name);
 }
 
+void Object_SetDescription(
+    const GAME_OBJECT_ID object_id, const char *const description)
+{
+    M_NAME_ENTRY *const entry = &m_NamesTable[object_id];
+    if (entry->description != NULL) {
+        Memory_FreePointer(&entry->description);
+    }
```
rr-: No need for the NULL check.

src/tr1/game/gameflow.c
```diff
@@ -563,7 +567,11 @@ static bool M_LoadObjectNames(
         { "pickup1", "O_PICKUP_OPTION_1" },
         { "pickup2", "O_PICKUP_ITEM_2" },
         { "pickup2", "O_PICKUP_OPTION_2" },
+        { "leadbar", "O_LEADBAR_ITEM" },
+        { "leadbar", "O_LEADBAR_OPTION" },
+        { "scion", "O_SCION_OPTION" },
         { NULL, NULL },
+        // clang-format on
```
rr-, line 574: Note: These guards are only useful if we do something against the autoformatter, such as aligning the items in columns.

src/tr1/game/inventory/inventory_ring.c
```diff
@@ -18,13 +19,25 @@
...
 
+#define INIT_EXAMINE_TEXT(text, x_pos, role_str, input_str)                    \
+    do {                                                                       \
+        char role[100];                                                        \
+        sprintf(role, GS(role_str), GS(input_str));                            \
+        text = Text_Create(x_pos, -100, role);                                 \
+        Text_AlignBottom(text, true);                                          \
+        Text_CentreH(text, true);                                              \
+        Text_Hide(text, true);                                                 \
+    } while (0)
```
rr-: Let's just make it a function.

src/tr1/game/ui/widgets/paginator.c
```diff
@@ -0,0 +1,228 @@
...
+    int32_t max_width =
+        MAX(self->page_content->count == 1 ? 0 : 100,
+            UI_Label_MeasureTextWidth(self->title));
+    int32_t max_nav_width = 0;
+    int32_t max_height = 0;
+    for (int32_t i = 0; i < self->page_content->count; i++) {
+        M_SelectPage(self, i);
+        max_width = MAX(max_width, UI_Label_MeasureTextWidth(self->text));
+        max_height = MAX(max_height, UI_Label_MeasureTextHeight(self->text));
+        if (self->bottom_stack != NULL) {
```
rr-, line 212: I think in this case, `UI_Label_MeasureTextWidth(self->page_label)` will default to 0, no?

rr- (commented):
Amazing

src/libtrx/game/objects/names.c
lahm86: Thank you, slipped my mind while reviewing 😄 Did this part a few days ago.

src/tr1/game/gameflow.c
lahm86: The autoformatter was putting this into columns after I added the extra entries so I added the guards - I assume there is a max line trigger for it?

src/tr1/game/inventory/inventory_ring.c
lahm86: Done.

src/tr1/game/ui/widgets/paginator.c
lahm86: This handles (silly) cases where you have a very short title, very short page entries, but many pages, such that the biggest element is the navigation itself. So we cycle through and may find `<- 49 / 50 ->` is the widest label. The page nav is hidden if there is only one page, and in that case bottom_stack is null.

I'm open to tidying this up though if you've any thoughts on an improved implementation.

aredfan (approved):

src/tr1/game/gameflow.c
rr-: Oh I see what you mean. LGTM.

src/tr1/game/ui/widgets/paginator.c
rr-: No it's just that if the underlying page nav label is an empty string it should still be 0. If it's filled with the pager, I'm fine with the current code.

rr- (approved):

walkawayy (approved):
Amazing

PR #1882, opened 2024-11-12
inventory: reset selected item on closing inventory

### LINKED ISSUE #1586: OG bug: One of the plaques stops spinning and its name disappear
**Save:** [savegame.zip](https://github.com/user-attachments/files/16714319/savegame.zip)

**Step to reproduce:**

1. Have both Mystic Plaques in Lara's inventory
2. Insert the first Mystic Plaque on the left side
3. Open Lara's inventory to view the other Mystic Plaque

This issue isn't present when Lara goes to the other slot to insert the plaque.  It's only present if the player opens the inventory after inserting the first plaque.

lahm86:
Resolves #1586.

#### Checklist

#### Description

Most of this is config noise, the fix is simply setting `g_InvKeysCurrent = 0;` after closing the inventory, otherwise the game will try to select a non-existent item if you go back in via Esc > Up (or whatever your bindings are).

I've left the option defaulting to `false`, in line with the other TR1 glitch fixing options, but LMK what you think about this.

The issue has the steps for testing TR2. In TR1, you can try this:

- `/play 9`
- `/tp 77`
- `/give rusty`
- `/give gold`
- Use the rusty key by pressing action at one of the keyholes
- Go to the other keyhole and _manually_ go into the inventory to select the gold key. Depending on the setting it will either be glitched out and act as the rusty key, or it will behave properly and you won't be able to use it here.


rr- (approved):

aredfan (approved):
Tested both scenarios with the option on and off, LGTM.
 
Thank you for working on this bugfix.

PR #1896, opened 2024-11-13
tr2/room: refactor floor data handling

### LINKED ISSUE #1883: Refactor FD data
Similar to what we did in #941 for TR1.
Requires me decompiling a bunch of functions related to testing triggers.

lahm86:
Resolves #1883.

#### Checklist

#### Description

This follows a similar approach as we did for TR1 in #941. Best viewed commit-by-commit to see the iterative changes.

As well as the main FD overhaul, there were a few snagging points that have been addressed here.
- `Item_Initialise` was previously called during `M_LoadItems` in `level.c` but this had to be shifted because there are floor data dependencies there. We now take a similar approach to TR1, so load the level data, then perform a "CompleteSetup"; this can be expanded later for things like injection processing.
- With the above, @aredfan found that if you start a level then go back to the title screen, it can crash. `g_LevelItemCount` wasn't being set to 0 here - not an issue originally because `Item_Initialise` (where the crash would occur) would have been skipped in that case.
- While the triggers that are meant to activate after pulling the dagger from the dragon are heavy, the call to test triggers came from Lara's extra anim routine. As we now pass the item, which determines the heavy state, this had to be amended to locate the dragon bones item and to pass that. The dragon control has already ended by the time the extra anim completes so we couldn't add the call anywhere there, and running it too early meant issues like the camera still being in cinematic mode while looking at the doors opening.

The dragon may need some extra checks based on where it lands e.g. if partially in the wall. I may change this to behave like the Torso in TR1, where it checks a range of surrounding sectors.

For testing, attached is a level with various trigger types, basic geometry, ladders, and death tiles. I didn't include a dragon for testing, that can be done in OG easily enough. For ceiling geometry, it's best to fly against it and hold down to see that Lara follows the structure properly.

[WALL.zip](https://github.com/user-attachments/files/17750165/WALL.zip)

There is a lot of scope now for sharing FD handling with TR1, but that can be tackled in the future. I may follow this up with some minor changes, like moving all FD macros to libtrx and renaming some enums, but I didn't want to touch TR1 at all in this PR.

### DIFF
--- a/src/tr2/global/types.h
+++ b/src/tr2/global/types.h
@@ -1522,34 +1522,6 @@ typedef struct __unaligned {
 } FRAME_INFO;
 
 typedef enum {
-    TO_OBJECT      = 0,
-    TO_CAMERA      = 1,
-    TO_SINK        = 2,
-    TO_FLIP_MAP    = 3,
-    TO_FLIP_ON     = 4,
-    TO_FLIP_OFF    = 5,
-    TO_TARGET      = 6,
-    TO_FINISH      = 7,
-    TO_CD          = 8,
-    TO_FLIP_EFFECT = 9,
-    TO_SECRET      = 10,
-    TO_BODY_BAG    = 11,
-} TRIGGER_OBJECT;
-
-typedef enum {
-    TT_TRIGGER     = 0,
-    TT_PAD         = 1,
-    TT_SWITCH      = 2,
-    TT_KEY         = 3,
-    TT_PICKUP      = 4,
-    TT_HEAVY       = 5,
-    TT_ANTIPAD     = 6,
-    TT_COMBAT      = 7,
-    TT_DUMMY       = 8,
-    TT_ANTITRIGGER = 9,
-} TRIGGER_TYPE;
-
-typedef enum {
     GF_S_PC_DETAIL_LEVELS      = 0,
     GF_S_PC_DEMO_MODE          = 1,
     GF_S_PC_SOUND              = 2,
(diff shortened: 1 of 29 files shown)


aredfan:
An issue I found that is currently confined to this PR - the `/title` command closes the game prematurely.

lahm86:
> An issue I found that is currently confined to this PR - the `/title` command closes the game prematurely.

Thanks, now addressed. The global item count wasn't being set to 0 for the title level. It wasn't an issue before because we were skipping `Item_Initialise` in that case, but that call now happens after everything else has loaded. So it was trying to load based on the count of whichever level you were on previously.

src/tr2/global/types.h
```diff
@@ -1521,34 +1521,6 @@ typedef struct __unaligned {
...
-    TT_PAD         = 1,
-    TT_SWITCH      = 2,
-    TT_KEY         = 3,
-    TT_PICKUP      = 4,
-    TT_HEAVY       = 5,
-    TT_ANTIPAD     = 6,
-    TT_COMBAT      = 7,
-    TT_DUMMY       = 8,
-    TT_ANTITRIGGER = 9,
-} TRIGGER_TYPE;
```
rr-, line 1550: As this file is autogenerated, we also need to mark these types as `// decompiled` in `progress.txt`, otherwise `just tr2-auto` will bring them back here.

rr- (approved):

aredfan (approved):

PR #1905, opened 2024-11-15
tr2/music_main: refactor music track handling

### LINKED ISSUE #1899: TR2X bug: some music triggers loop and others don't repeat
An example is the trigger at the zip line in Great Wall. Although the trigger isn't one-shot, in OG it only plays once if you stand on the tile, whereas in TR2X it loops until you move away.
One-shot music triggers behave fine.

Another issue is if you complete the assault course and get a best time, then complete and beat the time again without triggering any other music tracks, Lara does not repeat the "best time yet" line at the end.

Regression from 0.2.

lahm86:
Resolves #1899.

#### Checklist

#### Description

This refactors the way music is checked when requested to play tracks. We introduce tracking, similar to TR1 - I think prior to 0.2, it was implicitly done by setting (or not setting) `g_CD_TrackID` before playing the music in various places, so it was difficult to keep track of that. That global is now redundant, although remains in place, along with the original `Music_Play` function (renamed).

Some tracks require to be enforced such as from flip effects (like completing the assault course) and mounting the skidoo. Tracked trigger tracks will not play in succession, provided the mode is the same when requesting to play them. So standing on a tile doesn't loop a track.

Attached is a test level which should behave the same played in OG as TR2X. So e.g. if you cross the "manual" secret track trigger, then collect a secret, the track will restart (similarly if you collect the secrets quickly before the track finishes each time). And if you cross the skidoo track trigger, then mount the skidoo itself, the music will restart, the same as OG. On the other foot, if you mount the skidoo first, wait for the track to completely finish, then cross the trigger for the same track, it doesn't play.

The assault course is a good place to test everything too, especially the delayed tracks there and ensuring Lara acknowledges beating the best time twice in a row. The test level can also be renamed to `boat.tr2` to check a level with no ambience. Renaming it to `house.tr2` will allow for example triggering the skidoo music, then ending the level, and ensuring the credits restart the track.

[WALL.zip](https://github.com/user-attachments/files/17781058/WALL.zip)

### DIFF
--- a/src/tr2/global/vars_decomp.h
+++ b/src/tr2/global/vars_decomp.h
@@ -12,7 +12,6 @@
 #define g_RhwFactor (*(float*)0x0046408C) // = 335544320.0f
 #define g_CineTrackID (*(int32_t*)0x004640B0) // = 1
 #define g_CineTickRate (*(int32_t*)0x004640B8) // = 0x8000
-#define g_CD_TrackID (*(int16_t*)0x004640BC) // = -1
 #define g_FlipEffect (*(int32_t*)0x004640C4) // = -1
 #define g_BoxLines (*(int32_t(*)[12][2])0x00464180)
 #define g_AssaultBestTime (*(int32_t*)0x004641F0) // = -1
--- a/docs/tr2/CHANGELOG.md
+++ b/docs/tr2/CHANGELOG.md
@@ -7,6 +7,8 @@
 - fixed being unable to go from surface swimming to underwater swimming without first stopping (#1863, regression from 0.6)
 - fixed pistols appearing in Lara's hands when entering the fly cheat during certain animations (#1874)
 - fixed Lara continuing to walk after being killed if in that animation (#1880, regression from 0.1)
+- fixed some music tracks looping while Lara remained on the same trigger tile (#1899, regression from 0.2)
+- fixed some music tracks not playing if they were the last played track and the level had no ambience (#1899, regression from 0.2)
 
 ## [0.6](https://github.com/LostArtefacts/TRX/compare/tr2-0.5...tr2-0.6) - 2024-11-06
 - added a fly cheat key (#1642)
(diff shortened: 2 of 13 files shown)


src/tr2/global/vars_decomp.h
```diff
@@ -12,7 +12,7 @@
 #define g_RhwFactor (*(float*)0x0046408C) // = 335544320.0f
 #define g_CineTrackID (*(int32_t*)0x004640B0) // = 1
 #define g_CineTickRate (*(int32_t*)0x004640B8) // = 0x8000
-#define g_CD_TrackID (*(int16_t*)0x004640BC) // = -1
+#define g_Legacy_CD_TrackID (*(int16_t*)0x004640BC) // = -1
```
rr-: Hi, this file is updated by progress.txt as well. Let's mark this variable with a `+` flag in progress.txt (without renaming it). This way, it'll be visible as `g_CD_TrackID` in IDA, while it'll be completely skipped in our codebase.

As for `Music_Legacy_Play`, we can safely remove its definitions altogether, considering how since 0.2 we're no longer trying to inject it (and thus are free to use our own signatures).

After these changes please make sure to run `tools/tr2/generate_funcs` to see all is compatible.
Apologies for not documenting this process better, but hopefully it won't be long until we can start freely refactoring the majority of the codebase.

lahm86: Thank you, and no - my apologies for not getting this bit right.

I've changed the variable back to OG and marked it as decompiled.

The legacy function is still in use though: below is with logging added, it's triggered when Lara completes the assault course (possibly elsewhere, but this one for sure). I'm assuming that flip effect does some comparison with the player's time and decides which track to play (22 is "gosh, that was my best time yet").

```
game/music/music_main.c 116 Music_Legacy_Play Legacy call for track: 22
game/music/music_main.c 149 Music_Play Playing track 22 (real: 19), mode: 0
```

image

aredfan (approved):

rr- (approved):

PR #1918, opened 2024-11-18
libtrx/room: move common FD handling to libtrx

lahm86:
#### Checklist

#### Description

This moves the common floor data parsing routines into libtrx, along with the relevant consts and macros.

The `GameBuf_Alloc` definition is also in libtrx now as it's needed in the parsing, but I've left the implementations separate for now. LMK if this is an OK approach for now.


src/libtrx/game/rooms/common.c
```diff
@@ -21,3 +43,127 @@ int32_t Room_GetAdjoiningRooms(
...
+            }
+
+            break;
+        }
+
+        default:
+            break;
+        }
+    } while (!FD_IS_DONE(fd_entry));
+}
```
rr-, line 156: If we're going to get significantly more data related to floor data triggers, maybe we could create a file of its own for it `game/rooms/floor_data.[ch]`?

aredfan (approved):

src/tr1/game/gamebuf.h
```diff
@@ -9,45 +11,6 @@
...
-    GBUF_SAMPLE_INFOS,
-    GBUF_SAMPLES,
-    GBUF_TRAP_DATA,
-    GBUF_CREATURE_DATA,
-    GBUF_VERTEX_BUFFER,
-} GAME_BUFFER;
-
 void GameBuf_Init(void);
 void GameBuf_Reset(void);
-void *GameBuf_Alloc(int32_t alloc_size, GAME_BUFFER buffer);
```
rr-, line 52: No need to repeat the function declarations that we already declare in libtrx.

src/tr2/game/gamebuf.h
```diff
@@ -1,62 +1,10 @@
...
-    GBUF_SPRITE_INFOS             = 46,
-    GBUF_NUM_MALLOC_TYPES         = 47,
-    // clang-format on
-} GAME_BUFFER;
+#include <stdint.h>
 
 void GameBuf_Init(int32_t cap);
 void __cdecl GameBuf_Reset(void);
 void __cdecl GameBuf_Shutdown(void);
-void *__cdecl GameBuf_Alloc(size_t alloc_size, GAME_BUFFER buffer);
```
rr-, line 61: No need to repeat the function declarations that we already declare in libtrx.

src/libtrx/game/rooms/common.c
lahm86: There are 15 other OG types used in TR3+. Equivalent switch block is here:

https://github.com/LostArtefacts/TR-Rando/blob/master/TRLevelControl/Build/TRFDBuilder.cs#L147

Maybe it would be an idea to move the `FT_TRIGGER` case to a separate function: it's quite chunky, while the others are one-liners. And then for the future do similar, or split into files?

src/tr1/game/gamebuf.h
lahm86: `GameBuf_Init` has a different signature in the two games, and TR2 has a `GameBuf_Free` while TR1 doesn't. I figured just to move the one function we need for now, so the others are only declared in the corresponding game header.

TR2 still refers to a global buffer variable too. Quite happy to make everything the same though in this PR, but maybe better as a separate task?

src/libtrx/game/rooms/common.c
lahm86: Moved trigger reading to its own function, the main loop is more readable now IMO.

src/tr1/game/gamebuf.h
rr-: Good call – no, let's tackle that separately.

rr- (approved):
LGTM

Issue #1494, opened 2024-09-06
Refactor room and object meshes


lahm86:
Parse the raw data into structures and refactor usage throughout, most notably in `output.c`. This will allow extending room structures to support additional attributes such as additional vertex data mentioned in #1493.

PR #1938, opened 2024-11-23
tr1/misc: create room mesh structure

lahm86:
Part of #1494.

#### Checklist

#### Description

This eliminates raw mesh data for rooms and replaces it with a complete structure for better readability, and improved injection support. While the definitions and parsing are handled in `libtrx`, this currently only applies to TR1, so TR2 remains as-is. Best viewed commit-by-commit.

One change that does touch both games is in renaming room static meshes from `MESH` to avoid confusion with the actual room mesh, and I've also introduced defined flags for these, again for improved readability.

Injection files for room fixes where new faces or vertices are added have had to be updated. Old versions only store the total mesh injection size, but we now need to know explicitly how many faces and vertices are being injected to know in advance how much memory to allocate. Old files will still work for other texture fixes like rotating faces, moving vertices etc, but the new ones will be needed for all cases moving forward.

This also adds a flag to room vertices so we can continue with the water edge wibble fix. Again, another improvement rather than packing the flag into the lighting value.

I have object meshes implemented as well, but the changes for that are more involved than this PR, so best to keep things separate for now. Once these are done, we can eliminate the semi-duplicated output functions (`M_DrawObjectGT3` and suchlike).


lahm86:
Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/9

aredfan (approved):

src/tr1/game/level.c
```diff
@@ -1026,8 +1022,9 @@ static size_t M_CalculateMaxVertices(void)
             MAX(max_vertices, *(g_Meshes[static_info->mesh_num] + 5));
     }
 
-    for (int32_t i = 0; i < g_RoomCount; i++) {
-        max_vertices = MAX(max_vertices, *g_RoomInfo[i].data);
+    for (int32_t i = 0; i < Room_GetTotalCount(); i++) {
+        const ROOM *room = Room_Get(i);
```
rr-: Nitpick: `const *const`

src/tr1/game/level.c
```diff
@@ -226,14 +223,15 @@ static void M_LoadRooms(VFILE *file)
         }
 
         // Static mesh infos
-        r->num_meshes = VFile_ReadS16(file);
-        if (!r->num_meshes) {
-            r->meshes = NULL;
+        r->num_static_meshes = VFile_ReadS16(file);
+        if (!r->num_static_meshes) {
```
rr-: Nitpick: I'd make it super explicit by using `== 0`

src/tr1/game/level.c
```diff
@@ -226,14 +223,15 @@ static void M_LoadRooms(VFILE *file)
...
         } else {
-            r->meshes = GameBuf_Alloc(
-                sizeof(MESH) * r->num_meshes, GBUF_ROOM_STATIC_MESHES);
-            for (int32_t j = 0; j < r->num_meshes; j++) {
-                MESH *mesh = &r->meshes[j];
+            r->static_meshes = GameBuf_Alloc(
+                sizeof(STATIC_MESH) * r->num_static_meshes,
+                GBUF_ROOM_STATIC_MESHES);
+            for (int32_t j = 0; j < r->num_static_meshes; j++) {
+                STATIC_MESH *mesh = &r->static_meshes[j];
```
rr-: Nitpick: `STATIC_MESH *const mesh`

rr- (approved):
This is brilliant!

walkawayy (approved):
This looks amazing.

src/libtrx/game/level/common.c
```diff
@@ -0,0 +1,80 @@
...
+            GameBuf_Alloc(sizeof(ROOM_VERTEX) * alloc_count, GBUF_ROOM_MESH);
+        for (int32_t i = 0; i < room->mesh.num_vertices; i++) {
+            ROOM_VERTEX *const vertex = &room->mesh.vertices[i];
+            vertex->pos.x = VFile_ReadS16(file);
+            vertex->pos.y = VFile_ReadS16(file);
+            vertex->pos.z = VFile_ReadS16(file);
+            vertex->shade = VFile_ReadU16(file);
+            vertex->flags = 0;
+        }
+    }
```
walkawayy, line 32: Are these outer parentheses just used so you can reuse variable names by reducing the scope of each block?

lahm86: Yes, pretty much. Short of separating the four property reads into functions, I think this is OK to group the logic together while still making scope clear.

PR #1959, opened 2024-11-26
tr1|tr2/objects/door: reset sector tilt on closed doors

### LINKED ISSUE #1958: TR2X bug: invisible collision in Catacombs
**Dev snapshot:** 0.6-215-g5a4f33d

____________________

1. `/play 13`
2. `/tp 25 5 42`

20241126_114756_Catacombs_of_the_Talion

lahm86:
Resolves #1958.

#### Checklist

#### Description

Before the FD refactor in both games, closing a door would reset the sector FD index to 0, and so that meant functions such as `Room_GetCeiling` would not interpret tilt values. I missed applying the equivalent reset in both #1433 and #1896 and so the result was that `NO_HEIGHT` values would be adjusted with tilt values, and hence produce something other than `NO_HEIGHT` and therefore potential invisible geometry.

Quite specific still for the issue itself: you need a door that's not on a portal sector in a room above, and there must be tilt geometry defined. This doesn't happen in OG TR1 AFAIK, so I've attached a test level.

[level1.zip](https://github.com/user-attachments/files/17924851/level1.zip)

### DIFF
--- a/docs/tr1/CHANGELOG.md
+++ b/docs/tr1/CHANGELOG.md
@@ -24,6 +24,7 @@
     Swedish, Turkish and possibly more.
 
     Importantly, Asian and Arabic languages remain unsupported at the moment.
+- fixed a potential invisible wall issue in custom levels with non-portal doors and certain geometry (#1958, regression from 4.3)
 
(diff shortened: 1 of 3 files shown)


aredfan (approved):
LGTM.  Thanks for the sameday bugfix. 👍

rr- (approved):

Issue #1967, opened 2024-11-28
Refactor injection support


lahm86:
Main tasks
- [x] Move handling fully to TRX
- [x] Refactor the file format to support changes more easily
- [x] Use compression
- [x] Improve applicability tests
- [ ] Extend support fully to TR2

**TR2 progress**

_No dependencies_
- [x] Item position/rotation edits
- [x] Object info
- [x] Room property edits e.g. #1939

_Dependencies_
- [x] Anim data
  - [x] Rewrite anim usage throughout to remove globals and create API
  - [x] Read data via libtrx as standard for both games
- [x] Texture pages/packing
  - [x] Requires tex page limit #1796 
- [x] Tex info injection
  - [x] Requires tex info limit #1795 
- [x] Room texture/face/vertex edits
  - [x] Requires `ROOM_MESH` type per #1938
- [x] Object mesh insertion and edits
  - [x] Requires `OBJECT_MESH` type per #1941

_Misc_
- [ ] Automatic 8-bit conversion of textures

PR #1968, opened 2024-11-28
tr2/inject: adjust injection format

lahm86:
Part of #1967.

#### Checklist

#### Description

This makes the TR2 injection header more adaptable to change in terms of adding support for future injection types. We now read into an array of counts rather than fixed counts at expected positions in the file. This means we don't necessarily need to increment the version each time something is added, instead the count of any new edit type will just default to zero for older files. Overall it will make incremental injection support neater IMO.

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/12

As we have not yet shipped injections for TR2, the four existing floor data fixes have been re-created, so it'd be good to double check these in-game.

I've also fixed an `int` usage in a couple of places in TR1 and TR2.


rr-:
@lahm86 what do you think about another approach – encode number of all edits as a single integer, and for each edit, begin with an edit type member?

lahm86:
> @lahm86 what do you think about another approach – encode number of all edits as a single integer, and for each edit, begin with an edit type member?

I might be misunderstanding, but do you mean packing the type and count into a single integer? e.g.
```c
info->counts[value & 0xFF] = value >> 8;
```

Or do you mean effectively doing away with the individual counts in the header? Unfortunately, we need those in advance to know about memory allocation in `level.c` (not quite in use yet for TR2, but it's present in TR1 e.g. allocate extra space for new animations).

rr-:
> > @lahm86 what do you think about another approach – encode number of all edits as a single integer, and for each edit, begin with an edit type member?
> 
> I might be misunderstanding, but do you mean packing the type and count into a single integer? e.g.
> 
> ```c
> info->counts[value & 0xFF] = value >> 8;
> ```
> 
> Or do you mean effectively doing away with the individual counts in the header? Unfortunately, we need those in advance to know about memory allocation in `level.c` (not quite in use yet for TR2, but it's present in TR1 e.g. allocate extra space for new animations).

I mean to do something like this:

```c
typedef enum { ModifyFloorData, …, } EditType;
typedef struct { EditType edit_type; void *priv; } Edit;
```

As for the counting, we could always loop twice, once to measure the memory consumption, and once to load the injections. IDK which design will feel more elegant from the perspective of the consumer code, but speaking stricly about the file format, I'd say this approach feels more canonical.

lahm86:
OK, I've done some re-organising, LMK what you think. I didn't opt for an `Edit` struct as the individual edit/injection format is so arbitrary we'd need to have many different structures defined for `*priv` (as in the injection tool). For example, a floor edit could be one of many different sub-types, and need different counts of data read/interpreted. IMO it's more efficient just to read and consume the data as part of the actual injection.

I have introduced the idea of a "block" though. The overall counts are aggregated while the file is initially loaded, and the main data skipped in this process. We then reset the file position ready for the actual injection logic later. In TR1 we have an aggregate object that gets passed between the injection and level modules, but with the new approach for TR2, `level.c` only needs to call e.g. `Inject_GetDataCount(IDT_ANIMATIONS);` to get the information it needs for memory allocation. This way we're not having to pass around too much.

So to clarify, this is the overall file layout:
- Magic
- Version 
- Main injection type (linked to config setting where applicable)
- Total number of blocks to follow
- Blocks
  - Type
  - Count (this is contextual, so could be the number of floor edits, or number of animations etc)
  - Length (of data to follow in bytes)
  - Data

Each block is then fully interpreted by specific module functions e.g. `M_FloorDataEdits` in the main injection loop, or its data is skipped over if it's an unknown type.

aredfan (approved):

rr- (approved):
I think it's looking really elegant.
I'm not sure about more involved edits such as "inject this .wad" (jk) but I suppose that's the job of the injector to compile it down to atomic pieces.

Issue #1613, opened 2023-11-11
Feature request: Rotate 3D pickups


aredfan:
<details>
  <summary>Click to see the table</summary>

| Level name               | Item # | Type            | Room | Rotate |
|--------------------------|--------|-----------------|------|--------|
| The Great Wall           | 17     | Guardhouse Key  | 34   | 90     |
| The Great Wall           | 92     | Small Medipack  | 57   | 180    |
| Opera House              | 82     | Circuit Board   | 83   | 90     |
| Opera House              | 118    | Ornate Key      | 107  | 90     |
| Offshore Rig             | 30     | Green Pass Card | 18   | 90     |
| Diving Area              | 58     | Red Pass Card   | 30   | 180    |
| Diving Area              | 94     | Blue Pass Card  | 52   | 180    |
| Diving Area              | 120    | Machine Chip    | 80   | 90     |
| Diving Area              | 124    | M16             | 80   | 270    |
| Wreck of the Maria Doria | 64     | Large Medipack  | 43   | 180    |
| Wreck of the Maria Doria | 149    | Circuit Breaker | 76   | 90     |
| Wreck of the Maria Doria | 183    | Rest Room Key   | 93   | 90     |
| The Deck                 | 82     | Flare           | 85   | 90     |
| Tibetan Foothills        | 0      | Large Medipack  | 1    | 180    |
| Tibetan Foothills        | 48     | Hut Key         | 70   | 90     |
| Tibetan Foothills        | 68     | Drawbridge Key  | 87   | 90     |
| Barkhang Monastery       | 31     | Rooftops Key    | 14   | 90     |
| Barkhang Monastery       | 97     | Flare           | 50   | 90     |
| Barkhang Monastery       | 176    | Main Hall Key   | 107  | 90     |
| Barkhang Monastery       | 202    | Trapdoor Key    | 153  | 90     |
| Catacombs of the Talion  | 9      | Tibetan Mask    | 2    | 90     |
| Catacombs of the Talion  | 47     | Small Medipack  | 16   | 180    |
| Catacombs of the Talion  | 112    | Tibetan Mask    | 67   | 180    |
| Ice Palace               | 90     | Large Medipack  | 62   | 180    |
| Ice Palace               | 93     | Small Medipack  | 65   | 180    |
| Ice Palace               | 151    | Large Medipack  | 143  | 270    |
| Temple of Xian           | 0      | Small Medipack  | 1    | 90     |
[...]

PR #1970, opened 2024-11-28
tr2/inject: add item and room editing support

### LINKED ISSUE #1716: OG bug: possible softlock when collecting the gong | Ice Palace
If Lara collects the gong from the edge of the pickup trigger then turns away instead of swimming forward - the door beyond the yetis won't open.

https://github.com/user-attachments/assets/626bea11-f0ea-4e1a-8e5a-9a6bb86aa836

Screenshot

### LINKED ISSUE #1939: OG bug: room 98 is dry despite having a shark in it | Wreck of the Maria Doria
I came across this issue while investigating #1936.

In Wreck of the Maria Doria, a room containing a shark has no water in it. This may explain why the shark is alive in TR2X versions prior to 0.3 whereas other sharks were already dead upon level load.

Screenshot

I made this video as proof of concept of what the room would look like from above if the water effect is fixed. First half is OG and the second half is a modified level. The fixed water effect is an improvement to what may be a dev oversight.

https://github.com/user-attachments/assets/d45cde4f-f28e-4c5f-95db-4a159d4739fe

I've tested the modified level in HW and SW modes and found no graphical issues. If it's possible to introduce this fix via injection then that'd be great. Thank you for taking the time to read this issue and for considering it.

### LINKED ISSUE #1963: OG bug: invisible collision in front of closed door
1. `/play 14`
2. `/tp 49 11 29`

20241126_183351_Ice_Palace

**Edit:** I found another case of collision in front of a closed door.

1. `/play 16`
2. `/tp 126`

20241127_152804_Floating_Islands

lahm86:
Resolves #1716.
Resolves #1939.
Resolves #1963.

#### Checklist

#### Description

This fixes a few more FD issues as detailed in the issues. I've bundled the door fixes for Ice Palace and Floating Islands in as FD fixes too, even though they involve repositioning items. Given the issue they cause, it feels more natural coming under the FD fix config setting.

The item editing function will also be used later for #1613.

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/13


rr- (approved):

aredfan (approved):
These fixes are working beautifully. Thank you very much for tackling these 3 issues. 👍

PR #1983, opened 2024-12-02
Ditch DirectX in favor of OpenGL; use SDL window; ditch Windows Registry

rr-:
#### Checklist

- [x] I have read the [coding conventions](https://github.com/LostArtefacts/TR1X/blob/master/CONTRIBUTING.md#coding-conventions)
- [ ] I have added a changelog entry about what my pull request accomplishes, or it is an internal change

#### Description

As a reminder:

- (NEW) F1/Shift-F1: toggle upscaling 1-2-3-4× ("scaler")
- F2/Shift-F2: increase or decrease viewport size ("sizer")
- F3: toggle bilinear filter
- F4: toggle perspective filter
- F7: toggle zbuffer
- F12: toggle SWR/HWR renderer

Known issues:

- [x] There are black rectangles around the window on some configurations
- [x] The game sometimes crashes when using the new scaler option (F1 / Shift-F1)
- [x] Scaler/sizer woes with HWR/SWR
- [x] Console fails to get drawn on non-1x-scaling
- [ ] Window isn't brought to front on launch
- [ ] Text doesn't respect aspect ratio
- [x] Aspect ratio issues
- [x] ZBuffer near plane clipping (looking closely at door objects, passport animation)
- [x] Alpha problems for UI in HWR
- [x] Alpha problems for UI in SWR
- [x] Alpha edge problems in textures in HWR
- [x] Ingame backgrounds do not work at all (neither SWR nor HWR)
- [x] 3D pickups overlay clips with HWR + ZBuffer ON
- [x] The controls UI doesn't display
- [x] The ammo counter doesn't display
- [x] Health/air bars don't show in HWR + ZBuffer ON
- [x] Health/air bars appear ugly in HWR
- [x] The health bar doesn't display when highlighting a medi pack in the inventory
- [x] If you use F5/F6 or flip through the passport pages - the OPTION text blinks. It's subtle but noticeable.
- [x] Green pixels appear in explosion sprites in Dragon's Lair (level file textures seem OK, doesn't happen in OG)
- [x] Fade out in the title screen clips the inventory items in HWR / ZBuffer ON
- [x] The end of level stats background uses the main menu image instead, with these steps (`/play 1`, `/endlevel`)
- [x] The end of level stats background looks bugged (`/play 1`, open then close the inventory, `/endlevel`, skip cutscene)
- [x] The textures keep zooming in when toggling the renderer with the bilinear filter on
- [x] Quitting the game with Alt+F4 goes to the inventory very briefly(?)
[...]

PR #1994, opened 2024-12-09
Config refactor

rr-:
#### Checklist

#### Description

Extracted from #1983.

Moves `Config_ApplyChanges` to be called *after* launching event listeners, enabling client games to detect changes to specific config properties. To facilitate this, clients should implement logic in their `ApplyChanges` method like this:

```c
CONFIG g_SavedConfig = { 0 };

void Config_ApplyChanges(CONFIG_CHANGE_TRIGGER trigger)
{
    g_SavedConfig = g_Config;
}
```

Then the event subscribers can test for individual properties by comparing `g_Config` and `g_SavedConfig`.

### DIFF
--- a/src/libtrx/config/common.c
+++ b/src/libtrx/config/common.c
@@ -41,7 +41,6 @@ bool Config_Write(void)
     };
     const bool updated = ConfigFile_Write(&args);
     if (updated) {
-        Config_ApplyChanges();
         if (m_EventManager != NULL) {
             const EVENT event = {
                 .name = "write",
@@ -50,6 +49,7 @@ bool Config_Write(void)
             };
             EventManager_Fire(m_EventManager, &event);
         }
+        Config_ApplyChanges();
     }
     return updated;
 }


src/libtrx/config/common.c
```diff
@@ -26,32 +51,14 @@ bool Config_Read(void)
...
-            const EVENT event = {
-                .name = "write",
-                .sender = NULL,
-                .data = NULL,
-            };
-            EventManager_Fire(m_EventManager, &event);
-        }
-    }
-    return updated;
+    return M_Write(CCT_WRITE);
```
walkawayy: Maybe it's needed for a future PR so forgive the question, but do we need to extract this code out to `M_Write` if only `Config_Write` calls it?

rr-: Good call – I initially had another call that used this, but it was no longer needed. Let me bring it back.

rr-: Actually I think it's sufficient to just reorder the calls, and we can forgo the trigger changes altogether simplifying the final solution.

aredfan (approved):

walkawayy (approved):

lahm86 (approved):

PR #2008, opened 2024-12-12
tr1/data: fix mesh positions in Obelisk of Khamoon

### LINKED ISSUE #2006: OG bug: Obelisk keys are positioned slightly too low
**Copied over from 2004:**

> There's a new issue with the Obelisk keys being partially buried in the drawbridges.
> 
> **Edit:** The sprites are also partially buried in the DOS version - which doesn't suffer from the clipping issue. It's an OG bug. 
> 
> **Sprite pickups**
> 
> Dev snapshot
> 1 1 Dev
> PR build
> 1 2 PR
> 
> **3D pickups**
> 
> Dev snapshot
> 2 1 PR
> PR build
> 2 2 PR

lahm86:
Resolves #2006.

#### Checklist

#### Description

This fixes the angle of the drawbridge when open so that Lara's shadow isn't occluded and the artefacts aren't embedded. The open door will also extend fully to the wall. The artefact holder mesh is also shifted down slightly to avoid its texture clashing with the bridge, and it's also pushed against the wall. The caveat is that pickups are slightly above the holder mesh, but should be exactly on the open bridge. I can't see a workaround for this as with the mesh in its original position, the texture pushes through the open bridge.

In room 57, the senet table is also lifted and the toppled chair is lowered.

The "frame edit" addition to the injector is a temporary measure, once we have a new file structure in place (and the changes to the engine in line with this) this will become neater and more general purpose. Currently you could only edit the first rotation of the first frame of an animation.

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/17


lahm86:
Updated to fix the chair in Return to Egypt and Temple of the Cat.

walkawayy (approved):

aredfan (approved):
LGTM, and thank you for going the extra mile to fix the TRUB issues. 👍

rr- (approved):
Looks great! Thank you.

PR #2037, opened 2024-12-14
tr1/objects/movable_block: add trap feature to pushblocks

### LINKED ISSUE #2035: Add option to make falling blocks deadly
If a falling pushblock lands on Lara, it should kill her outright rather than making her clip to the top of it when the floor height is changed. This should be configurable via a gameflow setting.

lahm86:
Resolves #2035.

#### Checklist

#### Description

This makes falling pushblocks more realistic if they land on top of Lara, killing her outright. We hide Lara's shadow to avoid it appearing on top of the pushblock, as the floor height change will continue to operate in case there are active enemies nearby. Additionally, a camera target item is spawned to prevent issues with the camera looking into the pushblock at Lara i.e. if we set the target to be the pushblock, the camera will push into the raised floor. This target also needs to be stored and continually set in the state routine as the camera is otherwise eager to return to normal.

### DIFF
--- a/src/tr1/game/objects/traps/movable_block.c
+++ b/src/tr1/game/objects/traps/movable_block.c
@@ -2,11 +2,14 @@
 
 #include "game/collide.h"
 #include "game/effect_routines/floor_shake.h"
+#include "game/effects/blood.h"
+#include "game/gameflow.h"
 #include "game/input.h"
 #include "game/items.h"
 #include "game/lara/common.h"
 #include "game/objects/common.h"
 #include "game/objects/vars.h"
+#include "game/random.h"
 #include "game/room.h"
 #include "game/sound.h"
 #include "global/const.h"
@@ -38,6 +41,8 @@ static bool M_TestDoor(ITEM *lara_item, COLL_INFO *coll);
 static bool M_TestDestination(ITEM *item, int32_t block_height);
 static bool M_TestPush(ITEM *item, int32_t block_height, DIRECTION quadrant);
 static bool M_TestPull(ITEM *item, int32_t block_height, DIRECTION quadrant);
+static bool M_TestDeathCollision(ITEM *item, const ITEM *lara);
+static void M_KillLara(const ITEM *item, ITEM *lara);
 
 static const OBJECT_BOUNDS *M_Bounds(void)
 {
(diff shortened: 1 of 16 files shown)


src/tr1/game/objects/traps/movable_block.c
```diff
@@ -213,6 +217,48 @@ static bool M_TestPull(ITEM *item, int32_t block_height, DIRECTION quadrant)
...
+
+    const int16_t target_num = Item_Spawn(lara, O_CAMERA_TARGET);
+    if (target_num != NO_ITEM) {
+        ITEM *const target = Item_Get(target_num);
+        target->rot.y = g_Camera.target_angle;
+        target->pos.y = lara->floor - WALL_L;
+        Lara_SetDeathCameraTarget(target_num);
+    }
+
+    return true;
```
rr-: This function is named like a getter but has significant side effects, let's make sure the name captures it. Maybe also let's put the death logic into its own function.

rr- (commented):
WDYT about drawing Lara's shadow at all times but instead do not alter the floor height if she gets squished? I'm only suggesting this to simplify the Lara's API.

src/tr1/game/objects/traps/movable_block.c
lahm86: Thanks, I've moved the logic around a little and updated the naming.

Re the shadow, I initially had it so that the floor height wasn't changed but my thinking was that if there are enemies nearby, they would walk through the block. We'd also need to have something in place to update the height if Lara is resurrected (although that is a cheat so less critical).

Perhaps we could have a flag in `ITEM` for whether or not the shadow is enabled, so it would get checked in `output` for any item type?

aredfan (approved):

src/tr1/game/objects/traps/movable_block.c
rr-: Thanks, those are very valid points I haven't considered. The flag sounds good, but the current approach is also fine.

rr- (approved):

src/tr1/game/objects/traps/movable_block.c
lahm86: I've gone with the flag approach, which I think is a bit neater.

rr- (approved):

Issue #2017, opened 2024-12-13
3d pickups are too dark


Richard-L:
PS1
![Tomb_Raider_II_-_Starring_Lara_Croft_USA_2024-12-12-23-23-21.png](https://github.com/user-attachments/assets/02b05729-9245-49de-a235-fbd394644eaf)

TR2X w 3d pickups enabled 
![image.png](https://github.com/user-attachments/assets/502385bb-09a7-48f4-aa2d-5b1e8c29a831)

The comparison is a little unfair but just to illustrate that some pickups are much too dark. I guess this is normal when showing them in 3D, and in fact this is an issue that's also very present in OG TR3.

What could be done to remedy it? Is there a smart way to calculate the lightness of the textures as they show in their given environment, and then boost them accordingly where needed?

Had anything in this regard been done for TR1X actually?

Thanks for tackling this, and all of your work, as always.

aredfan:
Not to mention some pickups are well camouflaged. 👇 

![20241213_074258_The_Cistern](https://github.com/user-attachments/assets/c40527a1-fdb9-4d7c-a6e5-d065e5bf4a15)

Richard-L:
Ah shoot. Was hoping it may not be a TR1 problem perhaps.

I wonder what options exist to make the textures of these pickups brighter, and whether that alone will be satisfactory.

Also makes me think of PS1's improved contrast levels which I wonder if they would be helpful.

lahm86:
All items have a shade (or two shades in TR2, although they're both generally the same value) and this is used to calculate sprite lighting. We could possibly re-use these shades for 3D, similar to how static meshes are lit. At the moment, it only uses the room's lighting.

Richard-L:
There's now the sparkle effect Lahm added. If at some point in the future someone's inclined to look at a lighting-only approach, here were the OpenLara attemps I'd posted:
> fwiw this problem is immediately evident the moment OG started using 3d pickups, so I'd say it's very okay to spend longer time on this. 
>
>![image](https://github.com/user-attachments/assets/30d4c7d5-0f8b-44c2-981f-39fd1433d21f)
> 
> Imo nothing should be added, the rendering of the items in-game would ideally be the solution itself. OpenLara, albeit not a fair comparison, still already improves it thanks to better item lighting:
>
>![image](https://github.com/user-attachments/assets/c414d6db-86aa-483d-9779-0ee0f8da2094)

Richard-L:
[...]

Issue #1646, opened 2023-11-05
PS1 contrast levels


Richard-L:
This was something Arsunt deliberated on for quite a while but eventually made it happen. He presented it like this:

![image](https://github.com/LostArtefacts/TRX/assets/4934209/ff322d05-8155-478e-88a3-51c7505eb1ad)

Arsunt:
> DirectX 5 has limited texture lighting formula. Every texture cannot be displayed lighter than it is. 
> But it can be displayed darker than it is. PS1 had full range of texture brightness, so PS1 is more contrast when it comes to dynamic lighting and bright rooms
> So I made 3 modes of lighting contrast:
> - Low: texture lighting is clamped in 0%-100% range (that's how behaved PC version with hardware renderer).
> - Medium: texture lighting is clamped in 0%-150% range (that's most comfort way to play in my opinion, compromise between Low and High).
> - High: texture lighting is clamped in 0%-200% range (that's how behaved PC version with software renderer and PS1 version).
> This is especially noticeable when you use flare or flashing with guns

TR2M before (DX5):
![DX5](https://github.com/user-attachments/assets/4122bf12-a8cf-4225-b9a3-a97f8f69442f)

TR2M after (DX7):
![DX7](https://github.com/user-attachments/assets/163d9095-25a8-49af-9f7c-a66566c04476)

TR2X:
![image](https://github.com/LostArtefacts/TRX/assets/4934209/af146ae4-481b-4dd8-ae42-898d777af60f)

PS1:
![image](https://github.com/LostArtefacts/TRX/assets/4934209/5a0ccb32-0520-4cbe-9064-154663b3fee2)

In some levels that are naturally dark it can be quite striking - compare the dragon:
PC: 
![pc-dragon](https://github.com/user-attachments/assets/badcd5e5-dac2-4d56-800a-0ae953b15dae)

PS1:
![ps1-dragon](https://github.com/user-attachments/assets/d2bbc851-7423-4de8-ad0f-e473bb93cc3b)

Richard-L:
Some of the above image links had broken and I fixed them.

While doing so, I found some examples of people trying TR2M's PS1 contrast levels in Golden Mask. The snow there looks too white, almost like a PS1 TR2-JP Tibet. Maybe due to these levels never meant to be played on the PS1? Something to perhaps keep an eye on for later.

![gm-pc-contrast1](https://github.com/user-attachments/assets/cd0a5e95-68cc-4e9b-ad50-9e683a02bbac)
[...]

PR #2045, opened 2024-12-15
tr1/output: use item shade to light 3D pickups

lahm86:
Part of #2017.

#### Checklist

#### Description

This uses the shade value on items to light 3D pickup objects, so they appear similarly lit to the sprite equivalent.

If this looks ok in TR1, we can apply similar to TR2 at a later date.

### DIFF
--- a/docs/tr1/CHANGELOG.md
+++ b/docs/tr1/CHANGELOG.md
@@ -37,6 +37,7 @@
 - fixed camera positions in Return to Egypt and Temple of the Cat (#1317, regression from 4.1)
 - fixed being able to see the flipmap in Natla's Mines when moving the boat (#2019)
 - fixed an invisible wall in Temple of the Cat, due to a wrongly positioned door (#2021)
+- improved lighting on 3D pickups (#2017)
 
 ## [4.6.1](https://github.com/LostArtefacts/TRX/compare/tr1-4.6...tr1-4.6.1) - 2024-11-25
 - added ability to disable saves completely by setting the save slot to 0 (#1954)
--- a/src/tr1/game/objects/common.c
+++ b/src/tr1/game/objects/common.c
@@ -187,8 +187,7 @@ void Object_DrawPickupItem(ITEM *item)
         item->interp.result.rot.y, item->interp.result.rot.x,
         item->interp.result.rot.z);
 
-    Output_CalculateLight(
-        item->pos.x, item->pos.y, item->pos.z, item->room_num);
+    Output_CalculatePickupLight(item);
 
     frame = object->frame_base;
     int32_t clip = Output_GetObjectBounds(&frame->bounds);
--- a/src/tr1/game/output.c
+++ b/src/tr1/game/output.c
@@ -724,6 +724,20 @@ void Output_CalculateObjectLighting(
     Output_CalculateLight(offset.x, offset.y, offset.z, item->room_num);
 }
 
+void Output_CalculatePickupLight(const ITEM *const item)
+{
+    if (item->shade >= 0) {
+        m_LsAdder = item->shade;
+        m_LsDivider = 0;
+        const int32_t distance = g_MatrixPtr->_23 >> W2V_SHIFT;
+        m_LsAdder += M_CalcFogShade(distance);
+        CLAMPG(m_LsAdder, 0x1FFF);
+    } else {
+        Output_CalculateLight(
+            item->pos.x, item->pos.y, item->pos.z, item->room_num);
+    }
+}
+
 void Output_DrawObjectMesh(const OBJECT_MESH *const mesh, const int32_t clip)
 {
     if (!M_CalcObjectVertices(mesh->vertices, mesh->num_vertices)) {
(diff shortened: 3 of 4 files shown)


rr-:
Thank you for taking it on, but I'm afraid this approach is still prone to not addressing the underlying issue. There are still many cases where the pickups blend in with the environment too much:

20241215_135309_rrm
20241215_135828_ebp
20241215_140007_omt

Plus it makes most of the pickups appear very flat:

20241215_135404_kra

I tried using the average of the 3D lighting + flat sprite-based lighting, but it looked less than great and didn't help with the pickups that blended with the textures.

I think a potential solution to this would be to keep the lighting, but add a small exclamation mark :warning: or a similar 2D icon above the pickups if they're anywhere near the center of the screen, within reasonable distance and within the line of sight.

20241215_140939_rpf
20241215_140936_yun
20241215_140928_jwl

then the problem can be reduced to finding an icon that feels compatible with the OG.
I can imagine the builders who will want to be extra petty and have their pickups blend in with the textures purposefully, will complain about this, but they can always forcefully disable the pickup aids.

lahm86:
No problem, I agree it does make things quite flat but I suppose that's the nature of static lighting (which works fine for sprites really, but not so much 3D).
Happy to go with the icon approach. Should we just close this PR and tackle it afresh?

Richard-L:
Fwiw I hate the ! because it's so TRR.

Any other dynamic lighting solutions?

rr-:
> Fwiw I hate the ! because it's so TRR.
> 
> Any other dynamic lighting solutions?

I suppose we _could_ go to the trouble of shading each item separately… but man, that's a lot of trouble.
The pickup icon aids look like a good separate feature regardless whether we personally adore them or no – it feels like a good QoL option. Plus I think that given a decent sprite, they _could_ feel natural and like a part of the OG.

I'd keep this PR open until we have settled on an alternative as the way I see it we're still brainstorming the final solution.

Richard-L:
@lahm86 sorry if I'm doing something wrong, but using the build from here https://github.com/LostArtefacts/TRX/pull/2045#issuecomment-2543725452 I get the same outcome.
image

walkawayy:
I've seen games do different things.

RE4: Remake light beam:

image

Horizon Forbidden West is similar but with icons on top of the lightbeam:

image

I always liked Tomb Raider Legend's shimmer light effect that runs across the item every few seconds:

image

AOD does something similar with a small green light flash (and a small hand in the bottom right shows when near an interactable when close):

image

Tomb Raider Underworld hovers the interact button's icon over items when close:

image

lahm86:
> @lahm86 sorry if I'm doing something wrong, but using the build from here [#2045 (comment)](https://github.com/LostArtefacts/TRX/pull/2045#issuecomment-2543725452) I get the same outcome. image

This is only relevant for TR1 currently, so that's to be expected.

lahm86:
I don't think we can improve on the "flat shade" by going through every item and tweaking the values. The data already has differing values that have been selected appropriately IMO for the sprites e.g. the medi below in Vilcabmamba retains a dark shade and the gold idol below is also lit appropriately.

image

I don't feel that tweaking these values would be of benefit: the lighting would still be flat on the 3D pickups. Unless there is a way perhaps to change the output logic e.g. item shade + room light. But then builders possibly have less control over it.

Richard-L:
fwiw this problem is immediately evident the moment OG started using 3d pickups, so I'd say it's very okay to spend longer time on this.
image

Imo nothing should be added, the rendering of the items in-game would ideally be the solution itself. OpenLara, albeit not a fair comparison, still already improves it thanks to better item lighting:
image

rr-:
> fwiw this problem is immediately evident the moment OG started using 3d pickups, so I'd say it's very okay to spend longer time on this. image
> 
> Imo nothing should be added, the rendering of the items in-game would ideally be the solution itself. OpenLara, albeit not a fair comparison, still already improves it thanks to better item lighting: image

The large medpack is barely visible, though.

Richard-L:
Agreed, but there is an improvement imo, and I just wanted to share it as an example.

EDIT: Maybe is this something that with an enabled #1646 can be remedied?

lahm86:
@Richard-L What do you think about the following approach? Keep an eye on the pickups as Lara gets close to them.

https://youtu.be/NtnJu2ofMB0

Richard-L:
So kind of you to ask for my opinion and even make a quick clip ❤️ 

I think this is super clever. It uses the ricochet sprites, right? It's subtle and nice, and I think suitable for this case.

Please forgive that I would probably still choose to deactivate it, just because I find TR more immersive without any hints that tell you "I'm a game you can pick this up", if that makes sense.

Perhaps at some point in the later future, something with just mesh reflection and luminance might be done.

lahm86:
> So kind of you to ask for my opinion and even make a quick clip ❤️
> 
> I think this is super clever. It uses the ricochet sprites, right? It's subtle and nice, and I think suitable for this case.
> 
> Please forgive that I would probably still choose to deactivate it, just because I find TR more immersive without any hints that tell you "I'm a game you can pick this up", if that makes sense.
> 
> Perhaps at some point in the later future, something with just mesh reflection and luminance might be done.

It uses the twinkle effect from when Lara is turning to gold, or turning lead to gold. It's just slightly larger in size as the original wasn't very clear. We thought it was quite in-keeping overall with the game. Glad you think it's a suitable approach too.

And it will be optional 😌 

Thanks for taking a look, and the feedback 👍

lahm86:
Closing this as #2077 has provided an alternative approach. The issue itself can remain open for any further investigation in terms of lighting.

PR #2059, opened 2024-12-17
tr2/flares: fix flare throwing priority

### LINKED ISSUE #1989: Feature request: Prevent Lara from automatically throwing away a flare while in the air
[FLEP](https://github.com/FlexiblePatcher/FlexiblePatcher) has a "safe flare throw" patch for custom TRNG levels that prevents Lara from automatically throwing away a used flare while jumping or falling, which would be a nice feature to have in TR2 since it helps prevent deaths where Lara prioritizes throwing away a flare over grabbing a ledge/ladder.

lahm86:
Resolves #1989.

#### Checklist

#### Description

This performs stricter checks for when a flare should be automatically thrown away to avoid Lara being unable to grab ledges. So if Lara is mid-jump, or the player has jump input pressed, a spent flare will stay in her hand until she's otherwise safe to throw it. If however Lara is in fast fall state, or swan-diving, a spent flare will still be thrown because she cannot grab anyway in these states.

If the player actively discards a flare while mid-air, it will continue to be respected. The bug is in the automatic decision to throw it.

Something else I think we should consider is reverting flare input debouncing, which was introduced [here](https://github.com/LostArtefacts/TR2X/pull/161) as a workaround for the console key conflict. There may be input-sensitive gameplay where you want to tactically throw a flare at the right moment e.g. dropping off a ladder onto a slope before jumping off the slope somewhere else. This would be quite difficult with debouncing. We can raise this separately however.

### DIFF
--- a/docs/tr2/CHANGELOG.md
+++ b/docs/tr2/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## [Unreleased](https://github.com/LostArtefacts/TRX/compare/tr2-0.7.1...develop) - ××××-××-××
 - added an option to use PS1 contrast levels, available under F8 (#1646)
+- fixed Lara prioritising throwing a spent flare while mid-air, so to avoid missing ledge grabs (#1989)
 
 ## [0.7.1](https://github.com/LostArtefacts/TRX/compare/tr2-0.7...tr2-0.7.1) - 2024-12-17
 - fixed a crash when selecting the sound option (#2057, regression from 0.6)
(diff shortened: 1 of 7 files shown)


rr- (commented):
Let's make a gameplay fix setting for it that defaults to enabled state.
As for the debouncing, I agree that it should be tackled separately.

lahm86:
> Let's make a gameplay fix setting for it that defaults to enabled state. As for the debouncing, I agree that it should be tackled separately.

Thanks, config setting now in place. I'll raise an issue for the debouncing.

aredfan (approved):

walkawayy (approved):
Very nice fix! I always hated when this happened.

rr- (approved):

PR #2077, opened 2024-12-19
tr1/pickup: add option for pickup aids

### LINKED ISSUE #2076: Feature request: add pickup highlighting
Re #2017, add an option to intermittently highlight pickups when Lara is nearby.

lahm86:
Resolves #2076.

#### Checklist

#### Description

This adds an option to highlight pickup items, by spawning twinkle effects around the items when Lara is nearby. This is provided with an injection file with a modified Midas twinkle sprite sequence. Custom levels can specify their own in slot 185.

Items will highlight if Lara is within a 1.25 to 8 tile range - any nearer or further and the twinkling will stop. The effect will spawn every 2.5-5 seconds per item, with a little added randomness mainly for pickups that are close together.

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/21


src/tr1/config.h
```diff
@@ -111,6 +111,10 @@ typedef struct {
         int32_t cntlr_layout;
     } input;
 
+    struct {
+        bool highlight_pickups;
+    } visuals;
```
rr-: Maybe it should be `gameplay.enable_pickup_aids`?

src/tr1/game/objects/general/pickup.c
```diff
@@ -76,9 +88,53 @@ static void M_GetAllAtLaraPos(ITEM *item, ITEM *lara_item)
...
+
+    const GAME_VECTOR pos = {
+        .x = item->pos.x + 20 * (Random_GetDraw() - 0x4000) / 0x4000,
+        .y = item->pos.y - ABS(frame->bounds.max.y - frame->bounds.min.y) - 10
+            - 10 * (Random_GetDraw() - 0x4000) / 0x4000,
+        .z = item->pos.z + 20 * (Random_GetDraw() - 0x4000) / 0x4000,
+        .room_num = item->room_num,
+    };
+    PickupTwinkle_Spawn(&pos);
+}
```
rr-: I don't think we should manipulate object state or spawn new effects in the draw routines, which should be mostly immutable with regard to the objects and items, and instead would opt to handle this in the control (or possibly collision) routine. By doing so, we can eliminate the need for the check for the current phase, and avoid having to reference the clock, by working with logical frames instead.

rr- (commented):
As far as I remember, submerged mid-air pickups slowly descend until they land on the floor. In the possible refactors we'd need to make sure to support this as well. Usually it's done by storing additional variables in `FX->counter` – but we already need that for the lifecycle bookkeeping, so I'd be OK with introducing `void *priv` similar to `item->priv` (that's expected to be not persisted across saves).

src/tr1/game/objects/general/pickup.c
lahm86: I forgot to touch on this. I had it initially in the collision routine, but because of the min distance checks [here](https://github.com/LostArtefacts/TRX/blob/develop/src/tr1/game/lara/control.c#L132), pickup collision doesn't necessarily run when the item is visible. I think we could introduce a control routine for pickups though as an alternative?

lahm86: Actually, this is trickier as pickup items aren't classed as active so aren't included in the `Item_Control` loop.

rr-: What if we count them as active?
Or rather than using its own FX, we just draw the sparkle mesh manually in pickup's drawing routines?

lahm86: We'd need to activate items on load depending on visibility, monitor triggered items, carried items etc. Not sure if it would impact saves. I can try to investigate tomorrow (I'm running short on time this evening).
Otherwise, potentially a different collision test approach for pickup items could be a more streamlined approach.

src/tr1/config.h
lahm86: Done

rr-: Thanks

lahm86:
I've changed the approach to go with item control rather than spawning effects in the draw routine, and the alternative of doing the checks from the collision routine. This is because of the way nearby rooms are calculated for collision (via immediate vis portals), meaning some scenarios would result in pickups not being highlighted.

Something to check for in tests is saving/loading before and after picking something up, to make sure no stray effects linger. The other case is where pickup items are triggered, such as in Natla's Mines. This seems OK for me, but would appreciate a second look.

Re drop items, I have for the time being disabled the effect if an item is falling. Because of the delay, you would only see the effect when it first appears in the air, then usually once the timer is up, the item has already landed. For longer drops, the item would likely be deemed too far away to have the effect until it gets closer to the ground. Maybe we could tackle it in a rev2 if we think it's still needed.

This is also rebased against the latest develop, and I've changed the various names to "pickup aid" throughout including the bin file.

aredfan (approved):

aredfan:
Thank you for working on this feature. I'm already looking forward to using the pickup aids in my future playthroughs. 👍

src/tr1/game/objects/general/pickup.c
```diff
@@ -50,15 +59,75 @@ static const OBJECT_BOUNDS m_PickUpBoundsUW = {
...
+        return;
+    }
+
+    const int32_t distance = Item_GetDistance(lara, &item->pos);
+    if (distance < AID_DIST_MIN || distance > AID_DIST_MAX) {
+        return;
+    }
+
+    int32_t timer = (int32_t)(intptr_t)item->priv;
+    if (timer <= 0 || (timer < AID_WAIT_MIN && (Random_GetDraw() % 7) == 0)) {
```
rr-: Thanks. My only nitpick is that the more common way to use `Random_GetDraw()` is by doing `Random_GetDraw() < 0x4000`, instead of using the modulo operator like `Random_GetDraw() % 7 == 0`.

rr- (approved):

src/tr1/game/objects/general/pickup.c
lahm86: Thanks, I've updated this to stick to the conventions.

rr- (approved):

PR #2089, opened 2024-12-20
tr1/carrier: fix initial sector selection for falling items

### LINKED ISSUE #2088: TR1X bug: falling drop items sometimes disappear
If a flyer drops a pickup item and the flyer's position is exactly at the height of vertical portal, the drop item can disappear when it lands, although it can still be collected.

lahm86:
Resolves #2088.

#### Checklist

#### Description

This ensures that if an item is dropped exactly at the height of a vertical portal, we get the room just above that portal first to ensure that later the item is placed in the correct room, and hence drawn there.

This is tricky to test, but I have confirmed the fix works by hard-coding a flyer's position to be exactly on a portal.

### DIFF
--- a/docs/tr1/CHANGELOG.md
+++ b/docs/tr1/CHANGELOG.md
@@ -1,4 +1,5 @@
 ## [Unreleased](https://github.com/LostArtefacts/TRX/compare/tr1-4.7...develop) - ××××-××-××
+- fixed falling pickup items not being drawn when they land in rare cases (#2088)
 
 ## [4.7](https://github.com/LostArtefacts/TRX/compare/tr1-4.6.1...tr1-4.7) - 2024-12-20
 - added support for Wayland in binary Linux builds (#1927)
--- a/src/tr1/game/carrier.c
+++ b/src/tr1/game/carrier.c
@@ -79,8 +79,11 @@ static void M_AnimateDrop(CARRIED_ITEM *const item)
 
     ITEM *const pickup = Item_Get(item->spawn_num);
     int16_t room_num = pickup->room_num;
-    const SECTOR *const sector =
-        Room_GetSector(pickup->pos.x, pickup->pos.y, pickup->pos.z, &room_num);
+    // For cases where a flyer has dropped an item exactly on a portal, we need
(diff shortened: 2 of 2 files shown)


rr-:
Doesn't the proposed approach keep the same issue, only 10 pixels lower/higher?

lahm86:
Once the item falls through the portal, for a few frames it will still be in the room above with this approach, but because the minimum room height is `STEP_L`, it will eventually reach the correct room below. An alternative would be to get the sector at the top of the pickup item by checking its bounds but in theory you could have ones higher than `STEP_L`, so I opted for 10 as we only need a minor nudge up.

lahm86:
It's loosely a similar issue to https://github.com/LostArtefacts/TRX/pull/628 where it was too hastily going into the room below for bridges.

rr- (approved):
Since it's a tricky fix, let's capture the context in a comment. Other than that LGTM.

aredfan (approved):

lahm86:
> Since it's a tricky fix, let's capture the context in a comment. Other than that LGTM.

Thanks. Comment added, and I've made a quick video in case we need to refer back to this.

https://youtu.be/QqIlZ9IOMZA

I'll note the hard-coded part in `mutant.c` here in case we need to test this again (specific to Sanctuary as above).

```diff
diff --git a/src/tr1/game/objects/creatures/mutant.c b/src/tr1/game/objects/creatures/mutant.c
index b6bb1e75..ce8632f8 100644
--- a/src/tr1/game/objects/creatures/mutant.c
+++ b/src/tr1/game/objects/creatures/mutant.c
@@ -118,6 +118,13 @@ void Mutant_FlyerControl(int16_t item_num)
         item->status = IS_ACTIVE;
     }
 
+    item->pos.x = 48640;
+    item->pos.y = -5120;
+    item->pos.z = 57856;
+    if (item->room_num != 10) {
+        Item_NewRoom(item_num, 10);
+    }
+
     CREATURE *flyer = item->data;
     int16_t head = 0;
     int16_t angle = 0;
```

rr- (approved):

rr- (approved):

PR #2109, opened 2024-12-24
Effects refactor

rr-:
#### Checklist

#### Description

Internal refactor.

The chaotic output from the decompilation phase has been reorganized into three distinct modules. I've also updated TR1 to align with these changes.

- **`effects`**: This module is dedicated to the management of effects. It does not deal with specific objects and does not interact with game mechanics.
- **`item_actions`**: Previously known as `g_EffectRoutines`, this module is responsible for implementing parts of the game mechanics closely linked to the trigger logic from animation commands. Even though it might be tempting to introduce new actions here for our purposes, I advise consulting with the TombEditor team before doing so.
- **`spawn`**: This module was previously a collection of various effect spawning functions. Whenever multiple objects require the spawning of the same effect (or an item, for that matter), the code for it should likely land here.

Finally, a major update involved transforming `Effect_ExplodingDeath` into a direct API within the `items` module as `Item_Explode`. This decision is somewhat debatable, as it involves the creation of `O_BODY_PART` – a unique object without its own meshes – placing it slightly within the game mechanics territory, which the `items` module ideally shouldn't deal with. However, I couldn't identify a better alternative.

I decided against placing `Blood_Spawn` within `blood.c` or similar locations. Though this was the approach we'd been trying to follow in TR1, I opted to revert it to utilize the new `spawn.c` file instead, as the previous solution felt half-baked.

It'd be good to test it for effect regressions (gunshots, blood, ricochets, twinkles, Dragon light, waterfall mist, water splashes, bubbles, flamethrowers, etc.) in both TR1 and TR2.


aredfan:
**List of issues:**

TR2X

- [x] Killing a xian statue while midair causes the game to crash.
- [x] Killing a xian statue while grounded causes artifacts. Can also cause crash.
- [x] Bartoli's Hideout room 6. Set Lara on fire and have her jump into the water. The other fire sprites disappear and there are persistent bubbles. Can also cause crash.
- [x] Similar to above but in the Dragon's Lair. Allow the dragon to set Lara on fire and have her jump into the water.  There's a persistent fire sprite on the water surface. Can also cause crash.
- [x] Near the start of The Deck. The flamethrower's fire sprites is persistent at the end of the range, also causes crash.
- [x] Indefinite freeze during TR2X demo mode (drydenj)

TR1X looks good to me.

drydenj:
Indefinite freeze during TR2X demo mode at this moment: 
g4d04928 crash1

rr-:
Distilled the bug to be  caused by 167bfb9. I'll later investigate why, as the changes look innocent with the naked eye.

src/tr2/game/effects.c
```diff
@@ -81,42 +104,42 @@ int16_t __cdecl Effect_Create(const int16_t room_num)
...
+            if (m_Effects[link_num].next_free == effect_num) {
+                m_Effects[link_num].next_free = effect->next_free;
                 break;
             }
-            link_num = g_Effects[link_num].next_free;
+            link_num = m_Effects[link_num].next_free;
         }
     }
 
+    room = Room_Get(effect->room_num);
```
lahm86: I think this should be `Room_Get(room_num);`.

rr-: Thank you – that was it.

rr-:
@aredfan this is now ready for re-testing.

lahm86 (approved):

aredfan (approved):
Everything LGTM. 👍 

Btw I found a crash in TR1X but it's also present in dev snapshot.  I'll write an issue for it shortly.

PR #2126, opened 2024-12-27
Introduce phasers to TR2 (part 1)

rr-:
#### Checklist

#### Description

Refactors certain parts of the game use phasers known from TR1X:
- cutscenes
- picture display
- credits
- end of the level stats
- final game stats

The proposed phaser API has been simplified compared to what we have in TR1X. The most significant change is the removal of the phaser switch logic. Instead of `Phase_Set`, we now have `PhaseExecutor_Run(PHASE *phase)`, which I think is more straightforward. However, I'm not entirely sure if this design will hold up. The'll get truly battle-tested once I start working on the game and inventory phasers, as these often interact with each other. I can see them potentially getting us in trouble when called mid-control routines, right before the draw call from the outer loop.

As a bonus, I revamped the end-of-level stats screen, final game stats screen, stopwatch stats dialog, and the assault course stopwatch stats dialog to utilize the UI framework. The assault course stats can display up to 10 records, though we only show 7 rows in the UI. Traditionally this sort of scenario is managed by `REQUESTER`s, but I'm trying to phase these out, so I needed to mimic this behavior within the UI framework. I had two options: modify the item texts on the fly, which isn't ideal as some texts are centered, some appear in two columns, etc. – or simply hide the widgets. I went with the latter, and it even proved to be efficient (I've tested it with 1000 rows without lag), but it isn't this scalable yet since `Text_Create` relies on a fixed buffer, risking a crash if we exceed the TEXTSTRING limit (eg. 128 allocated texts, visible or not). A solution would be to make the TEXTSTRING array a variable size, or dynamically remove and create texts for the hidden UI widgets. I lean toward the latter approach.


lahm86:
Looks great. Just a couple of issues I've noticed:
- If you spam action/enter as soon as you start the game, the title screen seems to retain the fade overlay from the license picture
20241227_091518_Laras_Home

- The end of level stats screen is missing the time taken (top row)
image

aredfan:
- [x] When the main menu ring closes, the backdrop gets a tiny bit brighter for a frame or so.
- [ ] ~~At the start of HSH, the fixed camera that is focused on enemies gets skipped.~~

rr-:
> Looks great. Just a couple of issues I've noticed:
> 
> * If you spam action/enter as soon as you start the game, the title screen seems to retain the fade overlay from the license picture
>   20241227_091518_Laras_Home
> * The end of level stats screen is missing the time taken (top row)
>   image

Both issues are fixed. I believe the latter also occurs in the develop.

> When the main menu ring closes, the backdrop gets a tiny bit brighter for a frame or so.

Fixed as well.

lahm86 (approved):

aredfan (approved):

PR #2130, opened 2024-12-27
tr2/camera: reset last camera on initialise

### LINKED ISSUE #2129: TR2X bug: HSH intro fixed camera can get stuck or skipped when using cheats

_________________

The fixed camera (the one focused on enemies in the courtyard) gets stuck with these steps:
1. On the main menu, execute `/play 18`
2. Let the intro play out fully

The fixed camera gets skipped with these steps:
1. `/play 17`
2. Use the skip level cheat
4. Let the intro play out fully (don't move Lara)
5. `/play 18`
6. Let the intro play out fully again

lahm86:
Resolves #2129.

#### Checklist

#### Description

I *think* the actual issue was that `g_Camera.last` held garbage data in previous builds when loading straight in to HSH, and so the test [here](https://github.com/LostArtefacts/TRX/blob/develop/src/tr2/game/room.c#L319) was allowing the timer to subsequently be set and then everything would play out normally. In recent builds, `g_Camera.last` initialises to `0` but this is a valid camera number (and is the only one in HSH), so because the level starts with an in-game cutscene, `Camera_Update` does not set `g_Camera.last` as it would normally and hence we get the odd behaviour when jumping between levels. When going to Dragon's Lair, then HSH, Lair will have set it to `-1` already.

I think this is specific to this scenario e.g. special start anim followed by an immediate fixed camera, but it's safer all round to just reset the last camera with normal initialisation IMO.

### DIFF
--- a/docs/tr2/CHANGELOG.md
+++ b/docs/tr2/CHANGELOG.md
@@ -10,6 +10,7 @@
 - fixed Lara prioritising throwing a spent flare while mid-air, so to avoid missing ledge grabs (#1989)
 - fixed software renderer not applying underwater tint (#2066, regression from 0.7)
 - fixed some enemies not looking at Lara (#2080, regression from 0.6)
+- fixed the camera getting stuck at the start of Home Sweet Home (#2129, regression from 0.7)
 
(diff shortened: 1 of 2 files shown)


aredfan (approved):
LGTM, and thank you. 👍

rr- (approved):
Thank you for quickly finding a fix for this.

Issue #1615, opened 2023-11-09
Feature request: Activate sunglasses option for Main Menu and inventory


aredfan:
We'll need the sunglasses option for some graphic settings such as LostArtefacts/TRX#1636.

PR #2158, opened 2024-12-30
TR2: add UI scaling (without UI to control it yet)

### LINKED ISSUE #1636: Add UI scaling option
Same as TR1X

### LINKED ISSUE #2012: Don't stretch UI text if non 4:3 aspect ratio is set
Sorry that I can't test if this is an OG PC bug.

Currently the text will stretch depending on the window size.

### LINKED ISSUE #2149: Fix bar scaling
Currently health bar and air bar are always drawn at the same size of around 100 pixels, regardless of the window size. We should apply the same scaling algorithm as in TR1.

rr-:
#### Checklist

#### Description

Fixes bar and text scaling, which resolves #2012 and #2149, and adds options for bar and text scaling which resolves #1636 – however, these options are presently only available in the JSON config and via the `/set` commands.

I didn't play with the repositioning the UIs to make it always look pretty – I'd like to tackle that in separate tickets. For the UI part we have #1615.


lahm86 (approved):
LGTM. Agreed it makes sense to tackle UI repositioning and suchlike separately.

aredfan (approved):
LGTM 👍

Richard-L:
Big improvement to have this, hoping for more bar love soon 🙏

PR #2181, opened 2024-12-31
stats: respect turbo cheat

### LINKED ISSUE #2167: Ingame timer doesn't respond to turbo speed
Observed in the 265 build.

rr-:
#### Checklist

#### Description

Resolves #2167.

Now that 5785b0e65 has eliminated the clock drift problems, we can return to simple frame-based timer increments.
I've compared the current and proposed approaches for normal game speed. There are some initial differences of up to 4 frames due to loading times, but they remain stable and don't increase over time.
I kept the original code behind `#ifdef` but I think we maybe should remove it after all. I'd like to defer the final decision to @lahm86.

### DIFF
--- a/docs/tr1/CHANGELOG.md
+++ b/docs/tr1/CHANGELOG.md
@@ -2,6 +2,7 @@
 - added an option for pickup aids, which will show an intermittent twinkle when Lara is nearby pickup items (#2076)
 - added an optional demo number argument to the `/demo` command
 - changed demo to be interrupted only by esc or action keys
+- changed the turbo cheat to also affect ingame timer (#2167)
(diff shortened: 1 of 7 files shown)


lahm86 (approved):
LGTM.
I think it's fine to remove the previous code, we always have the repo history if we need to consult back for any reason.

aredfan:
Issue in TR2X:

- [x] Turbo speed 2 isn't faster than 1 when used for the game's stopwatch in the story levels, for example Floating Islands. The assault course timer isn't affected by this issue.

rr-:
@aredfan fixed.

aredfan (approved):

PR #2183, opened 2024-12-31
tr2/phaser: sync clock at phase start/resume

rr-:
#### Checklist

#### Description

Fixes a bug found by Lahm where playing an FMV causes the clock to not update. This results in the next phase trying to catch up on frames it thinks were missed. Translated to a real world example, this means that playing the Great Wall FMV would cause the game to start with Lara already having descended the opening slopes. The solution is to reset the clock's last tick value at the start of a phase. I've also applied this reset to the resume operation, e.g. when returning from a nested phase to an outer phase (for example whenever the player closes the inventory and goes back to the game).


lahm86 (approved):
Thank you

aredfan (approved):

PR #2195, opened 2025-01-02
Merge config structures and code

### LINKED ISSUE #2194: TRX: merge config code
Right now, client games define their own config structures within their own code, and libtrx doesn't know about them. This design is slowing us down because we have to keep creating new abstractions that aren't very solid. Instead, we should create a good common config API, with the pluggable parts being reduced to the config structure alone, and `map.def` files, and a sanitizing function. The side effects from config changes should be handled through either an additional single pluggable function or, ideally, using the event-driven programming model we've already set up.

rr-:
#### Checklist

#### Description

Resolves #2194 and moves the config implementation to libtrx, while trying to keep the specialized code to minimum. 

While I hoped to remove the customized loading routines, unfortunately we do still require them in order to keep backwards compatibility with old TR1X versions, which requires logic that cannot be captured via the .def files. Eventually we can remove these settings, as this logic is only relevant to the first launch after an upgrade from an old TR1X version.


aredfan:
TR2X issue (**Edit:** same thing happens with TR1X as well):

I get this popup when opening the config tool.



If I delete TR2X.json5 then launch TR2X so it creates a new one, then open the config tool - I get this popup.

rr-:
Since I haven't modified the config tool in this PR and the problem is specific to it, I suspect it might occur in version 0.8 as well. Can we confirm this? Thanks.

rr-:
> TR2X issue (**Edit:** same thing happens with TR1X as well):
> 
> I get this popup when opening the config tool.
> 
> 
> 
> If I delete TR2X.json5 then launch TR2X so it creates a new one, then open the config tool - I get this popup.
> 
> 

Thanks – fixed. The problem was that the game was saving the screenshot format as `"jpeg"` instead of `"jpg"`. While the game accepts both, the config tool does not, and we should always use `"jpg"`. Previously, it was left to chance which one would end up used, and it worked only by fluke.

The issue should go away on its own after relaunching the newest snapshot once.

lahm86 (approved):

aredfan (approved):

PR #2201, opened 2025-01-02
tr1/config: group settings semantically

rr-:
#### Checklist

#### Description

We no longer throw all of our settings into the top-level struct, and instead group them semantically.
While they don't fully reflect the layout proposed in the config tool, it already offers a solid improvement, and we can revisit this once we move towards 100% ingame config.


lahm86:
Looks good, the only issue is the renaming of `layout` and `cntlr_layout` means players will have to re-select that. Could we add a check in `M_LoadLegacyOptions`?

rr-:
> Looks good, the only issue is the renaming of `layout` and `cntlr_layout` means players will have to re-select that. Could we add a check in `M_LoadLegacyOptions`?

It should work as-is, as these values are read with specialized routines that use hardcoded key names, and I didn't change them.

lahm86:
> > Looks good, the only issue is the renaming of `layout` and `cntlr_layout` means players will have to re-select that. Could we add a check in `M_LoadLegacyOptions`?
> 
> It should work as-is, as these values are read with specialized routines that use hardcoded key names, and I didn't change them.

It doesn't seem to be working for me e.g. if I have `"layout": 1` in the current config and load with this build, it reverts to `"keyboard_layout": 0`

The actual assignments are retained, sorry I probably wasn't clear on that. The issue is the layout reverts to the default as the selected one.

rr-:
> > > Looks good, the only issue is the renaming of `layout` and `cntlr_layout` means players will have to re-select that. Could we add a check in `M_LoadLegacyOptions`?
> > 
> > 
> > It should work as-is, as these values are read with specialized routines that use hardcoded key names, and I didn't change them.
> 
> It doesn't seem to be working for me e.g. if I have `"layout": 1` in the current config and load with this build, it reverts to `"keyboard_layout": 0`
> 
> The actual assignments are retained, sorry I probably wasn't clear on that. The issue is the layout reverts to the default as the selected one.

Thanks, I misunderstood the original comment. I've added fallback readers.

lahm86 (approved):

aredfan (approved):

PR #2202, opened 2025-01-02
libtrx/anims/types: introduce frame size for animations

lahm86:
#### Checklist

#### Description

This reduces repeated shifts on the interpolation property by introducing a frame size property on the anim struct. This only applies to TR2.

Bones refactor after this PR.

### DIFF
--- a/src/libtrx/include/libtrx/game/anims/types.h
+++ b/src/libtrx/include/libtrx/game/anims/types.h
@@ -52,7 +52,8 @@ typedef struct __PACKING {
 #elif TR_VERSION == 2
     int16_t *frame_ptr;
 #endif
-    int16_t interpolation;
+    uint8_t interpolation;
+    uint8_t frame_size;
     int16_t current_anim_state;
     int32_t velocity;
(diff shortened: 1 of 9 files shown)


rr- (commented):
Question, can we make an artificial `frame_size` for TR1 with the value of 1? (I'm looking at the differences in `Item_GetFrames`)
(Would that even make sense?)

lahm86:
> Question, can we make an artificial `frame_size` for TR1 with the value of 1? (I'm looking at the differences in `Item_GetFrames`) (Would that even make sense?)

Sure. This is what we do in the rando reader library, it just defaults to 0. Will do that.

lahm86:
That's done. I think we may be able to work out the equivalent frame size for TR1 once we get into frame parsing properly later. But for now at least, the struct is similar for both.

Adding for later reference, I _think_ for TR1 it will simply be `2 * num_meshes + 9`. It varies in TR2+ because of superpacking.

aredfan (approved):

rr- (approved):

PR #2206, opened 2025-01-03
tr2/lara/control: fix testing triggers too early

### LINKED ISSUE #2205: TR2X bug: Lara triggers the rat early when climbing out of the water

_________________

Venice, room 57. 

I've been doing frame-by-frame analysis and have found the rat spawns too early.  The screenshots below shows the exact moment the rat is spawned.

**Latest:**

**0.6:**

In the 'latest' screenshot, Lara is spawning the rat too early, the trigger happens in an earlier frame of Lara's climbing animation.

This issue affects the camera behaviour during demo 1, but only very slightly and doesn't cause a desync.

lahm86:
Resolves #2205.

#### Checklist

#### Description

This puts Lara's calls to test triggers before her animation updates, to be in line with the original game, which cached the trigger index from earlier calls.

This will obviously affect anything Lara triggers so worth a few tests/comparisons I'd say. I'm also wondering if there is a similar issue in TR1 since we refactored the floor data in the same way. We can perhaps create a test level to mimic this part of Venice in TR1 to make comparing with TombATI easier, but that can be tackled on its own.

### DIFF
--- a/docs/tr2/CHANGELOG.md
+++ b/docs/tr2/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## [Unreleased](https://github.com/LostArtefacts/TRX/compare/tr2-0.8...develop) - ××××-××-××
 - added Linux builds and toolchain (#1598)
+- fixed Lara activating triggers one frame too early (#2205, regression from 0.7)
 
 ## [0.8](https://github.com/LostArtefacts/TRX/compare/tr2-0.8...tr2-0.8) - 2025-01-01
 - completed decompilation efforts – TR2X.dll is gone, Tomb2.exe no longer needed (#1694)
--- a/src/libtrx/include/libtrx/game/collision.h
+++ b/src/libtrx/include/libtrx/game/collision.h
@@ -63,7 +63,6 @@ typedef struct __PACKING {
     int16_t facing;
     int16_t quadrant;
     int16_t coll_type;
-    int16_t *trigger; // TODO: linked to g_TriggerIndex, so to be eliminated
     int8_t x_tilt;
     int8_t z_tilt;
     int8_t hit_by_baddie;
--- a/src/tr2/game/lara/control.c
+++ b/src/tr2/game/lara/control.c
@@ -201,7 +201,6 @@ void Lara_HandleAboveWater(ITEM *const item, COLL_INFO *const coll)
     coll->old_anim_num = item->anim_num;
     coll->old_frame_num = item->frame_num;
     coll->radius = LARA_RADIUS;
-    coll->trigger = NULL;
 
     coll->slopes_are_walls = 0;
     coll->slopes_are_pits = 0;
@@ -253,6 +252,7 @@ void Lara_HandleAboveWater(ITEM *const item, COLL_INFO *const coll)
 
     Lara_Animate(item);
 
+    Room_TestTriggers(item);
     if (!g_Lara.extra_anim && g_Lara.water_status != LWS_CHEAT) {
         Lara_BaddieCollision(item, coll);
         if (g_Lara.skidoo == NO_ITEM) {
@@ -262,7 +262,6 @@ void Lara_HandleAboveWater(ITEM *const item, COLL_INFO *const coll)
 
     Item_UpdateRoom(item, -LARA_HEIGHT / 2);
     Gun_Control();
-    Room_TestTriggers(item);
 }
 
 void Lara_HandleSurface(ITEM *const item, COLL_INFO *const coll)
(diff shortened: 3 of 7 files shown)


rr-:
> Resolves #2205.
> 
> #### Checklist
> * [x]  I have read the [coding conventions](https://github.com/LostArtefacts/TR1X/blob/master/CONTRIBUTING.md#coding-conventions)
> * [x]  I have added a changelog entry about what my pull request accomplishes, or it is an internal change
> * [x]  I have added a readme entry about my new feature or OG bug fix, or it is a different change
> 
> #### Description
> This puts Lara's calls to test triggers before her animation updates, to be in line with the original game, which cached the trigger index from earlier calls.
> 
> This will obviously affect anything Lara triggers so worth a few tests/comparisons I'd say. I'm also wondering if there is a similar issue in TR1 since we refactored the floor data in the same way. We can perhaps create a test level to mimic this part of Venice in TR1 to make comparing with TombATI easier, but that can be tackled on its own.

Yes let's definitely check TR1.

rr- (commented):
The changes look good to me. @lahm86 I didn't replicate your analysis, so just to confirm – I noticed we update Lara's position after `Lara_Animate`. Does the new code adhere to the OG in this aspect as well?
Thanks.

lahm86:
> The changes look good to me. @lahm86 I didn't replicate your analysis, so just to confirm – I noticed we update Lara's position after `Lara_Animate`. Does the new code adhere to the OG in this aspect as well? Thanks.

Actually, I've realised a mistake. The important thing is to test triggers just before we run Lara's collision routine, because that's where `g_TriggerIndex` would have been set in OG. I got mixed up thinking it would be prior to `Lara_Animate` originally.

Apologies @aredfan if you've already looked at this, the build is updated again.

rr- (approved):

aredfan:
> Apologies @aredfan if you've already looked at this, the build is updated again.

No worries, I was out all morning.  This PR fixes the demo 1 issue, thank you. 👍 

I found an issue with the airlock where the crank wheel is moving a little faster than the door, when Lara opens it.

**PR build:**
PR

**Dev snapshot:**
Dev snapshot

rr-:
> > Apologies @aredfan if you've already looked at this, the build is updated again.
> 
> No worries, I was out all morning. This PR fixes the demo 1 issue, thank you. 👍
> 
> I found an issue with the airlock where the crank wheel is moving a little faster than the door, when Lara opens it.
> 
> **PR build:** PR
> 
> **Dev snapshot:** Dev snapshot

I'm starting to think that it might not be possible to recover the OG behavior 100% consistently with every possible scenario, as the `Room_GetHeight` calls were quite literally all over the place and I think it's very likely that some places did _not_ call it even though they should. Let's wait for Lahm's reply, though.

lahm86:
I'm not sure that we can fix this in the code. The problem is the door and the wheel knob have "matching" animations, so when Lara is turning the wheel the door is also animating, it's just idle; however, the door has 186 frames and a framerate of 62 while its idle (which is weird) and the wheel has 140 frames and a rate of 1. We'd need to test the triggers some frames before Lara activates it, which is impossible because the earliest we can do so is when the "switch" (wheel) is activated.

I do think this is fixable with an injection for the door's frames (once we get to that stage) so IMO we could come back to this one.

I think overall fixing the early triggering like in Venice is important because that affects enemies and suchlike spawning compared to OG. I think this door is a super niche case, especially given how disjointed the data is.

rr-:
> I'm not sure that we can fix this in the code. The problem is the door and the wheel knob have "matching" animations, so when Lara is turning the wheel the door is also animating, it's just idle; however, the door has 186 frames and a framerate of 62 while its idle (which is weird) and the wheel has 140 frames and a rate of 1. We'd need to test the triggers some frames before Lara activates it, which is impossible because the earliest we can do so is when the "switch" (wheel) is activated.

@lahm86 I think that we may be one frame behind in calling `Switch_Trigger` because we stopped checking it after Lara's collision routines, which include `Switch_Collision` that can activate or deactivate the switch. With the patch, we now have to wait an extra frame in case of a switch state change until we call `Room_TestTriggers` again. Would checking the triggers twice, both before and after the collisions, be a potential solution?

lahm86:
I think that works, but I am a bit confused as to why I must admit 😄 
I'm now thinking `Item_UpdateRoom` could be the problem where `g_TriggerIndex` would have been set.
I'll let you try the new build, to my eyes it seems OK (and Venice too).

lahm86:
New changes pushed to perform trigger tests in switches, keyholes and puzzle holes, plus after item pickups. So this is as opposed to a default second trigger test after the collision routine.

rr- (approved):
Looks cool.

aredfan (approved):
The trigger and crank wheel issues are resolved.  Everything else LGTM.

Thank you for your hard work. 👍

PR #2212, opened 2025-01-03
tr2/lara/control: fix testing triggers too early

lahm86:
#### Checklist

#### Description

This supersedes #2206 as we found further regressions were introduced there with triggered camera behaviour and pads. https://github.com/LostArtefacts/TRX/commit/d3717d10507af638a68361dd0cc2609d8335bddb has been reverted.

We now store Lara's current sector prior to the collision routine call, so this effectively mimics how `g_TriggerIndex` used to work. It would be set in the first `Collide_GetCollisionInfo` call per frame for Lara, and these calls all stem from her collision routines. We then pass this sector at the end of Lara's control to `room.c` to test the triggers. This allows all dependencies to run properly, like setting `item->floor` for testing pads, getting gun status correct for refreshing the camera etc.

The original issue in Venice should still be fixed with this, and the airlock doors.


aredfan (approved):
LGTM 👍

rr- (approved):

PR #792, opened 2023-03-29
draw: prevent random draw changes in pause screens

lahm86:
Resolves #767.

#### Checklist

- [x] I have read the [coding conventions](https://github.com/rr-/Tomb1Main/blob/master/CONTRIBUTING.md#coding-conventions)
- [x] I have added a changelog entry about what my pull request accomplishes, or it is an internal change

#### Description

Fixes lightning and gun flash continuing to animate in the inventory, pause and statistics screens. Every other effect that calls `Random_GetDraw` for positioning has been tested and none is affected currently, nor impacted by the change.

- Bubbles
- Gunshot (ricochet)
- Splash
- Twinkle (Midas touch)
- Waterfall

As mentioned in the issue, the introduction of the status enum may be useful for naming the `g_OverlayFlag` values; this variable is only used in `game.c` so it can be removed as a global. My feeling is that this is best tackled in its own issue.

PR #2216, opened 2025-01-04
Move phase executor and the picture phase to libtrx

rr-:
#### Checklist

#### Description

- Moves phase executor and common types to libtrx
- Renames TR1 `PHASE` enum to `PHASE_ENUM` to avoid naming conflict with libtrx's `PHASE` which is a structure
- Makes TR1 use the new `PHASE_CONTROL` structure (with `PHASE_ACTION action` in place of `bool end`)
- Moves TR2 picture phase to libtrx
- Removes TR1 picture phase in favor of the merged TRX one
- Upgrades TR1 pause mode to use the new executor API (but doesn't yet move it to libtrx)
- Upgrades TR1 photo mode to use the new execotur API (but doesn't yet move it to libtrx)
- Changes TR1 pause mode UI to use the UI framework
- Ports the requester module the UI framework (beta, works fine for the pause screen)
- Adds a Frame UI widget that adds a border around its child with given padding and margin (TODO: use this the window widget)
- Fixes aspect ratio issues for the UI module in TR1 (didn't matter until now, where we need to center our first dialog)
- Introduces `Game_IsPlaying()` as a helper for pause and similar functions
- Moves `random.c` to libtrx
- Introduces `Random_FreezeDraw()` to disable advancing the random accumulator in the pause draw routines and similar


aredfan:
**TR1 issues:**

- [x] The lightning effect remains active during pause/photo mode. (**Edit:** gun flames are also affected)

`/play 5` then `/tp 47 18 40`

(#792 might be relevant)

__________________

- [x] I get a black screen for loading screens and credits when trying these scenarios:

-Select gym via the main menu
-Select new game via the main menu
-`/play 15` then use the level skip cheat (the final stats backdrop is unaffected)

rr-:
@aredfan fixed both issues.

src/tr1/game/random.c
```diff
@@ -30,8 +30,7 @@ int32_t Random_GetDraw(void)
     // Allow draw RNG to advance only during initial game setup (for such things
     // as caustic initialisation) and normal game play. RNG should remain static
     // when the game output is paused e.g. inventory, pause screen etc.
-    PHASE phase = Phase_Get();
-    if (phase == PHASE_NULL || phase == PHASE_GAME) {
+    if (Game_IsPlaying()) {
```
lahm86: Underwater caustics aren't working unfortunately, we relied on `PHASE_NULL` previously to allow rand to advance during startup.

aredfan:
> @aredfan fixed both issues.

Thank you 👍 

I found one more issue - the fade out of loading screens/credits is now slower if the image is skipped while still fading in.

In dev snapshot, the fade out is quicker if the image is skipped while still fading in, whereas fade out is slower if the image has fully faded in before getting skipped.

rr-:
@aredfan @lahm86 addressed

lahm86 (approved):

aredfan (approved):

PR #2224, opened 2025-01-06
tr2: migrate to anim/frame testing and switching utilities

lahm86:
#### Checklist

- [x] I have read the [coding conventions](https://github.com/LostArtefacts/TR1X/blob/master/CONTRIBUTING.md#coding-conventions)
- [x] I have added a changelog entry about what my pull request accomplishes, or it is an internal change
- [x] I have added a readme entry about my new feature or OG bug fix, or it is a different change

#### Description

This replaces the majority of direct `anim_num` and `frame_num` setting and testing to use `Item_SwitchToAnim`, `Item_TestFrameEqual` and `Item_TestFrameRange` in TR2. Some of these utilities have been moved to TRX, but the ones that access `g_Anims` directly remain in the separate games (these will eventually be moved too once shared anim storage is in place).

There are a few leftover areas that still use direct anim and frame manipulation, but this is mainly flare and gun arm control where there is no associated item.

This also resolves #1602 as it made sense to do so rather than migrate and retain the original bug. This has its own commit.

This touches many different things, which I'll post separately below. I've tested and everything is looking OK, but as always extra tests are much appreciated.

PR #2230, opened 2025-01-07
libtrx/anims/common: move ANIMs to TRX

lahm86:
#### Checklist

#### Description

This migrates `ANIM` storage to libtrx and introduces accessors rather than using globals. We have three functions essentially for getting an `ANIM`:
- `Anim_GetAnim(n)` - gets an anim by global index
- `Item_GetAnim(item)` - gets the given item's current anim
- `Object_GetAnim(obj, n)`  gets the `nth` anim of the given object

Animations are read commonly in libtrx for both games, plus TR1 injections. Frame setup is still as it was, but this will be refactored later separately. All anim and frame switching/testing functions are also now in libtrx.

### DIFF
--- a/src/libtrx/game/level/common.c
+++ b/src/libtrx/game/level/common.c
@@ -214,6 +214,41 @@ void Level_ReadObjectMeshes(
     Vector_Free(unique_indices);
 }
 
+void Level_ReadAnims(
+    const int32_t base_idx, const int32_t num_anims, VFILE *const file,
+    int32_t **frame_pointers)
+{
+    for (int32_t i = 0; i < num_anims; i++) {
+        ANIM *const anim = Anim_GetAnim(base_idx + i);
+#if TR_VERSION == 1
+        anim->frame_ofs = VFile_ReadU32(file);
+        const int16_t interpolation = VFile_ReadS16(file);
+        ASSERT(interpolation <= 0xFF);
+        anim->interpolation = interpolation & 0xFF;
+        anim->frame_size = 0;
+#else
+        const int32_t frame_idx = VFile_ReadS32(file);
+        if (frame_pointers != NULL) {
+            (*frame_pointers)[i] = frame_idx;
+        }
+        anim->frame_ptr = NULL; // filled later by the animation frame loader
+        anim->interpolation = VFile_ReadU8(file);
+        anim->frame_size = VFile_ReadU8(file);
+#endif
+        anim->current_anim_state = VFile_ReadS16(file);
+        anim->velocity = VFile_ReadS32(file);
+        anim->acceleration = VFile_ReadS32(file);
+        anim->frame_base = VFile_ReadS16(file);
+        anim->frame_end = VFile_ReadS16(file);
+        anim->jump_anim_num = VFile_ReadS16(file);
+        anim->jump_frame_num = VFile_ReadS16(file);
+        anim->num_changes = VFile_ReadS16(file);
+        anim->change_idx = VFile_ReadS16(file);
+        anim->num_commands = VFile_ReadS16(file);
+        anim->command_idx = VFile_ReadS16(file);
+    }
+}
+
 void Level_ReadAnimBones(
     const int32_t base_idx, const int32_t num_bones, VFILE *const file)
 {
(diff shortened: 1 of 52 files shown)


src/tr1/game/objects/creatures/statue.c
```diff
@@ -49,7 +49,7 @@ void Statue_Initialise(int16_t item_num)
     Item_Initialise(centaur_item_num);
 
     Item_SwitchToAnim(centaur, CENTAUR_REARING_ANIM, CENTAUR_REARING_FRAME);
-    centaur->current_anim_state = g_Anims[centaur->anim_num].current_anim_state;
+    centaur->current_anim_state = Item_GetAnim(centaur)->current_anim_state;
```
walkawayy, line 52: Am I crazy or is this doing nothing? It's setting `centaur->current_anim_state` to `centaur`'s `current_anim_state`?

src/libtrx/game/level/common.c
```diff
@@ -214,6 +214,41 @@ void Level_ReadObjectMeshes(
...
+{
+    for (int32_t i = 0; i < num_anims; i++) {
+        ANIM *const anim = Anim_GetAnim(base_idx + i);
+#if TR_VERSION == 1
+        anim->frame_ofs = VFile_ReadU32(file);
+        const int16_t interpolation = VFile_ReadS16(file);
+        ASSERT(interpolation <= 0xFF);
+        anim->interpolation = interpolation & 0xFF;
+        anim->frame_size = 0;
+#else
```
walkawayy, line 229: Are these `else` consistent with all later TRs? Is it worth explicitly saying `#if TR_VERSION == 2` to set up for potentially a `TR3X` or later that is different?

src/tr2/decomp/flares.c
```diff
@@ -208,22 +208,23 @@ void Flare_Create(const bool thrown)
...
+        anim_idx = 1;
     } else if (frame < LF_FL_IGNITE) {
-        anim_base = g_Objects[O_LARA_FLARE].anim_idx + 2;
+        anim_idx = 2;
     } else if (frame < LF_FL_2_HOLD) {
-        anim_base = g_Objects[O_LARA_FLARE].anim_idx + 3;
+        anim_idx = 3;
     } else {
-        anim_base = g_Objects[O_LARA_FLARE].anim_idx + 4;
+        anim_idx = 4;
```
walkawayy, line 221: Is it worth adding enums for 1-4 for flares as well as pistols? I guess the frame enums in the if test kinda tell you what these `anim_idx` are setting.

src/tr1/game/objects/creatures/statue.c
lahm86: It's setting the item's state to that of the animation, which is 6 in this case so it's in the rearing state.

src/libtrx/game/level/common.c
lahm86: TR3 uses the same format as TR2. Arguably, TR1 is the same in terms of struct size, it's just that frame size was introduced in TR2 to deal with the way rotations are packed.
TR4 and 5 have extra (unrelated) properties (lateral speed and acceleration IIRC).

src/tr2/decomp/flares.c
lahm86: I think we should. I will follow it up separately though if we're OK with that as I've just realised I didn't make use of `Anim_TestAbsFrameRange` in `Gun_Pistols_SetArmInfo` in #2224. Makes sense to do these together.

src/libtrx/game/level/common.c
rr-: It's OK – we'll have to go through every single `if TR_VERSION` when thinking to support more games anyway.

src/tr2/decomp/flares.c
rr-: Agreed and it's OK to do this in a follow-up PR

rr- (approved):
LGTM. Thanks. Can't wait to see the ANIM_FRAME arrive to TR2 as well!

walkawayy (approved):

aredfan (approved):

PR #2232, opened 2025-01-08
Simplify clock APIs; migrade fade effects to the new API

rr-:
#### Checklist

#### Clock API changes

This pull request simplifies the `CLOCK_TIMER` APIs. I've removed logical and draw frames, opting to use seconds (as decimals) throughout. Historically, this method was compromised by the turbo cheat, whose capability to alter the game pace made real-time measurements meaningless, but I managed to circumvent this problem with some tricks. `CLOCK_TIMER` is now split into two types: `CLOCK_TIMER_REAL`, which uses real-world time for tasks like player input debouncing or calculating rendered FPS, and `CLOCK_TIMER_SIM` (simulation), which adjusts for the turbo cheat speed multiplier. The APIs include the following functions:
- `ClockTimer_Sync`: reset the timer state to now,
- `ClockTimer_PeekElapsed`: return elapsed seconds since the last sync,
- `ClockTimer_TakeElapsed`: return elapsed seconds since the last sync, and immediately sync again,
- `ClockTimer_CheckElapsed`: check if the given time has passed since the last sync,
- `ClockTimer_CheckElapsedAndTake`: check if the given time has passed since the last sync, and sync again if so.

Coincidentally, this fixes also #2231.

#### Fader API changes

The fader module has been updated to operate in seconds rather than logical frames as well. Since it was made to utilize the new timers, there was no need for `Fader_Control` anymore and it was removed. This means that the current draw value, and whether a fader is active, is determined solely by the elapsed time, making it entirely immutable. In addition, the multiple factory functions have been combined into a single `Fader_Init` function, which abandons `FADER_ARGS` in favor of plain arguments with the debuff being determined automatically to cut down on verbosity for API users. The call sites can opt for `Fader_InitEx`, which still supports the full `FADER_ARGS`. I've also replaced manual `Output_DrawBlackRectangle` calls with a dedicated `Fader_Draw` function, which further reduces complexity in usage.

All TR1 phases have transitioned to the new fader system, resulting in the removal of the old `Output_Fade*` APIs. This update went smoothly, except for the pause phase, which was a bit peculiar – it depended on the overlay fade effect maintaining its state across different phases. Specifically, a fade-out effect that began on the pause screen would continue into the game after a phase change. To avoid managing global state across phases, I decided to complete the fade within the pause phase itself. The downside is that it prevents player interaction during this short period. An alternative approach that could prevent this regression would be for the game phase to transfer its internal fader to the pause phase using `PHASE_PAUSE_ARGS`.

#### Testing

- All fade effects and their durations
- Turbo speed
- 30/60 FPS (I expect it to have zero influence, so it's more of a sanity check)
- Holding arrows in load/save game dialogs
- Holding reset/unbind keys in the controls dialog
- Underwater caustics
- Health bar blinking
- Text flashing
- FPS counter
- Demo timer (waiting without inputs)
- Picture durations
- Inventory ring animation speeds
- 3D pickups UI speed


aredfan:
TR1X issues:

- [x]  Initiate the demo mode from the main menu, then press action to skip.  When you're back on the main menu, initiate the demo mode again and you'll get a black screen.
- [x] The eidos screen fades out as soon as it fades in.
- [x] On the main menu, the fade out is starting early when selecting load game, new game, gym, or exit to title.  In the latest dev snapshot, the fade out would start when the inventory ring have closed.
- [x] If you select load game or exit to title while ingame. the fade out is faster.  This also happens if you select new game from the gym level. In dev snapshot, you can see the inventory ring close fully during fade out whereas in the PR build, it's a black screen before that point.
- [x] The credits are showing for about 6.5 seconds between fades instead of 7.5.
- [x] In demo mode, if you have the FPS counter enabled, it overlaps with the health bar.
- [x] The fades in the final stats screen are a bit fast compared to the credits.

rr-:
@aredfan addressed all

rr-:
@aredfan I have proposed a different fade setup for TR1 in the latest commit, which makes ingame fades to black work the same as the main menu ones. Fades to transparency are working as they were. LMK how this change feels, if it's yay or nay.

aredfan:
> @aredfan I have proposed a different fade setup for TR1 in the latest commit, which makes ingame fades to black work the same as the main menu ones. Fades to transparency are working as they were. LMK how this change feels, if it's yay or nay.

I have to say I do like this change because it's visually neater to allow the ring to close before fading out, albeit a tiny bit slower when loading a savegame or exiting to the title. Regardless, yay is my answer. 😄 

- [x] I found a new TR1X issue, when the main menu is fading out, you can see the backdrop persist but only just.

Screenshot

rr-:
@aredfan addressed – I changed the fader to ensure it has drawn the target state before I consider it deactivated.

Richard-L:
Does the build here update when you say you've addressed something?  https://github.com/LostArtefacts/TRX/pull/2232#issuecomment-2576496609

I might have an hour for some testing later tonight.

rr-:
> Does the build here update when you say you've addressed something? [#2232 (comment)](https://github.com/LostArtefacts/TRX/pull/2232#issuecomment-2576496609)
> 
> I might have an hour for some testing later tonight.

No. Edit: I've rebased it (which means I've updated that other PR with the changes from here)

aredfan (approved):
> @aredfan addressed – I changed the fader to ensure it has drawn the target state before I consider it deactivated.

Thank you. Everything LGTM. 👍

lahm86 (approved):

PR #2243, opened 2025-01-09
libtrx/anims/frames: migrate TR1 frame handling to TRX

lahm86:
#### Checklist

#### Description

This moves TR1 frame handling into TRX; we currently retain the previous packed rotations approach, I will handle expanding these separately after TR2 has been moved across similarly. I'm taking a more discrete approach with frames, because of their complexity.

For assigning frame pointers to anims/objects, this has been changed with TR2 in mind. TR2 doesn't contain the number of meshes for each frame, unlike TR1, so we will need to get this from the object. And we would need to scan at some point over the objects to assign the pointers, so I hope that the approach I've proposed looks acceptable. Also, `frame->nmeshes` will be eliminated once TR2 support is in, hence no renaming of this for now.

Any early feedback on the overall approach would be welcome; I realise this deviates a bit from our other anim loading procedures. I feel because frames are so intricate (and will become more so with TR2 support added) they need their own special handling, so IMO it's best to deal with raw data in the level and injection modules and then process them fully only once thereafter.


rr- (commented):
Thanks for including us early on. I've gone through your design and think it's pretty solid. One tweak I'd suggest is to have `Anim_InitialiseFrames` focus just on allocation, like the other routines. We could then add specialized APIs like `Anim_GetFrameCount(…)` and `Anim_LoadFrames` (or similar) that it needs. This should help make the stormy parts of the code more isolated in their appearance and let us navigate them more easily.

lahm86:
Thank you. I've separated out the allocation and loading (separate commit, will likely squash once everything is good). I've opted to have the anim frame count calculated twice though rather than storing it, WDYT? It was probably not a good idea to allocate the count array on the stack anyway as previously, but I suppose we could either have a static array with the counts, or have a `frame_count` property in `ANIM`? I'm not sure we would ever need it other than for loading however.

aredfan (approved):

rr- (commented):
Yes I forgot to comment that we likely should avoid allocating the anims on the stack.

src/libtrx/include/libtrx/game/anims/common.h
```diff
@@ -7,7 +7,10 @@ void Anim_InitialiseChanges(int32_t num_changes);
 void Anim_InitialiseRanges(int32_t num_ranges);
 void Anim_InitialiseCommands(int32_t num_cmds);
 void Anim_InitialiseBones(int32_t num_bones);
+void Anim_InitialiseFrames(const int16_t *data, int32_t data_length);
+void Anim_LoadFrames(void);
```
rr-: Actually, I was thinking the opposite:
```
int count = ...;
Initialise(count);
Load(..., ...);
```

lahm86: Thanks, updated.

rr- (approved):

src/libtrx/include/libtrx/game/anims/common.h
rr-: Thanks :relieved:

PR #2244, opened 2025-01-09
libtrx/anims/frames: expand TR1 frame rotations

lahm86:
#### Checklist

#### Description

This parses TR1 frame mesh rotations fully once on level load, which means we eliminate the need to pass packed values to `matrix.c` and have it extract the individual values on every call.

I'm not sure if I like the name `Matrix_RotYXZ_16`. I can see potential though to change `Matrix_RotYXZ`'s signature in a follow-up task to accept an `XYZ_16` instead of the individual values.

The extraction macros are a little more complex ready for TR2, as it won't always have two `int16_t`s so we can't cast to `int32_t` and use the previous approach. Similarly, the `M_ParseMeshRotation` function returns the number of values read because of what's to come with TR2.

This also removes `nmeshes` from `ANIM_FRAME` - now taken from the related object.

### DIFF
--- a/src/libtrx/game/anims/frames.c
+++ b/src/libtrx/game/anims/frames.c
@@ -11,6 +11,9 @@ static int32_t M_GetAnimFrameCount(int32_t anim_idx);
 static OBJECT *M_GetAnimObject(int32_t anim_idx);
 static int32_t M_ParseFrame(
     ANIM_FRAME *frame, const int16_t *data_ptr, int16_t mesh_count);
+static void M_ParseMeshRotation(XYZ_16 *rot, const int16_t **data);
+static void M_ExtractRotation(
+    XYZ_16 *rot, int16_t rot_val_1, int16_t rot_val_2);
 
 static int32_t M_GetAnimFrameCount(const int32_t anim_idx)
 {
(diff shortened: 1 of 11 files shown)


src/libtrx/game/anims/frames.c
```diff
@@ -5,11 +5,17 @@
 #include "game/objects/common.h"
 #include "log.h"
 
+#define EXTRACT_ROT_X(r) ((r & 0x3FF0) << 2)
+#define EXTRACT_ROT_Y(r1, r2) ((((r1 & 0xF) << 6) | ((r2 & 0xFC00) >> 10)) << 6)
+#define EXTRACT_ROT_Z(r) ((r & 0x3FF) << 6)
```
rr-: Let's forgo these macros and put the code directly.

src/libtrx/game/anims/frames.c
```diff
@@ -54,17 +61,36 @@ static int32_t M_ParseFrame(ANIM_FRAME *const frame, const int16_t *data_ptr)
...
+    const int16_t rot_val_2 = *data_ptr++;
+    rot->x = EXTRACT_ROT_X(rot_val_2);
+    rot->y = EXTRACT_ROT_Y(rot_val_2, rot_val_1);
+    rot->z = EXTRACT_ROT_Z(rot_val_1);
+    return 2;
+#else
+    ASSERT_FAIL();
+    return 0;
+#endif
+}
```
rr-, line 96: This function could take a double pointer removing the need for a return value.

src/tr1/math/matrix.h
```diff
@@ -20,7 +20,7 @@ void Matrix_RotX(int16_t rx);
 void Matrix_RotY(int16_t ry);
 void Matrix_RotZ(int16_t rz);
 void Matrix_RotYXZ(int16_t ry, int16_t rx, int16_t rz);
-void Matrix_RotYXZpack(int32_t rots);
+void Matrix_RotYXZ_16(const XYZ_16 *rotation);
```
rr-: Let's just call it Rot or RotXYZ as the order doesn't matter anymore.

rr-: I'm also ok with XYZ16, as underscores are used in a different way in our function names, and this nane would clearly distinguish this function's signature from, say, TranslateRel which take individual components still.

src/libtrx/game/anims/frames.c
walkawayy: It's kinda nice to have the nice defined names though. Because that code is ugly and dense to read. 😆

walkawayy (approved):

aredfan:
- [x] The game crashes when trying to load Lara's Home or Tomb of Qualopec.

src/libtrx/game/anims/frames.c
lahm86: I've create a `M_ExtractRotation` function. For TR2 rots with two values, we will still need to utilise this but its words are reversed so a function will avoid repetition when the time comes.

src/libtrx/game/anims/frames.c
lahm86: Done

src/tr1/math/matrix.h
lahm86: I've gone with `Matrix_RotXYZ16`, agreed it keeps it clear.

lahm86:
@aredfan Thanks, fixed

rr- (approved):
LGTM. Kickass changes. Thank you.

aredfan (approved):

Issue #1906, opened 2024-11-16
TR2X bug: the mercenary hits the skidoo late


aredfan:
**Dev snapshot:** 0.6-90-gf4162df (latest)

___________________

In the third demo, observe where the blood splatter ends up in each of the screenshots below.

___________________

**OG:**
![1 3 OG](https://github.com/user-attachments/assets/840fe52f-f7d0-4a3a-bb75-45f1702df69c)

**0.6:**
![1 2 0 6](https://github.com/user-attachments/assets/85e7e0a8-3f66-4d8c-a0f4-4db3b23a36ca)

**Dev snapshot:**
![1 1 Dev](https://github.com/user-attachments/assets/6cfc24e1-d01e-4320-b943-86a6a5b1ecd8)
The mercenary hits Lara a _tiny_ bit late as opposed to the first 2 screenshots.
_______________________

**OG:**
![2 3 OG](https://github.com/user-attachments/assets/a54faf3a-1473-4d81-a560-14fd437ebc99)

**0.6:**
![2 2 0 6](https://github.com/user-attachments/assets/f12fd96a-eb86-4c26-9661-610b3653484d)

**Dev snapshot:**
![2 1 Dev](https://github.com/user-attachments/assets/a68cfa4f-dfa6-4f45-9b8f-80ec723f6717)

rr-:
Regression introduced in f166dcb65afda2279a247641dc601bb9128c223f.

rr-:
Item 32 is activated on frame 1086 prior to that change, whereas it gets activated on frame 1087 from the affecting change onwards. Here's the patch that I used to measure this:

```patch
diff --git a/src/tr2/decomp/decomp.c b/src/tr2/decomp/decomp.c
index 304a2dcc..06b45b98 100644
--- a/src/tr2/decomp/decomp.c
+++ b/src/tr2/decomp/decomp.c
@@ -1070,9 +1070,11 @@ void __cdecl CutscenePlayerGen_Initialise(const int16_t item_num)
     item->dynamic_light = 0;
 }
 
+int32_t g_Frames = 0;
 int32_t __cdecl Level_Initialise(
     const int32_t level_num, const GAMEFLOW_LEVEL_TYPE level_type)
 {
+    g_Frames = 0;
     g_GameInfo.current_level.num = level_num;
     g_GameInfo.current_level.type = level_type;
 
diff --git a/src/tr2/game/game.c b/src/tr2/game/game.c
[...]

PR #2246, opened 2025-01-10
libtrx/anims/frames: migrate TR2 frame handling to TRX

lahm86:
#### Checklist

#### Description

This moves TR2 frame handling into TRX. We retain the original (super)packed rotations for the time being, this will be tackled on its own. Apologies that this is mostly one fairly large commit - everything was so dependent I couldn't see a way to split it up further.

There is a workaround added for objects like the prayer wheel, which do not have any animations. Because we set the object's frame base while iterating animations, the wheel remained with a null pointer. The workaround is to store the frame offset while reading the level, then later point it to the matching animation (and to clarify, the wheel points to the flare pack's animation - we will fix this in the future with injections).

I've removed `frame_base` from the savegame (well, we're now skipping reading and just writing a 0) in line with TR1. This is set on each frame anyway and I've confirmed by saving in TR2X while using a gun for example, then loading it in OG, plus vice-versa. I'd appreciate some other tests around saving/loading while Lara's arms are busy with guns/flares.

Once this lands, we can migrate TR2's `BOUNDS16` to match TR1, which means we can streamline things like `M_ParseFrame`.

### DIFF
--- a/src/libtrx/game/anims/frames.c
+++ b/src/libtrx/game/anims/frames.c
@@ -4,18 +4,22 @@
 #include "game/gamebuf.h"
 #include "game/objects/common.h"
 #include "log.h"
+#include "utils.h"
 
 static ANIM_FRAME *m_Frames = NULL;
 
-static int32_t M_GetAnimFrameCount(int32_t anim_idx);
+static int32_t M_GetAnimFrameCount(int32_t anim_idx, int32_t frame_data_length);
 static OBJECT *M_GetAnimObject(int32_t anim_idx);
+static ANIM_FRAME *M_FindFrameBase(uint32_t frame_ofs);
 static int32_t M_ParseFrame(
-    ANIM_FRAME *frame, const int16_t *data_ptr, int16_t mesh_count);
+    ANIM_FRAME *frame, const int16_t *data_ptr, int16_t mesh_count,
+    uint8_t frame_size);
 static void M_ParseMeshRotation(XYZ_16 *rot, const int16_t **data);
 static void M_ExtractRotation(
     XYZ_16 *rot, int16_t rot_val_1, int16_t rot_val_2);
 
-static int32_t M_GetAnimFrameCount(const int32_t anim_idx)
+static int32_t M_GetAnimFrameCount(
+    const int32_t anim_idx, const int32_t frame_data_length)
 {
     const ANIM *const anim = Anim_GetAnim(anim_idx);
 #if TR_VERSION == 1
@@ -23,8 +27,11 @@ static int32_t M_GetAnimFrameCount(const int32_t anim_idx)
         ((anim->frame_end - anim->frame_base) / (float)anim->interpolation)
         + 1);
 #else
-    ASSERT_FAIL();
-    return 0;
+    uint32_t next_ofs = anim_idx == Anim_GetTotalCount() - 1
+        ? (unsigned)(sizeof(int16_t) * frame_data_length)
+        : Anim_GetAnim(anim_idx + 1)->frame_ofs;
+    return (next_ofs - anim->frame_ofs)
+        / (int32_t)(sizeof(int16_t) * anim->frame_size);
 #endif
 }
 
@@ -41,15 +48,40 @@ static OBJECT *M_GetAnimObject(const int32_t anim_idx)
     return NULL;
 }
 
-static int32_t M_ParseFrame(
-    ANIM_FRAME *const frame, const int16_t *data_ptr, int16_t mesh_count)
+static ANIM_FRAME *M_FindFrameBase(const uint32_t frame_ofs)
 {
-#if TR_VERSION > 1
-    ASSERT_FAIL();
-    return 0;
-#else
-    const int16_t *const frame_start = data_ptr;
+    const int32_t anim_count = Anim_GetTotalCount();
+    for (int32_t i = 0; i < anim_count; i++) {
+        const ANIM *const anim = Anim_GetAnim(i);
+        if (anim->frame_ofs == frame_ofs) {
+            return anim->frame_ptr;
+        }
+    }
 
+    return NULL;
+}
+
+static int32_t M_ParseFrame(
+    ANIM_FRAME *const frame, const int16_t *data_ptr, int16_t mesh_count,
+    const uint8_t frame_size)
+{
+    const int16_t *const frame_start = data_ptr;
+#if TR_VERSION > 1
+    frame->bounds.min_x = *data_ptr++;
+    frame->bounds.max_x = *data_ptr++;
+    frame->bounds.min_y = *data_ptr++;
+    frame->bounds.max_y = *data_ptr++;
+    frame->bounds.min_z = *data_ptr++;
+    frame->bounds.max_z = *data_ptr++;
+    frame->offset.x = *data_ptr++;
+    frame->offset.y = *data_ptr++;
+    frame->offset.z = *data_ptr++;
+
+    frame->mesh_rots =
+        GameBuf_Alloc(sizeof(int16_t) * (frame_size - 9), GBUF_ANIM_FRAMES);
+    memcpy(frame->mesh_rots, data_ptr, sizeof(int16_t) * (frame_size - 9));
+    data_ptr += MAX(0, frame_size - (data_ptr - frame_start));
+#else
     frame->bounds.min.x = *data_ptr++;
     frame->bounds.max.x = *data_ptr++;
     frame->bounds.min.y = *data_ptr++;
@@ -69,9 +101,8 @@ static int32_t M_ParseFrame(
         XYZ_16 *const rot = &frame->mesh_rots[i];
         M_ParseMeshRotation(rot, &data_ptr);
     }
-
-    return data_ptr - frame_start;
 #endif
+    return data_ptr - frame_start;
 }
 
 static void M_ParseMeshRotation(XYZ_16 *const rot, const int16_t **data)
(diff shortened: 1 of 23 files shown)


src/libtrx/game/level/common.c
```diff
@@ -215,23 +215,18 @@ void Level_ReadObjectMeshes(
...
         const int16_t interpolation = VFile_ReadS16(file);
         ASSERT(interpolation <= 0xFF);
         anim->interpolation = interpolation & 0xFF;
         anim->frame_size = 0;
 #else
-        const int32_t frame_idx = VFile_ReadS32(file);
-        if (frame_pointers != NULL) {
-            (*frame_pointers)[i] = frame_idx;
-        }
-        anim->frame_ptr = NULL; // filled later by the animation frame loader
```
rr-, line 234: I don't think we need this `#ifdef` now? I mean the combination of little-endian + that TR1 assertion tells me the TR2 code should work as well.

src/tr2/game/level.c
```diff
@@ -847,6 +839,11 @@ bool Level_Load(const char *const file_name, const int32_t level_num)
 {
     BENCHMARK *const benchmark = Benchmark_Start();
 
+    for (int32_t i = 0; i < O_NUMBER_OF; i++) {
+        OBJECT *const object = Object_GetObject(i);
+        object->loaded = false;
+    }
```
rr-, line 845: One thing we can worry about later is letting the player use the dev console even during the EIDOS/Legal screen when the game starts up. For that, we'd need the O_ALPHABET object to be injected, and ideally stay loaded throughout. But like I said, we can deal with this later as we're nowhere near close to be able to support this use case.

src/libtrx/game/level/common.c
lahm86: Thanks. You're right, now removed 👍

src/tr2/game/level.c
lahm86: That's a good point. I was just wary of stale pointers lingering and potentially causing issues, and we do unload objects in TR1 as well so it matches up.
I think it sounds doable though for the future, perhaps something like a persistent flag on the object.

rr- (approved):

src/libtrx/game/anims/frames.c
```diff
@@ -41,15 +48,40 @@ static OBJECT *M_GetAnimObject(const int32_t anim_idx)
...
+    frame->bounds.max_z = *data_ptr++;
+    frame->offset.x = *data_ptr++;
+    frame->offset.y = *data_ptr++;
+    frame->offset.z = *data_ptr++;
 
+    frame->mesh_rots =
+        GameBuf_Alloc(sizeof(int16_t) * (frame_size - 9), GBUF_ANIM_FRAMES);
+    memcpy(frame->mesh_rots, data_ptr, sizeof(int16_t) * (frame_size - 9));
+    data_ptr += MAX(0, frame_size - (data_ptr - frame_start));
+#else
```
walkawayy, line 84: Is this if test in the code below in the else still needed? Because the if above already checks for `TR_VERSION > 1`
```
    #if TR_VERSION == 1
    mesh_count = *data_ptr++;
    #endif
```

walkawayy (approved):

src/libtrx/game/anims/frames.c
lahm86: I'll be tidying up this function after this, it's a bit untidy atm until we get TR2 rotations expanded and `BOUNDS_16` working the same as TR1. The `#if`s will become neater then. I could remove now but would re-introduce it more or less immediately anyway.

aredfan:
I've tested with swapping saves between TR2X and OG, all LGTM.

I also found a small issue in the PR build that isn't present in latest dev snapshot, and similar to #1906. It looks like the mercenary's shot is hitting Lara in a different spot which causes the blood splatter to shift to another part of the screen.

PR:
1 1 PR

Dev snapshot:
1 2 Dev

PR:
2 1 PR

Dev snapshot:
2 2 Dev

lahm86:
Thank @aredfan. I think this is a nasty bug, and I believe OG might actually be incorrect. Tagging @rr, @walkawayy for thoughts on this.

I've done some logging on the dev snapshot and this PR, and you're exactly right, the blood is spawning in a different place. Everything else matches up.

**PR**
Goon pos: 40523 1427 51005
Lara pos: 40209 1349 48485
Goon gun spark pos: 40342 919 49703
_Lara blood pos: 40245 802 48588_

**Dev**
Goon pos: 40523 1427 51005
Lara pos: 40209 1349 48485
Goon gun spark pos: 40342 919 49703
_Lara blood pos: 40303 736 48392_

The issue I believe is in [`Spawn_GunHit`](https://github.com/LostArtefacts/TRX/blob/develop/src/tr2/game/spawn.c#L148). On logging (in both builds), the joint passed to `Collide_GetJointAbsPosition` is 17 (so that's the result of `Random_GetControl() * 25 / 0x7FFF`). Now this is _wrong_ because Lara only has 15 joints 😨 So that function is going through all of Lara's joints and then blindly moving on twice more to whatever is next in memory in terms of bones and frames. Because we have changed to read exactly the amount of frames needed, we're now getting something different. TR2 frames are padded, but because we are now reading from frame offset per anim, and reading only what we need, the result of blindly reading too far in things like `Collide_GetJointAbsPosition` is that we spill into another frameset rather than first hitting padding and then potentially another frameset.

I don't think we can match OG here if we are going to have structured frames in all honesty. Once we get to expanding frames, the padding is even more meaningless because we don't know what to assign it as.

walkawayy:
Hm so this has the potential to affect where blood spawns in normal gameplay as well, not just demos? It does seem like a bug that `Random_GetControl() * 25 / 0x7FFF` can produce a joint number bigger than Lara's total. Can we change this formula to make sure it's between 0-15 or whatever it needs to be? The point of the function seems to be that it randomly spawns blood on one of Lara's joints, so it doesn't seem like an issue to me to correct the function to pick a proper joint number. Maybe rr will have more thoughts.

Btw it also looks like it could affect TR1 since it uses the same formula:
https://github.com/LostArtefacts/TRX/blob/3362551b7be0416b0cab463062e5ca6c7a862a94/src/tr1/game/spawn.c#L172

lahm86:
Yeah, it's a TR1 issue too. But because frames aren't padded there we're still getting the same results as before, but still, we are reading into data we shouldn't be.
I've tried clamping the value to 15 and when Lara's standing still, you end up with seemingly "stickier" blood splats, as in they are precisely at whichever joint they spawn at. It's better with the following, we can likely play around with the values if we decide to fix this.

```diff
diff --git a/src/tr2/game/collide.c b/src/tr2/game/collide.c
index 5c76d797..0eb61e83 100644
--- a/src/tr2/game/collide.c
+++ b/src/tr2/game/collide.c
@@ -579,7 +579,9 @@ void Collide_GetJointAbsPosition(
     Matrix_RotYXZsuperpack(&mesh_rots, 0);
 
     const int16_t *extra_rotation = item->data;
-    for (int32_t i = 0; i < joint; i++) {
+    int32_t abs_joint = joint;
+    CLAMP(abs_joint, 0, object->mesh_count);
+    for (int32_t i = 0; i < abs_joint; i++) {
         const ANIM_BONE *const bone = Object_GetBone(object, i);
         if (bone->matrix_pop) {
             Matrix_Pop();
diff --git a/src/tr2/game/spawn.c b/src/tr2/game/spawn.c
index ba64ac86..1521a59a 100644
--- a/src/tr2/game/spawn.c
+++ b/src/tr2/game/spawn.c
@@ -143,9 +143,13 @@ int16_t Spawn_GunHit(
     const int32_t x, const int32_t y, const int32_t z, const int16_t speed,
     const int16_t y_rot, const int16_t room_num)
 {
-    XYZ_32 vec = {};
+    XYZ_32 vec = {
+        .x = 50 * (Random_GetDraw() - 0x4000) / 0x4000,
+        .y = 50 * (Random_GetDraw() - 0x4000) / 0x4000,
+        .z = 50 * (Random_GetDraw() - 0x4000) / 0x4000,
+    };
     Collide_GetJointAbsPosition(
-        g_LaraItem, &vec, Random_GetControl() * 25 / 0x7FFF);
+        g_LaraItem, &vec, Random_GetControl() * LM_NUMBER_OF / 0x7FFF);
     Spawn_Blood(
         vec.x, vec.y, vec.z, g_LaraItem->speed, g_LaraItem->rot.y,
         g_LaraItem->room_num);
```

rr-:
Don't the additional Random() calls cause the OG demos to desync?

lahm86:
It would change visuals only as it's `Random_GetDraw()` rather than control, I think? We could add a check to only do it in normal gameplay. But in any case, the blood splats will differ from OG anyway if we clamp to 15. Just something to consider because the bug aredfan found before at this spot led to discovering the early FD triggering issue, so it's definitely a benefit comparing demos across builds. If we proceed, we'd need to mark a new baseline build for comparing demos in the future.

rr-:
> It would change visuals only as it's `Random_GetDraw()` rather than control, I think? We could add a check to only do it in normal gameplay. But in any case, the blood splats will differ from OG anyway if we clamp to 15. Just something to consider because the bug aredfan found before at this spot led to discovering the early FD triggering issue, so it's definitely a benefit comparing demos across builds. If we proceed, we'd need to mark a new baseline build for comparing demos in the future.

I can't think of any other way out of this. Let's proceed, as long as the rest of the demos appear normal.

lahm86:
Will PR separately and rebase this one later.

lahm86:
Rebased, @aredfan, if you fancy doing a check. Develop build below on left, PR on right.

image

aredfan (approved):
> Rebased, @aredfan, if you fancy doing a check. Develop build below on left, PR on right.

Np and happy to check, I can confirm the position Lara gets shot matches your screenshots. I understand the demo will be slightly different visually and that's perfectly fine.

LGTM, and thank you again. 👍

lahm86:
> > Rebased, @aredfan, if you fancy doing a check. Develop build below on left, PR on right.
> 
> Np and happy to check, I can confirm the position Lara gets shot matches your screenshots. I understand the demo will be slightly different visually and that's perfectly fine.
> 
> LGTM, and thank you again. 👍

Thanks. This was a great find, good to get it fixed 😌

PR #2254, opened 2025-01-11
tr1|tr2: fix Lara blood spawn position

### LINKED ISSUE #2253: OG bug: blood spawned on Lara from gun hits uses illegal data
When Lara is hit by a gunshot, the blood splats that spawn can use data that is taken from bones/frames that are beyond Lara's data. This gives a quasi-random effect, but is dangerous and should instead use proper random position spawning near a legal bone.
See #2246.

lahm86:
Resolves #2253.

#### Checklist

#### Description

This fixes assigning a blood splat to a bone beyond Lara's data and introduces proper random positioning for it. This only affects blood spawned on Lara from gunshots; enemy blood is dealt with differently as are bites from creatures.

Let me know if you think the randomness itself looks OK in-game. A good test is to stand still in front of a gun goon and just monitor where the splats spawn. One of Lara's meshes is chosen at random, its world position obtained and then a random value between -64 and 63 applied to each coordinate.

This shouldn't affect TR1 demos as there are no gun enemies in those. It will change the TR2 demos but only in as much as the blood appearing differently e.g. in Tibet, and as a result anything that calls `Random_GetDraw`. There should be no desync issues as we are not changing the control random variable. If we prefer though, I can remove the randomness in TR2 if it's demo mode.

### DIFF
--- a/docs/tr1/CHANGELOG.md
+++ b/docs/tr1/CHANGELOG.md
@@ -11,6 +11,7 @@
 - fixed Lara activating triggers one frame too early (#2208, regression from 4.3)
 - fixed wrong underwater caustics speed with the turbo cheat (#2231)
 - fixed 1-frame UI flicker on pause screen exit confirmation
+- fixed blood spawning on Lara from gunshots using incorrect positioning data (#2253)
 - improved pause screen compatibility with PS1 (#2248)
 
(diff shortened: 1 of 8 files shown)


aredfan (approved):
LGTM, and thank you. 👍

rr- (approved):
Code LGTM. I didn't check the visuals.

PR #2255, opened 2025-01-11
tr2: migrate to common BOUNDS_16

lahm86:
#### Checklist

#### Description

This replaces TR2's `BOUNDS_16` to use the same as TR1, so we have two `XYZ16`s rather than individual values for each component. No logical changes, this is pretty much just renaming properties.

TR2 frame rot expansion next, which should get us to equal handling for both games.


aredfan (approved):

rr- (approved):

lahm86:
Found an issue with smashable windows and bells, they're a bit awkward to shoot now. Investigating.

lahm86:
Fixed. Would appreciate a glance over and test to confirm. Will squash the separate commit into the main one later.

The original code relied on the layout of `BOUNDS_16` for smashable/bell testing. This is the only instance like this, everything else just uses the properties as they come.

aredfan (approved):
The windows/bells behaviour is working as expected. LGTM.

rr- (approved):

PR #2266, opened 2025-01-12
libtrx/anims/frames: expand TR2 frame rotations

lahm86:
#### Checklist

#### Description

This parses TR2 frame mesh rotations fully once on level load, which means we eliminate the need to pass superpacked values to matrix.c and have it extract the component values on every call.

I've used pass by value for `Matrix_RotXYZ16` to make things easier to read and manage, and I've added a separate commit for TR1 to do the same.

Testing should be straight-forward - general item animation, including 3D pickups (ground and HUD) and skyboxes, should appear normal. It would be good to give Lara some particular attention - `Matrix_RotYXZsuperpack` used to advance the pointer on each call, but also accepted an index value to jump to a particular mesh, so this meant deciphering all of this into direct array access instead. It's pretty similar to TR1 now, but worth some attention.

TR1 is a more straight-forward change, so a quick check here should be all that's needed.

### DIFF
--- a/src/tr1/game/overlay.c
+++ b/src/tr1/game/overlay.c
@@ -400,7 +400,7 @@ static void M_DrawPickup3D(DISPLAY_PICKUP *pu)
         -(frame->bounds.min.x + frame->bounds.max.x) / 2,
         -(frame->bounds.min.y + frame->bounds.max.y) / 2,
         -(frame->bounds.min.z + frame->bounds.max.z) / 2);
-    Matrix_RotXYZ16(&frame->mesh_rots[0]);
+    Matrix_RotXYZ16(frame->mesh_rots[0]);
 
     Object_DrawMesh(obj->mesh_idx, 0, false);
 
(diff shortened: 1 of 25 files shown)


aredfan (approved):

src/tr1/game/overlay.c
```diff
@@ -415,7 +417,7 @@ static void M_DrawPickup3D(DISPLAY_PICKUP *pu)
         }
 
         Matrix_TranslateRel(bone->pos.x, bone->pos.y, bone->pos.z);
-        Matrix_RotXYZ16(&frame->mesh_rots[i]);
+        Matrix_RotXYZ16(mesh_rots[i]);
```
rr-: Can we use ``frame->mesh_rots` directly?

lahm86: Sure thing, done. I've left functions like `Lara_Draw` with the variables, as these get re-assigned at various stages for arms and what-not. LMK if that looks OK.

rr-: Yes these are fine, I meant only trivial cases. Thanks.

rr- (approved):

PR #2279, opened 2025-01-13
tr2/inventory_ring/control: default to gym option until launched

### LINKED ISSUE #2192: TR2X bug: exit to title doesn't open to the passport object
**Regression since:** [0.8](https://github.com/LostArtefacts/TRX/releases/tag/tr2-0.8)

_________________

When we quit the gym or the story levels in order to return to the main menu - the first highlighted object would be the passport. This is OG behaviour.

The regression is that the first highlighted object is Lara's Home when Exit to Title is used.

lahm86:
Resolves #2192.

#### Checklist

#### Description

Not exactly elegant, but it covers all situations (I think/hope) where you go from the inventory to some other level and then return; it should match OG now. So demos playing before having visited the gym should return back to highlighting the gym option, but anything else, the passport.

### DIFF
--- a/docs/tr2/CHANGELOG.md
+++ b/docs/tr2/CHANGELOG.md
@@ -10,6 +10,7 @@
 - fixed excessive braid movement when dead in windy rooms (#2265, regression from 0.8)
 - fixed item counter shown even for a single medipack (#2222, regression from 0.3)
 - fixed item counter always hidden in NG+, even for keys (#2223, regression from 0.3)
+- fixed the passport object not being selected when exiting to title (#2192, regression from 0.8)
 
 ## [0.8](https://github.com/LostArtefacts/TRX/compare/tr2-0.8...tr2-0.8) - 2025-01-01
 - completed decompilation efforts – TR2X.dll is gone, Tomb2.exe no longer needed (#1694)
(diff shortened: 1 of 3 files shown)


rr- (commented):
Hmm isn't it enough to just override `g_GymInvOpenEnabled` to `false` somewhere in `Level_Initialise`?
Eg
```c
void Level_Initialise(…)
{
    if (g_CurrentLevel != LV_GYM) {
        g_GymInvOpenEnabled = false;
    }
}
```
Later on I'd imagine this global becomes a member of `g_GameFlow` struct.

lahm86:
Good point, I wasn't considering globals so didn't think about this. Better than introducing a new function for now anyway.

rr- (approved):
LGTM, although I haven't tested.

aredfan (approved):
In OG, the gym would be highlighted again if Lara dies in the gym. Post credits open to the gym as well.  In both cases, the menu opens to the passport instead on the PR build. This might warrant a mention in the changelog.

Other than that, LGTM, and thank you. 👍

lahm86:
Thanks, @aredfan. I've pushed changes to make it consistent with OG. I think we should get it like-for-like for now.

aredfan (approved):

rr- (approved):

PR #2281, opened 2025-01-13
tr1|tr2: replace angle consts with shared TRX consts

lahm86:
#### Checklist

#### Description

This replaces all `PHD_DEGREE`, `PHD_90` instances with the new common consts in TRX, which follow the `DEG_` naming pattern.

The one I haven't touched is `PHD_HALF` in TR2 because I'm not sure this name is accurate. It's used in the TR2 SWR and has a value of 256.

I can squash these commits if we prefer, maybe one for each game. Left open for now as it might be easier to review.


rr- (approved):

rr-:
Thank you. I'm fine both with leaving the commits as-is, and with squashing them into a single a commit.

rr- (commented):
Actually @lahm86 please hold on – `PHD_ONE`  in some contexts is used not as an angular value, but as an arbitrary integer reference unit. Let's leave it as `PHD_ONE` in those instances. Thanks.

lahm86:
> Actually @lahm86 please hold on – `PHD_ONE` in some contexts is used not as an angular value, but as an arbitrary integer reference unit. Let's leave it as `PHD_ONE` in those instances. Thanks.

Sorry about that. Pushed two separate commits with hopefully the restores. LMK if OK. I will squash later I think for neatness.

rr- (approved):
Thank you and apologies for the hasty approval earlier.

aredfan (approved):

PR #2287, opened 2025-01-13
camera: fix Lara's vision range

### LINKED ISSUE #2276: Improve look down angle to support OG behaviour
In light of https://github.com/LostArtefacts/TRX/issues/1338 there was:

https://github.com/user-attachments/assets/f02fad61-ed70-444f-abec-334e511a2466

If there should be a way to restore the OG look angle, whilst also taking care of the camera flipping on the way back, it would be ideal.

rr-:
Resolves #2276.

#### Checklist

#### Description

The look camera looks at Lara with some offset, and when the player releases the look key, `Camera_Update` sets `g_Camera.target` to immediately point back right at Lara. Meanwhile `Camera_Apply` calculates the pitch and yaw by checking the angles between the camera position and the target position. The problem is that when the player releases the look key, these differences are too big and briefly point in weird directions. To fix the original issue, it's enough to smooth the position reset over a few frames. The impact on the existing camera work should be minimal apart from the 5 frames after releasing the look button in both games.

### DIFF
--- a/docs/tr1/CHANGELOG.md
+++ b/docs/tr1/CHANGELOG.md
@@ -7,13 +7,14 @@
 - changed the pause screen to wait before yielding control during fade out effect
 - changed the compass and final stats to use two columns, similar to TR2 (doesn't apply to end-of-level "bare" stats)
 - changed the fix for transparent eyes on wolves to use black instead of off-white (#2252)
+- fixed blood spawning on Lara from gunshots using incorrect positioning data (#2253)
+- fixed the upside-down camera fix to no longer limit Lara's vision (#2276, regression from 4.2)
 - fixed being unable to load some old custom levels that contain certain (invalid) floor data (#2114, regression from 4.3)
 - fixed a desync in the Lost Valley demo if responsive swim cancellation was enabled (#2113, regression from 4.6)
 - fixed the game hanging when Lara is on fire and enters the fly cheat on the same frame as reaching water (#2116, regression from 0.8)
 - fixed Lara activating triggers one frame too early (#2208, regression from 4.3)
 - fixed wrong underwater caustics speed with the turbo cheat (#2231)
 - fixed 1-frame UI flicker on pause screen exit confirmation
-- fixed blood spawning on Lara from gunshots using incorrect positioning data (#2253)
 - fixed being able to use keys and puzzle items in keyholes/slots that have already been used (#2256, regression from 4.0)
 - fixed textures animating during demo fade-outs (#2217, regression from 4.0)
 - fixed waterfall mist not animating during demo (#2218, regression from 3.0)
--- a/docs/tr2/CHANGELOG.md
+++ b/docs/tr2/CHANGELOG.md
@@ -22,6 +22,7 @@
 - fixed item counter shown even for a single medipack (#2222, regression from 0.3)
 - fixed item counter always hidden in NG+, even for keys (#2223, regression from 0.3)
 - fixed the passport object not being selected when exiting to title (#2192, regression from 0.8)
+- fixed the upside-down camera fix to no longer limit Lara's vision (#2276, regression from 0.8)
 
 ## [0.8](https://github.com/LostArtefacts/TRX/compare/tr2-0.8...tr2-0.8) - 2025-01-01
 - completed decompilation efforts – TR2X.dll is gone, Tomb2.exe no longer needed (#1694)
--- a/src/libtrx/include/libtrx/game/camera/types.h
+++ b/src/libtrx/include/libtrx/game/camera/types.h
@@ -20,7 +20,6 @@ typedef struct {
     int16_t actual_angle;
 #endif
     int16_t target_elevation;
-    int16_t box;
     int16_t num;
     int16_t last;
     int16_t timer;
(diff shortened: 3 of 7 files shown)


Richard-L:
Thanks for choosing to work on this.

It's a definite and noticeable improvement. When comparing to PS1 it's already superior.

I assume you know there's still a certain hitch, ie the camera doesn't move as clean on the way out as it moves on the way in. I believe this is because people will first press `look` and the camera will center behind Lara to initiate looking. When releasing `look` this "neutral state" isn't reassumed of course, and the camera will flip back towards its previous angle. But I feel a bit wrong to be too pedantic about it. And I'm not sure I'd have another solution idea. Had you considered to make the side motion kick in a little later on the "return path"?

rr-:
> Thanks for choosing to work on this.
> 
> It's a definite and noticeable improvement. When comparing to PS1 it's already superior.
> 
> I assume you know there's still a certain hitch, ie the camera doesn't move as clean on the way out as it moves on the way in. I believe this is because people will first press `look` and the camera will center behind Lara to initiate looking. When releasing `look` this "neutral state" isn't reassumed of course, and the camera will flip back towards its previous angle. But I feel a bit wrong to be too pedantic about it. And I'm not sure I'd have another solution idea. Had you considered to make the side motion kick in a little later on the "return path"?

I'll have to ask for a direct frame-to-frame comparison video as I really don't see what you mean.

Alternatively you can try this build where the duration of the adjustment is increased to 10 frames (decreasing it to 2 led to visual problems even I could see). https://home.wind.garden/serve/pub/TR2X-0.8-198-g10c048a-dirty-Windows.zip

walkawayy:
> > Thanks for choosing to work on this.
> > It's a definite and noticeable improvement. When comparing to PS1 it's already superior.
> > I assume you know there's still a certain hitch, ie the camera doesn't move as clean on the way out as it moves on the way in. I believe this is because people will first press `look` and the camera will center behind Lara to initiate looking. When releasing `look` this "neutral state" isn't reassumed of course, and the camera will flip back towards its previous angle. But I feel a bit wrong to be too pedantic about it. And I'm not sure I'd have another solution idea. Had you considered to make the side motion kick in a little later on the "return path"?
> 
> I'll have to ask for a direct frame-to-frame comparison video as I really don't see what you mean.
> 
> Alternatively you can try this build where the duration of the adjustment is increased to 10 frames (decreasing it to 2 led to visual problems even I could see). https://home.wind.garden/serve/pub/TR2X-0.8-198-g10c048a-dirty-Windows.zip

Ok I think I get what he's saying. When you press `look` and nothing else, the camera moves into a slightly different, higher position. So when you look forward:

1. Neutral position with no inputs means the camera is `CAMERA_TYPE` `CAM_CHASE`:

20250113_232603_St_Francis_Folly


2. You press `look` and the camera moves up slightly into `CAMERA_TYPE` `CAM_LOOK`:

20250113_232633_St_Francis_Folly

3. You start holding `forward` and the camera glides from the `look` position (step 2) until it hits its limit in the overhead view:

20250113_232729_St_Francis_Folly

4. When you let go of `look`, the camera is not only moving back behind Lara to `CAMERA_TYPE` `CAM_CHASE`, you are also losing the elevated camera from `CAM_LOOK`.

Now I personally don't think this looks bad at all. It matches the OG's angle nicely. I think it's just how the game camera works. Even in slowdown turbo mode, the camera doesn't really hitch. It just drops downward when you let go of `look` because it is changing `CAMERA_TYPE`s. Personally, I think this PR is good IMO.

walkawayy (approved):

aredfan (approved):

rr-:
>When you let go of look, the camera is not only moving back behind Lara to CAMERA_TYPE CAM_CHASE, you are also losing the elevated camera from CAM_LOOK.

But that's the OG behavior – it immediately switches to CAM_CHASE and sets the target to Lara. The problem with the OG is that this change is too sudden and causes the angle to go negative for a moment.

Here's a section view of what I mean.
O - Camera Origin
T - Camera Target
L - Lara

1. Neutral position
    ```
    O
        LT
    ```

2. Lara presses look
    ```
    O
            T
        L
    ```

2. Lara starts looking down
    ```
      O

        L

          T
    ```


4. Lara looks way down
    ```
         O


        L


        T
    ```
    The camera angle is 92°.

5. Lara lets go of look
    ```
        O


       LT
    ```
    The camera target is set immediately to Lara. No smoothing here involved in the OG - it's a sudden jerky change.
    This jolt causes the angle between the camera origin and target to be briefly -88° (as it's "on the other side").

6. Camera stabilizes itself
    ```
    O
        LT
    ```

The camera distances in the diagrams above may vary, I didn't verify them.
What this PR does is to add smoothing over a few frames at step 5.

lahm86 (approved):

Richard-L:
Thanks for this presentation, it's lovely :)

But it's missing the third dimension which I meant. Sorry I didn't have time to make a video last night. I tried, but needed to slow down the parts I meant and couldn't lose more sleep into editing it properly. You can see it when you load TR2's gym, press look, look down to the max, and then release. The side swing of the camera on its way back (because it's angled and not directly behind Lara), is what I meant. But again I don't think anything can be done about that, and probably isn't even worth it as it'd introduce camera return delay which in some tight game situations might be undesirable.

Walkaway also explained it nicely. Walkaway I'd like to apologise to you for hastily loving your previous solution, to now making wind about this where it's probably misplaced 🙏

rr-:
Yeah I now see what you mean. I think it's acceptable. I'd rather not change this any further to reduce the risk of causing unintended regressions. We can revisit this later.

PR #2288, opened 2025-01-14
libtrx/anims/commands: structure animation commands

lahm86:
#### Checklist

#### Description

This parses animation commands into structures and moves their storage to the animations to which they belong. This is similar to frame handling in loading, so we read and store the raw `int16_t` array, append anything we need in `inject.c` (TR1) and then interpret the data as a whole at the end of the loading process.

I see scope soon to try to merge `Item_Animate` in both games, but there are some differences yet with how effects/actions are handled. It was tempting to tackle it here, but thought it best to keep this fairly simple as just a storage/interpretation change.

Something else I will be following up in the injection refactor is how we handle the temporary data in the two `level.c` modules, for things like commands, frames and floor data. I realise there is a difference currently as TR2 doesn't have a `LEVEL_INFO` yet. So the static variables there are temporary for now.

### DIFF
--- a/src/libtrx/game/anims/commands.c
+++ b/src/libtrx/game/anims/commands.c
@@ -1,0 +1,72 @@
+#include "benchmark.h"
+#include "game/anims.h"
+#include "game/gamebuf.h"
+
+static void M_ParseCommand(ANIM_COMMAND *command, const int16_t **data);
+
+static void M_ParseCommand(ANIM_COMMAND *const command, const int16_t **data)
+{
+    const int16_t *data_ptr = *data;
+    command->type = *data_ptr++;
+
+    switch (command->type) {
+    case AC_MOVE_ORIGIN: {
+        XYZ_16 *const pos = GameBuf_Alloc(sizeof(XYZ_16), GBUF_ANIM_COMMANDS);
+        pos->x = *data_ptr++;
+        pos->y = *data_ptr++;
+        pos->z = *data_ptr++;
+        command->data = (void *)pos;
+        break;
+    }
+
+    case AC_JUMP_VELOCITY: {
+        ANIM_COMMAND_VELOCITY_DATA *const data = GameBuf_Alloc(
+            sizeof(ANIM_COMMAND_VELOCITY_DATA), GBUF_ANIM_COMMANDS);
+        data->fall_speed = *data_ptr++;
+        data->speed = *data_ptr++;
+        command->data = (void *)data;
+        break;
+    }
+
+    case AC_SOUND_FX:
+    case AC_EFFECT: {
+        ANIM_COMMAND_EFFECT_DATA *const data =
+            GameBuf_Alloc(sizeof(ANIM_COMMAND_EFFECT_DATA), GBUF_ANIM_COMMANDS);
+        data->frame_num = *data_ptr++;
+        const int16_t effect_data = *data_ptr++;
+        data->effect_num = effect_data & 0x3FFF;
+        data->environment = (effect_data & 0xC000) >> 14;
+        command->data = (void *)data;
+        break;
+    }
+
+    default:
+        command->data = NULL;
+        break;
+    }
+
+    *data = data_ptr;
+}
+
+void Anim_LoadCommands(const int16_t *data)
+{
+    BENCHMARK *const benchmark = Benchmark_Start();
+
+    const int32_t anim_count = Anim_GetTotalCount();
+    for (int32_t i = 0; i < anim_count; i++) {
+        ANIM *const anim = Anim_GetAnim(i);
+        if (anim->num_commands == 0) {
+            continue;
+        }
+
+        anim->commands = GameBuf_Alloc(
+            sizeof(ANIM_COMMAND) * anim->num_commands, GBUF_ANIM_COMMANDS);
+        const int16_t *data_ptr = &data[anim->command_idx];
+        for (int32_t j = 0; j < anim->num_commands; j++) {
+            ANIM_COMMAND *const command = &anim->commands[j];
+            M_ParseCommand(command, &data_ptr);
+        }
+    }
+
+    Benchmark_End(benchmark, NULL);
+}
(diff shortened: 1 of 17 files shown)


src/libtrx/game/anims/commands.c
```diff
@@ -0,0 +1,72 @@
...
+
+static void M_ParseCommand(ANIM_COMMAND *const command, const int16_t **data)
+{
+    const int16_t *data_ptr = *data;
+    command->type = *data_ptr++;
+
+    switch (command->type) {
+    case AC_MOVE_ORIGIN: {
+        XYZ_16 *const pos = GameBuf_Alloc(sizeof(XYZ_16), GBUF_ANIM_COMMANDS);
+        command->data = (void *)pos;
```
rr-: Let's move `command->data` assignments near `break;` for a cosmetic change – this will make the switch cases look more uniform.

src/libtrx/include/libtrx/game/anims/types.h
```diff
@@ -15,6 +16,22 @@ typedef struct {
...
+typedef struct {
+    int16_t fall_speed;
+    int16_t speed;
+} COMMAND_VELOCITY_DATA;
+
+typedef struct {
+    int16_t frame_num;
+    int16_t effect_num;
+    ANIM_COMMAND_ENVIRONMENT environment;
+} COMMAND_EFFECT_DATA;
```
rr-: Let's prefix with `ANIM_` this and the previous struct.

src/tr1/game/inject.c
```diff
@@ -499,8 +499,9 @@ static void M_AnimData(INJECTION *injection, LEVEL_INFO *level_info)
         level_info->anim_change_count, inj_info->anim_change_count, fp);
     Level_ReadAnimRanges(
         level_info->anim_range_count, inj_info->anim_range_count, fp);
-    Level_ReadAnimCommands(
-        level_info->anim_command_count, inj_info->anim_cmd_count, fp);
+    VFile_Read(
+        fp, level_info->anim_command_data + level_info->anim_command_count,
+        inj_info->anim_cmd_count * sizeof(int16_t));
```
rr-: Maybe we could create a set of functions in the Level module, similar to what we do for frames? Even if the only thing they'd do would be to `Memory_Alloc`, `VFile_Read`, and `Memory_Free`.

lahm86:
Forgot to mention testing points.
- SFX from animations, like footsteps, enemy noises, switch pulls etc
- Different SFX in water, like Lara/enemies (TR2) in one-click water
- The move origin command is handled for things like Lara climbing 2/3 clicks. So at the end of the anim, her shadow should snap on top of the block she's climbing per OG
- Special commands like Larson ending the level in ToQ, the guardian doing the same in Ice Palace
- Worth checking a few flip effects, like LQ pistons, Natla's Mines "earthquake" (near cowboy)

src/tr2/game/lara/control.c
```diff
@@ -746,66 +739,47 @@ void Lara_Animate(ITEM *const item)
...
-                    ItemAction_Run(action_id, item);
-                }
+            const ANIM_COMMAND_ENVIRONMENT type = data->environment;
+            if (type == ACE_ALL
+                || (type == ACE_LAND
+                    && (g_Lara.water_surface_dist >= 0
+                        || g_Lara.water_surface_dist == NO_HEIGHT))
+                || (type == ACE_WATER && g_Lara.water_surface_dist < 0
+                    && g_Lara.water_surface_dist != NO_HEIGHT)) {
+                Sound_Effect(data->effect_num, &item->pos, SPM_ALWAYS);
```
rr-, line 760: Offtop – can we use `SPM_UNDERWATER` to simplify the checks above?
Or alternatively, introduce a getter to retrieve Lara (or an `ITEM`) current environment.

lahm86: Yes, I think this can be made much simpler, similar to `Item_PlayAnimSFX` in TR1. I think I'll aim to make a common function for this in both games to tidy it all up (and this will help get us near a common `Item/Lara_Animate`). But I think tackling it on its own would be best rather than changing logic here, WDYT?

src/libtrx/game/anims/commands.c
lahm86: Done

src/libtrx/include/libtrx/game/anims/types.h
lahm86: Done

src/tr1/game/inject.c
lahm86: Good call, shifted now to the common level module.

src/tr2/game/lara/control.c
rr-: Yes definitely a separate effort so as not to muddle up this PR.

src/libtrx/game/anims/commands.c
```diff
@@ -0,0 +1,75 @@
...
+    command->type = *data_ptr++;
+
+    switch (command->type) {
+    case AC_MOVE_ORIGIN: {
+        XYZ_16 *const pos = GameBuf_Alloc(sizeof(XYZ_16), GBUF_ANIM_COMMANDS);
+        pos->x = *data_ptr++;
+        pos->y = *data_ptr++;
+        pos->z = *data_ptr++;
+
+        command->data = (void *)pos;
```
rr-, line 18: Let's remove the added line breaks too :D sorry for being a nitpick.

src/libtrx/game/level/common.c
```diff
@@ -260,14 +263,21 @@ void Level_ReadAnimRanges(
...
-        int16_t *const cmd = Anim_GetCommand(base_idx + i);
-        *cmd = VFile_ReadS16(file);
-    }
+    VFile_Read(file, m_AnimCommands + base_idx, sizeof(int16_t) * num_cmds);
+}
+
+void Level_LoadAnimCommands(void)
+{
+    Anim_LoadCommands(m_AnimCommands);
+    Memory_FreePointer(&m_AnimCommands);
```
rr-, line 280: That's great!

rr- (approved):

aredfan (approved):

PR #2290, opened 2025-01-14
libtrx/items/common: move Item_PlayAnimSFX to TRX

lahm86:
#### Checklist

#### Description

This merges all variations of playing SFX from `Lara_Animate` and `Item_Animate` in both games into a common function in TRX.

TR2 had a bit of a hack in `Item_Animate` for Lara's back guns, which don't have a world position, so it was setting their position there before playing their SFX. I've moved this to `gun_rifle.c` rather than having item modification while playing SFX.

Other than regular SFX from animations, like Lara hitting walls, switches in action, enemy noises etc, these are the main things to check over. There's no need to check every possible sound effect, it's more the different possible environments:
- Lara footsteps on land and in water
- Using shotgun, harpoon, M16 and grenade launcher in TR2 (and with Lara in motion, sound should follow her)
- Firing harpoon (should always play regardless of environment)
- Water creature sounds should only play underwater
- Enemies in shallow water e.g. Tiger at start of GW
- TR1 Lara's footsteps in shallow water - infamous wading test level is [here](https://github.com/LostArtefacts/TRX/pull/1728), but can also check in Tihocan room 99

### DIFF
--- a/src/libtrx/include/libtrx/game/rooms/enum.h
+++ b/src/libtrx/include/libtrx/game/rooms/enum.h
@@ -1,6 +1,14 @@
 #pragma once
 
 typedef enum {
+    RF_UNDERWATER = 0x01,
+    RF_OUTSIDE = 0x08,
+    RF_DYNAMIC_LIT = 0x10,
+    RF_NOT_INSIDE = 0x20,
+    RF_INSIDE = 0x40,
+} ROOM_FLAG;
+
+typedef enum {
     SMF_NON_COLLIDABLE = 1 << 0,
     SMF_VISIBLE = 1 << 1,
 } STATIC_MESH_FLAG;
--- a/src/libtrx/game/items.c
+++ b/src/libtrx/game/items.c
@@ -1,7 +1,10 @@
 #include "game/items.h"
 
 #include "game/const.h"
+#include "game/lara/common.h"
 #include "game/objects/common.h"
+#include "game/rooms.h"
+#include "game/sound/common.h"
 #include "utils.h"
 
 #include <stddef.h>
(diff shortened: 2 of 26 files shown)


src/libtrx/include/libtrx/game/rooms/enum.h
```diff
@@ -1,5 +1,15 @@
 #pragma once
 
+typedef enum {
+    RF_UNDERWATER = 0x01,
+#if TR_VERSION > 1
+    RF_OUTSIDE = 0x08,
+    RF_DYNAMIC_LIT = 0x10,
+    RF_NOT_INSIDE = 0x20,
+    RF_INSIDE = 0x40,
+#endif
```
rr-: Let's not `#if` these flags; they can simply be left as unused in TR1.

rr- (approved):

aredfan:
- [x] TR1X issue - you can hear the crocodile above the water, not below, when it attacks Lara, [save_tr1_00.zip](https://github.com/user-attachments/files/18414335/save_tr1_00.zip)

lahm86:
Thanks, @aredfan. Fixed.

I've also added a separate commit that removes `water_creature` flag from TR2 objects and instead does an object lookup test, similar to other use cases in the codebase. So the same test is now done for TR1 water enemies, TR2 water enemies plus the underwater propellers in Rig/Diving Area, and the submarine at the start of 40F.

aredfan (approved):
> Thanks, @aredfan. Fixed.
> 
> I've also added a separate commit that removes `water_creature` flag from TR2 objects and instead does an object lookup test, similar to other use cases in the codebase. So the same test is now done for TR1 water enemies, TR2 water enemies plus the underwater propellers in Rig/Diving Area, and the submarine at the start of 40F.

Thank you, and all these things LGTM. 👍

lahm86:
@rr- LMK if the latest commit is OK. A side task I think is that we should rename `O_GENERAL` to `O_MINISUB` and separate out the shared control with `O_DRAWBRIDGE` (see #2294) but this can be tackled separately.

rr- (approved):

rr-:
Re: minisub, I have mixed feelings about this, since it's reused in different levels IIRC as many different objects. That said, I don't have solutions.

lahm86:
> Re: minisub, I have mixed feelings about this, since it's reused in different levels IIRC as many different objects. That said, I don't have solutions.

I believe it's only used in 40F in the OG levels, but we can double check this. But in any case, we can investigate and track it in the issue above.

PR #2292, opened 2025-01-14
Move cutscene phaser to libtrx

rr-:
#### Checklist

#### Description

Almost there – only game phase left.

Offtop: in TR2 we do `Level_Initialise` once, when we play the game. In TR1, to accommodate gameflow sequencer actions that may happen _after_ the level has finished loading its data, we have split this process into "start" and "loop" stages. It's hard to remove this, as the custom actions are all very different – set the cinematic camera position, set its angle, do a flipmap, …do a mesh swap. So for now I've kept this logic, eg. initialise the level in the "start" sequencer action, and then only pretend to initialise it in the "loop" action inside the phaser (to stay forward compatible with TR2) eg initialise it only if it's different from `g_CurrentLevel`. (It's not perfect). I'd like to remove it eventually (with the help of LUA?). 

#### Testing

- All cinematics
- Infamous T-Pose bug on the last cinematic frame
- In-game cinematic cameras such as scion grab, t-rex death, midas transformation (no need to check all)
- Console
- Fades
- Exit fade in TR2
- Pause
- Photo mode
- **Story so far** in TR1


lahm86:
Looks good. The only issue I can see is that in TR2X, post-cutscene end-of-level stats are missing.

rr-:
> Looks good. The only issue I can see is that in TR2X, post-cutscene end-of-level stats are missing.

Thank you – fixed.

aredfan (approved):

lahm86 (approved):

PR #2304, opened 2025-01-15
Tidy game strings

rr-:
#### Checklist

#### Description

- Improves the game string linter tool by making it understand the relationships between projects:
    - it now detects unused game strings
    - it now reports game strings not reachable from client games (eg. a string defined in TR2 .def files is not visible for TR1, and it's now reported)
    - it also reports if a game string is defined in libtrx, but only a single game uses it
- Removes unused game strings
- Removes tombpc.dat game strings solution, moves completely to the game flow
- Tidies names of certain shared game strings
- Changes the OSD messages in TR2 to be no longer hardcoded
- Removes `O_*_ITEM` and `O_*_OPTION` object strings in favor of `INV_ITEM_*` game strings

#### Testing

We'd need to check if the object names in all levels are OK, and if the inventory ring menus have all text. Since the changes were done in a semi-automated way I don't expect to see many regressions.

#### Next steps

As a next step I'd like to extract the game string JSON to a separate file in TR2 by introducing a `game_string_table` module in TRX. After that, I'd like to incorporate that code in TR1.


aredfan (approved):

src/tr2/game/gameflow/gameflow_new.c
```diff
@@ -22,7 +23,10 @@ static void M_LoadObjectString(const char *const key, const char *const value)
 {
     const GAME_OBJECT_ID object_id =
         ENUM_MAP_GET(GAME_OBJECT_ID, key, NO_OBJECT);
-    if (object_id != NO_OBJECT) {
+    if (object_id == NO_OBJECT) {
+        LOG_ERROR("Invalid object id: %s", key);
+    } else {
+        LOG_ERROR("Renaming object id %s to %s", key, value);
```
lahm86: Just checking if this was only for debugging as it produces a lot of additional logging. If not, should it be `LOG_INFO`?

rr-: Yeah let's not check this in.

rr-: Fixed

lahm86 (approved):

PR #2307, opened 2025-01-15
libtrx/phase/phase_inventory: handle null ring on exit

### LINKED ISSUE #2306: TRX bug: using a keyhole/puzzle slot without any key items crashes the game
Regression in https://github.com/LostArtefacts/TRX/commit/944642e3ea5e06f26e836b1aed1d5929773afff4

lahm86:
Resolves #2306.

#### Checklist

#### Description

The inventory ring may be null when the phase exits if Lara has tried to use a keyhole or puzzle slot but doesn't have any key items, so this handles that case rather than asserting that the ring is not null.

I've noticed Lara doesn't say "no" in TR1 in this case, but she does in TR2 (and this is the case in current release). I can't remember if that was a change we introduced on purpose or is it a separate bug?

### DIFF
--- a/src/libtrx/game/phase/phase_inventory.c
+++ b/src/libtrx/game/phase/phase_inventory.c
@@ -45,9 +45,10 @@ static PHASE_CONTROL M_Control(PHASE *const phase, int32_t num_frames)
 static void M_End(PHASE *const phase)
 {
     M_PRIV *const p = phase->priv;
-    ASSERT(p->ring != NULL);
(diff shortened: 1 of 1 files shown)


walkawayy:
We did change how the inventory items function in #1267 for TR1 when removing `g_InvChosen`.

aredfan (approved):
Good catch. 👍 

The key/puzzle holes, midas, detonator, gong - all LGTM.

rr- (approved):

PR #1941, opened 2024-11-24
tr1/misc: create object mesh structure

lahm86:
Resolves #1494.

#### Checklist

- [x] I have read the [coding conventions](https://github.com/LostArtefacts/TR1X/blob/master/CONTRIBUTING.md#coding-conventions)
- [x] I have added a changelog entry about what my pull request accomplishes, or it is an internal change

#### Description

This eliminates raw mesh data for objects and replaces it with a complete structure for better readability, and improved injection support. Similar to #1938, this only applies to TR1 currently, although the groundwork has been laid in `libtrx` for TR2. Best viewed commit-by-commit, although note that this work was done incrementally so some intermediate functions/logic were added to support the new and old approach at the same time. There is some improvement remaining, such as making common reader functions for vertices and faces between rooms and objects, but I've left a `TODO` to follow this up, so that this PR doesn't impact anything done already in #1938.

This has a wide impact so will need plenty testing. I've tried a couple of TRLEs as well as OG, but the more the better. Areas impacted:
- Collision with enemies, including shooting
- Drawing Lara and Bacon Lara
- Drawing Lara's hair, and its collision
- Drawing any other animating object
- Drawing gun flare
- Drawing shadows
- Drawing static meshes
- Drawing 3D pickups (on ground and HUD)
- Drawing skyboxes
- Save crystal and Midas touch reflections
- Twinkle effect on Lara's meshes while using Midas touch, and turning to gold
- Mesh effects such as exploding eggs and body parts
- Savegame handling of Lara's mesh pointers e.g. save with Magnums equipped or holstered, and ensure on load Lara's meshes are accurate. Saves from older TR1X versions and legacy ATI saves should continue to work.
- Lara's mesh swaps:
  - Gym outfit
  - Angry Lara when shooting
  - Lara's hands, torso and thighs depending on selected gun and status
  - Getting mauled by the t-rex
  - Turning to gold
  - Tomb of Tihocan cutscene (Lara's pistols should be holstered)

PR #2323, opened 2025-01-17
tr2: structure room and object meshes

lahm86:
#### Checklist

#### Description

This migrates TR2 to use the same structured format for room and object meshes as TR1. This is likely best viewed per commit, `output.c` in particular as some functions have been moved around so the cumulative view isn't very readable.

This brings to light a few improvements we can make, such as potentially adopting an `item.mesh_ptrs` array, similar to `g_Lara` to allow us to eliminate the repeated drawing routines (skidoo, xian guards). But that's its own task for later.

#### Testing
This more or less touches everything that can be drawn. Linking #1941 for relevance, but a few additional points:
- Saving and loading with Lara's meshes in different swap states, like flare in hand, angry face etc
- Mesh swaps in cutscenes, like Diving Area, as well as in-game cutscenes (like Lara's hands/hips in HSH)
- Room sprites (I think the only ones are in the gym basement)
- Flares
- Skidoo track animation
- Xian guards turning to/from jade
- Bartoli sphere of doom
- Inventory background and items
- Vertices on water edges should not move
- Rooms with flickering lights (e.g. Wreck rm 31)
- Bartoli sunset effect (I have left it running myself for ~20 mins to test, but feel free to check too)
- HWR with/without z-buffer
- SWR


aredfan (approved):

walkawayy (approved):

src/tr2/game/render/hwr.c
```diff
@@ -599,35 +589,31 @@ static const int16_t *M_InsertObjectG3_Sorted(
             continue;
         }
 
-        const RGB_888 *const color = &g_GamePalette16[color_idx >> 8];
+        const RGB_888 *const color = &g_GamePalette16[face->texture >> 8];
```
rr-: Maybe we could make it an union and still refer to it as `face->color_idx`?

rr- (approved):
King

src/tr2/game/render/hwr.c
lahm86: Good idea, shall I push here or do separately and update TR1 as well?

rr-: It's OK to push here and do an update to TR1 as well while at it.

lahm86: Ok, will do so tomorrow morning 👌

lahm86: @rr- It turns out this would be overly complicated with the current TR1 injection module, which does things like "get the texture on face N of mesh K of model M and use it on face O of mesh J of model P. We'd have to introduce separate ways of saying if it's flat, use the color_idx, else texture_idx, and so it's a lot of extra code.
Maybe during the injection format refactor we can come up with something better here.

For now, I've restored creating explicit `color_idx` variables for clarity in the flat face rendering functions.

rr- (approved):

src/tr2/game/render/hwr.c
rr-: C unions are more or less guaranteed to have the same value in all members eg. assigning to color idx should also update texture idx and vice versa. So they're like variable aliases. Having that in mind, is it still not feasible atm?

PR #2311, opened 2025-01-16
Fix sprite statics conflict

rr-:
#### Checklist

- [x] I have read the [coding conventions](https://github.com/LostArtefacts/TR1X/blob/master/CONTRIBUTING.md#coding-conventions)
- [x] I have added a changelog entry about what my pull request accomplishes, or it is an internal change
- [x] I have added a readme entry about my new feature or OG bug fix, or it is a different change

#### Description

Resolves #2310.

#### Testing

- Look for missing static objects
- Look for missing sprites
- Check if animated sprites still work (the algae at the bottom of the Tihocan lake)
- TR1 and TR2

PR #2339, opened 2025-01-19
level: read objects and static objects in TRX

lahm86:
#### Checklist

#### Description

This moves object and static object reading to TRX, and tidies up related injections. For now, static sprites remain in the individual level modules, but I will follow this up with a separate PR as I have an idea for allowing sprites and statics to effectively have the same slot (ref #2311).

This also interprets static object flags (collidable and visible) directly on read to improve usage readability. I don't think there are any invisible static meshes in the original levels, but there are test levels [here](https://github.com/MontyTRC89/Tomb-Editor/pull/913) with the different types. Testing otherwise should be straight-forward, it will be immediately obvious if objects aren't loading properly.


src/libtrx/include/libtrx/game/objects/ids.h
```diff
@@ -6,3 +6,59 @@ typedef enum {
...
+    STATIC_DEBRIS_3       = 43,
+    STATIC_DEBRIS_4       = 44,
+    STATIC_DEBRIS_5       = 45,
+    STATIC_DEBRIS_6       = 46,
+    STATIC_DEBRIS_7       = 47,
+    STATIC_DEBRIS_8       = 48,
+    STATIC_DEBRIS_9       = 49,
+    STATIC_NUMBER_OF      = 50,
+    // clang-format on
+} GAME_STATIC_ID;
```
rr-: I think we can safely get rid of this enum type tbh and introduce MAX_STATICS instead.  WDYT?

rr- (approved):

src/libtrx/include/libtrx/game/objects/ids.h
lahm86: Good call, the types are never referenced so it makes sense to remove. Done.

aredfan (approved):

rr- (approved):

Issue #2309, opened 2025-01-16
TR1X Bug: Item and Ammo counter digit values between 2-5 are bugged when 'Fix sprite animations' is enabled


drydenj:
Does not occur in 4.4
Occurs in 4.5, 4.6, 4.7, 4.7.1, and dev releases thereafter.

When 'Fix sprite animations' is enabled, a bug will trigger after progressing to level 10 from either level 8 and 9, where item and ammo values with digits between 2-5 in any column appear to cycle upwards and loop between this range of numbers quickly and indefinitely.

To reproduce: 
1: /play 8 OR /play 9
2: /play 10
3: /give all
4: Equip the shotgun and observe the ammo counter

This behaviour is observed both in the inventory and open play UI.
Special items such as keys are also affected (/give key *5)
The stopwatch stats UI appears to be unaffected, though I only checked that in the latest dev -454-g66b5a13.

PR #2340, opened 2025-01-19
objects: refactor static mesh and room sprite interpretation

lahm86:
#### Checklist

#### Description

This creates firmer distinction between 3D static meshes and 2D room sprites, and so builds upon #2311 to allow the two to co-exist in similar slot numbers. Reading sprites is also moved to TRX, aside from injections currently, but this will be tackled under that refactor later.

I think this should also make it easier/clearer in the future when we need to extend `GAME_OBJECT_ID` - we'd just need to adapt the sprite sequence slot tests and introduce a fixed `OG_O_NUMBER_OF` const or similar.

Test level below with an animated room sprite in slot 0, a regular room sprite in slot 1 and a 3D static mesh in slot 0. Also worth double checking Cistern/Tihocan sprites, TR2's basement sprites and regular statics.
[level1.zip](https://github.com/user-attachments/files/18470313/level1.zip)

### DIFF
--- a/src/tr1/game/collide.c
+++ b/src/tr1/game/collide.c
@@ -346,7 +346,8 @@ bool Collide_CollideStaticObjects(
 
         for (int j = 0; j < r->num_static_meshes; j++) {
             const STATIC_MESH *const mesh = &r->static_meshes[j];
-            const STATIC_INFO *const sinfo = &g_StaticObjects[mesh->static_num];
+            const STATIC_OBJECT_3D *const sinfo =
+                &g_StaticObjects3D[mesh->static_num];
             if (!sinfo->collidable) {
                 continue;
             }
--- a/src/tr1/game/collide.c
+++ b/src/tr1/game/collide.c
@@ -347,7 +347,7 @@ bool Collide_CollideStaticObjects(
         for (int j = 0; j < r->num_static_meshes; j++) {
             const STATIC_MESH *const mesh = &r->static_meshes[j];
             const STATIC_OBJECT_3D *const sinfo =
-                &g_StaticObjects3D[mesh->static_num];
+                Object_GetStaticObject3D(mesh->static_num);
             if (!sinfo->collidable) {
                 continue;
             }
--- a/src/tr1/game/inject.c
+++ b/src/tr1/game/inject.c
@@ -445,7 +445,8 @@ static void M_TextureData(
             object->mesh_idx = mesh_idx + level_info->sprite_info_count;
             object->loaded = 1;
         } else if (object_id - O_NUMBER_OF < MAX_STATIC_OBJECTS) {
-            STATIC_INFO *object = &g_StaticObjects[object_id - O_NUMBER_OF];
+            STATIC_OBJECT_3D *object =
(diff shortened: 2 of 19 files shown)


src/tr1/game/collide.c
```diff
@@ -346,7 +346,8 @@ bool Collide_CollideStaticObjects(
 
         for (int j = 0; j < r->num_static_meshes; j++) {
             const STATIC_MESH *const mesh = &r->static_meshes[j];
-            const STATIC_INFO *const sinfo = &g_StaticObjects[mesh->static_num];
+            const STATIC_OBJECT_3D *const sinfo =
+                &g_StaticObjects3D[mesh->static_num];
```
rr-: Let's use `Object_GetStaticObject3D`.

src/tr1/game/inject.c
```diff
@@ -440,18 +440,18 @@ static void M_TextureData(
         const int16_t mesh_idx = VFile_ReadS16(fp);
 
         if (object_id < O_NUMBER_OF) {
-            OBJECT *object = &g_Objects[object_id];
+            OBJECT *const object = &g_Objects[object_id];
```
rr-: And the other getter functions

src/tr2/game/level.c
```diff
@@ -23,6 +23,7 @@
 #include <libtrx/game/level.h>
 #include <libtrx/log.h>
 #include <libtrx/memory.h>
+#include <libtrx/utils.h>
```
rr-: I'm not sure we need this include – we don't seem to need anything from it in the new code.

rr- (approved):
LGTM apart from minor code cleanups.
Terrific to see this distinction – trying to figure the purposes of relevant structures for the ticket mentioned in the OP was a big headache.

src/tr2/game/level.c
lahm86: Thank you. I had a temporary need for `ABS` in https://github.com/LostArtefacts/TRX/pull/2340/commits/999b9a3f1ec4dce6118e80fd57b10ed96d88a7c9, and then promptly forgot to remove this in the shift to TRX.

src/tr1/game/collide.c
lahm86: Thanks, updated to use the getters throughout.

aredfan:
I've found an issue that doesn't happen in the latest dev snapshot, it's similar to #2309 

1. `/set sprite 1`
2. `/play 9`
3. `/play 10`
4. type `I J K L` into the console, using capitals

lahm86:
> I've found an issue that doesn't happen in the latest dev snapshot, it's similar to #2309
> 
> 1. `/set sprite 1`
> 2. `/play 9`
> 3. `/play 10`
> 4. type `I J K L` into the console, using capitals

Great catch! Thank you, fixed.
Imagine re-introducing that bug so soon 🤦

aredfan (approved):
> Great catch! Thank you, fixed. Imagine re-introducing that bug so soon 🤦

Anytime, and you're very welcome. I think this PR looks good. 👍

rr- (approved):

PR #2361, opened 2025-01-22
Game flow follow-up work

### LINKED ISSUE #2360: TR1X bug: story so far in Folly is missing its opening FMV
The check in the `GF_PlayAvailableStory` is too eager and finishes it early, whereas it should only finish at `START_GAME`.

rr-:
#### Checklist

#### Description

- Resolves #2360 
- Fixes some tiny memory leaks in TR1 and TR2
- Moves `"demos"` to their own key in the TR2 gameflow JSON
- Does internal refactors to TR1 in preparation to merge via TRX (very preliminary work)
- Merges `start_cine` TR1 sequence event to `start_game`
- Merges `stop_game` TR1 sequence event to `loop_game` (they were always placed together)
- Merges `loop_cine` TR1 sequence event to `loop_game`
- Makes the `level_num` in the `play_level` TR2 sequence event optional
- Removes camera position sequence events from TR1 as they seem to be no longer required

More refactors to come…


aredfan:
**TR2X issues:**

- [x] The fmv before demo 3 seems to play with these steps: 

1. have fmvs enabled
2. /`play 1` (can be any level or demo)
3. `/demo 3`

- [x] In the final cutscene, the sound of the shotgun is cut off more early when compared to a recent dev snapshot. Load the attached save file and eliminate the last enemy [savegame.zip](https://github.com/user-attachments/files/18518430/savegame.zip)
- [x] This might be related to the prior issue - all the cutscenes are playing the CUT1 soundtrack, albeit with a delay.

rr-:
@aredfan LMK if the issues persist.

lahm86 (approved):
LGTM

aredfan:
> @aredfan LMK if the issues persist.

Initially the PR build didn't fix my latter 2 issues, so I've removed the audio folder from my TR2 directory and that actually fixed the issues.  I think the audio folder is now given priority over the music folder.  Previously, the music folder is given priority.

The issues with the audio folder preceded this PR.

(Apologies for my mistake.)

rr-:
> > @aredfan LMK if the issues persist.
> 
> Initially the PR build didn't fix my latter 2 issues, so I've removed the audio folder from my TR2 directory and that actually fixed the issues. I think the audio folder is now given priority over the music folder. Previously, the music folder is given priority.
> 
> The issues with the audio folder preceded this PR.

Thank you for this. You're absolutely right, that's related to the memory leak fix in this PR which didn't stop on the first matching backend and instead would continue to try all others. This is now fixed.

aredfan (approved):

PR #2363, opened 2025-01-23
Add support for optional dynamic lighting

### LINKED ISSUE #2357: Feature request: add optional dynamic gun/explosion lighting
Add an option to allow dynamic light for guns and explosions in TR1, similar to TR2. Also, make the feature itself optional in TR2.

lahm86:
Resolves #2357.

#### Checklist

#### Description

Adds dynamic lighting to TR1 gun flashes (Lara + enemies) and explosions (i.e. where the explosion sprite is used). This touches TR2 a little in that there is a shared function for Lara's gun flashes, so worth a quick test there too, including when Lara is on the armed skidoo. This also makes the feature itself optional in TR2.

Also fixed a potential div by zero in lighting output (unreleased).


aredfan:
Looks great! 👍 

The only issue I noticed is if you have `Gun/explosion flash` enabled and `Shotgun flash` disabled (which is related to drawing flame) - there's no flash for the shotgun. Fwiw the TR2 shotgun has a flash despite not drawing flame.

src/libtrx/config/map_tr1.def
```diff
@@ -86,6 +86,7 @@ CFG_BOOL(g_Config, gameplay.enable_uw_roll, true)
 CFG_BOOL(g_Config, input.enable_buffering, false)
 CFG_BOOL(g_Config, gameplay.enable_lean_jumping, false)
 CFG_BOOL(g_Config, gameplay.enable_target_change, true)
+CFG_BOOL(g_Config, gameplay.enable_gun_flash, true)
```
rr-: We already have `visuals.enable_shotgun_flash`, let's make it similar. For example, `visuals.enable_gun_lighting` or something. My proposal sounds poor, but the gist of it is to make these two options clearly discernible – that one is about sprites, and the other is about dynamic lighting.

I wouldn't limit this option to TR1, and would let the players also disable gun flashes and explosions in TR2.

lahm86:
Thank you, @aredfan, @rr-. The issue with the TR1 shotgun flash should be fixed now, and I have reworked this to make the feature itself optional in TR2 as well. I think `visuals.enable_gun_lighting` reads fine, I couldn't think of anything better than this without getting too wordy. I think the important thing is it distinguishes it from flare, Bartoli, dragon etc lighting which would be strange to disable.

aredfan (approved):
LGTM, and thank you. 👍

rr- (approved):

PR #2377, opened 2025-01-25
lara: introduce common hit frame logic

lahm86:
#### Checklist

#### Description

This replaces several duplicated switch statements for checking if Lara is in her hit animation with a common routine in TRX. It also renames related properties in `LARA_INFO` and elsewhere as applicable.

For testing, we can check that Lara and her braid are drawn properly while being hit by an enemy (and equally when not hit). Cutscenes too can be checked, although I don't think she would ever be hit here in OG, but the logic has changed slightly regardless. TR1 had left and right mixed up (in name only) as compared to TR2 (and WadTool), so worth a check that these correspond (looks OK to me). Anims 125-128 in WadTool for reference.

image
image


src/tr1/game/savegame/savegame_bson.c
```diff
@@ -773,11 +773,11 @@ static bool M_LoadLara(
...
-    lara->spaz_effect_count = JSON_ObjectGetInt(
-        lara_obj, "spaz_effect_count", lara->spaz_effect_count);
-    int spaz_effect = JSON_ObjectGetInt(lara_obj, "spaz_effect", 0);
-    lara->spaz_effect = spaz_effect && g_Config.gameplay.enable_enhanced_saves
-        ? Effect_Get(spaz_effect)
+    lara->hit_effect_count =
+        JSON_ObjectGetInt(lara_obj, "hit_effect_count", lara->hit_effect_count);
+    const int32_t hit_effect = JSON_ObjectGetInt(lara_obj, "hit_effect", 0);
+    lara->hit_effect = hit_effect && g_Config.gameplay.enable_enhanced_saves
+        ? Effect_Get(hit_effect)
```
rr-, line 780: This causes the saves to no longer have backwards compatibility, but I don't think this property is all that important?

lahm86: I missed that. You're right, but I think if Lara is in this state when saved, reloading the game would restore it as the enemy would still be at her.
Otherwise we could just write the new property names but retain the old ones for reading older saves for a period, and also check for the new names. I don't think it warrants a version update, WDYT?

rr-: I'd gladly lose backwards compatibility here due to the unfortunate naming, but we need to make sure the game doesn't do anything crazy if we load a save that has these properties active.

lahm86: It looks acceptable to me. The only instance where we lose an effect is if you save before this build when Lara has been hit by a mutant missile, then when you load she is in her normal state. But if it's an enemy hitting her, the game catches up anyway and sets her hit state.

rr- (approved):
Thank you for this. Nice screenshots BTW.

aredfan (approved):

PR #2380, opened 2025-01-25
tr2/lara/cheat: reset gun status if undrawing weapons

### LINKED ISSUE #2376: TR2X bug: undrawing a rifle type weapon ends up in hand when using the fly cheat
**Regression since:** [0.3](https://github.com/LostArtefacts/TR2X/releases/tag/0.3)
(There's no fly cheat in 0.2 and prior versions.)

_______________________

**Steps:**
1. have Lara holding a rifle type weapon
2. undraw the rifle
3. when the rifle is on Lara's back - use the fly cheat before her arm returns to its usual posture

I have missed this bug during initial tests when a similar issue was tackled last year. I've also made sure to check all the previous TR2X versions so I can confirm the bug's been persistently present since 0.3.

lahm86:
Resolves #2376.

#### Checklist

#### Description

This fixes guns potentially appearing in Lara's hand while entering the fly cheat and the weapon undraw animation is still running.

### DIFF
--- a/docs/tr2/CHANGELOG.md
+++ b/docs/tr2/CHANGELOG.md
@@ -39,6 +39,7 @@
 - fixed Skidoo snow wake effects at slow speeds (#2324, regression from 0.6)
 - fixed software renderer skybox occlusion issues (#2343, regression from 0.7)
 - fixed gunflare from bandits in Tibetan levels spawning too far from their guns (#2365, regression from 0.8)
+- fixed guns sometimes appearing in Lara's hands when entering the fly cheat while undrawing weapons (#2376, regression from 0.3)
 - improved rendering to achieve a slight performance boost in big rooms (#2325)
 
 ## [0.8](https://github.com/LostArtefacts/TRX/compare/tr2-0.8...tr2-0.8) - 2025-01-01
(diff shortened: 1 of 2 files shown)


rr- (approved):
Thanks – LGTM and it solves the original issue.
I noticed one more thing, but I don't think we need to solve it as it's really minor – if you time the fly cheat just right, the holsters sound is played twice:

https://github.com/user-attachments/assets/6abdd9d2-eda7-44db-91d5-30a2e9cfe7ea

lahm86:
I think we could resolve all of these fly cheat/gun issues by enforcing Lara out of the draw/undraw animations, but it would be quite jarring. At least for the most part she will complete putting the gun away neatly. This one looks quite tricky, maybe we'd have to be more thorough in our checks and look at frame numbers.

rr-:
> I think we could resolve all of these fly cheat/gun issues by enforcing Lara out of the draw/undraw animations, but it would be quite jarring. At least for the most part she will complete putting the gun away neatly. This one looks quite tricky, maybe we'd have to be more thorough in our checks and look at frame numbers.

I don't think it's worth it at the moment – we can raise an issue for this so that it doesn't get lost, but I'd say it's super low priority.

aredfan (approved):
I've checked all the rifles and everything looks good.  I also made sure to check the harpoon while Lara is in water for good measure.

Thank you for working on fixing this issue. 👍

PR #2381, opened 2025-01-25
Game flow rework (pt 6)

rr-:
#### Checklist

#### Description

This one is significantly smaller – I only move the demos in TR1 to their own key, removing the `"demo"` boolean property from individual levels. This now matches TR2 and I'd like to follow-up with cutscenes and the title level, and eventually the Gym (which likely will prove to be the hardest, as it's put _before_ regular levels.)
The TR1 GF level reader is refactored to have a similar structure to TR2.


lahm86 (approved):
LGTM.

As an aside, perhaps we should add a guard when starting a demo to verify that the level actually has demo data; currently it crashes if it's missing, but that's the same on develop.

rr-:
I'll address this and #2383 here.

rr-:
@lahm86 done

lahm86 (approved):

rr-:
(Just rebasing on top of newest develop.)

aredfan (approved):

walkawayy (approved):

PR #2388, opened 2025-01-25
Game flow rework (pt 7)

rr-:
#### Checklist

#### Description

This moves the cutscenes in TR1 to its own key. There are several things that needed to be done to make this work:

- Removes `GFS_EXIT_TO_CUTSCENE`
- Introduces `GFS_PLAY_CUTSCENE` from TR2
- Introduces `GFL_DUMMY` so that we don't have to change the level count in the main level list (needed for the legacy saves), and at the same time we don't need to provide all the usual information such as paths and whatnot for those levels. I'll later try to remove this type altogether with some heuristics in the savegame reader, but not now.
- Technically fixes some insignificant memory leaks when shutting down both TR1 and TR2.


lahm86:
The end of level stats don't show after the Qualopec cutscene. They appear fine after Tihocan and Atlantis cutscenes though.

lahm86:
Sorry, another issue is if you save in Qualopec, reload, then finish the level, the cutscene doesn't show (but the stats do).

rr-:
Thanks @lahm86 fixed both issues.

lahm86 (approved):
Thanks. LGTM

aredfan (approved):

PR #2394, opened 2025-01-26
packer: refactor texture packing

lahm86:
#### Checklist

#### Description

This moves the packer module to TRX and carries out some refactoring, mainly to update to standards we have for using `int32_t`, `const` etc. I've also adjusted the way it passes around rectangles, previously we were creating these on the heap so there was a lot of alloc/free calls required. The way virtual data is handled is also minimised for better efficiency. This gives a modest boost to injection loading overall.

I've also introduced 8-bit handling in preparation for TR2, the idea being we will pack the 32-bit textures just like TR1, and then replicate the same in the 8-bit pages by matching against the existing colour palette. This means we won't need to worry about changing the depth table with new colours, at least for an initial implementation.

Worth testing that TR1 levels all load OK, and injected textures look normal (Lara's braid, the explosion sprites, font sprites and Larson's gun in Qualopec).

### DIFF
--- a/src/libtrx/game/packer.c
+++ b/src/libtrx/game/packer.c
@@ -1,12 +1,9 @@
 #include "game/packer.h"
 
-#include "global/const.h"
-#include "global/types.h"
-#include "global/vars.h"
-
-#include <libtrx/log.h>
-#include <libtrx/memory.h>
-#include <libtrx/utils.h>
+#include "game/output.h"
+#include "log.h"
+#include "memory.h"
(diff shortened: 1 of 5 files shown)


aredfan (approved):

lahm86:
Updated to improve performance in 8-bit, so we create a palette LUT before packing to avoid too many look ups while moving a tex info. Thanks @rr-.

rr- (approved):

PR #2396, opened 2025-01-26
Game flow rework (pt 9)

rr-:
#### Checklist

#### Description

- **TR1: fixes "story so far" feature (broken in e04f955)**
- TR1: renames `play_synced_audio` event to `play_music` and its argument, `audio_id`, to `music_track`
- TR1: fixes docs on cutscene sequences
- TR1: fixes broken formatting of documentation on level types
- TR1: reorders sequence events and level properties according to importance
- TR1: clarifies that `level_stats` in Gym only does a fade-out
- TR1: drops `level_id` argument from `level_stats` events
- TR1: replaces `exit_to_level` event with `level_complete`
- TR2: introduces `play_music` event
- TR2: introduces `exit_to_title` event
- TR2: splits `level_stats` event from `level_complete`
- TR2: replaces `game_complete` event with `game_stats`
- TR2: removes `game_complete_track` in favor of `play_music` event
- TR2: removes `level_complete_track` in favor of `play_music` events
- TR2: replaces `DisplayCredits()` hardcoded sequence with JSON gameflow events
- TR2: adds fade-out effect to the Gym (via a `level_stats` event)

#### Next steps

I'd like to think of a way to merge both code bases via TRX next.
The biggest remaining difference between TR1 and TR2 is the presence of a separate `load_level` event in the TR1 game flow sequencer, which dictates how events such as `give_item` are internally implemented. Consolidating both code bases might require addressing this first. For now the leading idea is to introduce a separate sequence that only gets executed after a level is initialized, but before the game is played. This is similar to TR2's approach, but with a welcome reduction of global state. For TR1, it eliminates the need for the `load_level` event, making it less confusing for builders. Later down the line, this could seamlessly become a sensible LUA hook.

We still have `GFL_CURRENT` and `GFL_DUMMY` in TR1 but I suppose we can live with this for time being.

I'd rather tackle extracting gym to its own sequence once we have the majority of the game flow code in TRX.


walkawayy (approved):
TR1 and TR2 LGTM. Tried new games, cutscenes, FMVs, gym, and story so far in TR1.

lahm86 (approved):
Looks great 👍

aredfan (approved):

PR #2398, opened 2025-01-27
tr1/lara/state: check for responsive state changes

### LINKED ISSUE #2397: TR1X bug: Lara can potentially never jump or stop swimming
If responsive jumping is enabled, but `lara_animations.bin` is not injected (e.g. file missing, not referenced in gf etc), Lara refuses to jump at all. Similar issue for swim cancel.

lahm86:
Resolves #2397.

#### Checklist

#### Description

This augments the config options for responsive jumping and swimming to verify that relevant state changes exist in Lara's animation set. This can be tested by removing/re-adding `lara_animations.bin` and toggling the two related options.

### DIFF
--- a/src/tr1/game/lara/state.c
+++ b/src/tr1/game/lara/state.c
@@ -45,8 +45,11 @@ void (*g_LaraStateRoutines[])(ITEM *item, COLL_INFO *coll) = {
 };
 
 static bool m_JumpPermitted = true;
+static bool m_HasResponsiveJumping = false;
+static bool m_HasResponsiveSwimming = false;
 
 static int16_t M_FloorFront(ITEM *item, PHD_ANGLE ang, int32_t dist);
+static bool M_HasResponsiveState(LARA_ANIMATION anim_idx);
 
 static int16_t M_FloorFront(ITEM *item, PHD_ANGLE ang, int32_t dist)
 {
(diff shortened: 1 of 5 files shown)


src/tr1/game/lara/state.c
```diff
@@ -62,6 +65,32 @@ static int16_t M_FloorFront(ITEM *item, PHD_ANGLE ang, int32_t dist)
...
+
+    return false;
+}
+
+void Lara_State_Initialise(void)
+{
+    m_EnableResponsiveJumping =
+        g_Config.gameplay.enable_tr2_jumping && M_HasResponsiveState(LA_RUN);
+    m_EnableResponsiveSwimming = g_Config.gameplay.enable_tr2_swim_cancel
+        && M_HasResponsiveState(LA_SWIM_FORWARD);
```
rr-: Would it be possible to keep the `g_Config` property check for runtime? This way, we can ensure that the `/set` command can adjust the setting without needing to reload the level. While I'm not sure if this is how it currently works – for instance, using `/set braid on` does not take effect when the braid is turned off during loading the level, as the injections are skipped in that case – we can tackle that issue separately. My main point is to keep things forwards compatible, having this on our minds.

rr- (commented):
Thanks!

src/tr1/game/lara/state.c
lahm86: That's a good point, thanks. Will adjust.

lahm86: Done. To confirm, switching the options at runtime does work in develop, so this retains that ability. Thanks again.

And yes, the braid issue could be tackled I think. We could always inject the required data, but create separate meshes for Lara's head and torso and store these somewhere. Then the option would determine which meshes to reference. I'll open a ticket.

rr- (approved):

aredfan (approved):

PR #2400, opened 2025-01-27
level: read animated textures in TRX

lahm86:
#### Checklist

#### Description

This gets TR2 using the same structured animated texture approach as TR1, and shifts reading the data to TRX. Each game still animates in its own output module because of slight differences there in tick handling.


aredfan (approved):

src/tr1/game/level.c
```diff
@@ -607,49 +607,20 @@ static void M_LoadBoxes(VFILE *file)
...
 
-static void M_LoadAnimatedTextures(VFILE *file)
+static void M_LoadAnimatedTextures(VFILE *const file)
 {
     BENCHMARK *const benchmark = Benchmark_Start();
-    m_LevelInfo.anim_texture_range_count = VFile_ReadS32(file);
-    size_t end_position = VFile_GetPos(file)
-        + m_LevelInfo.anim_texture_range_count * sizeof(int16_t);
+    const int32_t data_size = VFile_ReadS32(file);
+    size_t end_position = VFile_GetPos(file) + data_size * sizeof(int16_t);
```
rr-: Nitpick: `const`

src/tr2/game/level.c
```diff
@@ -541,10 +541,15 @@ static void M_LoadBoxes(VFILE *const file)
 static void M_LoadAnimatedTextures(VFILE *const file)
 {
     BENCHMARK *const benchmark = Benchmark_Start();
-    const int32_t num_ranges = VFile_ReadS32(file);
-    g_AnimTextureRanges = GameBuf_Alloc(
-        sizeof(int16_t) * num_ranges, GBUF_ANIMATING_TEXTURE_RANGES);
-    VFile_Read(file, g_AnimTextureRanges, sizeof(int16_t) * num_ranges);
+    const int32_t data_size = VFile_ReadS32(file);
+    size_t end_position = VFile_GetPos(file) + data_size * sizeof(int16_t);
```
rr-: Nitpick: `const`

src/tr2/game/level.c
```diff
@@ -541,10 +541,15 @@ static void M_LoadBoxes(VFILE *const file)
...
+
+    const int16_t num_ranges = VFile_ReadS16(file);
+    LOG_INFO("%d animated texture ranges", num_ranges);
+    Output_InitialiseAnimatedTextures(num_ranges);
+    Level_ReadAnimatedTextureRanges(num_ranges, file);
+
+    VFile_SetPos(file, end_position);
```
rr-, line 553: Hmm isn't this function identical to TR1?

lahm86: It is, we also have others more or less identical like objects and static meshes. But TR1 currently has the extra injection counts in some others; I'm hoping eventually we can get everything common shifted to the TRX reader. So with that difference, I was thinking it'd be cleaner for now to keep the similar base private functions in the two level modules and just move the common "intensive" reading out to TRX.
But equally, I'm happy to shift the identical ones over now if that's preferred overall.

rr-: No it's OK to keep them around, whichever is more comfortable for you :relieved:

rr- (approved):

PR #2403, opened 2025-01-28
output: remove texture info globals

lahm86:
#### Checklist

#### Description

This removes `g_ObjectTextures`, `g_SpriteTextures`, `g_ObjectTextureCount` and `g_LabTextureUVFlag` in favour of common accessors. LMK about the last commit for the renderer if this is an acceptable approach. We would have to shift the routine for setting up the UV adjustments anyway ahead of injection work to allow for additional tex infos to be added, so thought it best to tackle now.

Just a quick test is needed I think in game to ensure textures appear OK, as well as sprites, and with bilinear toggled and SWR etc.

### DIFF
--- a/src/libtrx/game/level/common.c
+++ b/src/libtrx/game/level/common.c
@@ -382,7 +382,7 @@ void Level_ReadObjectTextures(
     }
 
     for (int32_t i = 0; i < num_textures; i++) {
-        OBJECT_TEXTURE *const texture = &g_ObjectTextures[base_idx + i];
+        OBJECT_TEXTURE *const texture = Output_GetObjectTexture(base_idx + i);
         texture->draw_type = VFile_ReadU16(file);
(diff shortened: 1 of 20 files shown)


rr- (approved):
LGTM. The UV adjustment code is fine too. BTW maybe we could allocate textures dynamically via gamebuf like level.c things?  The pixel data will be held in GFX Surface anyway so it shouldn't use up too much space.
Offtop: do we know what the Lab Texture UV Flag is for?

lahm86:
Sounds good. I was thinking we could tackle raising the various limits next too, so shifting to game buf would be nice to do in the same task. I think we'd need to increase the game memory though to match TR1 as we're currently only on 7.5MB compared with 128MB.

I _think_ the lab UV flag is related to the bleeding issue with bilinear filter on, but I admit I don't fully understand how it works.

aredfan (approved):

PR #2405, opened 2025-01-28
Raise TR2 texture limits

### LINKED ISSUE #1795: Feature request: raise texture info limit
The OG texture info limit is 2048, and sprite texture limit is 512; these should be raised to match TR1X.

### LINKED ISSUE #1796: Feature request: raise texture page limit
The OG texture page limit is 16, so should be raised to match TR1X.
I don't think this limit is checked/enforced (unlike other limits) but the game can crash if more than 16 pages are present.

lahm86:
Resolves #1795.
Resolves #1796.

#### Checklist

#### Description

This raises the following limits:
- Object textures (2048 => unlimited) - TR1 and TR2
- Sprite textures (512 => unlimited) - TR1 and TR2
- Texture pages (32 => 128) - TR2

The internal `game_buf` now also matches TR1 at 128 MB. Object and sprite textures are now allocated in game memory as opposed to keeping two permanent arrays at the maximum size. Texture pages are already using game memory. As an aside, there is a bit of a disjoint in the way we initially process texture pages, mainly because of injections in TR1, so I'll aim to get this behaving more commonly in a follow-up PR.

I've also separated the texture storage/handling from `output.c`, making split modules here instead. I'm aiming also to get texture pages moved from globals into the new texture module. The split is in its own commit, so this may be best reviewed overall per-commit.

I've tested a level with everything at the maximum and it loads fine. For regular levels, there should be no change.


aredfan:
Found an issue - the game crashes when I play The Deck (10) or Catacombs of the Talion (13), then use the level skip cheat.

lahm86:
Thanks @aredfan, fixed.

@rr- I've added an `Output_UnloadBackground` call when unloading a level, I think we were using the previous level's object texture in some cases, so because we're now clearing out the array this was a null. Maybe this call would be better in `M_End` in the stats phase? Or perhaps flushing it regardless when unloading the level is best as we don't necessarily know the GF sequences that have preceded.

aredfan (approved):

src/libtrx/include/libtrx/game/output/const.h
```diff
@@ -1,14 +1,8 @@
...
-    #define MAX_OBJECT_TEXTURES 8192
-    #define MAX_SPRITE_TEXTURES 1024
-    #define MAX_TEXTURE_PAGES 128
-#else
-    #define MAX_OBJECT_TEXTURES 2048
-    #define MAX_SPRITE_TEXTURES 512
-    #define MAX_TEXTURE_PAGES 32
-#endif
+#define MAX_OBJECT_TEXTURES 8192
+#define MAX_SPRITE_TEXTURES 1024
```
rr-: Do we still need these limits, considering we now allocate these dynamically?

rr- (approved):

src/libtrx/include/libtrx/game/output/const.h
lahm86: Thank you, great point. I've removed the limits for the texture infos; I'll look at potentially doing the same for pages in the next PR. I updated the lab texture array too here to use game_buf as well.

rr-:
> @rr- I've added an `Output_UnloadBackground` call when unloading a level, I think we were using the previous level's object texture in some cases, so because we're now clearing out the array this was a null. Maybe this call would be better in `M_End` in the stats phase? Or perhaps flushing it regardless when unloading the level is best as we don't necessarily know the GF sequences that have preceded.

Apologies – I missed your comment earlier. I think that normally, the module that loads the backgrounds should also be the one to unload them. The `O_INV_BACKGROUND` does stand out as a special case, but unloading it from `Level_Unload()` could unintentionally impact other types of backgrounds, like image-based ones. This might not cause noticeable issues to the player, but it does result in a bit chaotic dependencies.

Having `Phase_End` always handle background unloading could be a better approach, but it's not perfect either. On reflection, even the current setup is a bit shaky – phasers using backgrounds assume these will not be changed, not even by nested phasers. To deal with this, they'd need to manage changes during suspend/resume routines, likely by checking what the current background is.

With all that in mind, how do you feel about adding a function to check the background type, then unloading it only if it's `INV_BACKGROUND_OBJECT` during the object/texture unloading process?

lahm86:
Thanks, I've added this check now. LMK if this looks OK.
@aredfan looks OK to me in-game, but would appreciate a second quick test if you get a chance.

aredfan (approved):
> @aredfan looks OK to me in-game, but would appreciate a second quick test if you get a chance.

Of course. 👍 

I can confirm all the OG levels are loading properly, LGTM.

rr- (approved):

docs/tr1/CHANGELOG.md
```diff
@@ -11,7 +11,8 @@
 - added exit fade-out effects (#2348)
 - added optional dynamic lighting for gun flashes and explosions, similar to TR2+ (#2357)
 - ⚠️ changed the game data to use a separate strings file for text information, removing it from the game flow file
-- changed the sprite limit from 512 to 1024
+- changed the sprite texture limit from 2048 to unlimited (within game's overall memory cap)
```
walkawayy, line 14: @lahm86 @rr- this first TR1 changelog entry should say `object texture limit` I believe. I think it's wrong in `develop` currently.

lahm86: You're right. I can fix in the next texture PR. Thanks!

PR #2414, opened 2025-01-29
tr1/inject: improve texture injection performance

lahm86:
#### Checklist

#### Description

This removes texture packing from the injection process. We instead inject the individual pages that come with the injections (which are packed optimally anyway, albeit with respect to themselves only). Packing will be re-introduced for unoptimized page groups at a later date when we overhaul the injection format.

I've also replaced the Colosseum door fix with a simple frame rotation adjustment, rather than replacing the full model. This in turn means one less texture page is injected here. Ref #1288 and https://github.com/LostArtefacts/TRXInjectionTool/pull/23.


aredfan (approved):

rr- (approved):
Did we keep the 8 bit palette matching?

lahm86:
> Did we keep the 8 bit palette matching?

That's still in the packer module, but unused as TR1 is promoted to 32-bit on read, and we expand the base palette if needed in the injection module (still TR1 only). Once TR2 injections are up to speed we'll certainly use the matching for 8-bit again.

PR #2417, opened 2025-01-31
tr1/data: fix underwater lever animation

### LINKED ISSUE #2416: TR1X - Underwater Switch Bug
As I was testing a custom level for someone, me and the builder have encountered a strange bug- They made a sequence where you use two underwater switches to open a door, with 2-3 bits being switched on per underwater switch. If you pull both levels nothing breaks, however if you only use one switch, for some reason Lara breaks and I am unable to use ACTION or READY WEAPON at all, and the look camera slows down. I am guessing it's related to the fact that the underwater switch is only changing some and not all bits.

lahm86:
Resolves #2416.

#### Checklist

#### Description

This adds a "hands-free" command at the end of Lara's underwater lever animation, which is in line with TR2. It prevents issues after using an underwater lever followed by entering the wading state. The additional check in Lara's state control also matches TR2, which is one that we missed before.

Test level below. If you pull the lever and then try to climb the block at the starting position, or draw guns, it will fail on the current release.
[level1.zip](https://github.com/user-attachments/files/18623440/level1.zip)


lahm86:
Ref https://github.com/LostArtefacts/TRXInjectionTool/pull/24

aredfan (approved):
LGTM 👍

rr- (approved):

PR #2428, opened 2025-02-01
output: move depthQ and gouraud tables to TRX

lahm86:
#### Checklist

#### Description

This moves the depthQ and gouraud tables into TRX, and also populates these tables for TR1 rather than skipping it; it obviously remains unused there in the rendering for now. I've also removed some unnecessary functions for palette and texture page loading, as these were just calling the TRX functions anyway.

We only really need to check TR2 SWR with this one, although a quick check that TR1 loads OK would be good.


lahm86:
^ Rebased.

rr- (commented):
Can we rename depth queue to LightTable or LightMap, and gouraud entry to LightMapInverse? Or something better though I'm AFK ATM so can't think of good suggestions. 

One mapping is used to remap the entire palette depending on the brightness (which is 0-31 inclusively): `remapped_palette = g_DepthQTable[brightness]`. The other is to get all shadings of a single color: `int shadings[] = g_GouraudTable[color]`. Maybe we could merge them, too.

lahm86:
Sure, I've gone with `LIGHT_MAP` and `SHADE_MAP` and updated throughout accordingly. I'm not sure about merging them as the sizes are different e.g. one is 32x256 and the other 256x32. LMK though if you have an idea for this.

aredfan (approved):

rr- (approved):

PR #2434, opened 2025-02-02
Add /cutscene and /gym console commands; introduce console registry

rr-:
#### Checklist

#### Description

This PR adds two new QoL commands: `/cut` (`/cutscene`) and `/gym` (`/home`), and reorganizes the command registration to use a new console registry module. The main reason for introducing the registry is to cut down on the repetitive work needed to add a new command. Currently, the process requires:

- To create a command `.c` file with the implementation.
- To create a command `.h` file with the headers (only a global variable).
- To declare the `.c` file in `meson.build`.
- To set up the variable with the command inside the .c file.
- To declare the command global variable in the .h file.
- To include the command `.h` file from within the setup.
- To reference the command global variable inside the list of supported commands.

Instead, this uses `__attribute__((__constructor__))` C extension (which is supported by gcc, MinGW and clang) that causes the decorated function to be executed (in random order) prior to `main()`. With a little macro help, it lets us set it up in a much nicer way, and now only the following steps are necessary:

- To create a command `.c` file with the implementation
- To declare the `.c` file in `meson.build`
- To invoke `REGISTER_CONSOLE_COMMAND(prefix, entrypoint_func)` macro at the bottom of the .c file


lahm86 (commented):
A great change 😌 

The new commands work fine in TR1; in TR2 issuing `cut` from the title screen doesn't seem to work, it immediately returns to the title screen. It does work fine from within a level. I'm not seeing any clues in the log file.

aredfan:
Very nice! 👍 

Another small issue in TR2X - using `/cut 1` while in the inventory also returns you to the main menu without playing the cutscene.

rr-:
@lahm86 @aredfan the issue was related to debouncing the input, eg this issue doesn't manifest if the player manages to hold the console confirmation button for a single frame.

Came up with a generic fix, that needs to be tested in relation to input buffering.

lahm86 (approved):
Input buffering looks good to me 👍

aredfan (approved):

PR #2436, opened 2025-02-02
objects: remove object globals

lahm86:
#### Checklist

#### Description

This replaces all instances of `g_Objects[id]` with `Object_GetObject(id)`, and moves the objects themselves to be stored in TRX. Similarly, the 2D and 3D globals are now removed (previous commits already switched to the getters for these). All related reading is also done in TRX now.

There are a couple of TODOs off the back of this, but I will tackle them separately
- refactoring the main object setup module in TR1 so we don't pass the object and instead the individual setups are responsible for it (like TR2).
- introducing a relative anim/frame accessor for items to tidy up several places where we have identical logic (e.g. Bacon Lara, skater boy control etc).


rr- (commented):
WDYT about shortening `Object_GetObject` to `Object_Get` (and `Object_GetStatic2D` or `Object_Get2DStatic`) in the spirit of `Item_Get`?

src/tr1/game/text.c
```diff
@@ -200,8 +200,9 @@ void Text_DrawText(TEXTSTRING *const text)
             goto loop_end;
         }
         Output_DrawScreenSprite2D(
-            sx, sy, 0, sh, sv, g_Objects[O_ALPHABET].mesh_idx + glyph->mesh_idx,
-            16 << 8, 0, 0);
+            sx, sy, 0, sh, sv,
+            Object_GetObject(O_ALPHABET)->mesh_idx + glyph->mesh_idx, 16 << 8,
```
rr-: This can be just `obj`.

rr- (commented):
Re: initialization refactor – let's put this one on hold IMO, as I think we can use the recently adopted `constructor` attribute for this to save us _tons_ of boilerplate.

All in all great changes. I didn't expect that it would be this small.

src/tr1/game/text.c
rr-: BTW it'd be good to settle on whether we prefer `obj` or `object`. I'm fine with either.

rr- (approved):

src/tr1/game/text.c
lahm86: Thanks, yes I think we should go with `obj` as we currently have more of that than `object`. I'll follow up with a separate PR as the renames will be more involved.

lahm86:
> Re: initialization refactor – let's put this one on hold IMO, as I think we can use the recently adopted `constructor` attribute for this to save us _tons_ of boilerplate.
> 
> All in all great changes. I didn't expect that it would be this small.

That's great, I'll hold off on that one for sure.

aredfan (approved):

PR #2440, opened 2025-02-03
objects: rename object variables

lahm86:
#### Checklist

#### Description

A follow-up from #2436, this uses `obj` as the standard for naming `OBJECT` variables. There were more of these than `object` so I went with that. I've also shortened `Object_IsObjectType` to `Object_IsType` for simplicity.

I wasn't sure about doing the same for where we have e.g. `GAME_OBJECT_ID object_id` to just become `obj_id`. Can easily add another commit if we want to go with this for consistency.


aredfan (approved):

rr- (approved):
Thanks. Let's go with `obj_id`.

lahm86:
Thanks, done. Had it prepared.

rr- (approved):

PR #2447, opened 2025-02-04
rooms: move room reading and storage to TRX

lahm86:
#### Checklist

#### Description

This replaces all instances of `g_RoomInfo[idx]` with `Room_Get(idx)`, replaces `g_RoomCount` with a getter in TRX, renames variables to a standard of `room` (mix of this and just `r` previously) and shifts room reading fully to TRX. I've re-organised the common level reader a little as the room mesh function is now private, so this part may be best viewed per-commit.


src/libtrx/game/rooms/common.c
```diff
@@ -96,6 +99,27 @@ static const int16_t *M_ReadTrigger(
...
+{
+    m_RoomCount = num_rooms;
+    m_Rooms = num_rooms == 0
+        ? nullptr
+        : GameBuf_Alloc(sizeof(ROOM) * num_rooms, GBUF_ROOMS);
+}
+
+int32_t Room_GetTotalCount(void)
+{
+    return m_RoomCount;
```
rr-, line 112: Maybe we could do `Room_GetCount`?

src/tr1/game/camera/common.c
```diff
@@ -465,7 +467,8 @@ static void M_SmartShift(
     const bool bad_bottom =
         M_BadPosition(test, ideal->y, ideal->z, ideal->room_num);
     if (!bad_bottom) {
-        camera_box = r->sectors[z_sector + (x_sector + 1) * r->size.z].box;
+        camera_box =
+            room->sectors[z_sector + (x_sector + 1) * room->size.z].box;
```
rr-: It'd be nice to have a getter for this, eg `Room_Get???Sector???(ROOM *room, int32_t x_sector, int32_t z_sector)`. I can't think of a good naming as we have the more complex `Room_GetSector()` already that operates on a world pos.

src/tr2/game/viewport.c
```diff
@@ -202,3 +202,13 @@ void Viewport_AlterFOV(const int16_t view_angle)
...
+
+int32_t Viewport_GetMaxX(void)
+{
+    return g_PhdWinMaxX;
+}
+
+int32_t Viewport_GetMaxY(void)
+{
+    return g_PhdWinMaxY;
+}
```
rr-, line 214: While this change is fine, there's something I've been wanting to discuss with you. I'm not really satisfied with how the Viewport API is structured in both games in its present form. In TR1 we have a lot of getters, while in TR2, we're overly dependent on using global variables. My original concept for TR2 was to use a function like `Viewport_Get()` that would return a constant struct containing all the necessary information. But unfortunately the game relies on even more variables than TR1 so this struct turned out to be big and ugly. For modifying the viewport, there are a couple of scenarios to consider:
- resizing the window, of course
- changing the FOV during cutscenes (happens on every frame)
- changing the FOV during cinematic animations (happens on every frame)
- changing the FOV when the players tweaks their settings
- drawing a 3D pickup in the corner (in which case the viewport is temporarily changed)

Can you think of any design that we might use to improve this? It's difficult for me to estimate the effect of fusing some of the game variables together – and the whole purpose of the `game_vars` struct was to avoid handling this issue at the time :D

src/libtrx/game/level/common.c
```diff
@@ -72,6 +94,74 @@ static void M_ReadFace3(FACE3 *const face, VFILE *const file)
     face->enable_reflections = false;
 }
 
+static void M_ReadRoomMesh(const int32_t room_num, VFILE *const file)
+{
+    ROOM *const room = Room_Get(room_num);
+    const INJECTION_MESH_META inj_data = Inject_GetRoomMeshMeta(room_num);
```
rr-, line 100: Having this dependency here is unfortunate – a better design would be to pass a dictionary of room number to injection mesh data via a parameter, like we do with `extra_pages` in `Level_ReadTexturePages`. But this would be way wordier.

lahm86: It's a bit of a hangover from TR1 injections, it'll likely change when we redo the format there. Potentially something passed to the module as a whole after inject_init with all the additional information we need here.

src/tr1/game/camera/common.c
lahm86: Yes, I think we can come up with something. I'll try it out

src/tr2/game/viewport.c
lahm86: Hmm, that's a tricky one as you say as sometimes making a struct or ton of getters/setters explodes the complexity. Maybe we can figure something out in the refactor though, happy to help look into it.

src/libtrx/game/rooms/common.c
lahm86: Done, thanks.

src/tr1/game/camera/common.c
lahm86: We now have `Room_GetWorldSector` for getting a room's sector given X/Z world coords, and `Room_GetUnitSector` for getting one by X/Z index.

aredfan (approved):

rr- (approved):
This is great stuff.

PR #2448, opened 2025-02-04
Add /debug to TR1 that draws room triggers

rr-:
#### Checklist

#### Description

A fun side project. TR1 only, as TR2 lacks the means to draw this sort of primitive yet.


walkawayy (approved):
Works really well and looks great! I didn't notice any issues.

lahm86 (approved):

aredfan:
I found a few issues:
- [x] The game stops responding when I load Return to Egypt if debug is enabled.
- [x] If I load Atlantean Stronghold then tp to room 24, the game stops responding with debug enabled.
- [ ] If the uw triggers are viewed from above water, the missing distortion effect gives this appearance.

20250205_080240_Laras_Home

rr-:
>- The game stops responding when I load Return to Egypt if debug is enabled.
>- If I load Atlantean Stronghold then tp to room 24, the game stops responding with debug enabled.

These got fixed.

>- If the uw triggers are viewed from above water, the missing distortion effect gives this appearance.

Unfortunately, there isn't much I can do about this without significant preprocessing – which, in my opinion, isn't worth the effort. The wobble effect applies to vertices in 2D screen space after the projection matrix, meaning it occurs quite late in the pipeline. Meanwhile, the room geometry vertices aren't associated with specific sectors (which is exactly what the preprocessing would need to address), so I can't easily look up the final vertex coordinates for particular sectors. Let's leave it as-is IMO.

aredfan (approved):

PR #2453, opened 2025-02-06
items: migrate to TRX

lahm86:
#### Checklist

#### Description

I'm going to do an incremental PR for this move, as it's a bit more involved. I'll add a comment with each increment for what to check, then once approved will move to the next stage and re-request review.


lahm86:
The first stage replaces the majority of `g_Items[idx]` calls with `Item_Get`. `items.c` in both games remains untouched as I want to try to move some common routines that use the array out to TRX, so will do that in following PRs.

The only logical changes at the moment are as follows:
- adjusting the check for Lara locking on to a new target in TR1
- simplifying a check for Lara's drawing in TR2 (while she's holding or using the M16)

aredfan (approved):

lahm86:
Second stage removes the global _level_ item count in favour of a getter, and adds a utility for such things as pods and statues to create an item _and_ increment the level count. This count is independent of the actual total item count, as we need to retain it for savegames.

Good things to test in various save/load states (TombATI saves would be good as well to check):
- the injected PSX mummy in Khamoon
- pods
- centaur statues
- Thor's hammer
- Skidoo drivers
- The dragon

aredfan (approved):
LGTM👍

walkawayy (approved):

src/tr2/game/lara/misc.c
```diff
@@ -1803,3 +1803,14 @@ void Lara_Extinguish(void)
...
+bool Lara_IsM16Active(void)
+{
+    if (g_Lara.weapon_item == NO_ITEM || g_Lara.gun_type != LGT_M16) {
+        return false;
+    }
+
+    const ITEM *const item = Item_Get(g_Lara.weapon_item);
+    return item->current_anim_state == 0 || item->current_anim_state == 2
+        || item->current_anim_state == 4;
+}
```
rr-, line 1816: Can we add a description to what this function does?

rr- (approved):
Thank you.

src/tr2/game/lara/misc.c
lahm86: Thanks, done. Added it to the header. It's sort of an awkward function but as the code was duplicated in four places, I felt it warranted separating out.

lahm86:
The last three commits above should wrap-up this PR for now. Item storage and reading is now fully in TRX, and I've adjusted object setup and item initialisation to be called commonly from TRX as well, because of the need for this to happen after injections and in the defined order.

A general test here would be good to ensure items activate (enemies, traps, doors etc) and that they are killed. And a sanity check that item state persists over save/load would be good. Thanks.

src/libtrx/game/items.c
```diff
@@ -60,6 +92,148 @@ int16_t Item_CreateLevelItem(void)
...
+        lara->target = nullptr;
+    }
+
+#if TR_VERSION == 1
+    item->hit_points = -1;
+    item->flags |= IF_KILLED;
+#else
+    if (item_num < m_LevelItemCount) {
+        item->flags |= IF_KILLED;
+    }
```
rr-, line 112: I wonder why we're doing this check in TR2+. I mean, why would the game treat dynamically created objects any different?

src/libtrx/game/level/common.c
```diff
@@ -822,3 +823,16 @@ void Level_LoadPalettes(LEVEL_INFO *const info)
     Memory_FreePointer(&info->palette.data_24);
     Memory_FreePointer(&info->palette.data_32);
 }
+
+void Level_LoadObjectsAndItems(void)
+{
+    // Object and item setup/initialisation must take place after injections
+    // have been processed. We also need to use a cached item count as
```
rr-, line 830: Let's avoid 'we' and use more robot wording.

rr- (approved):
It's unfortunate that there are places that need to mess with `m_PrevItemActive` (and need `Item_SetPrevActive`) but it is what it is.

src/libtrx/game/items.c
lahm86: Yes, it seems a bit odd, but I figured safer for now to keep it as-is. I'll need to study elsewhere to see if the flag is used in different ways possibly first.

lahm86:
> It's unfortunate that there are places that need to mess with `m_PrevItemActive` (and need `Item_SetPrevActive`) but it is what it is.

Yes, it is a shame. I _believe_ this is only needed for the body bag feature; I'm not sure if we want to retain that as it seems only required for performance reasons, but that's a separate task to investigate anyway. Would much prefer not to expose this property if we could.

rr- (approved):

aredfan (approved):

PR #2457, opened 2025-02-07
sound: migrate sound data to TRX

### LINKED ISSUE #2458: TR1X bug: secret and speech sounds can fail to play
If a level has been compiled without the secret and enemy speech sound effects, but the player has the option enabled for these sounds not to kill any playing music track, the sounds do not play at all. It should fall back to the external files in this case.

### LINKED ISSUE #2460: OG bug: levels with no SFX data cannot be loaded
Currently TR1 will exit to desktop (with a message) if a level has no SFX data; OG TR2 crashes to desktop, but TR2X has already fixed this.

lahm86:
Resolves #2458.
Resolves #2460.

#### Checklist

#### Description

I'll aim to do another incremental PR here, similar to #2453.


lahm86:
The first stage moves sound sources over to TRX. These are fairly straight-forward to test, you can see where they are located in trview and we just need to double check that the flip status matches up (e.g. the first waterfall sound in Lost Valley should stop after the flipmap).

I also renamed these from sound effects to sound sources, which is the TombEditor name, and I think helps distinguish from sound effects that are triggered and not continuous.

aredfan:
LGTM

rr- (approved):

lahm86:
The sample LUT and sample infos themselves are now in TRX. The common level reader here needs a few additional arguments to support TR1 injections, but this will be tidied up during the injection refactor.

I've also included the fix for #2458, test levels attached.
SFX missing: [level1.zip](https://github.com/user-attachments/files/18719377/level1.zip)
SFX present: [level1.zip](https://github.com/user-attachments/files/18719388/level1.zip)

Otherwise, for testing, this touches all sound effects so good to do a general check to ensure nothing odd is happening. TR1 sound injections should still work (PSX uzis, guns in the gym and explosions).

rr- (approved):

aredfan:
LGTM 👍

lahm86:
The final stage moves music track flags to TRX, so that's all the music/sound globals taken care of. I opted to put the array in `music.c` despite their use being heaviest in `room.c`; this does add quite a few extra lines for setting the flags, but IMO this module is best placed for them logically. But LMK what you think.

Things to check are one-shot music triggers behaving (+save/load) and the assault course delayed tracks in TR2.

src/libtrx/include/libtrx/game/music/const.h
```diff
@@ -0,0 +1,3 @@
+#pragma once
+
+#define MAX_CD_TRACKS 64
```
rr-: Let's name this `MAX_MUSIC_TRACKS`.

lahm86: Thanks, done. Added the commit and changelog entries for #2460 as well.

aredfan (approved):

PR #2496, opened 2025-02-13
game-buf: rewrite to use arena allocator

rr-:
#### Checklist

#### Description

A proposal on dropping the fixed-size game buffer capacity, which saves us some 8 ms at the game startup, and should theoretically slightly reduce the game's memory footprint. It changes the game buffer to grow in increments of 5 MB at a time rather than allocating a fixed-size very large buffer.

### DIFF
--- a/src/libtrx/game/game_buf.c
+++ b/src/libtrx/game/game_buf.c
@@ -1,52 +1,28 @@
 #include "game/game_buf.h"
 
 #include "enum_map.h"
-#include "game/shell.h"
-#include "log.h"
 #include "memory.h"
 
-static size_t m_Cap = 0;
-static size_t m_MemUsed = 0;
-static size_t m_MemFree = 0;
-static char *m_MemBase = nullptr;
-static char *m_MemPtr = nullptr;
+static MEMORY_ARENA_ALLOCATOR m_Allocator = {
+    .default_chunk_size = 1024 * 1024 * 5,
+};
 
-void GameBuf_Init(const size_t cap)
+void GameBuf_Init(void)
 {
-    m_Cap = cap;
-    m_MemBase = Memory_Alloc(cap);
-    GameBuf_Reset();
 }
 
 void GameBuf_Reset(void)
 {
-    m_MemPtr = m_MemBase;
-    m_MemFree = m_Cap;
-    m_MemUsed = 0;
+    Memory_ArenaReset(&m_Allocator);
 }
 
 void GameBuf_Shutdown(void)
 {
-    Memory_FreePointer(&m_MemBase);
-    m_Cap = 0;
-    m_MemFree = 0;
-    m_MemUsed = 0;
+    Memory_ArenaFree(&m_Allocator);
 }
 
 void *GameBuf_Alloc(const size_t alloc_size, const GAME_BUFFER buffer)
 {
     const size_t aligned_size = (alloc_size + 3) & ~3;
-
-    const char *const buffer_name = ENUM_MAP_TO_STRING(GAME_BUFFER, buffer);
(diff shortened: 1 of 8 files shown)


aredfan (approved):

src/libtrx/game/game_buf.c
```diff
@@ -1,52 +1,28 @@
...
-#include "log.h"
 #include "memory.h"
 
-static size_t m_Cap = 0;
-static size_t m_MemUsed = 0;
-static size_t m_MemFree = 0;
-static char *m_MemBase = nullptr;
-static char *m_MemPtr = nullptr;
+static MEMORY_ARENA_ALLOCATOR m_Allocator = {
+    .default_chunk_size = 1024 * 1024 * 5,
```
walkawayy, line 7: This is 5 MB but your PR description says 10 MB.

src/libtrx/memory.c
```diff
@@ -51,3 +52,52 @@ char *Memory_DupStr(const char *const string)
...
+}
+
+void Memory_ArenaReset(MEMORY_ARENA_ALLOCATOR *const allocator)
+{
+    MEMORY_ARENA_CHUNK *chunk = allocator->first_chunk;
+    while (chunk != nullptr) {
+        chunk->offset = 0;
+        chunk = chunk->next;
+    }
+    allocator->current_chunk = allocator->first_chunk;
```
walkawayy, line 114: I think there might be an issue with this code but I'm not sure. From a quick look:

It looks like `Memory_ArenaReset` will set the offset of all chunks to 0 and keep the chain, but `Memory_ArenaAlloc` reaches the end of the current alloc buffer and makes a new one every time. It won't check that a "next" already exists, and will make a new chunk and then set that as the new next, breaking the chain and causing a memory leak.

It should also ensure that the amount requested is in fact smaller than the new chunk, otherwise it will make something too small and then whatever wanted it will trash memory as it goes over the "chunk" size.

walkawayy (changes requested):

src/libtrx/game/game_buf.c
```diff
@@ -1,52 +1,28 @@
...
-    m_MemFree = 0;
-    m_MemUsed = 0;
+    Memory_ArenaFree(&m_Allocator);
 }
 
 void *GameBuf_Alloc(const size_t alloc_size, const GAME_BUFFER buffer)
 {
     const size_t aligned_size = (alloc_size + 3) & ~3;
-
-    const char *const buffer_name = ENUM_MAP_TO_STRING(GAME_BUFFER, buffer);
```
lahm86, line 40: I suppose we could get rid of `GAME_BUFFER` altogether now as its only purpose was to log on failure. Or would you prefer to keep it for potential other uses? E.g. maybe logging of memory used with a console command, although that would need it stored somewhere.

src/libtrx/game/game_buf.c
rr-: I updated it accordingly as 10 MB felt excessive.

src/libtrx/memory.c
rr-: Thanks, that's a great catch. I'll update accordingly.

src/libtrx/game/game_buf.c
rr-: I was thinking of removing this enum too, but even if we're not using it in the proposed code form, I think keeping it around might prove useful, for instance if we want to check how much memory each buffer occupies.

src/libtrx/memory.c
rr-: @walkawayy updated.

lahm86 (approved):

walkawayy (approved):

Issue #1328, opened 2024-04-20
Passport page flipping is slow


Richard-L:
The passport flip animation needs to finish before a new flip input can be registered. Before you could cancel a flip at any time for faster inputs.
The cancelling is possible only after a certain time, about half of the flip animation perhaps, in v3.1.1 and OG. But it's a problem in v4.0.3.
Sadly can't test v4.0 as the download is instantly removed as a virus for me again...

The best solution would be that inputs can be registered immediately, and any animations cancelled accordingly.

rr-:
I tried to replicate this but was unsuccessful. The most recent significant change to the passport code occurred between versions 3.0 and 3.1, so I'd expect the last version with a noticeable change to be 3.0, rather than 3.1 as the version 3.1 should be similar to what we see in 4.0.3.

walkawayy:
I tested as far back as 2.13.2, and I don't see any difference in behavior. You can inconsistently flip slightly faster in all versions if the input is timed. But it's not really skipping and it's not that much faster.

The difference in behavior may be related to the slight input lag added from the animation interpolation IMO.

Richard-L:
That's a good thought. Sorry I didn't test back myself and thanks you did so!

I did feel if it wouldn't just be desirable to make it skipable full stop. I still stumble upon this when navigating the passport. Think it's much nicer without any forced input delay whatsoever here.

walkawayy:
@Richard-L how weird is this lol. Rudimentary attempt at removing the delay as a proof of concept.

[TR1X No Delay Inv v1.zip](https://github.com/LostArtefacts/TR1X/files/15301944/TR1X.No.Delay.Inv.v1.zip)

Richard-L:
@walkawayy It's fantastic, please merge, and thank you so much ❤️

Richard-L:
@walkawayy just to feed back on this again as I'm missing it not being merged all the time 😇: the only issue left to address would be that on using quickload/-save, the passport open anim is skipped too which wasn't the intention with this issue - that is to only allow skipping page flipping upon manual input.

When using quick save: The passport animates in and flips one page before the save menu appears.
When using quick load: The passport animates in and doesn't flip a page before the save menu appears.

In your above build, these animations are both skipped unfortunately.

For a memory aid:

https://github.com/user-attachments/assets/222ad0dd-cd25-4222-88aa-a504a8c69325

PR #2502, opened 2025-02-14
Refactor TR1 passport code

rr-:
#### Checklist

#### Description

This updates the TR1 code to align more closely with the TR2 code. We keep `PASSPORT_MODE_BROWSE` to allow players to "confirm" entry into a passport page (for instances where the page initially appears blank and requires pressing the Action key to display, like loading, saving, or starting a new game). Additionally, we keep the `mode` property, instead of inferring the mode from the active page like in TR2, to support navigation to and from the savegame-specific actions menu (eg. "Select level" and "Story so far" features).

This prepares TR1 for #1328.

### DIFF
--- a/src/tr1/game/option/option_passport.c
+++ b/src/tr1/game/option/option_passport.c
@@ -32,6 +32,7 @@ typedef enum {
 } PASSPORT_TEXT;
 
 typedef enum {
+    PAGE_UNDETERMINED = -1,
     PAGE_1 = 0,
     PAGE_2 = 1,
     PAGE_3 = 2,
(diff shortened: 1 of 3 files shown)


lahm86:
- If Lara dies and the passport displays, if you select `Load Level` but then decide you want to either `Restart Level` or `Exit to Title`, pressing Esc from the load page has no effect.

I also think there is a problem with story so far, but it seems to be on develop too - will raise separately.

aredfan:
- [x] Press F5 in the gym and the passport shows an arrow briefly when flipping pages.
- [x] If no saves are present, opening the passport manually via the inventory shows an arrow briefly when flipping pages. This happens on the main menu as well.
- [x] If no saves are present and Lara dies in-game, the passport shows an arrow briefly when flipping pages,
- [x] Open the passport manually via the inventory, select load/save game to show the dialog, then deselect and this also closes the passport instead of returning to the blank state. This happens on the main menu as well, including the new game dialog if enabled.

For the top 3 issues, the `Game` text under the passport is disappearing too quickly.

rr-:
@aredfan @lahm86 fixed, but it'll need testing TR2, too, as I had to rearrange the code a bit.

lahm86 (approved):
LGTM (built locally)

aredfan (approved):

PR #2505, opened 2025-02-14
option/passport: react faster to player inputs

### LINKED ISSUE #1328: Passport page flipping is slow

rr-:
#### Checklist

#### Description

Resolves #1328. Builds on top of #2502.
Note that this is not optional, and I'm not sure what to think about it, as it might be quite invasive to some players who like to mash buttons. LMK your opinion.


lahm86 (approved):
This looks good IMO. I tried mashing inputs and can't see any detrimental effects, so IMO it's fine that this isn't optional.

aredfan (approved):

rr-:
@aredfan @lahm86 I've rebased on newest develop + added a setting whether to enable this, which is on by default. I'd appreciate a sanity recheck.

aredfan (approved):

lahm86 (approved):

PR #2513, opened 2025-02-16
room: move flipmap vars and logic to TRX

lahm86:
#### Checklist

#### Description

This moves all variables related to flipmaps to TRX, along with the main logic to flip the map. We also had a comment previously in `Room_FlipMap` to check if item/effect assignment was required; I tested this and it is - without it, items disappear after the flipmap. So that comment is removed.

Things to test:
- general flipmap activation/deactivation
- saves in various flip states
- flip effects, like rushing water in The Cistern
- built-in flipmaps: lightning in Folly, Pyramid; cabin in Natla's Mines; lava bowl in Ice Palace
- dead enemies in flipped rooms in TR2 should disappear (like Living Quarters)
- pushblocks should still work, like Barkhang room 150 after flipping


src/tr1/game/item_actions/chain_block.c
```diff
@@ -8,20 +8,21 @@
...
 
-    if (g_FlipTimer == 0) {
+    int32_t flip_timer = Room_GetFlipTimer();
+    if (flip_timer == 0) {
         Sound_Effect(SFX_SECRET, nullptr, SPM_NORMAL);
     }
 
-    g_FlipTimer++;
-    if (g_FlipTimer == 55) {
+    Room_SetFlipTimer(++flip_timer);
```
rr-: Let's avoid preincrementing the arguments.

src/tr1/game/item_actions/flood.c
```diff
@@ -23,5 +24,5 @@ void ItemAction_Flood(ITEM *item)
         Sound_Effect(SFX_WATERFALL_BIG, &pos, SPM_NORMAL);
     }
 
-    g_FlipTimer++;
+    Room_SetFlipTimer(++flip_timer);
```
rr-: Perhaps Room_IncrementTimer(int32_t frames);?

src/tr1/game/level.c
```diff
@@ -594,15 +594,9 @@ bool Level_Initialise(const GF_LEVEL *const level)
...
     GF_SetCurrentLevel((GF_LEVEL *)level);
-    g_FlipEffect = -1;
 
     Overlay_HideGameInfo();
 
-    g_FlipStatus = 0;
-    for (int32_t i = 0; i < MAX_FLIP_MAPS; i++) {
-        g_FlipMapTable[i] = 0;
-    }
-
```
rr-, line 605: Is it safe to remove this?

src/tr1/game/savegame/savegame_legacy.c
```diff
@@ -665,8 +665,12 @@ bool Savegame_Legacy_LoadFromFile(MYFILE *fp, GAME_INFO *game_info)
     }
 
     M_ReadLara(&g_Lara);
-    M_Read(&g_FlipEffect, sizeof(int32_t));
-    M_Read(&g_FlipTimer, sizeof(int32_t));
+    int32_t flip_effect;
+    int32_t flip_timer;
+    M_Read(&flip_effect, sizeof(int32_t));
```
rr-, line 670: Let's use M_ReadS32 as a safer alternative.

rr- (commented):
Thank you 😌

src/tr1/game/level.c
lahm86: Yes, this has been merged into `Room_InitialiseFlipStatus`, which we call from `Level_ReadRooms`. So it just happens a little later, but still before the level actually finishes loading.

src/tr1/game/savegame/savegame_legacy.c
lahm86: We don't have this in TR1's legacy save module. Should we introduce it from TR2?

src/tr1/game/item_actions/flood.c
lahm86: Thanks, I've introduced this and tidied things up. In some cases we check frame numbers so that's been tweaked to take into account the local copies of the value, which are also now const for clarity.

src/tr1/game/item_actions/chain_block.c
lahm86: Fixed, coupled with update below.

src/tr1/game/savegame/savegame_legacy.c
rr-: Ideally, yes,  but since this would be about changing the entire file, let's do that separately.

rr- (approved):

aredfan (approved):

PR #2527, opened 2025-02-17
tr1/objects: use new constructor-based setup

rr-:
#### Checklist

#### Description

This pull request eliminates the majority of object headers, retaining only those that provide a significant public API, like `Pierre_Reset` or `BaconLara_InitialiseAnchor`. It restructures the `Setup` calls into pointers within the `OBJECT` type, which are assigned during program startup via a new `REGISTER_OBJECT` macro. This macro generates a distinct constructor function for each evaluation. Object-specific functions such as `Collision`, `Draw` are now private.

I'll do a similar change to TR2 next.

#### Considerations

`MovableBar` used `Cog_Control`, which makes little sense.  I changed it to duplicate the logic of the cog module under its own namespace. This is a poor solution, but I couldn't think of anything better.

The transition wasn't the smoothest, as we've had a few checks dependent on comparing `control_func` or `collision_func`. To make it possible for these functions to become private, I switched to using raw ID checks. However, in a subsequent pull request, I plan to implement inversion of control and introduce savegame-related function pointers within the `OBJECT` type:
- `process_item_after_load_func`
- `process_item_before_load_func`
- `process_item_before_save_func`

These names are a bit wordy, and I'd appreciate thoughts on this.
While this will let us eliminate most of the object ID-based logic from the savegame code, TR2 vehicles are more involved:
```c
        switch (item->object_id) {
        case O_BOAT:
            M_Write(item->data, sizeof(BOAT_INFO));
            break;

        case O_SKIDOO_FAST:
            M_Write(item->data, sizeof(SKIDOO_INFO));
            break;

        case O_LIFT:
            M_Write(item->data, sizeof(LIFT_INFO));
            break;
        }
```
so even with this approach, some degree of hardcoding the object IDs would remain.

I'd also like to rename `initialise_func` to `initialise_item_func` because having `setup_func` and `initialise_func` can be confusing due to their similar names. This change will make it clearer that each function serves a different purpose.


aredfan (approved):

lahm86:
The non-flying mutants don't spawn - flyers are fine, as are ones that hatch in Atlantis. So e.g. Egypt mummies, Sanctuary non-flyers etc.

lahm86:
Re:
- `process_item_after_load_func`
- `process_item_before_load_func`
- `process_item_before_save_func`

Perhaps we could use a single function with an enum param e.g.:
```c
typedef enum {
   PRE_SAVE,
   PRE_LOAD,
   POST_LOAD,
} SAVE_EVENT_TYPE;
```
And then e.g. `process_save_event(SAVE_EVENT_TYPE type, int16_t item_num)`.

aredfan:
> The non-flying mutants don't spawn - flyers are fine, as are ones that hatch in Atlantis. So e.g. Egypt mummies, Sanctuary non-flyers etc.

To reproduce the bug described by Lahm, you'll need to load a level with grounded mutants which will appear normally, then reload or load another level with grounded mutants again for the bug to appear.

rr-:
Thanks @lahm86 @aredfan fixed. The object copy done in mutant init functions would reset the `setup_func` pointer to that of a flyer's.

Great idea to use an argument to indicate the mode instead of cluttering the `OBJECT` with multiple callbacks. I would keep the `_func` suffix, though, to show it's a function pointer. I know it sounds like Hungarian notation, but if we drop `_func`, we'd have to rename `draw_func` to just `draw`. This could create confusion, as the name `draw` might mistakenly suggest a boolean property, similar to `is_drawn`.

aredfan (approved):

lahm86 (approved):

walkawayy (approved):
LGTM. I like getting rid of most headers. What's the biggest advantage of the switch to `__attribute__((constructor))`? Just that we don't need all the overhead in the old `setup` file?

rr-:
> LGTM. I like getting rid of most headers. What's the biggest advantage of the switch to `__attribute__((constructor))`? Just that we don't need all the overhead in the old `setup` file?

Correct :)

PR #2529, opened 2025-02-18
pathing: move pathing data to TRX

lahm86:
#### Checklist

#### Description

This moves box, overlap and zone data to TRX, which were the last parts of the level data that were stored globally. I have also updated TR2 boxes to use world rather than sector units to reduce the amount of repeated shifting needed, and to help bring the camera and box modules closer in line with each other in both games.

Probably best viewed by commit.

I've been testing this quite a lot as pathfinding is so finicky, and indeed I did have some problems along the way; hopefully these have all been addressed. Fortunately, if pathfinding is off, it's usually very obvious e.g. demos will play out very badly. The other area impacted for testing is the camera, so this should behave normally.


src/libtrx/game/level/common.c
```diff
@@ -713,10 +713,10 @@ void Level_ReadPathingData(VFILE *const file)
         box->top = VFile_ReadS32(file);
         box->bottom = VFile_ReadS32(file);
 #else
-        box->left = VFile_ReadU8(file);
-        box->right = VFile_ReadU8(file);
-        box->top = VFile_ReadU8(file);
-        box->bottom = VFile_ReadU8(file);
+        box->left = VFile_ReadU8(file) << WALL_SHIFT;
+        box->right = (VFile_ReadU8(file) << WALL_SHIFT) - 1;
```
rr-, line 717: I believe this might lead to _slight_ inconsistencies. For instance, original `Box_ValidBox` implementation expects that the `right` property does not have `1` subtracted, and we maintain a non-strict inequality in the proposed changes. At the same time, the TR1 code also utilizes a non-strict version, and given how similar the math is between both games, I'm assuming the `right` and `bottom` level data values _do_ have `1` subtracted in TR1.
All in all let's double-confirm this minor difference with some debugger and then we can decide which version to go with. Fortunately, there aren't that many usages.

lahm86: Apologies, I forgot to talk about this, and thanks for raising it.

TR1 data does have 1 subtracted from right and bottom in all cases in the OG levels, and this is also how TE writes TR1 boxes. The original camera code for TR2 did also subtract one, and also in other places in `box.c`. While I couldn't see any issues in game, perhaps we should add the 1 back in for those other cases on TR2. I think it's a literal edge case where things may be affected, but then perhaps the question too is if TR1 should have the 1 added to match.

Can do some more investigation tomorrow - I'll try to see if I can forge an outcome on the edge.

lahm86: I wasn't able to find anything conclusively about the +1 snag, so I decided to just reinstate it for the cases in TR2 `box.c` where it previously was using full sector units. I think we should perhaps re-address this when we come to merge the modules more fully in the future, and I've marked comments as reminders. It'll save clouding this PR with potential regressions in this respect, when it should just focus on the data shift to TRX, and the standardizing of the box values. WDYT?

rr-: That's the best of the two worlds. Thank you.

rr- (approved):

aredfan (approved):

PR #2534, opened 2025-02-20
objects: handle savegames by themselves

rr-:
#### Checklist

#### Description

This refactor changes the system so that each object in the game can set itself up when a game is loaded, instead of relying on one central piece of code to do it for all objects. This makes the code easier to work with because each object knows how to handle its own setup. It's done by introducing a new callback, `handle_save_func` in `OBJECT` with an additional argument, `SAVEGAME_STAGE` that tells whether it's called before loading the item, after loading the item, or before saving the item to the savegame file.

#### Considerations

Unfortunately, we still do have three specific objects references left hardcoded in TR2's savegame code, and these are the boat, the skidoo, and the lift. This is because unlike other objects that simply change level geometry or flip some internal flags, these require reading and writing additional state to the savegame data. I'd rather not touch this until we have BSON saves in place.

I've coupled the objects module to rely on the savegame module solely for the `SAVEGAME_STAGE` enum, which creates a soft cyclic dependency between them. I think that's acceptable.

Additionally, there's one event, `SAVEGAME_STAGE_BEFORE_SAVE`, which exists solely for the save crystal object. The crystal needs to be removed after the player confirms the save slot, but also needs to stay when the player chooses not to save. Rather than passing the outcome whether the save was done or not all the way to `GF_ShowInventory(INV_SAVE_CRYSTAL_MODE)`, the crystal chooses a different approach and handles this with this pseudo-event. It's a bit on the ugly side, considering it's literally the only object in the game that could benefit from this enum member, but I couldn't think of anything better. This approach also allows `GF_ShowInventory` to become asynchronous at some point (even if it currently is not).

#### Testing

We need to test saving and loading all objects affected by these changes - both prior to "consuming" and after "consuming" an object. Here's a handy table.

| Object            | TR1 | TR2 | Notes                                                 |
| ---               | --- | --- | ---                                                   |
| Baldy             | ✅  |     | Speech needs to trigger once                          |
| Cowboy            | ✅  |     | Speech needs to trigger once                          |
| Larson            | ✅  |     | Speech needs to trigger once                          |
| Pierre            | ✅  |     | Speech needs to trigger once                          |
| Mutant pods       | ✅  |     | Exploded pods need to stay invisible                  |
| Save crystals     | ✅  |     | Consumed crystals disappear                           |
| Sliding pillar    | ✅  |     | (Qualopec) Level geometry stays as expected           |
| Push blocks       | ✅  | ✅  | Level geometry stays as expected                      |
| Pickups           | ✅  | ✅  | Collected pickups disappear                           |
| Puzzle holes      | ✅  | ✅  | Used puzzle holes cannot be reused                    |
| Dragon            |     | ✅  | Dragon needs to stay as a skeleton after death        |
| Black skidoo      |     | ✅  | The skidoo needs to be mountable after driver's death |
| Aquatic mine      |     | ✅  | (Bartoli's Hideout)                                   |
| Smashable windows |     | ✅  | Smashed windows needs to stay invisible and passable  |

### DIFF
--- a/src/libtrx/include/libtrx/game/objects/types.h
+++ b/src/libtrx/include/libtrx/game/objects/types.h
@@ -3,6 +3,7 @@
 #include "../anims/types.h"
 #include "../collision.h"
 #include "../items/types.h"
+#include "../savegame.h"
 #include "../types.h"
 
 #include <stdint.h>
@@ -59,6 +60,7 @@ typedef struct OBJECT {
         const ITEM *item, int32_t x, int32_t y, int32_t z, int16_t height);
     int16_t (*ceiling_height_func)(
         const ITEM *item, int32_t x, int32_t y, int32_t z, int16_t height);
+    void (*handle_save_func)(ITEM *item, SAVEGAME_STAGE stage);
 #if TR_VERSION == 1
     const OBJECT_BOUNDS *(*bounds_func)(void);
     bool (*is_usable_func)(int16_t item_num);
--- a/src/libtrx/include/libtrx/game/savegame.h
+++ b/src/libtrx/include/libtrx/game/savegame.h
@@ -2,6 +2,12 @@
 
 #include <stdint.h>
 
+typedef enum {
+    SAVEGAME_STAGE_BEFORE_LOAD,
+    SAVEGAME_STAGE_AFTER_LOAD,
+    SAVEGAME_STAGE_BEFORE_SAVE,
+} SAVEGAME_STAGE;
+
 // Remembers the slot used when the player starts a loaded game.
 // Persists across level reloads.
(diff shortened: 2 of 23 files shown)


aredfan:
Issues (testing in progress):

- [x] Have Lara collect the scion in Tomb of Qualopec, then save and reload and the scion reappears on the altar.

rr-:
> Issues (testing in progress):
> 
> * [ ]  Have Lara collect the scion in Tomb of Qualopec, then save and reload and the scion reappears on the altar.

Thanks – this is fixed.

lahm86 (approved):
LGTM. Found #2535 but probably best tackled separately as it's OG.

aredfan (approved):

PR #2538, opened 2025-02-20
Fix Jade Dragon disappearance in Bartoli's Hideout

### LINKED ISSUE #2536: TR2X bug: jade secret persists after detonation in Bartoli's Hideout
**Dev snapshot:** 0.9.2-38-g8456e59

______________

1. `/play 3`
2. `/tp det` x2
3. use the all items cheat
4. detonate the bomb
5. fly to the secret in the destroyed building

Image

rr-:
#### Checklist

#### Description

Resolves #2536.

The original game has a unique logic for the Jade Dragon, and only the jade variant alone, where the dragon removes itself immediately upon activation. So in terms of game mechanics, when Lara uses the detonator, it triggers the Dragon pickup, and that makes it disappear.

To me, it's a very bizarre approach, so rather than restoring it, I went with making pickups use the `IF_INVISIBLE` flags whenever their host room gets flipped, and not allowing Lara to see or interact with them if this flag is set. This method feels more conventional to me.

That said, I find it hard to believe that this is the only pickup in both games that is placed in a flippable room, requiring such intricate trigger-based solution from the OG code, and that this scenario is not already handled in some other way. Also, I'm unsure if my usage of the flag is good, and would appreciate @lahm86's opinion on this. We hardly ever seem to use `IF_INVISIBLE` anywhere beyond initial item setup.

Similar to the newest changes done in regard to the savegame↔specific object code decoupling, I've seen that the room flip logic also references certain objects, and I introduced a new method to `OBJECT` called `handle_flip_func` to address this.

#### Testing

- A quick confirmation if the push blocks and sliding pillars work as intended with flipped rooms.
- Possibly a test level that lets you flip rooms that contain pickups on and off would be good to know for sure it's working as intended.
- Also it would be good to programmatically verify whether there are any more pickup items in either game that are triggered by something.

### DIFF
--- a/src/tr2/game/room.h
+++ b/src/tr2/game/room.h
@@ -25,8 +25,3 @@ int32_t Room_GetCeiling(const SECTOR *sector, int32_t x, int32_t y, int32_t z);
 
 void Room_TestTriggers(const ITEM *item);
 void Room_TestSectorTrigger(const ITEM *item, const SECTOR *sector);
-
-// TODO: eliminate
-int16_t Room_Legacy_GetDoor(const SECTOR *sector);
-void Room_Legacy_TestTriggers(const int16_t *fd, bool heavy);
-void Room_Legacy_TriggerMusicTrack(int16_t value, int16_t flags, int16_t type);
--- a/docs/tr2/symbols.txt
+++ b/docs/tr2/symbols.txt
@@ -3121,11 +3121,11 @@ typedef enum {
 0x00414B70  0x0198  SECTOR *__cdecl Room_GetSector(int32_t x, int32_t y, int32_t z, int16_t *room_num);
 0x00414D10  0x0168  int32_t __cdecl Room_GetWaterHeight(int32_t x, int32_t y, int32_t z, int16_t room_num);
 0x00414E80  0x0265  int32_t __cdecl Room_GetHeight(const SECTOR *sector, int32_t x, int32_t y, int32_t z);
-0x00415100  0x00E7  void __cdecl Camera_Legacy_RefreshFromTrigger(int16_t type, const int16_t *data);
-0x004151F0  0x0690  void __cdecl Room_Legacy_TestTriggers(int16_t *data, int32_t heavy);
+0x00415100  0x00E7  void __cdecl Camera_RefreshFromTrigger(int16_t type, const int16_t *data);
+0x004151F0  0x0690  void __cdecl Room_TestTriggers(int16_t *data, int32_t heavy);
 0x004158D0  0x0055  int32_t __cdecl Item_IsTriggerActive(ITEM *item);
 0x00415930  0x023D  int32_t __cdecl Room_GetCeiling(const SECTOR *sector, int32_t x, int32_t y, int32_t z);
-0x00415B90  0x004E  int16_t __cdecl Room_Legacy_GetDoor(const SECTOR *sector);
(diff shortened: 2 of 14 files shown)


lahm86:
I've checked throughout OG TR1 and 2 and the only other place items are triggered is in Natla's Mines - the pistols, and one of the fuses.

The current implementation doesn't seem to fix the Bartoli's issue though. I tried inverting the condition, and it does fix it, but it then also causes all other pickup items in a flip room to disappear. I _think_ this is needed (same in TR1):
```diff
diff --git a/src/tr2/game/objects/general/pickup.c b/src/tr2/game/objects/general/pickup.c
index 3d4720aa3..1e0bdb8ae 100644
--- a/src/tr2/game/objects/general/pickup.c
+++ b/src/tr2/game/objects/general/pickup.c
@@ -231,8 +231,10 @@ static void M_Setup(OBJECT *const obj)
 
 static void M_HandleFlip(ITEM *const item, const ROOM_FLIP_STATUS flip_status)
 {
-    if (flip_status == RFS_UNFLIPPED) {
-        item->flags |= IF_INVISIBLE;
+    if (flip_status == RFS_FLIPPED) {
+        if (Item_IsTriggerActive(item)) {
+            item->flags |= IF_INVISIBLE;
+        }
     } else {
         item->flags &= ~IF_INVISIBLE;
     }
```

Test level here with a pushblock, regular items and a secret set up like Bartoli's.
[WALL.zip](https://github.com/user-attachments/files/18890341/WALL.zip)

rr-:
@lahm86 @aredfan updated. Thanks a lot to Lahm for helping me create a good solution.

Testing:
- TR1: the Fuse pickup in Natla's Mines is not visible until Lara activates the conveyor belt
- TR1: the Pistols pickup in Natla's Mines is not visible until Lara lowers the porta-cabin
- TR2: the Jade Dragon secret pickup in Bartoli's Hideout is visible until Lara explodes the building using the detonator box

src/tr2/game/room.h
```diff
@@ -25,8 +25,3 @@ int32_t Room_GetCeiling(const SECTOR *sector, int32_t x, int32_t y, int32_t z);
 
 void Room_TestTriggers(const ITEM *item);
 void Room_TestSectorTrigger(const ITEM *item, const SECTOR *sector);
-
-// TODO: eliminate
```
lahm86, line 29: Thanks for tidying this up, I forgot to come back to this 😄

lahm86 (approved):

aredfan:
This might be a separate issue because it's also present in the latest dev snapshot - the jade dragon persists with these steps.

1. `/play 3`
2. `/tp sec` x2
3. pick up the jade dragon
4. `/tp det`
5. use the all items cheat and activate the detonator
6. fly to the room where the secret was
7. `/save 1`
8. `/load 1`

**Edit:** The fuse and pistols in Natla's Mines LGTM.

rr-:
@aredfan thank you :) Fixed, but it'll require retesting all scenarios from scratch. The logic is rather delicate unfortunately.

aredfan (approved):
> @aredfan thank you :) Fixed, but it'll require retesting all scenarios from scratch. The logic is rather delicate unfortunately.

Anytime 👍

I tested all the scenarios again including Lahm's test level, and your fix checks out in all cases. LGTM.

lahm86 (approved):

Issue #2036, opened 2024-12-14
OG bug: various issues with pushblocks


lahm86:
Future task to bring pushblock implmentation into line with TR1. Following issues/PRs for reference. This may be better suited as individual tickets, but keeping this here for the time being as a tracker for investigation.

- [ ] https://github.com/LostArtefacts/TRX/issues/619
- [ ] https://github.com/LostArtefacts/TRX/pull/1050
- [ ] https://github.com/LostArtefacts/TRX/issues/1079
- [ ] https://github.com/LostArtefacts/TRX/issues/2035

PR #2542, opened 2025-02-21
rooms: move height vars and functions to TRX

### LINKED ISSUE #2541: Feature: add support for no floor in TR1
Add support for custom levels to use the `disable_floor` GF sequence in TR1, similar to TR2's Floating Islands.

lahm86:
Resolves #2541.

#### Checklist

#### Description

This primarily replaces `g_HeightType` and `g_GF_NoFloor` with function calls in TRX, but also moves `Room_GetHeight` and `Room_GetCeiling` together as well for consistency. There remains some height specific logic in the separate modules, for example altering floor height, but that's linked with #2036 so can be handled separately there.

It seemed sensible to offer `disable_floor` in TR1 as part of this as it's quite a straight-forward feature.

For testing, we can check floor and ceiling heights are OK, including slopes, plus the abyss in Floating Islands. You can also try an abyss in TR1; one I tried was setting the value to 4096 in Great Pyramid and then dropping off the starting ledge. It doesn't really make sense in OG environments though, so it's definitely more of a custom level feature.


src/tr2/game/game_flow/sequencer_events.c
```diff
@@ -200,7 +200,7 @@ static DECLARE_GF_EVENT_HANDLER(M_HandleDisableFloor)
 {
     GF_COMMAND gf_cmd = { .action = GF_NOOP };
     if (seq_ctx != GFSC_STORY) {
-        g_GF_NoFloor = (int16_t)(intptr_t)event->data;
+        Room_SetNoFloorHeight((int16_t)(intptr_t)event->data);
```
rr-: Maybe `SetAbyssHeight`? or `SetVoidHeight`. I'm also perfectly fine with the current naming.

rr- (approved):
LGTM and thank you.

src/tr2/game/game_flow/sequencer_events.c
lahm86: Thanks, I think `Abyss` is a good choice; `NoFloor` was a bit awkward.

rr- (approved):

aredfan (approved):
Overall LGTM.

A potential issue I noticed is when I use the `disable_floor` in Great Pyramid, there's also invisible collision around the top platform that Lara will bump into if she flies around.  Is that to be expected?

lahm86:
Thanks @aredfan. That's a good point, and could cause issues in TR2 customs as well; the hard-coding of 16384 as the absolute abyss maximum causes this, as the height at the top of the platform in GP causes things to overflow (note there is other collision here just beyond the scion platform and near the sloped ceiling, but that's OG). I've adjusted it now so it calculates the absolute floor, based on the difference between 16384 and OG floater abyss value. It'd be worth flying around near the roof in Floater to see if everything looks ok with this, and see how GP looks too. Thanks!

aredfan (approved):
Thank you for explaining.  The issue is fixed, and the Floater and GP levels both looks good. 👍

src/libtrx/game/rooms/common.c
```diff
@@ -511,12 +512,14 @@ SECTOR *Room_GetSkySector(
 
 void Room_SetAbyssHeight(const int16_t height)
 {
-    m_AbyssHeight = height;
+    m_AbyssMinHeight = height;
+    m_AbyssMaxHeight = height == 0 ? 0 : m_AbyssMinHeight + 30 * STEP_L;
```
rr-: Hmm, what does `30 * STEP_L` signify? Asking cause it doesn't match 16384 (16 tiles).
Also would be good to explain what is the min and max role in the calculations, as it's become quite confusing to me. I always thought it's a simple plane.

lahm86: My bad, I got my numbers muddled. I've set it now to `26 * STEP_L`, so this is the difference between 16384 and Floater abyss height of 9728. This allows time when Lara is falling between setting her HP to 0 and the inventory showing, so that we don't hear Lara hit the floor. Although I've noticed that in OG, you can _just_ hear the crunch SFX begin as the inventory opens, but we can perhaps address that separately.
We also clamp to just under the max height; in TombEditor in theory you could build a room this low.
I've added comments, hopefully they clear things up.

rr- (approved):

PR #2546, opened 2025-02-22
tr2/flares: fix flare throw reset test

### LINKED ISSUE #2545: TR2 bug: issues affecting Lara
I've ran into these 2 issues while testing #2544 but I cannot find a way to reliably reproduce them, thus it'll be difficult to pinpoint the regressions at this time.

- savegame 0 : Lara is stuck and keeps moving her arm as if she's throwing away a flare.
- savegame 1: Lara won't interact with a zipline handle

[savegame.zip](https://github.com/user-attachments/files/18925097/savegame.zip)

lahm86:
Resolves #2545.

#### Checklist

#### Description

This reverts to testing open upper end frame range when Lara throws a flare. I think the problem can happen because throwing a flare doesn't use the conventional state change system, and relies on frame numbers in `g_Lara`, and so I think it all depends on exactly what Lara is doing when throwing a flare. Unfortunately, I don't think we can recover from the saves provided in the issue as-is in the codebase - we would need to introduce a guard in Lara's main control to check for this, because at this stage, the flare control has finished. In-game, entering and exiting the fly cheat can be used to fix the problem.

It's hard to test this one. I suppose we can spend a decent amount of time disposing flares in various states, unless we can find a way to reproduce it exactly.

### DIFF
--- a/docs/tr2/CHANGELOG.md
+++ b/docs/tr2/CHANGELOG.md
@@ -2,6 +2,7 @@
 - added a `/cheats` console command
 - added a `/wireframe` console command (#2500)
 - fixed smashed windows blocking enemy pathing after loading a save (#2535)
+- fixed a rare issue whereby Lara would be unable to move after disposing a flare (#2545, regression from 0.9)
 
 ## [0.9.2](https://github.com/LostArtefacts/TRX/compare/tr2-0.9.1...tr2-0.9.2) - 2025-02-19
 - fixed secret rewards not handed out after loading a save (#2528, regression from 0.8)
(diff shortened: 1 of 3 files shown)


rr- (approved):
LGTM although it'd be great to confirm if it's indeed the right fix by at least provoking the problem programmatically.

aredfan (approved):
Thank you. I'm 60% sure this issue is fixed but it's hard to know for sure.

lahm86:
I was able to recreate the problem by hardcoding `g_Lara.flare_frame` to go beyond the animation's range; the fix does seem to resolve the problem here. I think we should be safe to merge this, but perhaps continue to keep an eye on it in general testing.

rr- (approved):

PR #2549, opened 2025-02-23
tr2/movable_block: fix various pushblock issues

### LINKED ISSUE #2036: OG bug: various issues with pushblocks

### LINKED ISSUE #2193: OG bug: falling pushblock uses enemy grunt SFX
The ID used is 70, which is the T-rex footstep in TR1, but this is a goon grunting in TR2.

lahm86:
Resolves #2036.
Resolves #2193.

#### Checklist

#### Description

Following on from #2542, I wanted to get this one done to allow shifting the alter floor height function to TRX (which I will do separately after this). Test level attached, and the linked TR1 issues that @walkawayy fixed originally are in #2036. The only thing I haven't done is make falling pushblocks deadly, but as that's a new feature I'll put it into its own ticket for another time.

[WALL.zip](https://github.com/user-attachments/files/18931149/WALL.zip)

### DIFF
--- a/docs/tr2/README.md
+++ b/docs/tr2/README.md
@@ -85,6 +85,7 @@ game with new enhancements and features.
     - fixed an invisible wall above stacked pushblocks if near a ceiling portal
     - fixed floor height issues with pushblocks poised to fall in various scenarios
     - fixed being unable to stack multiple pushblocks over multiple rooms
+    - fixed falling pushblocks using the enemy grunt sound effect
 - improved the animation of Lara's braid
 
 #### Cheats
--- a/docs/tr2/README.md
+++ b/docs/tr2/README.md
@@ -81,8 +81,10 @@ game with new enhancements and features.
 - fixed smashed windows blocking enemy pathing after loading a save
 - fixed Lara getting stuck in a T-pose after jumping/falling and then dying before reaching fast fall speed
 - fixed several issues with pushblocks
+- fixed several issues with pushblocks:
     - fixed an invisible wall above stacked pushblocks if near a ceiling portal
     - fixed floor height issues with pushblocks poised to fall in various scenarios
+    - fixed being unable to stack multiple pushblocks over multiple rooms
 - improved the animation of Lara's braid
 
 #### Cheats
--- a/docs/tr2/README.md
+++ b/docs/tr2/README.md
@@ -80,6 +80,8 @@ game with new enhancements and features.
 - fixed the game crashing if a cinematic is triggered but the level contains no cinematic frames
 - fixed smashed windows blocking enemy pathing after loading a save
 - fixed Lara getting stuck in a T-pose after jumping/falling and then dying before reaching fast fall speed
+- fixed several issues with pushblocks
+    - fixed an invisible wall above stacked pushblocks if near a ceiling portal
 - improved the animation of Lara's braid
 
 #### Cheats
--- a/docs/tr2/README.md
+++ b/docs/tr2/README.md
@@ -82,6 +82,7 @@ game with new enhancements and features.
 - fixed Lara getting stuck in a T-pose after jumping/falling and then dying before reaching fast fall speed
 - fixed several issues with pushblocks
     - fixed an invisible wall above stacked pushblocks if near a ceiling portal
+    - fixed floor height issues with pushblocks poised to fall in various scenarios
 - improved the animation of Lara's braid
 
 #### Cheats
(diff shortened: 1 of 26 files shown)


rr- (approved):
Code LGTM on my mobile. How close are we to TR1? As far as I understand the code for pushblocks should now be mostly consistent apart from cosmetics.

lahm86:
> Code LGTM on my mobile. How close are we to TR1? As far as I understand the code for pushblocks should now be mostly consistent apart from cosmetics.

Yes, just cosmetics mainly that differ for the main control/collision logic. The killer block feature will be a direct copy I think, and then the only other difference will be the way we use `m_MovableBlockBounds` - we'll eventually need to migrate all TR2 objects to use `obj->bounds_func` rather than a raw array.

aredfan:
I'm afraid the test level crashes for me.

lahm86:
> I'm afraid the test level crashes for me.

Oh, you might need to disable the item rotation fix in the config.

aredfan:
> Oh, you might need to disable the item rotation fix in the config.

That was indeed the cause of the crash, apologies, my bad.

- [x] I came across this issue - play the test level, go into the next room, save and reload, and you'll get invisible collision below 3 of the pushable blocks.

lahm86:
Thanks, @aredfan. Looks like it's present in TR1 too - equivalent level attached. I will try to work it out.

[LEVEL1.zip](https://github.com/user-attachments/files/18932058/LEVEL1.zip)

aredfan:
Thank you @lahm86, the previous issue is fixed. There's a new regression in the first room if you stack the pushables blocks, then save and reload, the top pushable block loses its collision.

20250223_183728_The_Great_Wall

lahm86:
Thanks again @aredfan, I noticed just after pushing and stopped typing my message to you 😄 I should've said it was still WIP, apologies.

Pushblocks ☠️

lahm86:
I've reverted the previous commit for now. I'll keep working on this and let you know if I find a solution.

@walkawayy can you remember if we had this issue back in the original PR? I can't see a mention of it, but trying the TR1 level above on TR1X 3.0 has the issue. I'm wondering if we just missed this case.

lahm86:
The latest commit I think fixes the problems identified, plus another issue I found in the test level when there are flipmaps involved. The test levels have been updated - so there is one for each game, and these have different types of pushblocks (having a combination of object IDs is worth testing).

So the new process unfortunately goes against our discrete item initialisation, flip map and save handling, but I can't think of a better way to approach pushblocks because of their complexity. We now process these collectively on level load, and in order of Y pos in the level, so to avoid altering the floor too early or too late.

I'm not too keen on the logic to be honest. I would be happy to cancel this PR until we can perhaps come up with a neater solution (maybe events?); it seemed like it was going to be a straight-forward bug fix to port from TR1, unfortunately we discovered the new issues.

LMK what you think - either way, I'm happy. If we are testing this, we'll need to check OG TR1 and TR2 blocks carefully as well.

rr- (commented):
Thanks. The changes to the apis indeed set us back.  I'll have to think of it when I get back,  but atm I'm not too keen on merging it in the current form, either. That's not to say they're bad or that I have something better, I just need time to process.

lahm86:
@rr- I tried to get the handling shifted to the room API but it ended up being even more clunky, as the room API then needs to know about the block's status/height which IMO is priv to the block module. I think this might be as good as we can get it sadly :-/

I did find a bug if you trigger a block to fall, then push it under another one, then trigger the second one to fall. After reloading, the second one wouldn't have collision. So that's now resolved.

WDYT? Worth another round of testing, or should we go back to the drawing board?

rr-:
Let's proceed. We can figure something out later. I'm inclined towards replacing callbacks with event managers, but that deserves its own initiative.

lahm86:
@aredfan, @walkawayy If you have time to test this again, that would be great. Thanks!

walkawayy (approved):
LGTM. I tested a bunch in your test levels including saving and loading.

aredfan (approved):

docs/tr2/README.md
```diff
@@ -80,6 +80,12 @@ game with new enhancements and features.
 - fixed the game crashing if a cinematic is triggered but the level contains no cinematic frames
 - fixed smashed windows blocking enemy pathing after loading a save
 - fixed Lara getting stuck in a T-pose after jumping/falling and then dying before reaching fast fall speed
+- fixed several issues with pushblocks
```
walkawayy, line 83: Oops double changelog entry I missed.

lahm86: Oops, thanks. I'll PR a fix tomorrow.

PR #2554, opened 2025-02-26
inventory: limit max item quantities to prevent crashing

### LINKED ISSUE #2497: Console command to give large quantities of items can crash the game
`/give 10000000000 magnum` might run (after soft locking the game for many seconds) but it also has a high chance to crash the game. The game may also crash when viewing the large quantity of item in the inventory. We should probably cap the number?

walkawayy:
#### Checklist

#### Description
1. Includes a commit to fix spacing so my local linter stops driving me insane. (TBD the Github linter likes the no space apparently)
2. Changes inventory item quantities to be stored as an `int32_t` since it seems everywhere else that's the quantity type.
3. Adds a `Inv_InsertAmmo` helper function similar to `Inv_InsertItem`. This allows us to clamp to the max quantity of `MAX_QTY 999999` (open to suggestions on this). `MAX_QTY` prevents the `give` command from causing an overflow which can crash the game, overflow the item quantity, etc. This number seemed big enough to be avoided during normal gameplay, but also give a massive amount if desired. Also allows you to type something like `give 1000000000 shotgun shells` to max out the item for testing without worrying about crashing the game.

At first I tried experimenting with checking max `int32_t` values on scan and then checking to make sure the scanned value + the current quantity fit in an `int32_t` size, but I think just capping it like this makes the code much simpler. You will only ever hit this max quantity with cheats anyway, so it's just a theoretical cap to prevent crashes.

Requires testing all sorts of `give x` ammo commands, picking up normal pickups while at max value, NG+, etc. I don't see any issues but I did notice 2 quirks for feedback:
1. In TR1, maxing out an item quantity removes the ammo string even in a non-NG+ save. I assume this is some hard coded reason for NG+.
2. `999,999` shotgun ammo is maxed out as `166,666` because of the shotgun technically having 6x the pellets per ammo. I don't think it's an issue but I wanted to note it.

### DIFF
--- a/src/libtrx/game/inventory.c
+++ b/src/libtrx/game/inventory.c
@@ -2,6 +2,7 @@
 
 #include "game/inventory_ring/vars.h"
 #include "game/objects/vars.h"
+#include "utils.h"
 
 bool Inv_AddItemNTimes(const GAME_OBJECT_ID obj_id, const int32_t qty)
 {
@@ -10,6 +11,12 @@ bool Inv_AddItemNTimes(const GAME_OBJECT_ID obj_id, const int32_t qty)
         result |= Inv_AddItem(obj_id);
     }
     return result;
+}
+
+void Inv_AddAmmo(AMMO_INFO *const weapon_ammo, int32_t qty)
+{
+    weapon_ammo->ammo += qty;
+    CLAMPG(weapon_ammo->ammo, MAX_QTY);
 }
 
 GAME_OBJECT_ID Inv_GetItemOption(const GAME_OBJECT_ID obj_id)
--- a/docs/tr1/CHANGELOG.md
+++ b/docs/tr1/CHANGELOG.md
@@ -2,6 +2,7 @@
 - added support for custom levels to use `disable_floor` in the gameflow, similar to TR2's Floating Islands (#2541)
 - changed the Controls screen to hide the reset and unbind texts when changing a key (#2103)
 - fixed the bear AI fix option being applied in the Vilcabamba demo (#2559, regression from 4.8)
+- fixed extremely large item quantities crashing the game (#2497, regression from 0.3)
 
 ## [4.8.3](https://github.com/LostArtefacts/TRX/compare/tr1-4.8.2...tr1-4.8.3) - 2025-02-17
 - fixed some of Lara's speech in the gym not playing in response to player action (#2514, regression from 4.8)
--- a/docs/tr2/CHANGELOG.md
+++ b/docs/tr2/CHANGELOG.md
@@ -5,6 +5,7 @@
 - fixed a rare issue whereby Lara would be unable to move after disposing a flare (#2545, regression from 0.9)
 - fixed flare pickups only adding one flare to Lara's inventory rather than six (#2551, regression from 0.9)
 - fixed play any level causing the game to hang when no gym level is present (#2560, regression from 0.9)
+- fixed extremely large item quantities crashing the game (#2497, regression from 0.3)
 
 ## [0.9.2](https://github.com/LostArtefacts/TRX/compare/tr2-0.9.1...tr2-0.9.2) - 2025-02-19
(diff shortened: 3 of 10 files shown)


walkawayy:
This linter will be the death of me. Why does it prefer this 💀 
```
-        TRY_OR_FAIL(VFile_TrySkip(file, num * size));                          \
+        TRY_OR_FAIL(VFile_TrySkip(file, num *size));                           \
```

rr-:
Inv_InsertItem refers to inserting an item to the inventory ring. I think a more appropriate name for InsertAmmo would be AddAmmo. WDYT? 

Maybe we could support /give max shotgun ammo for some syntax sugar?

walkawayy:
> Inv_InsertItem refers to inserting an item to the inventory ring. I think a more appropriate name for InsertAmmo would be AddAmmo. WDYT?
> 
> Maybe we could support /give max shotgun ammo for some syntax sugar?

Good idea on the function name. Ya I will look into supporting the displayed max quantity of shotgun ammo instead of the underlying ammo. I will also try to look into TR1 hiding the ammo count outside of NG+.

walkawayy:
I believe all issues are resolved. I fixed the ammo string not displaying for large quantities in TR1. I made a `TODO` for the shotgun ammo max quantity. I think it will require a libtrx merge to make the `shotgun_ammo` common. I also reverted the lint.

lahm86 (approved):

walkawayy:
@aredfan I will wait for your approval before merging. Take as long as you need, no rush.

src/libtrx/game/inventory.c
```diff
@@ -12,6 +13,14 @@ bool Inv_AddItemNTimes(const GAME_OBJECT_ID obj_id, const int32_t qty)
     return result;
 }
 
+void Inv_AddAmmo(AMMO_INFO *const weapon_ammo, int32_t qty)
+{
+    weapon_ammo->ammo += qty;
+    // TODO Check for shotgun_ammo when merged to libtrx and clamp to MAX_QTY
+    // * 6.
```
rr-: I believe this can be already done using Lara_GetInfo().

aredfan:
> @aredfan I will wait for your approval before merging. Take as long as you need, no rush.

@walkawayy  no problem and thank you. The only issue I found happens if Lara doesn't have weapons in her inventory. If I use the give command for ammo such as `/give shells`, the limit appears to be 1,999,998 instead of 999,999. This affects all ammo in TR1/2 except for grenades.

walkawayy:
Oops I fixed a TR2 bug where I had copied Uzi clip constants into the Magnum functions.

@aredfan maybe @lahm86 can correct me, but I don't think it's a bug actually. The way the ammo works underneath the hood is a little odd. Uzis for example:
- Each Uzi clip you pickup in TR1 is technically a quantity of `1`.
- However, that 1 pickup is technically 100 bullets composed of 2 `UZI_AMMO_CLIP`s of 50 each. I think this is to simulate Lara's dual weapons where she would technically need a clip in each weapon in real life.
- For display purposes, the game outputs the ammo clips' ammo strings as the quantity * 2., so the 1 picked up clip shows as 2. (`InvRing_ShowItemQuantity`) That's why the displayed number in inventory is `999,999` * 2 aka `1,999,998`. Lara technically has `999,999` clips.
- This applies for all ammo, even the shotgun where `NUM_SG_SHELLS` is `2` even though that ammo is a little different where `SHOTGUN_AMMO_CLIP` is `6` because Lara fires 6 pellets. Similar to that shotgun shots fired statistics issue.

The main purpose of the PR is to set a realistic cap that will never be reached in normal gameplay and avoid crashes, so the exact number isn't that important imo.

aredfan (approved):

walkawayy:
@lahm86 is it ok if I merge?

lahm86:
It LGTM. Maybe hang on for rr's approval?

rr- (approved):

PR #2562, opened 2025-02-27
tr1/objects/bear: apply AI fix in bear control

### LINKED ISSUE #2559: TR1X bug: fix bear ai is affecting demos

_____________

Prior to 4.8, the `Fix bear AI` option would be forced off for the demo modes. This seems to no longer be the case because demo 1 is being affected by the option if enabled.

#1370 for reference.

lahm86:
Resolves #2559.

#### Checklist

#### Description

This moves the bear AI fix to the control routine as otherwise when the demo alters the setting, its effect has already taken place in the bear setup function. It's not ideal calling it on every frame a bear is active, but I suppose a benefit is that it does allow changing the setting in-game with the console without having to reload.

### DIFF
--- a/src/tr1/game/objects/creatures/bear.c
+++ b/src/tr1/game/objects/creatures/bear.c
@@ -58,11 +58,6 @@ static void M_Setup(OBJECT *const obj)
     obj->collision_func = Creature_Collision;
     obj->shadow_size = UNIT_SHADOW / 2;
     obj->hit_points = BEAR_HITPOINTS;
-    if (g_Config.gameplay.fix_bear_ai) {
-        obj->pivot_length = 0;
-    } else {
-        obj->pivot_length = 500;
-    }
     obj->radius = BEAR_RADIUS;
     obj->smartness = BEAR_SMARTNESS;
     obj->intelligent = 1;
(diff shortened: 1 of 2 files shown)


src/tr1/game/objects/creatures/bear.c
```diff
@@ -76,6 +71,7 @@ static void M_Setup(OBJECT *const obj)
 
 static void M_Control(const int16_t item_num)
 {
+    Object_Get(O_BEAR)->pivot_length = g_Config.gameplay.fix_bear_ai ? 0 : 500;
```
rr-: Let's avoid assignments to a pointer from a function call, eg let's introduce an `obj` variable on which we later execute state-altering operations.

src/tr1/game/objects/creatures/bear.c
```diff
@@ -76,6 +71,8 @@ static void M_Setup(OBJECT *const obj)
 
 static void M_Control(const int16_t item_num)
 {
+    OBJECT *const obj = Object_Get(O_BEAR);
```
rr-: One last nitpick - can we use `item->object_id`?

rr- (approved):

src/tr1/game/objects/creatures/bear.c
lahm86: Of course, my bad. I was a bit hasty with this 😄

aredfan (approved):
LGTM, and thank you. 👍

Issue #2461, opened 2025-02-08
Mesh problems with fly cheat and saves


rr-:
TR1:

1. `/play 0`
2. `/save 1`
3. `/load 1`
4. `/fly`. Observe how Lara gets holsters
5. `/save 1`
6. `/load 1`. The holsters now render the wrong mesh.

![Image](https://github.com/user-attachments/assets/ffc27a81-0729-4afa-95db-3ce5daa323d6)

@aredfan can I ask for an impact assessment? (eg if it's unreleased, if it happens with tr2 etc.) Thanks.

aredfan:
> [@aredfan](https://github.com/aredfan) can I ask for an impact assessment? (eg if it's unreleased, if it happens with tr2 etc.) Thanks.

@rr- yes of course.

The TR1X bug is a regression since [4.7](https://github.com/LostArtefacts/TRX/releases/tag/tr1-4.7). I've checked the gym/HSH levels in TR2 and the bug isn't present there.

lahm86:
Similar issues:
- `/play 2`
- use `I` for all weapons. Shotgun is on Lara's back
- `/fly`
- Shotgun disappears

And:
- `/play 2`
- use `I` for all weapons.
- Draw Magnums
- `/fly`
- Pistols appear in holsters briefly

And:
- `/play 2`
- use `I` for all weapons.
- Draw Magnums and then undraw them
- `/fly`
- Pistols appear in holsters

PR #2571, opened 2025-03-01
tr1/lara/cheat: reset meshes in fly cheat

### LINKED ISSUE #2565: TR1X bug: Lara's model doesn't reset if resurrected after being eaten by the trex

__________________

Image

**Edit:** I think this bug happens in all cases of Lara getting killed, not just death by trex.

For example, have Lara draw magnums, then reduce her hp to 0 and resurrect.

### LINKED ISSUE #2572: TR1X bug: Lara can end up with weapons in her hands

_____________

1. `/play 1`
2. use the all weapons cheat
3. draw either pistols/shotgun/uzis/magnums
4. use the fly cheat, then immediately press shift to disable fly
5. Lara will have weapons in her hands

I think this issue was tackled in TR2X some time ago which is why it's not affected at this time.

lahm86:
Resolves #2565.
Resolves #2572.

#### Checklist

#### Description

This restores Lara's meshes if she is resurrected with the fly cheat. The only scenario we can't efficiently cover is the following:
- draw magnums or uzis
- draw shotgun
- die and resurrect

In this case the holstered guns will be pistols because we don't store the last holstered gun type, and looking it up would involve a lot of mesh comparison which I don't feel is worth it, plus for things like the T-Rex and Midas deaths, the information is lost anyway. If the last gun is not the shotgun, the correct holster meshes will be restored normally.

Otherwise, other scenarios should hopefully work as expected 🤞 Worth checking #2461 too.

### DIFF
--- a/src/tr1/game/lara/cheat.c
+++ b/src/tr1/game/lara/cheat.c
@@ -139,6 +139,12 @@ bool Lara_Cheat_EnterFlyMode(void)
         return false;
     }
 
+    g_Lara.request_gun_type = LGT_UNARMED;
+    if (g_LaraItem->hit_points <= 0) {
+        g_Lara.gun_status = LGS_ARMLESS;
+        Lara_InitialiseMeshes(GF_GetCurrentLevel());
+    }
+
     if (g_Lara.water_status != LWS_UNDERWATER || g_LaraItem->hit_points <= 0) {
         g_LaraItem->pos.y -= STEP_L;
         g_LaraItem->current_anim_state = LS_SWIM;
@@ -193,6 +199,7 @@ bool Lara_Cheat_ExitFlyMode(void)
         g_Lara.torso_rot.y = 0;
     }
     g_Lara.gun_status = LGS_ARMLESS;
+    Lara_InitialiseMeshes(GF_GetCurrentLevel());
     Console_Log(GS(OSD_FLY_MODE_OFF));
     return true;
 }
--- a/src/tr1/game/lara/common.c
+++ b/src/tr1/game/lara/common.c
@@ -655,30 +655,40 @@ void Lara_RevertToPistolsIfNeeded(void)
 void Lara_InitialiseMeshes(const GF_LEVEL *const level)
 {
     const RESUME_INFO *const resume = Savegame_GetCurrentInfo(level);
-
-    if (resume != nullptr && resume->flags.costume) {
-        for (LARA_MESH mesh = LM_FIRST; mesh < LM_NUMBER_OF; mesh++) {
-            Lara_SwapSingleMesh(mesh, mesh == LM_HEAD ? O_LARA : O_LARA_EXTRA);
-        }
-        return;
-    }
+    const bool use_costume = resume != nullptr && resume->flags.costume
+        && Object_Get(O_LARA_EXTRA)->loaded;
 
     for (LARA_MESH mesh = LM_FIRST; mesh < LM_NUMBER_OF; mesh++) {
-        Lara_SwapSingleMesh(mesh, O_LARA);
+        Lara_SwapSingleMesh(
+            mesh, mesh == LM_HEAD || !use_costume ? O_LARA : O_LARA_EXTRA);
     }
 
-    LARA_GUN_TYPE holsters_gun_type =
-        resume != nullptr ? resume->holsters_gun_type : LGT_UNKNOWN;
-    LARA_GUN_TYPE back_gun_type =
-        resume != nullptr ? resume->back_gun_type : LGT_UNKNOWN;
+    LARA_GUN_TYPE back_gun_type = g_Lara.back_gun_type;
+    LARA_GUN_TYPE holsters_gun_type = g_Lara.holsters_gun_type;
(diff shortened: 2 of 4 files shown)


aredfan:
Everything looks good, thank you. 👍 

I did find a bug with Lara holding a shotgun in one hand, but I don't think it's related to this PR as it's already on develop, will raise an issue shortly.

aredfan (approved):

lahm86:
> Everything looks good, thank you. 👍
> 
> I did find a bug with Lara holding a shotgun in one hand, but I don't think it's related to this PR as it's already on develop, will raise an issue shortly.

Thanks, should be addressed. Easiest just to do it here as it's all so related.

aredfan (approved):
> Thanks, should be addressed. Easiest just to do it here as it's all so related.

Your fix did the trick. Thank you. 😄

src/tr1/game/lara/cheat.c
```diff
@@ -25,6 +26,39 @@
...
+
+    if (back_gun_type != LGT_UNARMED) {
+        Gun_SetLaraBackMesh(back_gun_type);
+    }
+
+    if (holsters_gun_type != LGT_UNARMED) {
+        Gun_SetLaraHolsterLMesh(holsters_gun_type);
+        Gun_SetLaraHolsterRMesh(holsters_gun_type);
+    }
+}
```
rr-: This logic is duplicated from somewhere isn't it? I think Lara_InitialiseMeshes. Any chance we can avoid duplication?

lahm86: Yeah, `Lara_InitialiseMeshes` uses `RESUME_INFO` though, so doesn't take into account guns she currently has, plus it results in the original bug with #2461. Maybe we could change the signature though.

rr- (approved):
LGTM but it'd be nice to merge the common parts, possibly through a new, third function.

lahm86:
> LGTM but it'd be nice to merge the common parts, possibly through a new, third function.

WDYT about getting rid of `resume->flags.costume` in favour of just testing if it's currently the gym level? I think this would make it easier to merge the functions as we don't really need `RESUME_INFO`; Lara's inventory is already initialised on the only other call to `Savegame_ApplyLogicToCurrentInfo`.

rr-:
> > LGTM but it'd be nice to merge the common parts, possibly through a new, third function.
> 
> WDYT about getting rid of `resume->flags.costume` in favour of just testing if it's currently the gym level? I think this would make it easier to merge the functions as we don't really need `RESUME_INFO`; Lara's inventory is already initialised on the only other call to `Savegame_ApplyLogicToCurrentInfo`.

This will make adding a potential option to change costumes at runtime more difficult; can we use a hybrid solution?

lahm86:
Sure, I'll take another look tomorrow 👍

lahm86:
Hybrid approach is now in place, so we just have a single function. This will need testing again I'm afraid, especially going between levels, and in Natla's Mines with gun loss. Also, going from the gym to a game level, or back to the gym via a save or otherwise, to ensure the costume swap behaves normally.

aredfan:
> Hybrid approach is now in place, so we just have a single function. This will need testing again I'm afraid, especially going between levels, and in Natla's Mines with gun loss. Also, going from the gym to a game level, or back to the gym via a save or otherwise, to ensure the costume swap behaves normally.

These scenarios and the linked issues LGTM.

The only issue I found is that Lara's holstered weapons revert to pistols between levels if she has magnums/uzis holstered and the shotgun equipped.

lahm86:
Thanks, @aredfan. Hopefully that's fixed now.

aredfan (approved):
> Thanks, @aredfan. Hopefully that's fixed now.

Can confirm, thank you. 👍

src/tr1/game/lara/common.c
```diff
@@ -655,31 +655,41 @@ void Lara_RevertToPistolsIfNeeded(void)
...
+        Lara_SwapSingleMesh(
+            mesh, mesh == LM_HEAD || !use_costume ? O_LARA : O_LARA_EXTRA);
     }
 
-    LARA_GUN_TYPE holsters_gun_type =
-        resume != nullptr ? resume->holsters_gun_type : LGT_UNKNOWN;
-    LARA_GUN_TYPE back_gun_type =
-        resume != nullptr ? resume->back_gun_type : LGT_UNKNOWN;
+    LARA_GUN_TYPE back_gun_type = g_Lara.back_gun_type;
+    LARA_GUN_TYPE holsters_gun_type = g_Lara.holsters_gun_type;
```
rr-, line 667: Is this swap ok with saved games? Eg having uzis holstered, saving, playing caves, and then loading the saved game will not revert the holstered guns back to pistols, right?

aredfan: Yep, I can confirm that whichever weapon Lara has holstered/equipped is being loaded correctly.

E.g. If Lara has magnums/uzis holstered and shotgun equipped, it does not revert to pistols when saving and reloading.

walkawayy (approved):

Issue #2515, opened 2025-02-16
Save states brick each other's loading


Richard-L:
https://github.com/user-attachments/assets/b6457525-65f9-4162-b246-915356d2b70c

See how:
1. Loading "Lost Valley 280" from the main menu works fine.
2. The same save state no longer loads if "Tomb of Qualopec 286" was loaded prior.

Files: [saves.zip](https://github.com/user-attachments/files/18815547/saves.zip)

rr-:
This is because the Lara from the Qualopec save has a Shotgun, and the Lost Valley save mistakenly replaces Shotgun with Shotgun Ammo, thinking Lara has it there as well.
Regression caused by the recent game flow consolidation.

rr-:
Actually this is because the Lara from the Qualopec has NG+ flag enabled.
@aredfan can I ask for a confirmation which is the first version in which this bug surfaces?

Steps to reproduce:

- Play NG+, `/save 2`
- `/play 3` without NG+, `/save 1`
- `/load 2`
- `/load 1`

The save should load normally, and instead we get a hard exit to title. It's also worth observing what happens to the Shotgun pickup.
Also, is this issue present in TR2?

Richard-L:
Thanks Dash.

Would love to have NG+ signalled somehow in the save/load dialogue. Don't know if a "(NG+)" tag in the name of the save or something would look okay (?).

aredfan:
> Actually this is because the Lara from the Qualopec has NG+ flag enabled. [@aredfan](https://github.com/aredfan) can I ask for a confirmation which is the first version in which this bug surfaces?
> 
> Steps to reproduce:
> 
> * Play NG+, `/save 2`
> * `/play 3` without NG+, `/save 1`
> * `/load 2`
> * `/load 1`
> 
> The save should load normally, and instead we get a hard exit to title. It's also worth observing what happens to the Shotgun pickup. Also, is this issue present in TR2?

@rr- Of course.  For TR1X, it's a regression since [2.8](https://github.com/LostArtefacts/TRX/releases/tag/tr1-2.8).

It doesn't look like the bug occurs for TR2X on the latest dev snapshot (0.9.1-15-g5f0e321).

PR #2573, opened 2025-03-02
tr2/game_flow: fix losing NG+ flag in saves

### LINKED ISSUE #2566: TR2X bug: ammo count is displayed in NG+ when loading a savegame
**Regression since:** [0.9.2](https://github.com/LostArtefacts/TRX/releases/tag/tr2-0.9.2)

______________

1. `/play 18`
2. use the level skip cheat and skip the credits
3. start new game
4. `/save 1`
5. `/load 1`

This bug also affect flares.

lahm86:
Resolves #2566.

#### Checklist

#### Description

This postpones the fix for #2515 in TR2 because of the way the original savegame is loaded; `S_LoadGame` is called first and sets the bonus flag, then puts the rest of the savegame data into a buffer. We were then resetting the flag in the GF, and later processing the save buffer. This works in TR1 because the later save processing captures the NG+ flag in BSON, and legacy reading reads again from the beginning. #2515 didn't apply to TR2 so I think this is currently the best approach until we refactor the savegame logic, as that's a much bigger endeavour.

Worth testing different kinds of saves, with and without the flag and after completing the game, or exiting to title early (and any other scenario you can think of 😄).

### DIFF
--- a/docs/tr2/CHANGELOG.md
+++ b/docs/tr2/CHANGELOG.md
@@ -9,6 +9,7 @@
 - fixed extremely large item quantities crashing the game (#2497, regression from 0.3)
 - fixed missing new game text in the passport when play any level is enabled (#2563, regression from 0.9)
 - fixed the play any level dialog not showing in the gym passport (#2564, regression from 0.9)
+- fixed losing the NG+ flag when loading a save that has it set (#2566, regression from 0.9.2)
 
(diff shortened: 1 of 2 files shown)


rr- (approved):

aredfan (approved):
LGTM, and thank you. 👍

PR #2585, opened 2025-03-06
tr1/output: add drawing mesh spheres to debug command

walkawayy:
#### Checklist

#### Description
Added drawing of object mesh spheres to the `/debug` console command. It's only for meshes (for now?).


walkawayy:
Applies wireframe mode during drawing which works well except for the fact that the color is ignored so the spheres are pretty opaque. I couldn't get using only `glPolygonMode(GL_LINE)` to work. This also ignores extremely large spheres like Lara's pistols, large medipack (radius of 2000+ ?!), and other large objects. Also let me know if it's ok to tie this into the existing `/debug` command or if you want it to be separate.

20250306_002642_City_of_Vilcabamba

aredfan:
Looks very cool.

One issue I found is the `/set sphere` command has no effect whereas `/set debug-triggers` affects both the spheres and triggers.

**Edit:** I noticed that the debug and wireframe modes cannot be enabled at the same time. In the latest dev snapshot, it's possible for both modes to be enabled at the same time.

Richard-L:
What is this used for?

src/tr1/game/output.c
```diff
@@ -93,6 +94,66 @@ static void M_CalcRoomVertices(const ROOM_MESH *mesh);
 static void M_CalcRoomVerticesWibble(const ROOM_MESH *mesh);
 static void M_CalcWibbleTable(void);
 
+static void M_DrawMeshSphere(const OBJECT_MESH *const mesh)
+{
+    if (mesh->radius >= WALL_L) {
```
lahm86: Maybe the condition would be better outside of this function if it's to be used potentially in the future with things other than object meshes. And in the same regard, maybe just passing the center and radius to it is enough.

src/tr1/game/output.c
```diff
@@ -93,6 +94,66 @@ static void M_CalcRoomVertices(const ROOM_MESH *mesh);
...
+    GFX_Context_SetWireframeMode(true);
+
+    const RGBA_8888 color = { .r = 255, .g = 255, .b = 255, .a = 128 };
+
+    S_Output_DisableTextureMode();
+    S_Output_DisableDepthWrites();
+    S_Output_SetBlendingMode(GFX_BLEND_MODE_NORMAL);
+
+    // More subdivisions means smoother spheres.
+    const int subdivisions = 12;
```
lahm86: A few nitpicks from here to the end of the function:
- `int` in place of `int32_t`
- var case e.g. `sinTheta` in place of `sin_theta`
- missing `const`s

src/tr1/game/output.c
```diff
@@ -600,6 +661,10 @@ void Output_DrawObjectMesh(const OBJECT_MESH *const mesh, const int32_t clip)
         M_DrawObjectFace4EnvMap(mesh->flat_face4s, mesh->num_flat_face4s);
         M_DrawObjectFace3EnvMap(mesh->flat_face3s, mesh->num_flat_face3s);
     }
+
+    if (g_Config.rendering.enable_debug_triggers) {
```
lahm86: Looks like this is the wrong config option.

src/tr1/game/output.c
walkawayy: Yes on the mesh.

About the center and radius pass, I agree that would be ideal. In my debug branch I had a version of the function like that. But I think first refactoring `SPHERE` to use `XYZ_32 pos` instead of standalone x/y/z would be better. If @rr- is on board with that I can do it.

rr-: Sounds good to me; wdyt Lahm?

src/tr1/game/output.c
```diff
@@ -93,6 +94,64 @@ static void M_CalcRoomVertices(const ROOM_MESH *mesh);
 static void M_CalcRoomVerticesWibble(const ROOM_MESH *mesh);
 static void M_CalcWibbleTable(void);
 
+static void M_DrawSphere(const XYZ_32 pos, const int32_t radius)
```
walkawayy, line 97: I wasn't sure what to do here about `XYZ_32` because `M_CalcVertex` takes `XYZ_16`. There's some type conversions throughout the codebase.

src/tr1/game/output.c
```diff
@@ -93,6 +94,64 @@ static void M_CalcRoomVertices(const ROOM_MESH *mesh);
...
+                color);
+            S_Output_DrawFlatTriangle(
+                &vertices[index_0], &vertices[index_2], &vertices[index_3],
+                color);
+        }
+    }
+
+    S_Output_SetBlendingMode(GFX_BLEND_MODE_OFF);
+    S_Output_EnableDepthWrites();
+    GFX_Context_SetWireframeMode(false);
```
rr-: This may cause issues when the user executes `/wireframe 1`, as it disables it for the rest of the pipeline. Even if this doesn't manifest right now, it's most likely due to the sequence of function calls. As such, I'd suggest changing this to restore the previous state.

src/tr1/game/output.c
```diff
@@ -93,6 +94,64 @@ static void M_CalcRoomVertices(const ROOM_MESH *mesh);
...
+            index++;
+        }
+    }
+
+    for (int i = 0; i < subdivisions; i++) {
+        for (int j = 0; j < subdivisions; j++) {
+            const int index_0 = i * (subdivisions + 1) + j;
+            const int index_1 = (i + 1) * (subdivisions + 1) + j;
+            const int index_2 = (i + 1) * (subdivisions + 1) + (j + 1);
+            const int index_3 = i * (subdivisions + 1) + (j + 1);
```
rr-: `int32_t`

src/tr1/game/output.c
```diff
@@ -600,6 +659,17 @@ void Output_DrawObjectMesh(const OBJECT_MESH *const mesh, const int32_t clip)
         M_DrawObjectFace4EnvMap(mesh->flat_face4s, mesh->num_flat_face4s);
         M_DrawObjectFace3EnvMap(mesh->flat_face3s, mesh->num_flat_face3s);
     }
+
+    // Ignore very large spheres (e.g., Lara's pistols).
+    if (g_Config.rendering.enable_debug_spheres && mesh->radius < WALL_L) {
+        M_DrawSphere(
+            (XYZ_32) {
```
rr-, line 667: This conversion is fine.

rr- (commented):
LGTM apart from a minor state issue, and two cosmetic things.

src/tr1/game/output.c
lahm86: Yeah, this looks like a good move 👍

src/tr1/game/output.c
walkawayy: Good call. Did you see my comment earlier about this? Is there a better way to handle this?
`Applies wireframe mode during drawing which works well except for the fact that the color is ignored so the spheres are pretty opaque. I couldn't get using only glPolygonMode(GL_LINE) to work.`

rr-: I haven't seen it. I can see about this tonight.

rr-: While I can't get the colors to show up as semitransparent no matter what I try in the `GL_LINE` mode, I noticed some issues with the depth buffer. A good area to test this is near the Scion pedestal in Qualopec. Removing the lines about toggling depth writes seems to help.

walkawayy:
I think it looks pretty good now. I also changed the code to draw spheres as black in wireframe mode because you couldn't see them when they were white.

aredfan (approved):

rr- (commented):
One last thing – I'd draw the large and absurd spheres regardless, as currently it seems to skip certain objects. Plus seeing how silly the bounding boxes are is kind of useful information on its own, IMO.

PR #2598, opened 2025-03-10
inject: restructure format and handling

### LINKED ISSUE #1967: Refactor injection support

lahm86:
Resolves #1967.

#### Checklist

#### Description

This will be an incremental PR (although admittedly the first increment is big). At present, this gets the new format in place and gets all existing injections in TR1 and TR2 working again with that format. This was difficult to break down into smaller pieces unfortunately, but future changes should be more manageable.

We now have:
- Much smaller files (zlib)
- "Chunk" format, so we can easily add/remove/change individual handlers in the future without affecting the entire file structure
- Applicability tests are now in place, so e.g. `caves_textures.bin` will automatically not apply to custom levels even if present in the gameflow
- Separated modules for clarity/abstraction

Best viewed by commit, and the bin file rebuild commit can be skipped over (although seeing the drop in size there is nice 😌).

Ref https://github.com/LostArtefacts/TRXInjectionTool/pull/27

**Testing**
Unfortunately we'll need to check a lot here: basically see the gameflow for the files we inject, and check in level that things are working. I don't think we need to explicitly check every room texture fix as there are far too many of those. If they're working in one place, they should be working elsewhere as the underlying logic hasn't changed, so a quick scan over should be enough. A good indicator too for testing is the log file e.g. play a few levels then search the log file for "inject" - any messages should be just info rather than any errors/warnings.

We can also test trying level-specific injections on invalid levels. These are the camera shifts, item rotations, floor data fixes and texture fixes. The log should produce a warning for these.


lahm86:
Next increments after initial testing:
- Handle 8-bit for TR2 SWR
- Test each type in TR2
- Finalise merging of readers in TR1/2 `level.c`

src/libtrx/game/inject/common.c
```diff
@@ -0,0 +1,377 @@
...
+static INJECTION_CHUNK M_ReadChunk(const INJECTION *injection);
+static void M_InitialiseBlock(VFILE *file);
+static bool M_IsRelevant(INJECTION_FILE_TYPE type);
+static bool M_IsApplicable(const INJECTION *injection);
+
+static void M_LoadFromFile(
+    INJECTION *const injection, const char *const filename)
+{
+    VFILE *const file = VFile_CreateFromPath(filename);
+    if (file == NULL) {
```
rr-: nullptr :)

rr- (approved):
Very elegant design. What are the next steps? I can think of palette matching and documenting the format for the TE people.

src/libtrx/game/inject/common.c
lahm86: Oops! Thanks.

lahm86:
> Very elegant design. What are the next steps? I can think of palette matching and documenting the format for the TE people.

Thanks. Yes, I'm going to see about getting 8-bit handled next so we can then test texture import in TR2. I was wondering about having it automatically generated, but I'm not sure yet, it might be slow.

aredfan:
Testing:

- [x] The option to add a purple tint to save crystals seem to have no effect.

lahm86:
Thanks @aredfan, crystals should be fixed.

aredfan (approved):

rr- (approved):

PR #2612, opened 2025-03-11
tr1/savegame: optimize scanning saved games

### LINKED ISSUE #1335: Passport gets stuck on animate-out after saving
[TR1X.4.0.3-28-g211d744-Windows.zip](https://github.com/LostArtefacts/TR1X/releases/download/latest/TR1X.4.0.3-28-g211d744-Windows.zip) on the left and last release on the right:

https://github.com/LostArtefacts/TR1X/assets/4934209/3c925ca6-fcc2-4744-a37b-ff446e905fa4

Notice how on the left the passport gets stuck for a few frames before plopping away at the end of the animation.

### LINKED ISSUE #2610: Passport open animation has delay
This comparison is a bit desperate but was a certain amount of labour, so forgive the imperfections. Also, I seem to be getting slightly different frame interpolation results between recordings (?). I definitely cut the source clips all at the same beginning frame.

https://github.com/user-attachments/assets/6e2deb0d-18be-41ef-a702-cbdbb99881e7

Back four are TR1X, front middle is PS1.

The main issue is: As of 1-2 months, the passport animates in, but then gets stuck for a few milliseconds before the page open animation initiates. This changes a bit per try. Typically the very first time, directly after game launch, the animation will be the slowest (top left), after which a more consistent delay (other three) manifests itself.

Still, it's too slow when compared against OG.

Source clip:

https://github.com/user-attachments/assets/fcca5a70-2252-4b25-980b-ea93541cbbb7

PS1 src:

https://github.com/user-attachments/assets/fb89e80e-db69-436c-976b-00287a992aae

rr-:
#### Checklist

#### Description

Resolves #2610. Resolves #1335.

With all the accumulated additions over time – case sensitivity resolving, for example, or guessing the game root directory – the previous approach grew to be quite horrible.
The downside is that the save file names are now case sensitive due to usage of `scanf()`, but I'm not sure how big of an issue this is.

### DIFF
--- a/docs/tr1/CHANGELOG.md
+++ b/docs/tr1/CHANGELOG.md
@@ -9,6 +9,7 @@
 - changed the `/pos` command to show `Demo` and `Cutscene` instead of `Level` when relevant
 - changed the `/pos` command to show demo and cutscene numbers starting at 1, in line with `/play`
 - changed the `/play` and `/pos` commands to always treat the gym level as the level 0 – even if it's not included
+- fixed delays when scanning available save games (#2610, #1335, regression from <3.0)
 - fixed several instances of the camera going out of bounds (#1034)
 - fixed issues with stacked, floating and flipmap pushblocks in custom levels
 - fixed issues with fixed cameras in 60 FPS shifting before settling on their target (#1186)
--- a/src/libtrx/benchmark.c
+++ b/src/libtrx/benchmark.c
@@ -1,7 +1,6 @@
 #include "benchmark.h"
 
 #include "log.h"
-#include "memory.h"
 
 #include <SDL2/SDL_timer.h>
 
@@ -36,12 +35,13 @@ static void M_Log(
     }
 }
 
-BENCHMARK *Benchmark_Start(void)
+BENCHMARK Benchmark_Start(void)
 {
-    BENCHMARK *const b = Memory_Alloc(sizeof(BENCHMARK));
-    b->start = SDL_GetPerformanceCounter();
-    b->last = b->start;
-    return b;
+    const Uint64 perf = SDL_GetPerformanceCounter();
+    return (BENCHMARK) {
+        .start = perf,
+        .last = perf,
+    };
 }
 
 void Benchmark_Tick_Impl(
@@ -58,5 +58,4 @@ void Benchmark_End_Impl(
     const char *const func, const char *const message)
 {
     Benchmark_Tick_Impl(b, file, line, func, message);
-    Memory_FreePointer(&b);
 }
--- a/src/libtrx/filesystem.c
+++ b/src/libtrx/filesystem.c
@@ -444,3 +444,28 @@ void File_CreateDirectory(const char *path)
 #endif
     Memory_FreePointer(&full_path);
 }
+
+void *File_OpenDirectory(const char *const path)
+{
+    ASSERT(path != nullptr);
+    char *full_path = File_GetFullPath(path);
+    DIR *path_dir = opendir(full_path);
+    Memory_FreePointer(&full_path);
+    return path_dir;
+}
+
+const char *File_ReadDirectory(void *const dir)
+{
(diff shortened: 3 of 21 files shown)


aredfan (approved):

lahm86 (approved):

rr-:
@Richard-L can you report how the dev snapshot feels? Thanks.

rr-:
@Richard-L @aredfan @lahm86  I made two additional large optimizations:

### Scanning only once

The game now scans the save directory only once at launch, updating the slot information only when the player saves the game. Unfortunately, this means that pasting save game files into the save directory while the game is running will no longer work and requires a game restart. Alternatively, the player can still trigger a rescan at runtime by fiddling with the slot count using the `/set` command.

### Improve slot information performance

The existing method of filling the slot information is quite sluggish, averaging around 0.15-0.18 ms **per slot**, and this can quickly add up with a large number of slots. To address this, I've included a backward-compatible footer at the end of the file containing the information essential to display the slot in easily accessible fields that can be retrieved without parsing the main BSON document. This change reduced the performance hit to <0.01 ms per slot. Filling information about 300 save game slots takes around 1.2 ms on my machine now. To observe the performance gains, the player needs to re-save their existing saves, though.

### Miscellaneous

I've also reduced two `fopen()` calls to one when trying to open a file and the file exists for further I/O ops reduction.

### Considerations

I think the second improvement is enough to let us restore rescanning on passport open.

Pros of continuing to rescan:
- we can remove and paste new savegame files while the game is minimized
- slightly less fragile code

Cons:
- it will be slow on network shaders and slow HDDs, still
- players with many save files will need to resave each file using the new game version to observe _any_ performance gains.

Another approach is to always rescan when the title screen loads, for example. LMK your thoughts. I've left a benchmarking code in the ScanSavedGames function so that @Richard-L can report his times too.

## Testing

Testing is quite complex as it impacts several scenarios:

- Saving when no prior saves exist
- Overwriting an existing save
- Saving in a vacant slot
- Saving using the `/save` command
- Conflicts with ATI saves (BSON should be preferred, both after saving and after fresh start)

Extra attention needs to be paid to the following:
- The save counters
- Correct reflection of changes in the load/save dialog after saving
- Highlighting the correct slot after saving, selecting a different slot and loading, etc.

lahm86:
I think I'd be in favour of doing the rescan in the title screen only. My feeling is the priority of the requirement to be able to move saves around while the game is open is less than that of being able to offer support to slower HDDs/potential network locations.

rr-:
Pushed a change so that the title inventory ring rescans the saved games. I'm not sure how it plays with highlighting the most recent slot, as I haven't tested it too heavily around this.

lahm86 (approved):

aredfan:
Issues (testing in progress):

**TR1X**
- [x] If no saves are present, start new game, then open the save dialog via F5 or the inventory - the saves dialog is blank the first time it's opened. The second time the saves dialog is opened, the bottom save slot is highlighted.
- [x] If no saves are present, start new game, then make a save via F5 or the inventory, then open the passport via the inventory and the load game page gets skipped over.

rr-:
> Issues (testing in progress):
> 
> **TR1X**
> 
> * [ ]  If no saves are present, start new game, then open the save dialog via F5 or the inventory - the saves dialog is blank the first time it's opened. The second time the saves dialog is opened, the bottom save slot is highlighted.
> * [ ]  If no saves are present, start new game, then make a save via F5 or the inventory, then open the passport via the inventory and the load game page gets skipped over.

Addressed both issues.

aredfan (approved):

Richard-L:
@rr- thanks so much for putting so much work into this 🙏 !!

The passport feels much smoother now, I can't seem to produce any mini freezes with it again. Great work :)

I tried saving and loading, etc. Didn't run into any issues.

One thing I noticed: when using level select, Lara would oftentimes start the level without any weapons. Should this be related here at all or a separate concern?

PR #2619, opened 2025-03-12
tr2: refactor sound effect loading

### LINKED ISSUE #2615: Feature: enable per-level SFX files
Currently the usage of `data/main.sfx` is globally fixed. Add the ability to define the global path in the gameflow, and allow individual levels to use different files (defaulting to the global if undefined).

lahm86:
Resolves #2615.

#### Checklist

#### Description

This adjusts the reading of `main.sfx` to allow support for offsets that aren't necessarily in order, such as when we later come to inject sound fixes. We do this by scanning the file once and noting positions and lengths of all SFX, then scan over the defined offset to select the correct data.

The game flow has been updated too to support the default SFX file being defined there, as well as allowing a different one to be used in each level. This can be tested using the Golden Mask or TR3 equivalent files (although the sounds won't make sense 😄).

### DIFF
--- a/src/tr2/game/level.c
+++ b/src/tr2/game/level.c
@@ -34,6 +34,8 @@
 #include <libtrx/utils.h>
 #include <libtrx/virtual_file.h>
 
+#define DEFAULT_SFX_PATH "data/main.sfx"
+
 typedef struct {
     int32_t game_index;
     int32_t file_index;
@@ -41,8 +43,8 @@ typedef struct {
 
 static int32_t M_CompareSampleOffsets(const void *a, const void *b);
 static void M_LoadFromFile(const GF_LEVEL *level);
-static void M_InitialiseSoundEffects(void);
-static void M_CompleteSetup(void);
+static void M_InitialiseSoundEffects(const char *const file_name);
+static void M_CompleteSetup(const GF_LEVEL *level);
 
 static int32_t M_CompareSampleOffsets(const void *const a, const void *const b)
 {
@@ -51,12 +53,12 @@ static int32_t M_CompareSampleOffsets(const void *const a, const void *const b)
     return entry_a->file_index - entry_b->file_index;
 }
 
-static void M_InitialiseSoundEffects(void)
+static void M_InitialiseSoundEffects(const char *const file_name)
 {
     BENCHMARK benchmark = Benchmark_Start();
     SAMPLE_ENTRY *entries = nullptr;
-    const char *const file_name = "data\\main.sfx";
-    const char *full_path = File_GetFullPath(file_name);
+    const char *full_path =
+        File_GetFullPath(file_name == nullptr ? DEFAULT_SFX_PATH : file_name);
     LOG_DEBUG("Loading samples from %s", full_path);
     MYFILE *const fp = File_Open(full_path, FILE_OPEN_READ);
     Memory_FreePointer(&full_path);
@@ -172,7 +174,7 @@ static void M_LoadFromFile(const GF_LEVEL *const level)
     Benchmark_End(&benchmark, nullptr);
 }
 
-static void M_CompleteSetup(void)
+static void M_CompleteSetup(const GF_LEVEL *const level)
 {
     BENCHMARK benchmark = Benchmark_Start();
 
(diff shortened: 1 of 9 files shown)


src/tr2/game/level.c
```diff
@@ -31,66 +31,85 @@
...
 
-static void M_InitialiseSoundEffects(void)
+static void M_InitialiseSoundEffects(const char *const file_name)
 {
     BENCHMARK benchmark = Benchmark_Start();
-    const char *const file_name = "data\\main.sfx";
     const char *full_path = File_GetFullPath(file_name);
     LOG_DEBUG("Loading samples from %s", full_path);
-    MYFILE *const fp = File_Open(full_path, FILE_OPEN_READ);
+    VFILE *const fp = VFile_CreateFromPath(full_path);
```
rr-: While I know that VFile has attractive APIs, this is sketchy, as it'll load the entire file into memory at once as VFile doesn't know how to stream internally. It seems to cause a 50% performance downgrade for this function, going from 2 ms to 3+ ms on each level load.

src/tr2/game/level.c
```diff
@@ -31,66 +31,85 @@
...
+            LOG_WARNING("Invalid sample offset %d", offset);
             continue;
         }
 
-        const size_t sample_data_size = 0x2C + aligned_size;
-        char *sample_data = Memory_Alloc(sample_data_size);
-        memcpy(sample_data, header, 0x2C);
-        File_ReadData(fp, sample_data + 0x2C, aligned_size);
+        const WAV_ENTRY entry = *(const WAV_ENTRY *)Vector_Get(entries, offset);
+        char *sample_data = Memory_Alloc(entry.size);
```
rr-: We could prevent double allocation for each sample by exposing the VFile's internal buffer. While this might improve performance, it does create some issues, as it's not the prettiest design, and we'd still end up reading all samples, not just those necessary for the level, so the benefits are uncertain.

rr- (commented):
I am experiencing a crash when exiting the game related to freeing the `sfx_path`:
```
#6  0x00007ffff7cc69ab in free () from /usr/lib/libc.so.6
#7  0x0000555555738944 in Memory_Free (memory=0x55555888fd60) at ../../../src/tr2/subprojects/libtrx/memory.c:43
#8  0x00005555557389cd in Memory_FreePointer (arg=0x55555653f8d0 <g_GameFlow+144>) at ../../../src/tr2/subprojects/libtrx/memory.c:53
#9  0x00005555556fbbea in GF_Shutdown () at ../../../src/tr2/subprojects/libtrx/game/game_flow/common.c:97
#10 0x00005555556db0a9 in Shell_Shutdown () at game/shell/common.c:480
#11 0x000055555571d855 in Shell_Terminate (exit_code=0) at ../../../src/tr2/subprojects/libtrx/game/shell/common.c:100
#12 0x000055555571dc60 in main (argc=1, argv=0x7fffffffd588) at ../../../src/tr2/subprojects/libtrx/game/shell/main.c:31
```
Most likely this is caused by the default settings holding a constant buffer, which gets repurposed as dynamically allocated strings rather on. To combat this, I'd leave it as `nullptr` and create a `#define` with the default path at the usage site in the event the string wasn't provided by the gameflow editor.

Personally, I think it's worth sticking to plain `File` over `VFile`.

src/tr2/game/level.c
rr-: Actually it's a triple allocation, as `Audio_Sample_LoadSingle` copies the data too :D

lahm86:
I've reverted back to using `FILE` and have defaulted the SFX path to `nullptr`.

I think there will still be a minor performance hit because we are caching the offsets, so inevitably that additional loop involves some extra time. On my end, it costs an extra ~0.8ms compared with develop. I can't really see a way to optimise this further, but would appreciate any further thoughts. We're being careful in the first loop to only cache the offsets so the footprint there is minimal, it's just unfortunate that we need to do so for every sample in the game.

aredfan (approved):

lahm86:
Thank you @rr- for providing a much neater solution to the SFX loading approach 😌

rr- (approved):

PR #2632, opened 2025-03-14
tr2/data: fix seaweed collision in Living Quarters

### LINKED ISSUE #2197: OG bug: seaweed collision prevents Lara from climbing out of water on one side | Living Quarters
20250102_081509_Living_Quarters
Screenshot

lahm86:
Resolves #2197.

#### Checklist

#### Description

Provides a new injection editor to edit static mesh properties, so allowing the fix for Living Quarters. I've added a comment in the issue about the similar TR1 problem, but here the issue I think we should fix in code because changing the data breaks the collision on the floor. If Lara's (effective) position is above the mesh, the collision test shouldn't apply.

Also fixes a bug in injections if we add a new data type but try to use that injection on an older version.

Ref https://github.com/LostArtefacts/TRXInjectionTool/pull/30.


rr- (approved):

walkawayy (approved):
You're too good

aredfan (approved):
Looks great, thank you. 👍

PR #2650, opened 2025-03-19
tr1/savegame: fix removing requester heading early

### LINKED ISSUE #2649: TR1X bug: missing headers in the save/load dialog
**Dev snapshot:** https://github.com/LostArtefacts/TRX/commit/2baeeabd23b52e8f7cbb2b2e2072001ab3f20f90

___________________

> The `Save Game`/`Load Game` headers in the save/load dialogs are missing.

The issue I reported in 2626 was already in the latest dev snapshot, thus it didn't regress in the PR.

lahm86:
Resolves #2649.

#### Checklist

#### Description

This moves the full savegame requester shutdown to the passport control when other text is being closed rather than when present saves are filled.

### DIFF
--- a/src/tr1/game/option/option_passport.c
+++ b/src/tr1/game/option/option_passport.c
@@ -176,7 +176,7 @@ static void M_RemoveAllText(void)
     }
     Requester_Shutdown(&m_SelectLevelRequester);
     Requester_Shutdown(&m_NewGameRequester);
-    Requester_ClearTextstrings(&g_SavegameRequester);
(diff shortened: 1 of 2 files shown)


rr- (approved):

rr-:
The idea behind the original change was so that `Requester_Init()` twice doesn't leave the previous `items` pointer leaking. That said, this module is something we're trying to move away from, and a potential leak should be microscopic, so it's not worth spending too much time here.

aredfan (approved):

PR #2661, opened 2025-03-21
tr1/savegame_bson: handle Lara animation shift

### LINKED ISSUE #2654: TR1X - Loading savegames from an older version causes crash
TR1X - Loading savegames from an older version causes crash

Noticed this with a build from current source 19-03-2025.
Also with official windows releases v4.8.3, 4.8.2, 4.8.1.

It writes two lines in the log about an issue with the header and about considering it a legacy save. Will add the exact message when I am back home. 

It does work properly when saving and loading games within the same version 4.8.x.

Did not have this problem with TR1X v4.2 and earlier.

lahm86:
Resolves #2654.

#### Checklist

#### Description

Prior to 4.8, Lara's legacy animation set could still be referenced from old savegames, but the case now is that only applicable animations are setup during level load, so after injections have reset the Lara object with the new values. This update shifts Lara's animation index to match her object if detected on load; a similar approach is already in place for TombATI saves.

### DIFF
--- a/docs/tr1/CHANGELOG.md
+++ b/docs/tr1/CHANGELOG.md
@@ -26,6 +26,7 @@
 - fixed the `/pos` command not showing demo and cutscene titles
 - fixed the embedded bats fix causing problems inside rooms with trapdoors (regression from 4.6)
 - fixed cutscene music looping (#2591, regression from 4.8)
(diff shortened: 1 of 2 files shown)


rr- (approved):
Wish we had unit tests for code like this, that is vulnerable to accidental refactors.

aredfan (approved):

Issue #1621, opened 2023-11-07
Feature request: Add Golden Mask support


aredfan:
A list of files contained within the `data` folder from the free downloadable Golden Mask pack.

| .  | Filename    | GM unique files | Conflict with TR2 filename | Notes                                                               |
|----|--------------|----------------|---------------------------|---------------------------------------------------------------------|
| 1  | credit00.pcx | ✅              |                           |                                                                     |
| 2  | credit01.PCX |                |                           |                                                                     |
| 3  | credit02.PCX |                |                           |                                                                     |
| 4  | credit03.PCX |                |                           |                                                                     |
| 5  | credit04.PCX |                |                           |                                                                     |
| 6  | credit05.PCX |                |                           |                                                                     |
| 7  | credit06.PCX |                |                           |                                                                     |
| 8  | credit07.PCX | ✅              | ✅                         |                                                                     |
| 9  | credit08.PCX |                |                           |                                                                     |
| 10 | credit09.pcx |                |                           |                                                                     |
| 11 | legalg.PCX   |                |                           | Duplicate of TR2's legal screen despite having a different filename. |
| 12 | level1.tr2   | ✅              |                           |                                                                     |
| 13 | level2.tr2   | ✅              |                           |                                                                     |
| 14 | level3.tr2   | ✅              |                           |                                                                     |
| 15 | level4.tr2   | ✅              |                           |                                                                     |
| 16 | level5.tr2   | ✅              |                           |                                                                     |
| 17 | MAIN.SFX     | ✅              | ✅                         |                                                                     |
| 18 | title.pcx    | ✅              | ✅                         | Two versions exist; Euro and US #1609                               |
| 19 | title.tr2    |  ✅              | ✅                          | Different sample data as a result of differing MAIN.SFX         |
| 20 | tombPC.dat   | ✅              | ✅                         | No longer required                                                             |

__________________

Here's a zip file containing all the Golden Mask data files in its entirety.

[data.zip](https://github.com/user-attachments/files/18585530/data.zip)

For the sake of transparency, the GM files originated from the [Core Design Fan Site](https://core-design.com/goodies_tr2_downloads.html), specifically the English version.

VGMFR:
This would be really great indeed, so I just wanted to +1 this feature request with this comment :)

lahm86:
Roadmap:

- [x] Allow specifying title screen background in gameflow
[...]

PR #2663, opened 2025-03-21
tr2/stats: determine secret count automatically

### LINKED ISSUE #1582: OG bug: end level stats hardcode the last two levels to not have secrets
This causes problem for TR-Rando. The code should instead read the gameflow.

lahm86:
Resolves #1582.
Part of #1621.

#### Checklist

#### Description

This removes the secret count from the game flow, which was only used to determine when a level had _no_ secrets, but in addition, the game hardcoded excluding the final two levels in the total secret count stats, as well as the default count being three. The previous global variable was only effectively used to hide the relevant row in the end level stats. The secret count is now calculated automatically, so a level can contain up to three secrets in total. This does introduce a similar problem as we have in TR1, whereby if you load a (legacy) save, the final stats will only show the maximum secrets collectible after that save. We may have a workaround later for this before the next release, which will be worked on in the BSON save migration, but for the time being this allows GM work to proceed.


rr- (approved):
Code LGTM. Haven't played with the game though.

aredfan (approved):

walkawayy (approved):

PR #2670, opened 2025-03-22
tr2/stats: allow two of each secret type

### LINKED ISSUE #2674: Allow two of each secret type per level
The savegame is currently limited to 8 bits for storing secret flags, so we can extend secret support to 6 per level for the time being. Initially, allow counting up to two of each type.

lahm86:
#### Checklist

#### Description

This fixes not triggering secret rewards if a level contains fewer than 3 secrets. Test level available for this with 2 secrets, and OG levels should work as normal. It also adds support for allowing up to two of each secret per level.

### DIFF
--- a/src/tr2/game/objects/general/pickup.c
+++ b/src/tr2/game/objects/general/pickup.c
@@ -15,6 +15,7 @@
 #include "game/output.h"
 #include "game/overlay.h"
 #include "game/room.h"
+#include "game/stats.h"
 #include "global/vars.h"
 
 #include <libtrx/config.h>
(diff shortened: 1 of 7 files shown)


src/tr2/game/objects/general/pickup.c
```diff
@@ -85,7 +85,7 @@ static void M_DoPickup(const int16_t item_num)
         && (g_SaveGame.current_stats.secret_flags & 1)
                 + ((g_SaveGame.current_stats.secret_flags >> 1) & 1)
                 + ((g_SaveGame.current_stats.secret_flags >> 2) & 1)
-            >= 3) {
+            >= g_SaveGame.current_stats.max_secret_count) {
```
rr-: Shouldn't this count the bits in the `secret_flags` to also work with 4+ secrets?

lahm86: Yes, we should probably make it more generic, although it would still be limited by the fact currently there are only 3 secret object types, and the flags are set by their type on pickup. Maybe we should rework it to allow duplicate secret types? E.g. 2 stone dragons, 1 jade, 1 gold.

walkawayy:
@lahm86 I don't see a test level attached? Did you forget to upload?

lahm86:
This has been updated to include the support for up to two of each secret type. The stats dialog display is not ideal when you go over 3 and there is a jump in type, but that's on the builder to allocate appropriate types for the additional ones, IMO. Test levels are updated with more scenarios. Secret rewards should trigger as expected, and OG dialog display should be unaffected.

rr- (approved):

aredfan (approved):

PR #2683, opened 2025-03-25
controls: fix console opening when remapping its key

### LINKED ISSUE #2641: Don't trigger console when remapping its keybind
Very tiny thing, sorry 😇

walkawayy:
Resolves #2641.

#### Checklist

#### Description
Fixed the console opening when remapping its key.

### DIFF
--- a/docs/tr1/CHANGELOG.md
+++ b/docs/tr1/CHANGELOG.md
@@ -32,6 +32,7 @@
 - fixed the embedded bats fix causing problems inside rooms with trapdoors (regression from 4.6)
 - fixed cutscene music looping (#2591, regression from 4.8)
 - fixed saves created before version 2.15 causing a crash on load (#2654, regression from 4.8)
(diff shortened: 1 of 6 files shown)


rr- (commented):
Wouldn't it make more sense to check if the input is in listening mode near the code that does the console open, rather than opening and closing it immediately?

walkawayy:
> Wouldn't it make more sense to check if the input is in listening mode near the code that does the console open, rather than opening and closing it immediately?

Probably. The TR1 controls menu is handled so differently but I can prob add a getter function for the key state like TR2 has.

walkawayy:
> Wouldn't it make more sense to check if the input is in listening mode near the code that does the console open, rather than opening and closing it immediately?

Ok should be good now I think.

rr- (approved):

lahm86 (approved):

aredfan (approved):

PR #2688, opened 2025-03-31
tr1/output/sprites: simplify projection matrix

rr-:
#### Checklist

#### Description

Removes a lot of poor code to achieve the original depth buffer resolution, in favor of a modern approach with a perspective-capable projection matrix. The code was mostly there, but the depth buffer components were poorly defined and thus not usable.  This PR fixes it. (Matching the depth buffer calculations with the original is important, because we still use both the original and modernized pipelines).

#### Testing

Since sprites are the only elements using the new z-buffer calculations, we need to examine how they behave in TR1 concerning camera distance, scene depth, and occlusion by other objects or surfaces.

### DIFF
--- a/data/tr1/ship/shaders/common.glsl
+++ b/data/tr1/ship/shaders/common.glsl
@@ -26,11 +26,11 @@ bool discardTranslucent(sampler2D tex, vec2 uv)
     return texel.a == 0.0;
 }
 
-bool discardTranslucent(sampler2DArray tex, vec3 uv)
+bool discardTranslucent(sampler2DArray tex, vec3 uvw)
 {
     // do not use smoothing for chroma key
     ivec2 size = textureSize(tex, 0).xy;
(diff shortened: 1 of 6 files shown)


aredfan:
I found these 2 issues:

- [x] Both the room sprites and sprite pickups flicker when bilinear filter is enabled.
- [ ] ~~In the busybody custom level, the sprite pickups appear black. This issue doesn't happen in the og levels.~~ Apologies my bad, this is already present in develop.

rr-:
> I found these 2 issues:
> 
> * [ ]  Both the room sprites and sprite pickups flicker when bilinear filter is enabled.
> * [ ]  ~In the busybody custom level, the sprite pickups appear black. This issue doesn't happen in the og levels.~ Apologies my bad, this is already present in develop.

Can I ask for a video? I can't reproduce this issue.
Edit: I realized this is specific to the test level itself. However, Lahm has confirmed that this is due to how it's built, rather than an issue with the code.

lahm86 (approved):

aredfan (approved):

PR #2693, opened 2025-04-03
tools: merge config tool projects

lahm86:
#### Checklist

#### Description

Some housekeeping to merge the config tool projects into a single solution, similar to the installer. I've also cleaned up namespaces and removed some redundancies. There should be no logical difference in the built executables.

The docker entrypoints for this and the installer have also been updated to remove the restore command, as this is implicit with the [publish command](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-publish#implicit-restore), so should be marginally faster to run as it's not having to do it twice.


rr-:
Can I ask to rebase to the latest develop which I forced pushed to, to remove the harmful room rendering changes?

rr- (approved):

aredfan (approved):

PR #2695, opened 2025-04-03
tr2/savegame: migrate to BSON save format

### LINKED ISSUE #2662: Feature: implement BSON saves
Add BSON saves similar to TR1. In addition, investigate the ability to include the total collectible secret count when legacy saves are used. This isn't applicable to TR1 because TombATI saves don't store the collected secret count per level, only the level that the save applies to, and therefore adjusting the end total isn't fundamental. All the data is however available in TR2 saves.

lahm86:
Resolves #2662.

#### Checklist

#### Description

This is a step towards BSON saves, initially tackling making the original logic more like TR1X. We still have a hybrid of new modules and the decomp/savegame module; I'll build on this PR incrementally with the goal being BSON plus legacy properly segmented.

I'd appreciate some initial testing of what's done so far, that is results in game should be the same for save/load, stats, and passport/slot display/counters.


rr- (approved):
From the quick tests everything seems to be in order. I haven't tested end game stats. The code also LGTM.

aredfan:
> I'd appreciate some initial testing of what's done so far, that is results in game should be the same for save/load, stats, and passport/slot display/counters.

LGTM 👍

lahm86:
The latest commits involve a lot of moving code around, but we're very close to being able to start the BSON module 😀 I'd appreciate another general save/load/stats test for TR2. It'd also be good to verify NG+ works, and in addition, for TR1 as well because I've moved this from being a global to a module var, common to both games. Thanks!

rr- (approved):

aredfan:
TR2X looks good including the NG+ flag. The NG+ saves for TR1X works as expected too. 👍

lahm86:
BSON saves are in, and I've rebased against develop to get everything up to date. I unfortunately hit a problem with skidoo drivers  - because of the way their object IDs are swapped, I've had to revert to calling `obj->handle_save_func` within the strategy rather than in the common module, otherwise the previous object ID is lost.

I'll continue with testing myself but wanted to mark this as ready to get more eyes looking for any issues 👀. Things to look out for:

- Carried items
- Lift
- Boat
- Skidoos (drivers and black skidoos should work too currently)
- Thrown flares
- One shot music
- NG+
- Angry monks
- Stats
- Inventory
- Flipmaps
- Golden Mask
- And anything else you can think of! Animation state, gun status etc

There is also a temporary uncompressed save file stored for debugging any issues we hit - it's `debug.json` in the saves folder (only the latest save). I will squash out the code for this before we finalise everything. Thanks!

aredfan:
Issues I noticed (this list may grow so keep an eye on it):

- [x] Lara doesn't stay on fire - `/play 1` `/tp 65`

I'll hijack this comment to add things I find too (Lahm):
- [x] - `/play 12`, `/give all`, save/load (item swapping foulplay)
- [x] Equip Shotgun, M16, Harpoon or GL; start firing; save; load a _different_ level, then reload the first save (likely crash)

lahm86:
@aredfan Thanks, updates pushed. Note that I've changed the format slightly so you'll need to delete your current (BSON) saves. I think we can afford to do this until we get it all working, rather than putting in extra code.

data/tr2/ship/cfg/TR2X_gameflow_level.json5
```diff
@@ -3,6 +3,7 @@
 
     "main_menu_picture": "data/images/title_eu.png",
     "savegame_fmt_legacy": "savegame.%d",
+    "savegame_fmt_bson": "save_tr2_%02d.dat",
```
rr-, line 7: These collide with the OG saves, can we make the pattern separate?

docs/tr2/README.md
```diff
@@ -310,6 +310,7 @@ as Notepad.
 - added Linux builds
 - added macOS builds
 - added .jpeg/.png screenshots
+- added BSON savegame support
```
rr-: I think the meaning can be vague to the average recipient, instead I'd focus on the benefits for the TRLE and Rando communities both here and in the changelog.

src/tr2/game/savegame/common.c
```diff
@@ -42,25 +194,441 @@ int32_t Savegame_GetSlotCount(void)
...
+        if (!strategy.allow_save || strategy.save_to_file_func == nullptr) {
+            continue;
+        }
+
+        char *file_name =
+            String_Format(strategy.get_save_file_pattern_func(), slot_idx);
+        char *full_path = String_Format("%s/%s", SAVES_DIR, file_name);
+        MYFILE *const fp = File_Open(full_path, FILE_OPEN_WRITE);
+        if (fp != nullptr) {
+            m_SaveCounter++;
```
rr-: Let's move this to the if outside the for loop as theoretically it can increment this counter once per each successful saving strategy.

src/tr2/game/game_flow/sequencer_events.c
```diff
@@ -114,23 +115,29 @@ static DECLARE_GF_EVENT_HANDLER(M_HandlePlayLevel)
 
     switch (seq_ctx) {
     case GFSC_SAVED:
-        ExtractSaveGameInfo();
+        const int16_t slot_num = Savegame_GetBoundSlot();
+        if (!Savegame_Load(slot_num)) {
+            LOG_ERROR("Failed to load save file!");
+            Game_SetCurrentLevel(nullptr);
+            GF_SetCurrentLevel(nullptr);
```
rr-: Are these two setters necessary?

rr- (commented):
Thank you. Looks terrific!

src/tr2/game/savegame/common.c
lahm86: Done, I've had to add a decrement check at the end too if the result isn't true - the idea is the strategies access it with `Savegame_GetCounter` so we need to have incremented it beforehand.

src/tr2/game/game_flow/sequencer_events.c
lahm86: Good point, it seems to be fine without it. I can follow up with TR1 (it's the same there) after this PR while doing some refactoring on the SG logic there to do the same.

rr- (approved):
Thank you. Code LGTM. I haven't tested the game.

lahm86:
^ Removed the temporary `debug.json` output in last commit.

aredfan (approved):
LGTM, and well done. 👍

Issue #2258, opened 2025-01-11
Texture resizes w bilinear filtering


Richard-L:
I had brought this up before, so I hope it's okay to post something I observed:

See how Duckstation/TR2X change from nearest neighbour to bilinear filtering (no edge blending).

DS

https://github.com/user-attachments/assets/3523e1f9-b23d-4f43-9625-892a311046ec

TR2X

https://github.com/user-attachments/assets/3644488a-21b1-48c8-89de-943543343d9f

Sadly TR2X is resizing the textures. TR textures are bad enough, and the super low-poly environments they're on don't help them. Their seamlessness is very hit or miss, to put it lightly. So to further damage this by resizing the textures (when filtered), seems quite costly to me.

If there's anything that can be done to reach Duckstation's no edge blending filtering, I think it would be worthwhile.

rr-:
For the record this is OG PC behavior.

Normally we could fix this using `GL_CLAMP_TO_EDGE`, but the problem is the game uses texture atlases which produces bad artifacts with the bilinear filter on. The solution is to add a fake border around each texture. 

Ref:
https://stackoverflow.com/a/35119455
https://gamedev.stackexchange.com/a/61800

This could potentially also help with the anisotropic filtering.

There are two ways I can imagine we could go around this.
The first is to literally offset all texture UVs and add those pixels by hand. It'll require us to support non-256×256 texture pages and has unknown performance hit on level loading.
The other is to use shaders, but this has an unknown performance hit on rendering which is already strained. I think Duckstation implements this here: https://github.com/stenzek/duckstation/blob/ce0689687c58020e9fd45bc3c6b0053b988b63b4/src/core/gpu_shadergen.cpp#L69-L96

CC @lahm86

lahm86:
I know that TombEditor generates padding around every texture, so whichever pixel is at UV bounds gets repeated. But as you say this means 256x256 tiles become inefficient with lots of dead space.I agree, shifting everything in OG would have a big impact on loading for sure. Texture packing is fast when we're just appending bits and pieces to end tiles, like for injections, but rebuilding the entire set on the fly is another story.
I personally feel it looks fine as is and is in-keeping with OG PC. But appreciate that there is a request for change.

rr-:
The TR1 part is now implemented in TR1X 4.10. What remains is TR2X.

PR #2707, opened 2025-04-06
Remove bilinear filter UV shift in TR1

rr-:
#### Checklist

#### Description

This addresses the TR1 aspect of issue #2258. The initial limitation stemmed from the need to circumvent the bilinear filter sampling pixels from neighboring textures in the texture atlas, which was previously managed by shifting the UV coordinates outwards by half a pixel. This pull request revises that approach by eliminating the offset, opting instead to clamp the texture UVs within the fragment shader. An OpenGL texture buffer, currently used for UVWs, can hold only 4 float values per data point. Since we use 3 for the UVW components, I created a separate buffer for the texture sizes. (An alternative would be to bump the profile to 4.x and adopt SSBOs instead, which allow much better structuring, but I'd like to try and stay at 3.3 for compatibility reasons.) On top of improving how toggle the filter looks, this should also marginally enhance anisotropic filtering appearance on NVIDIA graphics cards.

Additionally this PR merges the UVW and texture atlas texture buffers between sprites and objects so that they refer to a single linear buffer.

~~Curiously I added some `constexpr` and `static_assert` statements as an experiment on whether it'd work, and everything seems to compile in order, including the Mac builds.~~ Nope, I looked at the wrong build – it fails, just as expected of this platform.

### Before
output

### After
output2


Richard-L:
Love you so much Dash ❤️ 

Can't test yet but travelling back home now.

aredfan:
Looks great! The only issue I noticed is the pretty pixels option has no effect regardless of whether bilinear filter is on/off.

rr-:
Do you think it would make sense for us to remove it altogether and force-enable it?

aredfan:
> Do you think it would make sense for us to remove it altogether and force-enable it?

Yes I think that would be a good move. The pretty pixels make the textures look more uniform, and it would bring TR1 into line with TR2+.

**Edit:** I just came across this issue - there seems to be a few texture issues when the trapezoid filter is disabled. It's not present in dev snapshot.

20250407_075328_City_of_Khamoon

rr-:
> > Do you think it would make sense for us to remove it altogether and force-enable it?
> 
> Yes I think that would be a good move. The pretty pixels make the textures look more uniform, and it would bring TR1 into line with TR2+.
> 
> **Edit:** I just came across this issue - there seems to be a few texture issues when the trapezoid filter is disabled. It's not present in dev snapshot.
> 
> 20250407_075328_City_of_Khamoon

Thanks – fixed.

I also removed the pretty pixels option. However, one point to consider is that this causes a questionable gameplay regression compared to the original game and we should think about whether we actually want this change. With the bilinear filter disabled, the seams on the push blocks are no longer visible to the player.  IMO it's acceptable, as the seams are still clearly visible with the bilinear filter on, plus it's kind of cheating anyway. If anything, I'd make an option that specifically targets push blocks to keep the seams on just for them. LMK your thoughts.

aredfan (approved):

lahm86 (approved):

PR #2709, opened 2025-04-07
tr1: refactor savegame logic

lahm86:
#### Checklist

#### Description

Following #2695, this brings TR1's overall save control more into line with TR2, and will later allow us to move several components to TRX. I've stopped short of that merger to hopefully avoid regressions (or catch them early).

Some NBs:
- Removed all dependencies in the strategies with `GAME_INFO`
- Removed storage of `bonus_level_unlock` as AFAIK this is redundant. The gameflow sequencer re-calculates it at the end of the level anyway, so the stored value is never actually used. I realised too I had made an error in TR2 with this, by storing it as a config profile property (I think I got mixed up with NG+), so both games are now the same with a simple function-scope var.
- The death count is now held in level stats rather than `g_GameInfo`
- The current TR1X version string will now be appended to the end of save files, like TR2. I'm not sure whether to increase the version for this as we don't interpret it on load (the idea is we can just use it by externally inspecting the file). LMK what you think.


rr- (commented):
Thanks. I'd bump the version and provide a changelog in the comment, mostly for documentation purposes.

rr- (approved):

aredfan (approved):

rr- (commented):
One more thing, I think it's OK to supply this string as part of the regular BSON payload rather than appending it at the end of the file. IMO we should limit those special fields to minimum, as this special structuring exists only for performance reasons.
Apologies for not having commented about this earlier.

lahm86:
> One more thing, I think it's OK to supply this string as part of the regular BSON payload rather than appending it at the end of the file. IMO we should limit those special fields to minimum, as this special structuring exists only for performance reasons. Apologies for not having commented about this earlier.

No problem, updated. Applied the same to TR2 as well as it's unreleased and a quick patch.

walkawayy:
Regarding bonus levels, I think the idea was to keep it unlocked on the player's profile after they get all secrets once even if they delete their saves, start a new game, etc. How does it work with this new PR?

lahm86:
> Regarding bonus levels, I think the idea was to keep it unlocked on the player's profile after they get all secrets once even if they delete their saves, start a new game, etc. How does it work with this new PR?

That's still the case for NG+ i.e. player can unlock it after completing the game. But saving the bonus flag (for all secrets) isn't needed as the gameflow will recalculate it every time you finish a level regardless. It's only used to decide whether to play the next level or exit to title, and is set just a few lines prior in the same function.

rr-:
> > Regarding bonus levels, I think the idea was to keep it unlocked on the player's profile after they get all secrets once even if they delete their saves, start a new game, etc. How does it work with this new PR?
> 
> That's still the case for NG+ i.e. player can unlock it after completing the game. But saving the bonus flag (for all secrets) isn't needed as the gameflow will recalculate it every time you finish a level regardless. It's only used to decide whether to play the next level or exit to title, and is set just a few lines prior in the same function.

We may be introducing a slight inconsistency by preserving the NG+ flag while not doing so for the bonus level flag. However, the choice to play NG+ or NG is accessible directly from the title screen, whereas bonus levels can only be played within the game after playing at least one level. With that in mind, I believe it's okay to keep it as is (eg not store the bonus flag in the profile), unless I'm forgetting some feature – such as the "play any level" option, but that one is currently associated with specific save files rather than the general profile, so this consideration doesn't apply there.

lahm86:
Yes, the select level feature crossed my mind too but as you say you would need to have a save anyway to get to be able to get to that point.
I'll merge for now.

PR #2710, opened 2025-04-07
Make fog distances customizable by the player

### LINKED ISSUE #634: Feature request: ability to control rendering distance in the config
Blocked by #633 

The rendering distance is currently defined per-level (or with a default for all levels) in the gameflow file. This obviously has benefits for custom level builders, but players who want to play OG should be able to tweak this without having to resort to editing the JSON.

We should retain the ability for the custom level builders to control it on a per-level basis, but at the same time, the config tool should let us override it. I imagine a checkbox, and two sliders: min and max rendering distance. To know whether to fall back to the gameflow definition it would suffice to have NULL values or missing JSON keys, but it's OK to also keep a separate artificial setting for the checkbox state if it's easier for the tool from #633.

rr-:
#### Checklist

#### Description

The builders can still override it in the game flow file. We are doing this in the cutscene 4.
The UI for this is poor, as it's not indicated anywhere if that is the case – in the event the builder does configure the fog, the settings remain editable by the player, but nothing happens.
Implementing this setting revealed a regression in 4.9 that caused the sprites to be immune to the fog.


lahm86 (approved):
LGTM

aredfan (approved):

PR #2712, opened 2025-04-08
items: handle null anims/frames for sprite objects

### LINKED ISSUE #2711: TR1X bug: crash in folly's 4th secret when 3d pickups are disabled

lahm86:
Resolves #2711.

#### Checklist

#### Description

This ensures that sprite objects (or anything with no animation index defined) are assigned a default animation and as such allows us to handle null frames when getting item bounds. When Lara's animations are injected, animation index 0 (which sprite objects would previously default to) is no longer valid, hence sprites would then be pointing to a null entry.

Essentially the fix is in `Level_ReadSpriteSequences` - by setting sprite objects to have anim index -1.

While this wouldn't currently affect TR2 as we don't shift Lara's animations, the same logic has been applied for consistency, and because it makes sense that sprites should not have animations/frames assigned.

Thanks @rr- for help with this.

### DIFF
--- a/docs/tr1/CHANGELOG.md
+++ b/docs/tr1/CHANGELOG.md
@@ -6,6 +6,7 @@
 - fixed mesh faces not being drawn under some circumstances (#2452, #2438)
 - fixed objects disappearing too early around screen edges (#2005)
 - fixed the trapezoid filter being toggled if Alt-F4 (either left or right) is used to close the game (#2690)
+- fixed a crash when 3D pickups are disabled and Lara crosses a trigger to look at a pickup item (#2711, regression from 4.8)
 - fixed trapezoid filter warping on faces close to the camera (#2629, regression from 4.9)
 - fixed Mac builds crashing upon start (regression from 4.9)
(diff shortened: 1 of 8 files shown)


rr- (approved):
Thank you :heart:

docs/tr1/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - fixed mesh faces not being drawn under some circumstances (#2452, #2438)
 - fixed objects disappearing too early around screen edges (#2005)
 - fixed the trapezoid filter being toggled if Alt-F4 (either left or right) is used to close the game (#2690)
+- fixed a crash when 3D pickups are disabled and Lara crosses a trigger to look at a pickup item (#2711)
```
rr-: Let's add an information that it's a regression from 4.8.

aredfan (approved):

docs/tr1/CHANGELOG.md
lahm86: Good catch, thanks.

PR #2714, opened 2025-04-08
tr1/output: make water color player-customizable

### LINKED ISSUE #1532: Feature request: ability to control water colour from the config
This is to log what was discussed on discord, and in some ways mirrors #634.

The level builders have control over water colour in the gameflow on a per-level basis, and this arrangement will continue to have priority over the config when using a mod.  However, the default water colour for all levels could be moved to config, with a drop-down selection of 3 options: TombATI, PS1 and DOS water colours.

__________________

| Config options | R    | G   | B   |
|----------------|------|-----|-----|
| TombATI        | 0.45 | 1.0 | 1.0 |
| PS1            | 0.3  | 1.0 | 1.0 |
| DOS            | 0.6  | 0.7 | 1.0 |

For reference: https://github.com/LostArtefacts/TR1X/blob/develop/GAMEFLOW.md#global-properties

rr-:
#### Checklist

#### Description

Resolves #1532.


lahm86 (commented):
Very nice :)
Perhaps it could be useful to allow discrete changes e.g. if the walk key is pressed, adjust only by 1 rather than 10.

rr-:
> Very nice :) Perhaps it could be useful to allow discrete changes e.g. if the walk key is pressed, adjust only by 1 rather than 10.

Good suggestion, thanks. Added.

lahm86 (approved):

aredfan (approved):
Thank you for working on this. The PR build works great including holding shift when changing values.

I was wondering if we could add these RGB values in documentation somewhere in case someone might need it. I've put together this table.

| .       | R   | G   | B   |
|---------|-----|-----|-----|
| TombATI | 115 | 255 | 255 |
| PS1     | 77  | 255 | 255 |
| DOS     | 153 | 179 | 255 |

PR #2717, opened 2025-04-08
savegame: move and rename various elements

lahm86:
#### Checklist

#### Description

This is a lot of renaming to bring some types into line with each other ready for creating a common savegame module. I also want to rename object IDs like shotgun ammo to match, so we can avoid too many `#if/else` later for functions like `Savegame_PersistGameToCurrentInfo`; I will do this separately next as this is big enough.

One gotcha here is the way TR2's used medipack stats were stored. I've changed this now to a double to match TR1, and updated reading legacy savefiles appropriately. I haven't uprevved the BSON though as it's unreleased, LMK if you think that's OK. It would just mean any saves since the BSON introduction wouldn't show accurate medi usage. That's probably the main thing to look for in testing, that the small/large usage shows accurately.


src/libtrx/include/libtrx/game/savegame/enum.h
```diff
@@ -0,0 +1,37 @@
...
+    // access to essential data, such as the level counter and the level title,
+    // without the need to parse the entire BSON document.
+    // Added TR2+ stats ammo hits/used, health packs used, distance travelled.
+    VERSION_7 = 7,
+
+    // Added the current TR1X version string.
+    VERSION_8 = 8,
+#elif TR_VERSION == 2
+    VERSION_0 = 0,
+#endif
```
rr-: TBH I'd use linear versioning across both games, considering we're aiming for a common TRX format. WDYT?

rr- (approved):
LGTM.

src/libtrx/include/libtrx/game/savegame/enum.h
lahm86: That makes sense, done 😌

aredfan (approved):

PR #2726, opened 2025-04-09
savegame: move main logic to TRX

lahm86:
#### Checklist

#### Description

This gets the majority of the savegame logic into TRX. The strategies are still separated for now per game, and there are still some dependencies we can't yet shift like the requesters and slot counts. `Savegame_ApplyLogicToCurrentInfo` is also still separated because of the different ways the two games handle losing weapons, ammo etc. Adjusting that can be done later, as again, this PR has grown big enough IMO.


src/libtrx/game/savegame/common.c
```diff
@@ -19,3 +181,459 @@ int32_t Savegame_GetBoundSlot(void)
...
+const SAVEGAME_INFO *Savegame_GetSavegameInfo(const int32_t slot_num)
+{
+    return &m_SavegameInfo[slot_num];
+}
+
+void Savegame_InitCurrentInfo(void)
+{
+    if (TR_VERSION == 2 && Game_IsBonusFlagSet(GBF_NGPLUS)) {
+        return;
+    }
```
rr-, line 305: Can we explain in the comment why we're doing this?

src/libtrx/game/savegame/common.c
```diff
@@ -19,3 +181,459 @@ int32_t Savegame_GetBoundSlot(void)
...
+    if (GF_GetGymLevel() != nullptr) {
+        Savegame_GetCurrentInfo(GF_GetGymLevel())->flags.available = 1;
+    }
+    if (GF_GetFirstLevel() != nullptr) {
+        Savegame_GetCurrentInfo(GF_GetFirstLevel())->flags.available = 1;
+    }
+
+#if TR_VERSION == 2
+    Game_SetBonusFlag(GBF_NONE);
+#endif
```
rr-, line 324: and this as well.

src/libtrx/include/libtrx/game/savegame/common.h
```diff
@@ -13,7 +37,31 @@ void Savegame_UnbindSlot(void);
...
-extern bool Savegame_Load(int32_t slot_num);
-extern bool Savegame_Save(int32_t slot_num);
+extern void Savegame_HighlightNewestSlot(void);
+extern void Savegame_ApplyLogicToCurrentInfo(const GF_LEVEL *level);
+
+#define REGISTER_SAVEGAME_STRATEGY(strategy_)                                  \
+    __attribute__((__constructor__)) static void M_Register(void)              \
+    {                                                                          \
+        Savegame_RegisterStrategy(strategy_);                                  \
+    }
```
rr-, line 67: Yessss

src/libtrx/include/libtrx/game/savegame/const.h
```diff
@@ -1,3 +1,5 @@
 #pragma once
 
 #define SAVEGAME_CURRENT_VERSION 8
+#define MAX_STRATEGIES 2
```
rr-: This can be a module-level constant

rr- (approved):
Looks great. I'd appreciate heavy testing around highlighting the newest slot.

src/libtrx/game/savegame/common.c
lahm86: I'm not entirely sure why this is in place, as removing it (and the second call below) doesn't appear to cause any problems. But rather than removing it here, I've added a `TODO` for when we implement NG options (which I'm happy to look at next), then it can be properly tested. At least currently, this matches what was in TR2's savegame module.

src/libtrx/include/libtrx/game/savegame/const.h
lahm86: Done, thanks.

aredfan (approved):

src/libtrx/game/savegame/common.c
rr-: Thanks!

rr- (approved):

PR #2756, opened 2025-04-13
tr2: fix collision issues with trapdoors and bridges

### LINKED ISSUE #2294: OG bug: drawbridge produces dynamic light
- /play 16
- /fly
- /tp 65 -1 65
- Roll

Use the door cheat to activate the drawbridge (may need to nudge Lara forward) and watch the right of the screen - dynamic light is produced. This object uses the same control as the 40F minisub.

### LINKED ISSUE #2752: OG bug: collision issues with trapdoors/bridges
Ref #606 - this applies to TR2 as well.

lahm86:
Resolves #2752.
Resolves #2294.

#### Checklist

#### Description

This moves the three bridges, drawbridge and trapdoor implementations to TRX, and makes the config option to fix collision issues available in TR2. This implicitly fixes #2294 as the drawbridge is no longer using the same control as the sub (O_GENERAL).

The TR2 test level for this has been updated to capture an embedded bridge (à la Lost Valley) and where a half tile slope together with a trapdoor is used (just to verify this continues to work), plus of course the thorough test scenarios we had implemented for TR1.

Something I noticed as an aside is that `O_TRAPDOOR_TYPE_3` doesn't have any object registration. I don't think it's used anywhere in TR1 or TR2 so  - should we allocate it as per the other two?


rr- (approved):
Thank you. I'd leave the trapdoor type 3 around as-is, as we want to go with an game-agnostic object mapper eventually anyway.

aredfan (approved):
LGTM. 👍 

I also noticed an issue that is also present in develop. In room 17, if Lara pushes the block onto the bridge, it loses its collision.

lahm86:
> LGTM. 👍
> 
> I also noticed an issue that is also present in develop. In room 17, if Lara pushes the block onto the bridge, it loses its collision.

Thanks. I forgot to mention that this room doesn't work yet in either game. Pushblocks with bridges, trapdoors (and lifts) can be tackled later (I will raise an issue).

Issue #2086, opened 2024-12-20
Lara doesn't lose water tint during water exit animations


Richard-L:
https://github.com/user-attachments/assets/6dc08022-5340-4d04-8a1d-96def29a339b

https://github.com/user-attachments/assets/e796d499-3f14-4d8f-9070-1d7d25013a85

I thougt there was an issue for this already but just couldn't find it.

Depending on the height of the animation, the effect of her tint plopping back to normal is more or less pronounced. In TR1X, the tint is reset immediately at the beginning of the animation. This seems like an improvement, but better yet might be to determine a frame in the animation where the tint changes again (?). Due to TR2's wading, the differences are more pronounced here.

Richard-L:
@aredfan posted this image on discord, and when looking at screenshots, the issue may be even more noticeable than with animations.
![20241221_075130_Laras_Home](https://github.com/user-attachments/assets/17bb2d62-b791-431c-b7db-b1c08da49417)

I wonder if the game knows the height to which Lara is underwater. At least the braid seems to know.

aredfan:
TR3 for reference, with only Lara's torso staying tinted above the waterline. Both games could be improved.

![Screenshot](https://github.com/user-attachments/assets/6137eb63-bdc9-4b16-94d4-002ca4969069)

Richard-L:
Ah very good idea to compare to TR3. Interesting that they've chosen a middle ground with the arms. I've never noticed this until now.

Here's a 10fps slowed down clip of her exiting water. It looks like the behaviour is similar to TR1, where the entire Lara model tint (ie just the arms in this case) is reset upon water exit animation. It's hidden quite well.

https://github.com/user-attachments/assets/5dd71997-7eab-4f94-b999-1855c2c6da78

There _is_ some kind of colour overlay happening though aside from the tint, with just the arms, where they're in or out of water:

![image](https://github.com/user-attachments/assets/ea9a7c32-a266-449b-aabb-b431d1549d25)

I wonder if this may only just be the semi-transparent white overlay from the ripples.

rr-:
A note from Discord: the fix for this issue should also cover enemy meshes.

Issue #879, opened 2023-06-08
OG bug: shadow of enemies/traps can turn opaque


aredfan:
Reported by Raildex on discord.  Unlike Lara whose shadow is always semi-transparent no matter the height, the winged mutant's shadow is fully opaque once a certain height is reached.  I've confirmed this to happen in the `Sanctuary of the Scion` and `Atlantis` levels.

<details><summary>Screenshots</summary>

![1](https://github.com/LostArtefacts/Tomb1Main/assets/94776985/4295be98-e6c0-499a-94da-c6719fb7efb6)
![2](https://github.com/LostArtefacts/Tomb1Main/assets/94776985/17583f17-dd83-47bd-bafe-6c655ee12f75)
![3](https://github.com/LostArtefacts/Tomb1Main/assets/94776985/6f3de5ae-2e96-4a4a-9f33-c1fac180c905)

</details>

Raildex:
Not only mutants have this issue
![Screenshot 2023-07-25 114913](https://github.com/LostArtefacts/Tomb1Main/assets/14938934/e52c82d6-4279-4d5c-bfe8-5b2828a5eab6)
![Screenshot 2023-07-25 114933](https://github.com/LostArtefacts/Tomb1Main/assets/14938934/103f0ecd-0e68-45d3-9427-7a96f8348c90)

aredfan:
A few more observations: before the trap is activated its shadow is opaque.

<details><summary>Screenshot</summary>

![20230727_071722_Tomb_of_Tihocan](https://github.com/LostArtefacts/Tomb1Main/assets/94776985/5b780358-2430-4478-a01f-aa987bffdfd3)

</details>

From this angle, the shadow is half-opaque and half-transparent.

<details><summary>Screenshot</summary>

![20230727_071806_Tomb_of_Tihocan](https://github.com/LostArtefacts/Tomb1Main/assets/94776985/60bb1f73-23a3-40db-afd2-18491df30dc1)

</details>

If Lara is on the surface after activating the trap, there is no shadow.

<details><summary>Screenshot</summary>

![20230727_072015_Tomb_of_Tihocan](https://github.com/LostArtefacts/Tomb1Main/assets/94776985/d186fafd-c035-4ec1-9b5d-9efa38beb626)

</details>

All of these behaviours are the same in TombATI.

PR #2757, opened 2025-04-13
creature: set TR1 enemy room number above position

### LINKED ISSUE #2753: TR1X - Creatures don't make any sounds when in 1 click high water
Tested with the Gorilla and Lion (Male and female), not tested with other creature entities. Happens as of latest stable and dev release (TR1X-4.9-132-g454d133).

[Example level](https://github.com/user-attachments/files/19724459/Level1.zip)

lahm86:
Resolves #2753.

#### Checklist

#### Description

This uses the same approach as TR2 to set an enemy's room number slightly above their position, so that one click water does not produce a tint and to allow sound effects to play. Re tint, #2086 can be tackled later for responsive tinting, this PR brings things into line with Lara in one-click water.

Test level in issue comment (original attachment has missing SFX data, but the bug is present nonetheless).

### DIFF
--- a/docs/tr1/CHANGELOG.md
+++ b/docs/tr1/CHANGELOG.md
@@ -13,6 +13,7 @@
 - fixed mesh faces not being drawn under some circumstances (#2452, #2438)
 - fixed objects disappearing too early around screen edges (#2005)
 - fixed the trapezoid filter being toggled if Alt-F4 (either left or right) is used to close the game (#2690)
+- fixed enemies in one-click high water appearing with a water tint, and not making any animation sounds (#2753)
 - fixed a crash when 3D pickups are disabled and Lara crosses a trigger to look at a pickup item (#2711, regression from 4.8)
 - fixed trapezoid filter warping on faces close to the camera (#2629, regression from 4.9)
 - fixed Mac builds crashing upon start (regression from 4.9)
--- a/src/libtrx/game/creature/common.c
+++ b/src/libtrx/game/creature/common.c
@@ -912,10 +912,13 @@ bool Creature_Animate(
         item->rot.x = 0;
     }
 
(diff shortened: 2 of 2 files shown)


rr- (approved):

aredfan (approved):
Testing LGTM. 👍 

I also noticed a minor bug is now introduced with the entities' shadow turning black in the water when viewed from above, not present in develop. I'm guessing this issue is already covered under #879. (Not a big deal, just making a note of it.)

lahm86:
> Testing LGTM. 👍
> 
> I also noticed a minor bug is now introduced with the entities' shadow turning black in the water when viewed from above, not present in develop. I'm guessing this issue is already covered under #879. (Not a big deal, just making a note of it.)

Is that will all enemies now? Or just the ones in one-click water? Hmm. @rr- I wonder if this is a sorting issue e.g. the room above which contains the enemy is first drawn, so its shadow is in that room, but then the room below, where its shadow actually is, is drawn later so the faces there are then on top of the transparent shadow faces?

aredfan:
> Is that will all enemies now? Or just the ones in one-click water?

Just for the enemies in one-click water when using the test level. In the screenshot below, the ape has opaque shadow while in water whereas the lioness has transparent shadow on land.

20250413_152031_Test_Level

rr-:
The most straightforward fix currently is to delay rendering of the shadows to be a part of the `Output_FlushTranslucentObjects` call. We already do that for the lightnings – we'd have to reintroduce `SHADOW_INFO` or a similar struct, remember the current `g_MatrixPtr`, queue it and then call  the current implementation of `Output_DrawShadow`.

Another place where this is problematic is the t-rex – `/play 3` `/tp t-rex`.  Depending on the viewing angle half of its shadow becomes opaque due to depth buffer shenanigans.

lahm86:
Sounds good, thanks. I guess we can tackle it under #879 as it's not really something introduced here, just a by-product.

PR #2763, opened 2025-04-13
ui: introduce UI v2

rr-:
#### Checklist

#### Description

This PR introduces a new UI framework (ui v2), living under game/ui2/, based on an immediate-mode approach.

#### Background

The original UI system (ui v1) was added to reduce the need for manual widget positioning in client code, which was painful to debug and maintain. While v1 helped with layout, it required a significant amount of boilerplate for each widget. Every element needed explicit struct definitions, state management, and a bunch of wrapper code to hook into the layout and rendering system.

#### The new approach
This PR explores a more flexible approach based on immediate-mode UI. In this model, widgets are re-issued every frame and built into a virtual tree that uses an arena allocator to avoid memory thrashing. Layout and rendering are done in a single pass, and widgets operate directly on user-owned state.

#### Example usage:

```c
static UI2_PROMPT_STATE m_Prompt = {0};

// at startup
UI2_Prompt_Init(&m_Prompt);

// on every frame
UI2_BeginStack(UI_STACK_VERTICAL);
UI2_Label("Enter your name:");
UI2_Prompt(&m_Prompt);
UI2_EndStack();

// on shutdown
UI2_Prompt_Free(&m_Prompt);
```

Instead of storing widget state inside the UI system, this design lets the caller manage state explicitly. It makes the widget code dramatically simpler – see `game/ui/widgets/spacer.c` and `game/ui2/widgets/spacer.c` to see an example of this. It also removes the need to store references to individual widgets in the client code.

#### Notes

Normally, immediate-mode systems use internal widget IDs to track widget state across frames. I believe that it adds complexity we don't need here, and instead I opted to require callers to pass in their own state (e.g. `UI2_PROMPT_STATE`). In practice, this seems to work well.

#### Performance concerns

One questionable bit is that we reconstruct every label using a `TEXTSTRING` twice on every frame. This means the `strcmp()` dance around every glyph to match the Unicode endpoints to sprite indices. To help with this, I added caching to `TEXTSTRING` glyph Unicode decomposition, and rendering 5000 characters on the screen takes ~1 ms. I think that's acceptable. Rendering the controls dialog takes 0.18 ms.

#### Progress

All of the v1-dependent code is ported to v2 within the PR.


aredfan:
The console is working as expected in both games. 👍

rr-:
Added a basic caching of glyph decomposition to the text module, and I got a ~60% reduction in runtime for this code:
```c
    BENCHMARK benchmark = Benchmark_Start();
    UI2_BeginAnchor(0.5, 0.5);
    UI2_BeginStack(UI2_STACK_VERTICAL);
    for (int i = 0; i < 128; i++) {
        char tmp[200] = "";
        for (int j = 0; j < 50; j++) {
            char tmp2[10];
            sprintf(tmp2, "%d ", i);
            strcat(tmp, tmp2);
        }
        UI2_Label(tmp);
    }
    UI2_EndStack();
    UI2_EndStack();
    Benchmark_End(&benchmark, nullptr);
```

```
old=2.82 ms     new=1.18 ms     (41.84%)
```

I think we can live with this. Eventually I would like to remove text.c and TEXTSTRINGs altogether so we can spend some more time on performance tweaks at that time IMO.

lahm86:
Looking great! I tried the benchmarking too and am getting 0.2ms on average (replaced the contents of `UI2_Console` with the test code).

rr-:
> Looking great! I tried the benchmarking too and am getting 0.2ms on average (replaced the contents of `UI2_Console` with the test code).

Thanks, doubly so for taking the time to do a benchmark. However for a clearer picture of performance, you might want to focus on the `UI2_EndScene` function, as that's where the main tasks like sizing, laying out widgets, and drawing happen :relieved:

aredfan:
I've ran into these issues while testing the latest PR builds. (9afb65d)

- [x] The toggle UI input has no effect for photo mode in both games.
- [x] TR2X: The photo mode UI appears to be missing its border and text, see screenshot below.
- [x] TR2X: While photo mode is open, it can also affect a small part of the console, where the `U` is missing in `UI disabled`.

Screenshot

rr-:
@aredfan thanks – the photo mode issues should be now fixed.

The newest patch also fixes text scaling and ports the new game dialog in both games.

aredfan:
@rr- I can confirm that the UI toggle is working again for photo mode in TR1, and it does make the TR2's photo mode transparent box hide or show. Thank you. 👍 

The other 2 issues persists and seem to affect the new game dialog as well. If I start up the game and select new game option, the transparent box appears for a split second then disappears. If I play a level then return to the title screen and select new game, the transparent box stays visible until I deselect it.

I've also tried deleting TR2X.json5 and launched the game with default config settings, and the issues remain, so we can rule that out at least.
______________

- [x]  TR1X: I noticed that both photo mode and new game dialog box have either sizing issue or gets misplaced when the text size is changed to a value other than 1.0.

rr-:
@aredfan thank you. I've fixed the TR1 UI scaling issue. I've tried to address the TR2 issue as well, but it's a bit more difficult as I can't reproduce it on my end. Please LMK if that worked. Lastly, the pause screen is now powered with the v2.

aredfan:
@rr- The text scaling and pause screen both LGTM for TR1X. I'm afraid the TR2X issues still persists - affecting photo mode, new game dialog and pause screen.

rr-:
> @rr- The text scaling and pause screen both LGTM for TR1X. I'm afraid the TR2X issues still persists - affecting photo mode, new game dialog and pause screen.

Thank you. The only difference I notice with version 1 is the order of the text and background draws being swapped. I've now updated the order in version 2 to match it. Hopefully we got it this time :relieved:

aredfan:
**Issues:**

- [x] TR1X: Regarding the new game and pause dialogs - the border around the header and selection looks dark as if positioned behind the transparent box.

rr-:
Pushed a fix. Also ported the stats dialog.
Points of interest:
- TR1 end of the level in gym (no stats)
- compass stats in both games
- compare widget appearance timings with develop (esp. the stats dialog - in the compass, in gym, and in regular levels)
- assault course with no records
- assault course with records (it can store up to 10 records, and while the dialog only shows 7, you can actually use arrow buttons to move around like in other dialogs)
- a bug fix to the PS1-style stats dialogs – the title bar is now properly brown rather than purple
- small tweaks to paddings, margins and sizes all across the ui v2 widgets
Considering that the latest changes affected most of the core widgets, I'd appreciate some retesting. Screenshot comparisons with the original TombATI would be ideal – while the UI don't need to be pixel-perfect, they need to be a good-enough visual match.

From the v1 widgets left to port, only the TR2 graphic options and controls dialog are left.

aredfan:
> Pushed a fix.

Can confirm, thank you. 👍 

New issues I've come across:

- [x] In both games, the final stats doesn't fade-in from black, or fade-out to black
- [x] TR1: The end of level stats doesn't fade-out to black

lahm86 (approved):

lahm86 (approved):

aredfan (approved):

PR #2769, opened 2025-04-16
tr1/output: fix calculating texture sizes

rr-:
#### Checklist

#### Description

Fixes faces such as this:
image
The original code only looked at 3 corners to establish the texture size, which wouldn't work with certain trapezoid shapes.

~~The fix is simple and it's to take the fourth corner into account. However, it has the potential of reintroducing the bug in other places. Specifically, the fourth corner is also used to establish the min/max sizes for triangles. If it contains garbage data, we might see weird lines around triangle faces. @lahm86 is there a simple way for us to see if `OBJECT_TEXTURE` is part of a 3-face or 4-face without iterating through every possible mesh in the level?~~
Aredfan found out that this approach does indeed cause issues. Updated to establish the relevant UV count.

Additionally the wibble effect applying also to sprites – they should remain unaffected.

### DIFF
--- a/src/libtrx/game/level/common.c
+++ b/src/libtrx/game/level/common.c
@@ -881,6 +881,7 @@ void Level_AppendObjectTextures(
 {
     for (int32_t i = 0; i < num_textures; i++) {
         OBJECT_TEXTURE *const texture = Output_GetObjectTexture(base_idx + i);
+        texture->uv_count = 4; // Default to 4 vertices
         texture->draw_type = VFile_ReadU16(file);
         texture->tex_page = VFile_ReadU16(file) + base_page_idx;
         for (int32_t j = 0; j < 4; j++) {
@@ -1189,6 +1190,26 @@ finish:
 
 void Level_LoadTextures(void)
 {
+    for (int32_t room_num = 0; room_num < Room_GetCount(); room_num++) {
+        const ROOM *const room = Room_Get(room_num);
+        for (int32_t j = 0; j < room->mesh.num_face3s; j++) {
+            const FACE3 *const face = &room->mesh.face3s[j];
+            OBJECT_TEXTURE *const texture =
+                Output_GetObjectTexture(face->texture_idx);
+            texture->uv_count = 3;
+        }
+    }
+
+    for (int32_t i = 0; i < Object_GetMeshCount(); i++) {
+        const OBJECT_MESH *const mesh = Object_GetMesh(i);
+        for (int32_t j = 0; j < mesh->num_tex_face3s; j++) {
+            const FACE3 *const face = &mesh->tex_face3s[j];
+            OBJECT_TEXTURE *const texture =
+                Output_GetObjectTexture(face->texture_idx);
+            texture->uv_count = 3;
+        }
+    }
+
     for (int32_t room_num = 0; room_num < Room_GetCount(); room_num++) {
         ROOM *const room = Room_Get(room_num);
         for (int32_t j = 0; j < room->mesh.num_face4s; j++) {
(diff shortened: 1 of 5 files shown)


lahm86:
> @lahm86 is there a simple way for us to see if OBJECT_TEXTURE is part of a 3-face or 4-face without iterating through every possible mesh in the level?

As far as I know, if the last UV has both elements as 0 then it's definitely meant to be a triangle texture. Using a triangle on a quad face would be on the builder IMO; but using a quad on a triangle face I think is "legal" in TE.

rr-:
> > @lahm86 is there a simple way for us to see if OBJECT_TEXTURE is part of a 3-face or 4-face without iterating through every possible mesh in the level?
> 
> As far as I know, if the last UV has both elements as 0 then it's definitely meant to be a triangle texture. Using a triangle on a quad face would be on the builder IMO; but using a quad on a triangle face I think is "legal" in TE.

Aredfan found out that going with just 4 corners at all times does indeed cause issues for 3-faces. Updated to establish the relevant UV count at the time of loading a level file.

aredfan:
I've come across a few issue:
- [x] The water animated texture appears and disappears in the examine test level, and Temple of the Cat, room 32.
- [x] The animated waterfall texture in cutscene 1 is affected as well.
- [x] I'm experiencing a crash with the following steps:

1. `/play 1`
2. `/play 2`

rr-:
> I've come across a few issue:
> 
> * [ ]  The water animated texture appears and disappears in the examine test level, and Temple of the Cat, room 32.
> * [ ]  The animated waterfall texture in cutscene 1 is affected as well.
> * [ ]  I'm experiencing a crash with the following steps:
> 
> 1. `/play 1`
> 2. `/play 2`

@aredfan hopefully  these should be fixed now.

aredfan (approved):
> @aredfan hopefully these should be fixed now.

I can happily confirm all these issues are fixed, thank you.

src/libtrx/game/level/common.c
```diff
@@ -1189,6 +1190,26 @@ void Level_ReadSamples(VFILE *const file)
 
 void Level_LoadTextures(void)
 {
+    for (int32_t room_num = 0; room_num < Room_GetCount(); room_num++) {
```
lahm86, line 1193: My only reservation with scanning faces is that if a triangle face has a rectangle texture, but the same rectangle texture is also used validly elsewhere on a rectangle face, then it would become corrupted on those other faces.

I think we could simplify this as follows. 

```c
for (int32_t i = 0; i < Output_GetObjectTextureCount(); i++) {
    OBJECT_TEXTURE *const texture = Output_GetObjectTexture(i);
    if (texture->uv[3].u == 0 && texture->uv[3].v == 0) {
        texture->uv_count = 3;
    }
}
```

Triangles always have 0/0 as their UV for the final vertex. It doesn't conflict with an actual XY atlas coord of 0,0 because of the way the values are packed.

https://github.com/MontyTRC89/Tomb-Editor/blob/master/TombLib/TombLib/LevelData/Compilers/Util/TexInfoManager.cs#L1500
https://github.com/LostArtefacts/TR-Rando/blob/master/TRLevelControl/Model/Common/TRObjectTextureVert.cs

E.g. Tex 6 in Caves is at atlas 0, XY `0, 0` while its UV values are `1, 1`

lahm86: Double checked and TE will always apply a triangle to a triangle face, so I think we are good as-is.

lahm86 (approved):

src/libtrx/game/level/common.c
rr-: Resolved with internal testing on Discord – the only possible scenario for a texture to be shared between quads and triangles is for it to come from the current iteration of the injection framework. And the only scenario where it'd cause issues is if the quad was a trapezoid, which is never the case with the current content. TombEditor is guaranteed to always produce a separate texture for a triangular face.

PR #2772, opened 2025-04-17
tr2/data: fix invalid portals in The Deck

### LINKED ISSUE #2393: OG bug: possible for barracuda to be visible or targeted through overlapped rooms | The Deck
**Steps A:**

1. `/play 10`
2. drop down through the opening on the left
3. eliminate the barracudas using a conventional weapon, and not a grenade gun
4. `/tp 51 4 48`
5. move the camera until you can see the dead barracudas 

Image

**Steps B:**

1. `/play 10`
2. `/kill goon`
3. `/tp 52 5 53`
4. take out a weapon and wait a bit until you see Lara locking onto invisible barracudas 

Image

__________________

A copy of Lahm's reply from #1853 which I believe may be relevant to the issue. 

> This is a strange one. It seems like a data issue - room 17 has a visibility portal into room 104 (which is clearly wrong, must have been a level editing bug), and in turn because that is linked to room 80, room 80 is drawn. The portal is marked by the waypoints here, just imagine it pointing downwards to room 104.
> 
> image
> 
> I'm not sure where the regression is if this doesn't happen in OG. Ideally we would fix the data and the regression itself.

lahm86:
Resolves #2393.

#### Checklist

#### Description

This disables the visibility portal from room 17 to room 104 (and vice versa) in The Deck to avoid Lara seeing or targeting enemies in the disconnected rooms.

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/37


rr- (approved):

aredfan:
I can confirm the visibility issue is resolved, thank you. 👍 

But Lara can still target the barracudas from the following positions.

- `/tp 52 5 53` (there's a good chance of getting a target lock from this position)
- `/tp 51 4 48`

lahm86:
> I can confirm the visibility issue is resolved, thank you. 👍
> 
> But Lara can still target the barracudas from the following positions.
> 
> * `/tp 52 5 53` (there's a good chance of getting a target lock from this position)
> * `/tp 51 4 48`

Thanks. My bad, I mistakenly thought the visible rooms were taken into account for gun targeting checks, but I was getting muddled with baddie collision. So it's working because LOS check doesn't take into consideration the visible rooms. I'll have a think about it.

lahm86:
This is complicated. Changing `LOS_Check` to be more strict with the rooms breaks other things badly, for example Lara standing on a ledge above an enemy with the M16, and she then isn't able to target, whereas she can on develop. I thought we could maybe rely on the previous frame's drawn room list, but this list gets reset for things like static collision (and that only populates immediate rooms, so again affects targeting at a distance).

My vote is we raise a second ticket for targeting and I'll adjust the changelog to say that this fixes the visibility issue only, as this will need some deeper investigation for a workaround. WDYT? Tagging @rr- too.

rr-:
> This is complicated. Changing `LOS_Check` to be more strict with the rooms breaks other things badly, for example Lara standing on a ledge above an enemy with the M16, and she then isn't able to target, whereas she can on develop. I thought we could maybe rely on the previous frame's drawn room list, but this list gets reset for things like static collision (and that only populates immediate rooms, so again affects targeting at a distance).
> 
> My vote is we raise a second ticket for targeting and I'll adjust the changelog to say that this fixes the visibility issue only, as this will need some deeper investigation for a workaround. WDYT? Tagging @rr- too.

That makes sense to me. It's totally fine to handle this separately if all broken portals have been removed and none are left. I agree that this is already a significant, standalone improvement. However, before we proceed with the merge, I'd suggest taking extra care to ensure that room flip maps aren't affecting the expected outcome.

aredfan (approved):
I have no problem with merging this PR as-is. It makes little sense to get bogged down without further research on the secondary issue. 👍 

> I'd suggest taking extra care to ensure that room flip maps aren't affecting the expected outcome.

I did a quick check - the flip maps still removes dead bodies.

lahm86:
Thanks. Raised #2774 for follow-up, and changelog updated here.

PR #2778, opened 2025-04-18
tr1/game-flow: fix story so far showing up in UB

### LINKED ISSUE #2611: TRUB should not have "story so far"
When you go into level select of a new TRUB save.

Bit sad there's nothing to show.

rr-:
#### Checklist

#### Description

Resolves #2611.

#### Testing

UB should no longer show this feature as there are no FMVs to play.
Another good corner case that's not present with the game flows we ship, is that FMVs that come _after_ the level shouldn't trigger the Story So Far to show up. Consider a game flow where the first level is structured as follows:

```
            "sequence": [
                {"type": "loop_game"},
                {"type": "play_fmv", "fmv_id": …},
                {"type": "level_stats"},
                {"type": "level_complete"},
            ],
```

If there are no cutscenes or FMVs before this sequence and a player saves their progress during the level, the Story So Far shouldn't show up. This is because the FMV has not yet occurred when the save is made.


aredfan (approved):

lahm86 (approved):
Good catch with FMVs that come after the game loop.

PR #2782, opened 2025-04-18
tr2/ui/graphic_settings: add further controls to graphics dialog

lahm86:
#### Checklist

#### Description

This makes several visual/rendering options available in TR2's graphic dialog, including adding support for enum types.

Still to add:
- [x] ui.text_scale
- [x] ui.bar_scale
- [x] rendering.scaler
- [x] rendering.sizer

Further adjustments
- [x] Adjust sizing for flexible layout (e.g. right arrow in render mode currently overlaps)
- [x] Check visible rows when text size increases


src/tr2/game/ui/dialogs/graphic_settings.c
```diff
@@ -194,6 +343,13 @@ static bool M_RequestChangeValue(const int32_t row_idx, const int32_t dir)
         *component = component_i;
         break;
     }
+    case COT_ENUM: {
+        const M_ENUM_LOOKUP enum_lookup = M_GetEnumEntry(option);
+        const M_ENUM_ENTRY *const next_entry =
+            &m_EnumOptions[option->misc][enum_lookup.position + delta];
+        *(int32_t *)option->target = next_entry->value;
```
rr-, line 411: Good call on handling non-linear enum values by working with their indices instead.

src/tr2/game/ui/dialogs/graphic_settings.c
```diff
@@ -27,7 +27,71 @@ typedef struct {
...
+};
+
+// Indexed by M_OPTION.misc
+static const M_ENUM_ENTRY *m_EnumOptions[] = {
+    m_TextureFilterOptions,
+    m_LightingContrastOptions,
+    m_RenderModeOptions,
+    m_AspectModeOptions,
+    nullptr,
+};
```
rr-: I believe we can get by without this enum if we choose to make `misc` a `void*` general pointer, that for the current entities will be cast to `(int32_t)(intptr_t)`, and for the newly introduced enums, it'll let them reference `M_ENUM_ENTRY[]` directly.

rr- (commented):
Re: the arrow,

Currently all values are put inside a `Resize` component that ensures the column is of the same width. But it's a bit dumb, eg if the child is too big, it'll cause the sort of behavior we're seeing. Ideally we should let it resize automatically. One possible refactor would be to go with two requester bodies, one for each column, that go inside a horizontal stack. Going with this approach alone is not enough once the dialog becomes scrollable, though, as it'll cause the columns (and the dialog) to expand and shrink as the player scrolls elements of different sizes into view. To combat this, we can follow the approach from the examine dialog, and always call all the rows, except wrap the rows not meant to be drawn in a `Resize(-1.0f, 0.0f)`. This will cause the children to not be shown, yet let all of them still participate in the layout pass and take up horizontal space they need regardless of whether they're shown or not.

rr-:
BTW I emboldened the `hide photo mode UI` option by mistake on Discord – personally I don't think it needs to be put in there, but I'll leave it to your judgment. 😌

Richard-L:
Small thing: can you make it so that pressing up/down on the first/last item in the list will loop to the end/start?

Otherwise just amazing to finally see these in-game 🙏

rr-:
> Small thing: can you make it so that pressing up/down on the first/last item in the list will loop to the end/start?
> 
> Otherwise just amazing to finally see these in-game 🙏

That's a solid suggestion. Should we apply this to savegame dialogs as well? IMO yes, behind a new `ui.enable_wraparound` setting that defaults to true.
(It won't work with the current codebase because these dialogs are still on the old code, but I'm thinking ahead to when we update them.)

src/tr2/game/ui/dialogs/graphic_settings.c
lahm86: Thanks, done 😌

lahm86:
Thanks, @rr-, @Richard-L. I'll work on the resizing next and will add a separate commit for wrapping.

Richard-L:
Thanks to you both 🙏

Personally would love the wrapping for save states too.

lahm86:
We should be good for checking this now. Thanks @rr- for helping 😌

src/tr2/game/ui/dialogs/graphic_settings.c
```diff
@@ -186,6 +388,16 @@ static bool M_RequestChangeValue(const int32_t row_idx, const int32_t dir)
...
         break;
+    case COT_DOUBLE: {
+        const float df = ceilf((delta / 100.0f) * 10) / 10;
+        *(double *)option->target += df;
+        break;
+    }
+    case COT_FLOAT: {
+        const float df = ceilf((delta / 100.0f) * 10) / 10;
+        *(float *)option->target += df;
+        break;
```
rr-, line 398: I think even with this approach, it can accumulate error over time. Maybe we could do something in the lines of
```c
*(float*)option->target = (round(*(float*)option->target * 10) + delta) / 10.0f;
```
?

lahm86: Thanks, updated. We still need to divide the delta alone as it's an int. I suppose the alternative could be having `delta_slow`/`delta_fast` as floats themselves.

rr-: Thanks. The delta can remain an int- if we multiply and divide by 10, we get precision of 0.1 for delta=1. Eg ((0.3*10)+1)/10 = 0.4 and thanks to doing this on integers we can ensure error to be minimal.

rr- (approved):

aredfan (approved):

PR #2786, opened 2025-04-18
objects/movable_block: fix angle resetting and snapping

### LINKED ISSUE #2776: OG bug: pushblock angle is not restored
Pushblock angles are normalised during collision tests, but are not restored afterwards. This isn't an issue in OG, but for ones with customised textures, it can cause textures to effectively flip on the fly.

Image

Image

lahm86:
Resolves #2776.

#### Checklist

#### Description

This restores pushblock angles after performing collision tests, and also snaps them to the grid on initialisation if a builder places them at non 90 degree angles to avoid instead snapping during collision.

### DIFF
--- a/docs/tr1/CHANGELOG.md
+++ b/docs/tr1/CHANGELOG.md
@@ -17,6 +17,7 @@
 - fixed enemies in one-click high water appearing with a water tint, and not making any animation sounds (#2753)
 - fixed the scale of the four keys in St. Francis' Folly (#2652)
 - fixed the panther at times not making a sound when it dies, and restored Skate Kid's death SFX (#2647)
+- fixed pushblocks being rotated when Lara grabs them, most noticeable if asymmetric textures have been used (#2776)
 - fixed a crash when 3D pickups are disabled and Lara crosses a trigger to look at a pickup item (#2711, regression from 4.8)
 - fixed trapezoid filter warping on faces close to the camera (#2629, regression from 4.9)
 - fixed Mac builds crashing upon start (regression from 4.9)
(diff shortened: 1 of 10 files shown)


aredfan:
Nice bugfix. Funnily enough there are push blocks with non-symmetric textures in Furnace, Kingdom and Nightmare.

~~- [ ] I've encountered a bug with a push block getting stuck in a specific position that isn't present in develop.~~

~~1. launch golden mask~~
~~2. `/play 4`~~
~~3. try moving the push block onto the gold square~~

**Edit:** It appears to be Lara being able to push/pull on one side not the other 3 sides.

lahm86:
Good catch as always @aredfan. I have an idea what's going on, will sort it out in the morning.

lahm86:
This should now be fixed. I have had to introduce a bit of extra code when drawing, allowing objects to define that their root bone should accept extra rotations. We do this because allowing the control/collision to change the original angle is essential to the blocks animating in the correct direction. What we do now though is add a counter-angle as an extra rotation, but because blocks only have one mesh, the additional drawing code was needed as there is effectively no bone. I think we can tidy this up as a separate task, have a common struct for rot flags between objects and bones, and eliminate some of the rotation duplication when drawing.

I've shared test levels, which have one of each block type and each at a different angle. There are also ones if you look in trview positioned at odd angles - the game should snap these to the grid. Otherwise, the block textures should never change, and of course be movable in each direction. This should persist with save/load as well.

Thanks for your patience.

aredfan (approved):
I ❤️ testing push blocks

rr- (approved):

PR #2788, opened 2025-04-19
Remove REQUESTER_INFO from tr1 in favor of UI

rr-:
#### Checklist

#### Description

Rewrites load/save and select level TR1 dialogs to use the ui framework.
The TR2 dialogs are still on the legacy `REQUESTER_INFO`.

#### Testing

The dialogs are known to not have a 100% pixel perfect match, but they seem good enough.
It's worth paying attention to the highlighting of the newest slot.
Since the PR involved doing significant changes to some of the core elements, a general sanity check for various dialogs (including graphic settings) would be appreciated.


lahm86 (commented):
The select level option is available from within the savegame dialog, and if you go to it, you then are unable to return or escape from the inventory.

But otherwise, LGTM.

rr-:
> The select level option is available from within the savegame dialog, and if you go to it, you then are unable to return or escape from the inventory.
> 
> But otherwise, LGTM.

Thank you – fixed.

lahm86 (approved):

aredfan (approved):

PR #2791, opened 2025-04-19
console: use room numbers for teleporting and update pos output

### LINKED ISSUE #2486: TRX bug: Lara sometimes ends up in incorrect rooms when using the teleport cheat
**TR1X regression since:** [3.0](https://github.com/LostArtefacts/TRX/releases/tag/tr1-3.0)
**TR2X regression since:** [0.3](https://github.com/LostArtefacts/TR2X/releases/tag/0.3)
In both cases, it's when the developer console was introduced.

________________

**Steps 1:**
1. launch tr1
2. `/play 0`
3. use `/tp 15` several times
4. Lara will either end up in rooms 15 or 13

**Steps 2:**
1. launch trub
2. `/play 2` (Temple of the Cat)
3. use `/tp 60` several times
4. Lara will either end up in rooms 60 or 47/50/51

**Steps 3:**
1. launch tr2
2. `/play 14`
3. use `/tp 127`
4. Lara will end up in room 126 according to the console logs

### LINKED ISSUE #2487: TRX bug: the console logs report incorrect room number when in flipped rooms
**TR1X regression since:** [3.0](https://github.com/LostArtefacts/TRX/releases/tag/tr1-3.0)
**TR2X regression since:** [0.3](https://github.com/LostArtefacts/TR2X/releases/tag/0.3)

______________________

**Steps 1:**

1. launch tr1
2. `/play 10`
3. `/tp 53 4 34`
4. `/pos` and observe it's room 54
5. `/flip`
6. `/pos` which should show room 60 but it's still room 54

**Steps 2:**

1. launch tr2
2. `/play 3`
3. `tp 35 3 18`
4. `/pos` and observe it's room 73
5. `/flip`
6. `/pos` which should show room 143 but it's still room 73

lahm86:
Resolves #2486.
Resolves #2487.

#### Checklist

#### Description

If teleporting to a specific room or item, the room number is already known so we now pass that to Lara's teleporting function. Teleporting to a position will continue to locate the room by position. I've also updated it to make use of the `no-space` fix to avoid Lara teleporting into these types of walls (example where it was common was Temple of the Cat room 75).

I've also added a guard for this approach if you try to `tp` to a flipped room when the flipmap is set. An example is Khamoon, room 54, set the flipmap so you're effectively in room 60 (although 60 is in 54's slot) then try to `tp 60`. Without the check, the game will hang as slot 60 is invalid in the current state. This worked OK previously as we were going by position.

The `/pos` command will also now check the flipmap status and report the alternative room when applicable (again as above, the old room index is technically correct, it's the slots that have swapped, but I think the change is good as it ties in with what you see in trview, TombEditor etc).

My only concern with the updates is that teleporting _by position_ may be less lenient than it was before when overlapping rooms are involved. But perhaps this is acceptable, LMK what you think.

### DIFF
--- a/docs/tr1/CHANGELOG.md
+++ b/docs/tr1/CHANGELOG.md
@@ -31,6 +31,7 @@
 - fixed Story So Far showing up even when there's nothing to play (#2611, regression from 2.10)
 - fixed Story So Far not playing the opening FMV, `cafe.rpl` (#2779, regression from 2.10)
 - fixed Lara at times ending up in incorrect rooms when using the teleport cheat (#2486, regression from 3.0)
+- fixed the `/pos` console command reporting the base room number when Lara is actually in a flipped room (#2487, regression from 3.0)
 - improved bubble appearance (#2672)
 - improved rendering performance
 - improved pause exit dialog - it can now be canceled with escape
--- a/docs/tr1/CHANGELOG.md
+++ b/docs/tr1/CHANGELOG.md
@@ -30,6 +30,7 @@
 - fixed wrong PS1-style title bar color for the end of the level stats dialog (regression from 4.9)
 - fixed Story So Far showing up even when there's nothing to play (#2611, regression from 2.10)
 - fixed Story So Far not playing the opening FMV, `cafe.rpl` (#2779, regression from 2.10)
+- fixed Lara at times ending up in incorrect rooms when using the teleport cheat (#2486, regression from 3.0)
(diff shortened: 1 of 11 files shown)


aredfan (approved):
I tested all the scenarios documented in the issues, and a few more I could think of.

Amazing work, and thank you.👍

rr- (approved):

PR #2807, opened 2025-04-20
tr2/flares: fix flicker in 60 fps

### LINKED ISSUE #2806: TR2X bug: flare lighting doesn't make Lara flicker - 60 fps
**Regression since:** [0.10](https://github.com/LostArtefacts/TRX/releases/tag/tr2-0.10)

rr-:
#### Checklist

#### Description

Resolves #2806. The OG mistakenly used the wrong arm (right instead of left) to determine the flare light position in
`Lara_GetJointAbsPosition_I`. At 60 FPS, this mistake caused Lara's hand position to appear significantly off from the expected location, which coincidentally led to the flicker effect disappearing. While fixing the bone did position Lara's hand correctly, it didn't bring back the missing flicker at 60 FPS. In fact, it caused the flicker effect to disappear at 30 FPS as well, making it static at both frame rates. It turned out that although the game does randomize the position of the flares' light source, it doesn't do so sufficiently. The primary cause of the flickering was actually the randomness introduced by using the incorrect bone. To restore the flicker effect to how it originally appeared, I had to enhance the randomness of the flare's light position.

### DIFF
--- a/docs/tr2/CHANGELOG.md
+++ b/docs/tr2/CHANGELOG.md
@@ -42,6 +42,7 @@
 - fixed pushblocks being rotated when Lara grabs them, most noticeable if asymmetric textures have been used (#2776)
 - fixed the boat briefly having an underwater hue when Lara first climbs on (#2787)
 - fixed destroyed gondolas appearing embedded in the ground after loading a save (#1612)
+- fixed flares flipped to the right when thrown (regression from 0.10)
 - fixed the camera going out of bounds in 60fps near specific invalid floor data (known as no-space) (#2764, regression from 0.10)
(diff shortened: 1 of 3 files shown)


aredfan (approved):
The flares flickering in 30/60 fps LGTM. Thank you for taking the time to fix this bug. 👍

lahm86 (approved):
Thanks 😌 
This also fixes another problem since 0.10, where the flare is flipped to the right on release. Worth adding this to the changelog too? Happy to raise a separate ticket if preferred.

image

rr-:
> Thanks 😌 This also fixes another problem since 0.10, where the flare is flipped to the right on release. Worth adding this to the changelog too? Happy to raise a separate ticket if preferred.
> 
> image

Thanks – added.

PR #2814, opened 2025-04-22
gym: calculate best time automatically

### LINKED ISSUE #2812: Lara dubs assualt time as "best" even when slower
TR2X-0.10-281-gfbc26fc-Windows.zip

Image

"Gosh, that was my best time yet" was triggered on the second finish time for me.

I completed the assault course once (Finish 1), closed the game to see if upon restart it would now retain the time (which it did), and then finished the course again, falling down and having to abort and track back along the way. Upon eventually finishing it I knew it was slower and thus surpised she would say the line. Was expecting to hear "...but perhaps I could've been faster" or similar.

lahm86:
Resolves #2812.

#### Checklist

#### Description

~~Writes the best time to the config to allow it to be restored on load. As it's unreleased, best to either add the time manually in your config file if you want to retain your current stats, otherwise delete the object from the JSON, or the file itself.~~

This removes the need to store the current best time as it is always the first completed entry's time (or -1 if no completions).

### DIFF
--- a/src/libtrx/config/file.c
+++ b/src/libtrx/config/file.c
@@ -363,8 +363,6 @@ bool ConfigFile_LoadAssaultStats(
             }
         }
     }
-    assault_stats->best_time =
-        JSON_ObjectGetInt(stats_obj, "best_time", assault_stats->best_time);
     assault_stats->total_attempts = JSON_ObjectGetInt(
         stats_obj, "total_attempts", assault_stats->total_attempts);
     return true;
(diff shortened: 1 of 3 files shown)


rr-:
Hold on – isn't it always `entries[0].time`? If that is the case, we could compute it ourselves.

lahm86:
> Hold on – isn't it always `entries[0].time`? If that is the case, we could compute it ourselves.

You're right, this could be simpler. Will tweak.

lahm86:
Updated

rr- (approved):
Thank you. Apologies for missing this in my original PR.

aredfan (approved):

PR #2826, opened 2025-04-22
tr2/shell: improve moving window around

rr-:
#### Checklist

#### Description

This PR enhances performance when moving the TR2 window by eliminating unnecessary operations. It may, or may not, also address issues related to dual head setups.

- Consolidates three separate window/display‑size APIs into a single `SHELL_SIZE` struct and four functions:
    - `Shell_GetWindowSize()`
    - `Shell_GetCurrentDisplaySize()`
    - `Shell_GetCurrentSize()` (auto‑picks display vs. windowed)
- Replaces all duplicate calls to get width/height scattered throughout tr1/tr2 with the new API.
- Enhances the viewport sync logic in `shell/common.c` :
    - introduces `m_ViewportSize` to remember last applied size, and only call `Viewport_Reset()` when dimensions truly change
    - changes `Output_ApplyLevelSettings` to only be called if relevant player's settings actually change, whereas previously it would unconditionally call `Viewport_Reset()`
    - switcheds `M_SyncFromWindow()` to take a `bool update_viewport` flag so that move events can skip unnecessary viewport resets
- Moved `Shell_IsFullscreen` definition to libtrx

#### Testing

- Check if TR1 has still desktop resolution as the first entry in the resolution list
- Move the TR2 window around, it should be slightly more performant
- Play with minimize, maximize, moving the window around, etc. No new bugs should have been introduced.
- Check the dual head setup, maybe it has improved things?

### DIFF
--- a/docs/tr2/CHANGELOG.md
+++ b/docs/tr2/CHANGELOG.md
@@ -55,6 +55,8 @@
 - fixed a crash on level load if an animation has no frames (#2746, regression from 0.8)
 - fixed a crash in custom levels with large rooms (#2749)
 - fixed flares missing the flicker effect in 60 FPS (#2806)
+- fixed the viewport not always in sync with the window (#2820)
+- improved performance when moving the window around
 - improved pause exit dialog - it can now be canceled with escape
 - removed the need to specify in the game flow levels that have no secrets (secrets will be automatically counted) (#1582)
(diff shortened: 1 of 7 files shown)


walkawayy (approved):
LGTM. Seems on par with the previous PR.

(TR2 still won't move to a secondary monitor)

aredfan (approved):

lahm86 (approved):

Issue #2820, opened 2025-04-22
Impossible to move TR2X window between screens on multi-monitor setup


Richard-L:
![Image](https://github.com/user-attachments/assets/4893594e-e1b2-4c29-8bca-f7a8823884fc)

I have two monitors:
- **Internal monitor of laptop**: 3072x1920px, 120hz (uses Win11's display scaling at 225%)
- **External monitor**, connected via Thunderbolt 4 hub and then DisplayPort, 1920x1080px, 100% scaling, 240hz.

Win11 is set to "extend" the displays. The external monitor is set as my "main display", so TR2X' window launches on it. When trying to move the window to the laptop's screen, the window jumps back to the main screen, and is severely oversized (I presume due to the DPI differences and scaling going on via Win11 between the displays).

The same occurs when using Windows shortcuts for quick app moving across screens like <kbd>WIN</kbd>+<kbd>SHIFT</kbd>+<kbd>LEFT</kbd>/<kbd>RIGHT</kbd>

Expected behaviour would be that the TR2X window assumes the "same" size and correct scaling, based on which display it finds itself on.

Ideally this would also work dynamically for fullscreen mode, even across different aspect ratio displays, and assume letterboxing/pillars where it find itself on a non-16:9 ratio.

I couldn't test whether the issue is the same for TR1X due to https://github.com/LostArtefacts/TRX/issues/2819

rr-:
>Ideally this would also work dynamically for fullscreen mode, even across different aspect ratio displays, and assume letterboxing/pillars where it find itself on a non-16:9 ratio.

Not supporting Win+Shift+Left/Right for fullscreen apps to my knowledge is a Windows limitation. My knowledge is not up to date on this matter, though (Win 10 times).

Richard-L:
I hate to use this line because it often sounds unkind, but: It works in other apps/games.

rr-:
Got it. I wonder how it's implemented for those fullscreen apps – if it's through a Windows-supplied window manager shortcut as I'd imagine it should, or if it's something that the games need to do by themselves.

Richard-L:
So after Lahm's suggestion here to set both values to 0, and TR1X' window back, I can in fact use the keyboard shortcuts to move the game from screen to screen, _both_ windowed and fullscreened 🙂 

So maybe that aspect is just a thing Win11 already does.

For TR2X, in my `TR2X.json5` I have:
```json5
"is_fullscreen": true,
  "is_maximized": false,
[...]

PR #2827, opened 2025-04-22
tr2/shell: fix support for dual monitors

rr-:
#### Checklist

#### Description

Addresses #2820.

When the window changes, the new state was saved to the config, but this triggered the config change callback, which then attempted to sync the changes back to the window. Although a debounce timer was used to reduce this effect, at least one unnecessary call still occurred. This would cause the window to snap to the primary display, as the APIs we use sanitize the window position. This PR implements a guard that ignores config change events during window events.

Additionally, there was an issue related to scaling. Previously, we used the main display size to determine the viewport state, which caused incorrect screen sizing when the game was moved to a display with a different resolution. This PR changes it to refer to the window's current display instead.

### DIFF
--- a/docs/tr2/CHANGELOG.md
+++ b/docs/tr2/CHANGELOG.md
@@ -15,6 +15,7 @@
 - added the current music track and timestamp to the savegame so they now persist on load (#2579)
 - added waterfalls to the savegame so that they now persist on load (#2686)
 - added support for aspect ratio-specific images (#1840)
+- added a guard to ensure the game always starts on a visible screen even after unplugging displays (#2819)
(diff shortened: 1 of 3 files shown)


walkawayy:
LGTM except for one minor remaining issue:

1. Move the window to the secondary monitor.
2. Fullscreen
3. Exit fullscreen

The window goes back to the primary display instead of staying on the secondary window that it exited fullscreen from.

lahm86 (approved):

aredfan (approved):

walkawayy (approved):

aredfan (approved):

PR #2860, opened 2025-04-26
tr2/objects/door: prevent Lara voiding in closed doors

### LINKED ISSUE #2848: OG bug: Lara can void near closing doors
If doors are placed on non-portal tiles and are timed to close, Lara can void if she's standing on the same tile. Doesn't occur in OG (AFAIK) but it can be replicated here for example by closing the door with the fly cheat.

Image

Cf #82.

lahm86:
Resolves #2848.

#### Checklist

#### Description

Applies the same closed-door check as TR1 for Lara when a door is not on a portal, and merges the two door modules into TRX as the overall logic is now identical.


rr- (commented):
Won't there be problems with us changing `portal_room.pit` and `.sky` to `NO_ROOM_NEG` in the context of TR1? I know we have been using that in TR2, but there are still some checks lying around that expect `NO_ROOM` there.

Also it's an `uint8_t` so the comparison with `-1` will always return 0.
Edit: no matter, it's just the assignment, which means it will promote it to 255 and the comparisons with `NO_ROOM` will continue to work.

src/libtrx/game/objects/general/door.c
```diff
@@ -171,22 +202,26 @@ static void M_Control(const int16_t item_num)
...
+            M_Shut(&data->d2);
+            M_Shut(&data->d1flip);
+            M_Shut(&data->d2flip);
         }
     }
 
+    M_Check(&data->d1);
+    M_Check(&data->d2);
+    M_Check(&data->d1flip);
+    M_Check(&data->d2flip);
```
rr-: Let's keep it as `door` IMO.

rr- (approved):
```c
        Lara_Push(
            item, coll,
            item->current_anim_state != item->goal_anim_state ? coll->enable_hit
                                                              : false,
            true);
```

In this part, can we do it like this?

```c
        Lara_Push(
            item, coll,
            coll->enable_hit && (item->current_anim_state != item->goal_anim_state,
            true);
```

lahm86:
Thanks @rr-, updated.

rr- (approved):

aredfan (approved):

PR #2878, opened 2025-04-27
objects: use common bounds and alignment checks

lahm86:
#### Checklist

#### Description

This mainly updates TR2 objects such as pickups, keyholes etc to use the same approach as TR1 for bounds testing, so we use `OBJECT_BOUNDS` on the object rather than the OG array of int16s.  `Lara_TestPosition` and `Lara_AlignPosition` functions are now also in TRX, so all usage here is common. TR2's aligning has a small difference compared with TR1, which I think would be OK to introduce there, but I'll first work out its significance and we can test that separately.

The following are affected by this change, so we'll need to check Lara's interaction with each. She should continue to snap into position when relevant e.g. if at a slight angle when using a lever, and hopefully there are no regressions otherwise.

- keyholes, puzzle slots, levers, switches, buttons, wheel doors
- pickups (regular, underwater, and special, like the ToQ scion)
- crystals
- midas touch (converting lead bars)
- pushblocks
- Bartoli's detonator
- Ice Palace gong
- zipline

### DIFF
--- a/src/libtrx/game/lara/common.c
+++ b/src/libtrx/game/lara/common.c
@@ -2,8 +2,9 @@
 
 #include "game/const.h"
 #include "game/item_actions.h"
+#include "game/lara/const.h"
 #include "game/matrix.h"
-#include "game/rooms/const.h"
+#include "game/rooms.h"
 
 void Lara_Animate(ITEM *const item)
 {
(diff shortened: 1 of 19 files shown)


rr- (approved):

lahm86:
> TR2's aligning has a small difference compared with TR1, which I think would be OK to introduce there, but I'll first work out its significance and we can test that separately.

It fixes #2879, so I'll follow this up with a separate PR to do the same in TR1.

aredfan (approved):

PR #2894, opened 2025-04-29
lara: merge common functions

lahm86:
#### Checklist

#### Description

This merges some of Lara's common functions into TRX, but also moves previous Item functions that were used exclusively by Lara  into her module as well for clarity. `Lara_MovePosition` will allow us to offer the walk to items feature in TR2 later.

I'd appreciate some checks for pickups and switches in particular (these call `Lara_MovePosition`), plus in TR1, with and without walk to items enabled. Other checks are quite general, such as Lara responding to geometry if for example mid-roll, or being nudged by walls while swimming etc.

Also, I'm happy to delay merging this until after 4.10 is out.

### DIFF
--- a/src/libtrx/game/lara/common.c
+++ b/src/libtrx/game/lara/common.c
@@ -243,3 +243,23 @@ void Lara_AlignPosition(const ITEM *const item, const XYZ_32 *const vec)
 
     lara->pos = new_pos;
 }
+
+bool Lara_IsNearItem(const XYZ_32 *const pos, const int32_t distance)
+{
+    const ITEM *const item = Lara_GetItem();
+    const XYZ_32 d = {
+        .x = pos->x - item->pos.x,
+        .y = pos->y - item->pos.y,
+        .z = pos->z - item->pos.z,
+    };
+    if (ABS(d.x) > distance || ABS(d.z) > distance || ABS(d.y) > WALL_L * 3) {
+        return false;
+    }
+
+    if (SQUARE(d.x) + SQUARE(d.z) > SQUARE(distance)) {
+        return false;
+    }
+
+    const BOUNDS_16 *const bounds = Item_GetBoundsAccurate(item);
+    return d.y >= bounds->min.y && d.y <= bounds->max.y + 100;
+}
--- a/src/libtrx/game/lara/common.c
+++ b/src/libtrx/game/lara/common.c
@@ -1,10 +1,15 @@
 #include "game/lara/common.h"
 
+#include "config.h"
 #include "game/const.h"
 #include "game/item_actions.h"
 #include "game/lara/const.h"
 #include "game/matrix.h"
 #include "game/rooms.h"
+
+#define M_MOVE_ANIM_VELOCITY 12
+#define M_MOVE_SPEED 16
+#define M_MOVE_ANGLE (2 * DEG_1) // = 364
 
 void Lara_Animate(ITEM *const item)
 {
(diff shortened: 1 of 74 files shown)


lahm86:
Just realised I was a few commits behind develop, updated now.

walkawayy (approved):
Everything LGTM in TR1 and 2. I tried TR1 with animated interactions on and off.

aredfan (approved):

src/libtrx/game/lara/common.c
```diff
@@ -243,3 +248,174 @@ void Lara_AlignPosition(const ITEM *const item, const XYZ_32 *const vec)
...
+    const int32_t x = lara_item->pos.x;
+    const int32_t y = height + lara_item->pos.y;
+    const int32_t z = lara_item->pos.z;
+
+    int16_t room_num = lara_item->room_num;
+    const SECTOR *const sector = Room_GetSector(x, y, z, &room_num);
+    lara_item->floor = Room_GetHeight(sector, x, y, z);
+    if (lara_item->room_num != room_num) {
+        const int16_t item_num = Item_GetIndex(lara_item);
+        Item_NewRoom(item_num, room_num);
```
rr-: Can we rename (in a separate commit) this function to `Item_UpdateRoom`? Thanks.

src/libtrx/game/lara/common.c
```diff
@@ -243,3 +248,174 @@ void Lara_AlignPosition(const ITEM *const item, const XYZ_32 *const vec)
...
+void Lara_ShiftCol(COLL_INFO *const coll)
+{
+    ITEM *const lara_item = Lara_GetItem();
+    lara_item->pos.x += coll->shift.x;
+    lara_item->pos.y += coll->shift.y;
+    lara_item->pos.z += coll->shift.z;
+    coll->shift.z = 0;
+    coll->shift.y = 0;
+    coll->shift.x = 0;
+}
```
rr-: Let's try to clean up her APIs a bit and move things that are used mostly inside the lara namespace to `misc.c` or something similar, and here leave only basic functions that are used publicly. Maybe `col_utils.c` for the utility functions needed by `col.c`.
I haven't checked the usages of this function `Lara_ShiftCol` – it's more of a general comment.

src/libtrx/game/lara/common.c
```diff
@@ -243,3 +248,174 @@ void Lara_AlignPosition(const ITEM *const item, const XYZ_32 *const vec)
...
+#define ITEM_ADJUST_ROT(source, target, rot)                                   \
+    do {                                                                       \
+        if ((int16_t)(target - source) > rot) {                                \
+            source += rot;                                                     \
+        } else if ((int16_t)(target - source) < -rot) {                        \
+            source -= rot;                                                     \
+        } else {                                                               \
+            source = target;                                                   \
+        }                                                                      \
+    } while (0)
```
rr-: Let's `#undef` this macro after we're done with the usage.
Or on the contrary – make this macro public. I think it's worth checking if we have more usages of this pattern (not the macro) across the code base – I think we have quite a lot.

src/tr2/decomp/decomp.c
```diff
@@ -77,54 +77,6 @@ void CutscenePlayer1_Initialise(const int16_t item_num)
...
-        src_pos->pos.x == dst_pos->pos.x &&
-        src_pos->pos.y == dst_pos->pos.y &&
-        src_pos->pos.z == dst_pos->pos.z &&
-        src_pos->rot.x == dst_pos->rot.x &&
-        src_pos->rot.y == dst_pos->rot.y &&
-        src_pos->rot.z == dst_pos->rot.z
-    );
-    // clang-format on
-}
-
```
rr-, line 127: This is such a bad function name lol. Thank you for clearing this up.

rr- (commented):
LGTM. A couple of cosmetic remarks.

src/libtrx/game/lara/common.c
lahm86: Thanks, I wasn't too sure about the line between common and misc, but this makes sense. Will update later.

src/libtrx/game/lara/common.c
lahm86: Separate commit added with this. I also removed the `if (item->room_num != room_num)` we had in so many places before calling this into the function itself as IMO it makes sense to do the check there rather than repeated throughout.

src/libtrx/game/lara/common.c
lahm86: `Lara_UpdateRoom` and `Lara_ShiftCol` are now in `misc.c` as they're only used by other Lara modules. We can perhaps see how thing pan out the more we move to TRX and see about potentially introducing other modules too.

src/libtrx/game/lara/common.c
lahm86: Made this public in `items/utils.h`.

lahm86:
@aredfan Would you mind doing a quick check again as the last commit touches on a few things? Just need to ensure items move between rooms OK mainly, a sanity check would be appreciated.

aredfan (approved):
> @aredfan Would you mind doing a quick check again as the last commit touches on a few things? Just need to ensure items move between rooms OK mainly, a sanity check would be appreciated.

No problem, I don't mind. I did another test for both games and this PR still looks good. 👍

src/libtrx/game/lara/common.c
rr-: Agreed – thank you.

src/libtrx/game/lara/common.c
```diff
@@ -1,11 +1,16 @@
...
+#include "config.h"
 #include "game/const.h"
 #include "game/item_actions.h"
 #include "game/lara/const.h"
 #include "game/matrix.h"
 #include "game/rooms.h"
 
+#define MOVE_ANIM_VELOCITY 12
+#define MOVE_SPEED 16
+#define MOVE_ANGLE (2 * DEG_1) // = 364
```
rr-: Nitpick, I'm trying to low-key introduce a convention for local defines that uses `M_…` syntax for their naming.

src/libtrx/game/objects/general/door.c
```diff
@@ -188,7 +188,7 @@ static void M_Initialise(const int16_t item_num)
         M_Shut(&door->d2flip);
 
         const int16_t prev_room = item->room_num;
-        Item_NewRoom(item_num, room_num);
+        Item_UpdateRoom(item_num, room_num);
```
rr-, line 191: I'm a bit worried about the early exit here if the rooms match. Doesn't this function rely on the side effects to happen even if the room matches?

src/libtrx/game/lara/misc.c
```diff
@@ -28,3 +28,29 @@ void Lara_GetCollisionInfo(const ITEM *const item, COLL_INFO *const coll)
         coll, item->pos.x, item->pos.y, item->pos.z, item->room_num,
         LARA_HEIGHT);
 }
+
+void Lara_UpdateRoom(const int32_t height)
```
rr-: Sorry about this back and forth but this may be now confused with `Item_UpdateRoom` signature which takes `room_num` as its integer argument, rather than height. Maybe something such as `Lara_UpdateRoomToHeight`, or `Lara_UpdateRoomWithOffset`…

src/tr1/game/objects/creatures/bacon_lara.c
```diff
@@ -80,7 +80,7 @@ static void M_Control(const int16_t item_num)
         item->rot.x = g_LaraItem->rot.x;
         item->rot.y = g_LaraItem->rot.y - DEG_180;
         item->rot.z = g_LaraItem->rot.z;
-        Item_NewRoom(item_num, g_LaraItem->room_num);
+        Item_UpdateRoom(item_num, g_LaraItem->room_num);
```
rr-, line 83: Same comment as door.c about forced side effects
(Pickups have that too but I don't think there's anything special about those)

src/tr2/decomp/skidoo.c
```diff
@@ -874,20 +874,16 @@ int32_t Skidoo_Control(void)
...
         return 0;
     }
 
     Skidoo_Animation(skidoo, collide, dead);
-    if (room_num != skidoo->room_num) {
-        Item_NewRoom(g_Lara.skidoo, room_num);
-        Item_NewRoom(g_Lara.item_num, room_num);
-    }
+    Item_UpdateRoom(g_Lara.skidoo, room_num);
+    Item_UpdateRoom(g_Lara.item_num, room_num);
```
rr-, line 886: Flagging it as a risky update too.

src/libtrx/game/lara/common.c
lahm86: Ok, thanks. Will bear in mind going forward too.

src/libtrx/game/objects/general/door.c
lahm86: `Item_Initialise` does this:

```c
ROOM *const room = Room_Get(item->room_num);
item->next_item = room->item_num;
room->item_num = item_num;
```

So an item is always linked to its room. The side effects of `Item_UpdateRoom` are purely to move these item links around; if the room number is the same, that work is unnecessary as it's already been done on initialisation, or by previous `Item_UpdateRoom` calls where the numbers have differed.

src/libtrx/game/lara/misc.c
lahm86: No problem, updated.

src/tr1/game/objects/creatures/bacon_lara.c
lahm86: Same as above, and for pickups I think when they're carried they're assigned `NO_ROOM` so the updating will always kick in in that case.

src/tr2/decomp/skidoo.c
lahm86: Lara and the skidoo always have the same room number - I don't feel this is risky, unless I'm perhaps missing something?

src/libtrx/game/objects/general/door.c
rr-: Just making sure – so there's no way this is initialized wrong, and the side effects of Item_UpdateRoom fix the links that were wrong, right? And the newly introduced `if` inside is safe to rely on data in the `item` in all of those cases.

I'm speaking from my memory but I remember some crashes when working with door.c that were easy to trigger due to some optimistic assumptions about its state. I'm just not sure if it's that. Just a double check kind of thing.

src/tr2/decomp/skidoo.c
rr-: I haven't fully followed the code, only saw there was no `if` and wanted us to pay extra attention to it if `room_num` of both items can be relied on in this case.

rr- (approved):
Thank you.

src/libtrx/game/objects/general/door.c
lahm86: Thanks. I've done some extra checks around doors, bacon Lara and the skidoo and it checks out. Interestingly, with Bacon Lara, prior to this, the code was always removing her from the current room, then re-adding her to the same room, so this is an improvement there. I've tested this extensively on a custom level too with different rooms, and it behaves as expected.

PR #2898, opened 2025-04-30
tr2/gun: prevent gun control in extra anim state

### LINKED ISSUE #2895: OG bug: Lara can draw weapons during in-game cutscenes
If you use the all items cheat at the start of Rig then draw pistols, Lara ends up with guns in her hands after the intro but is in an unarmed state. Fixing this would also be useful for TRLEs where a level has a starting animation, and Lara has not lost her weapons.

lahm86:
Resolves #2895.

#### Checklist

#### Description

Not sure if this should be optional - my feeling is not, but LMK what you think.

### DIFF
--- a/docs/tr2/CHANGELOG.md
+++ b/docs/tr2/CHANGELOG.md
@@ -8,6 +8,7 @@
 - fixed Lara being killed if she enters the void in a level that uses the `disable_floor` sequence in the game flow (#2874, regression from 0.10)
 - fixed flame emitter 23 in room 6 not being deactivated when the lever in room 1 is used (#2851)
 - fixed Lara snapping to face forwards if she has a slight angle and action is pressed after using an airlock door (#2215)
+- fixed Lara being able to equip guns and flares during in-game cutscenes (#2895)
 - fixed the game crashing on unknown sequencer events
 - improved the `/set` console command to display available options if given an unknown argument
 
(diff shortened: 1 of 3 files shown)


rr- (approved):

rr-:
>Not sure if this should be optional - my feeling is not, but LMK what you think.

Agreed.

aredfan:
I noticed that if you use a weapon hotkey during  in-game cutscenes, Lara will eventually draw the weapon once the animation is over.

Also the fix affects flares too, I just wanted to check if that's intended.

rr-:
> I noticed that if you use a weapon hotkey during in-game cutscenes, Lara will eventually draw the weapon once the animation is over.
> 
> Also the fix affects flares too, I just wanted to check if that's intended.

Good catch, I'd vote for blocking this input if there's no immediate output. 
The OG behavior is only useful to speed running and even that not really as I don't think there are any stars that rely on this.

lahm86:
Hotkeys are now blocked (well the request gun type is always reset).
Flares remain blocked too because the special animation is arbitrary, so if a builder has designed one in a specific way, having Lara's arm move to a flare pulling out could break the feel of the scene.

aredfan (approved):
Thank you @lahm86, LGTM. 👍 

(It might be worth mentioning the flares part in the changelog.)

PR #2901, opened 2025-05-01
items: move further functions to TRX

lahm86:
#### Checklist

#### Description

This moves the following to TRX:
- `Item_GetHeight` (used by pushblocks when shifting the floor up/down)
- `Item_Spawn` (used by TR1 enemy drops to spawn pickups, and when Lara is killed by a falling pushblock to create a fixed camera target)
- `Item_Test3DRange` (used by TR1 mutant bullets to check if Lara is within damage range)
- `Item_Explode` - mutants, pods, kill cheat etc

`Item_GetWaterHeight` has also been removed (TR1) as it's is not referenced anywhere in the codebase.

TR2's `Item_IsSmashable` is also replaced with an object type check - so a quick check of smashable windows and bells in Tibet would be good.

Exploding items probably warrant the most attention, as TR1 is now using the same code as TR2. The only real difference is that lighting is calculated and a resultant shade is applied to the exploded parts. I'll admit I can't actually see a difference in-game but LMK what you think. The other thing to check is that Torso's explosion looks correct as it uses slightly different speed values.

The check to not apply extra rotation on TR1's exploded parts remains. I think we could probably spend time investigating the OG crash that removing this stems from, but probably best as a separate task.

There are just a few more functions left to port, and after that I plan to look into potentially splitting TRX's items module into smaller related components.


rr-:
>The check to not apply extra rotation on TR1's exploded parts remains. I think we could probably spend time investigating the OG crash that removing this stems from, but probably best as a separate task.

To the best of my recollection, this is a very old bug fix originating from TombATI's runtime patches. It was designed to address a crash that occurred in Tihocan when the centaur statues would explode. Given our improved understanding of the codebase, the issue with the rotation pointers was likely due to null values, and the crash was probably caused by trying to dereference extra rotations that weren't there.

src/libtrx/game/items.c
```diff
@@ -569,3 +607,94 @@ bool Item_IsTriggerActive(ITEM *const item)
...
+
+    Output_CalculateLight(item->pos, item->room_num);
+
+    const ANIM_FRAME *const best_frame = Item_GetBestFrame(item);
+
+    Matrix_PushUnit();
+    g_MatrixPtr->_03 = 0;
+    g_MatrixPtr->_13 = 0;
+    g_MatrixPtr->_23 = 0;
+    g_MatrixPtr->_23 = 0;
```
rr-: Matrix_PushUnit already initializes these to 0 :)

rr- (approved):
I wonder if we should introduce a `is_smashable` flag on `OBJECT_INFO` or `ITEM_INFO` even. But that's LUA territory, so let's leave it as is.

src/libtrx/game/items.c
lahm86: Nice, thanks.

lahm86:
> > The check to not apply extra rotation on TR1's exploded parts remains. I think we could probably spend time investigating the OG crash that removing this stems from, but probably best as a separate task.
> 
> To the best of my recollection, this is a very old bug fix originating from TombATI's runtime patches. It was designed to address a crash that occurred in Tihocan when the centaur statues would explode. Given our improved understanding of the codebase, the issue with the rotation pointers was likely due to null values, and the crash was probably caused by trying to dereference extra rotations that weren't there.

That sounds like it could be the problem, thanks for the info. I think we can try removing it and doing some testing, I'll raise it separately.

aredfan (approved):

Issue #2873, opened 2025-04-27
TR2X bug: misplaced bubbles


lahm86:
Bubbles can briefly appear below Lara when swimming. Report from Lore on Discord.
https://discord.com/channels/1071085441077481544/1157995722914922507/1365806728314359830

aredfan:
Regression since: [0.10](https://github.com/LostArtefacts/TRX/releases/tag/tr2-0.10)

Steps to reproduce:

1. download the attached save and add it to your save folder [save_tr2_00.zip](https://github.com/user-attachments/files/19982390/save_tr2_00.zip)
2. `/load 1`
3. `/play 2`
4. `/load 1`
5. pay attention to the bubbles close to the top of the screen

PR #2905, opened 2025-05-01
Further items shift

### LINKED ISSUE #2904: OG bug: NG+ affects demo mode
Enemy HP is doubled in demo mode, so they don't play out as expected. Found by aredfan, fixed in TR2.

lahm86:
Resolves #2904.

#### Checklist

#### Description

This moves the remaining item functions, excluding the following:
- `Item_Control` - in TR1 this calls `Carrier_AnimateDrops` directly. I'll have a think about this and how we can merge that module with TR2's item drops.
- `Item_ClearKilled` - this is for the body bag feature in TR2. I'm not sure if we should either make this available in TR1 (although TE doesn't support the related trigger type there) or just move it anyway in the knowledge it's only for TR2. The other option is removing it altogether as AFAIK it was only in place for performance reasons, but that's perhaps too far a shift from OG.

Otherwise for testing, the related moves are for item initialisation and frame/bounds handling, so we shouldn't see any issues with how items spawn and animate.

This also fixes enemy HP in TR1's demo mode when NG+ is set (fix was in TR2 already).


aredfan (approved):

src/libtrx/game/interpolation.c
```diff
@@ -473,9 +473,3 @@ void Interpolation_Interpolate(void)
         M_InterpolateBraid(m_WorldRate, lara_item);
     }
 }
-
-void Interpolation_RememberItem(ITEM *item)
-{
-    item->interp.prev.pos = item->pos;
-    item->interp.prev.rot = item->rot;
-}
```
rr-, line 481: I hoped for a moment that not calling this in TR2 was the source of #2873, but bubbles are `EFFECT`s and here we talk `ITEM`s.

src/libtrx/game/items.c
```diff
@@ -690,3 +768,103 @@ int32_t Item_Explode(
...
+    target->prop = (b1)->prop + ((((b2)->prop - (b1)->prop) * frac) / rate);
+
+    BOUNDS_16 *const result = &m_InterpolatedBounds;
+    CALC(result, &frames[0]->bounds, &frames[1]->bounds, min.x);
+    CALC(result, &frames[0]->bounds, &frames[1]->bounds, max.x);
+    CALC(result, &frames[0]->bounds, &frames[1]->bounds, min.y);
+    CALC(result, &frames[0]->bounds, &frames[1]->bounds, max.y);
+    CALC(result, &frames[0]->bounds, &frames[1]->bounds, min.z);
+    CALC(result, &frames[0]->bounds, &frames[1]->bounds, max.z);
+#undef CALC
```
rr-, line 802: Let's add a blank line below

src/libtrx/game/items.c
```diff
@@ -690,3 +768,103 @@ int32_t Item_Explode(
...
+        return &m_NullBounds;
+    }
+
+    if (frac == 0) {
+        return &frames[0]->bounds;
+    }
+
+#define CALC(target, b1, b2, prop)                                             \
+    target->prop = (b1)->prop + ((((b2)->prop - (b1)->prop) * frac) / rate);
+
```
rr-: Let's remove this blank line

src/libtrx/game/items.c
```diff
@@ -690,3 +768,103 @@ int32_t Item_Explode(
...
+            denominator =
+                anim->frame_end + key_frame_span - second_key_frame_num;
+        }
+#else
+        const int32_t second_key_frame_num2 =
+            (cur_frame_num / key_frame_span + 1) * key_frame_span;
+        if (second_key_frame_num2 > anim->frame_end) {
+            denominator += anim->frame_end - second_key_frame_num2;
+        }
+#endif
```
rr-, line 836: It'd be nice to consolidate this. I wonder if it's possible?

rr- (approved):

src/libtrx/game/items.c
lahm86: I initially tried TR1 using TR2's code here and, while I only played around Vilcabamba for a little bit, it seemed fine. We could try it separately I think and see how we get on.

PR #2933, opened 2025-05-03
Add a help console command

### LINKED ISSUE #2917: Feature request: Add /help command
Makes the dev console a bit less dev.

I imagine it to work as follows – if the player just types `help`, it lists the available commands  in a comma-separated list, and if they type `help drain` it shows the help for the specific command. The help text should likely be provided by the commands themselves, and as such the `REGISTER_CONSOLE_COMMAND` needs to be modified to include a `GAME_STRING_ID` that we'd `GameString_Get` at runtime.

rr-:
#### Checklist

#### Description

Resolves #2917.

- Splits the regex-based aliasing system to separate registration points. Sometimes (such as in the case of the easy config commands) this means registering the same entrypoint with different prefixes and help strings. I decided to keep the `end-?level` and `next-?level` in their regex form for backwards compatibility and not to cause them to take too much space in the listing by using up 4 entries for the same command.
- Updates the registration macro to allow to specify multiply registration points inside the same file.
- Adds a `/help` command that lists the commands or displays a detailed help for the specified one.
- The easter-egg command doesn't appear in the listing or in the help as intended.

The system is simple: one command, one description. No programmatic way to explain what argument it takes, beyond what we provide in the command's description. Aliases are duplicated in the list. We could mark them somehow as duplicate (maybe with a syntax such as `[play, level]`), but IMO that belongs to a separate ticket.

#### Testing

I'd appreciate checking each command's description to confirm they all make sense.


walkawayy (commented):
Looks good except for `end-?level` and `next-?level` are displayed weirdly with the `help` command. I think my strings files are ok. Screenshots below:

image

image

rr-:
> Looks good except for `end-?level` and `next-?level` are displayed weirdly with the `help` command. I think my strings files are ok. Screenshots below:
> 
> image
> 
> image

I addressed that in the description, but I agree it's not optimal.

aredfan (approved):

rr-:
> > Looks good except for `end-?level` and `next-?level` are displayed weirdly with the `help` command. I think my strings files are ok. Screenshots below:
> > image
> > image
> 
> I addressed that in the description, but I agree it's not optimal.

Actually I just found a nice way to solve it :relieved:  I created four aliases, but only two of them get the help message, so typing all four variants will still work, but the dashed variants will not get listed in `/help`.

lahm86 (approved):

PR #2761, opened 2025-04-13
tr2/camera: determine room for in-game cutscenes

lahm86:
Resolves #2754.

#### Checklist

- [x] I have read the [coding conventions](https://github.com/LostArtefacts/TRX/blob/master/CONTRIBUTING.md#coding-conventions)
- [x] I have added a changelog entry about what my pull request accomplishes, or it is an internal change
- [x] I have added a readme entry about my new feature or OG bug fix, or it is a different change

#### Description

Standalone cutscene levels already determine the room for cinematic positions, but in-game cutscenes did not and hence camera interpolation clamping could yield bad results for wide ranging cinematics. The reason this wouldn't be an issue in a custom level using a different exe is that we perform that additional post-interpolation clamping; and additionally the OG in-game cutscenes don't venture too far from where they start, so this hasn't been an issue so far.

Worth checking Offshore Rig start, dragon dagger pull, and HSH start/finish.

PR #2943, opened 2025-05-05
tr2/room: remove cinematic room initialiser

lahm86:
#### Checklist

#### Description

`Room_InitCinematic` was used to mark all cutscene level rooms as ready to be drawn, and to also set the outside flag on each room. Since #2761, the camera's room is calculated per frame anyway, which means room drawing can calculate the visible rooms normally and there is no need to rely on the preset. ~~Injections have been created to add the outside flag to all cutscene rooms.~~ Testing shown that the outside flag is not required for the OG cutscenes.

The four cutscenes should play out as expected. Note this doesn't affect in-game cutscenes, like Rig/HSH.

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/43 (_cancelled_)


rr- (approved):
Do we understand why we need the outside flag to begin with?

lahm86:
> Do we understand why we need the outside flag to begin with?

Logically, I don't think it's needed as there are no skyboxes in the cutscenes, but `room_draw.c` uses it such as below, so I figured safer for now to keep it in place. If builders were making cutscenes, they can add the flag anyway in TE if it's needed.
https://github.com/LostArtefacts/TRX/blob/0351429d87af2c0d7b54e84654dd51a8777f4ee6/src/tr2/game/room_draw.c#L477

rr- (approved):
Thanks. In the case the cutscenes render fine without the injections, maybe we could forgo adding them?

aredfan (approved):

lahm86:
> Thanks. In the case the cutscenes render fine without the injections, maybe we could forgo adding them?

They look good to me without. @aredfan, if you omit the injections, do they look OK for you too?

aredfan (approved):
> They look good to me without. @aredfan, if you omit the injections, do they look OK for you too?

Yes, I can confirm the cutscenes still look good without the injections. 👍

lahm86:
> > They look good to me without. @aredfan, if you omit the injections, do they look OK for you too?
> 
> Yes, I can confirm the cutscenes still look good without the injections. 👍

Thank you. I've squashed out the injections, cancelled the linked PR and updated the commit message.

PR #2955, opened 2025-05-06
carrier: merge carried item logic

### LINKED ISSUE #2952: OG bug: various carried item issues
1. Every intelligent creature can pick up items, but not all of them can drop them when they're killed.
    - Small spiders
    - Fish
    - Sharks
    - Scuba divers
    - Xian guards
    - Dragon
    - Skidoo driver
2. Both eel types should be excluded from being able to pick up as they can't be killed (current OG behaviour).
3. Items carried by birds that are killed with the grenade launcher mid-flight will spawn mid-air and can't be obtained.

This can all be solved by merging TR1's carrier logic with TR2.

### LINKED ISSUE #2954: TR1X bug: animated drop items are not interpolated
Regression since 4.0.

### LINKED ISSUE #2956: TR1X bug: drop items assigned to ToQ mummy spawn early
If save/load is used, the items spawn regardless of status. Regression since 4.6.

### LINKED ISSUE #2957: TR1X bug: using kill all can crash when a Qualopec mummy with carried items is present
Regression since 4.6.

lahm86:
Resolves #2952.
Resolves #2954.
Resolves #2956.
Resolves #2957.

#### Checklist

#### Description

This moves TR1's carrier logic to TRX and updates TR2 to use the same approach. Game flow defined drops remain a TR1-only feature, as these are required for the OG setup only, and the TR2 approach we had introduced previously is already compatible with TR2 itself. All enemies in TR2 can now drop, but the two eel types remain excluded as they can't be killed. Falling items now animate in 60 FPS as well.

I'll share a test level, but checking OG in both games would be good too. Checking current TR2X saves and legacy saves would be ideal as well.

### DIFF
--- a/docs/tr1/CHANGELOG.md
+++ b/docs/tr1/CHANGELOG.md
@@ -13,6 +13,7 @@
 - fixed health bar in top center position covering inventory text
 - fixed select level dialog not reacting to the menu back key (#2918, regression from 4.9)
 - fixed carried items falling from flying enemies not animating in 60 FPS (#2954, regression from 4.0)
+- fixed items carried by the Qualopec mummy spawning early after save/load (#2956, regression from 4.6)
 
 ## [4.10.1](https://github.com/LostArtefacts/TRX/compare/tr1-4.10...tr1-4.10.1) - 2025-04-30
 - fixed water caustics appearance (#2896, regression from 4.10)
--- a/docs/tr1/CHANGELOG.md
+++ b/docs/tr1/CHANGELOG.md
@@ -14,6 +14,7 @@
 - fixed select level dialog not reacting to the menu back key (#2918, regression from 4.9)
 - fixed carried items falling from flying enemies not animating in 60 FPS (#2954, regression from 4.0)
 - fixed items carried by the Qualopec mummy spawning early after save/load (#2956, regression from 4.6)
+- fixed potential memory corruption if `/kill all` is used with a Qualopec mummy that is carrying items (#2957, regression from 4.6)
 
 ## [4.10.1](https://github.com/LostArtefacts/TRX/compare/tr1-4.10...tr1-4.10.1) - 2025-04-30
 - fixed water caustics appearance (#2896, regression from 4.10)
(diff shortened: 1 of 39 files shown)


walkawayy (approved):
Looks great!

aredfan:
I noticed that `/kill all` causes the game to crash when using the tr1 test level, it only happens if tr2 item drops are enabled.

Although not a regression introduced in this PR, I thought it's worth mentioning.

aredfan:
- [x] I found just one bug - the dragon seem to drop the pickup prematurely with these steps.

1. launch the tr2 test level
2. `/save 1`
3. `/load 1`
4. activate the dragon **edit:** nvm you can see the pickup in bartoli's body, there's no need to activate the dragon

____________________

Oops I spoke too soon. I believe I have found another bug.

- [x] The `/kill` command seem to be buggy as Lara have to get really close to the dragon for it to be effective.
- [x] The `/kill` command also destroys the black skidoo, previously it would only kill the rider.
- [x] In the tr1 test level, if you save and reload, the mummy that stares at Lara drops the pickup prematurely. This bug is also present in the latest dev snapshot.

lahm86:
Thanks, @aredfan. The `/kill` commands should now be fixed, along with the mummy dropping prematurely. Still looking into the dragon early drop problem and the crash (also mummy-related, but having no luck so far).
I have a feeling the early drop for the mummy has been a problem since we introduced the TR2 style drops in 4.6. If you have time, could you double check if that's the case (just for the changelog)?

rr- (approved):

aredfan:
> Thanks, @aredfan. The `/kill` commands should now be fixed, along with the mummy dropping prematurely. Still looking into the dragon early drop problem and the crash (also mummy-related, but having no luck so far). I have a feeling the early drop for the mummy has been a problem since we introduced the TR2 style drops in 4.6. If you have time, could you double check if that's the case (just for the changelog)?

Anytime. 👍

I can confirm the mummy's early drop bug is a regression since 4.6, just checked.

lahm86:
Thanks, @aredfan. Updated the changelog for the mummy.
The dragon early spawn should also be fixed too.
I'll look again at the mummy crash.

lahm86:
Fixed 😌 That was a well-hidden problem, explanation is in the commit message.

aredfan (approved):
> Fixed 😌 That was a well-hidden problem, explanation is in the commit message.

Well done and thank you. Everything LGTM 👍

walkawayy (approved):
Nice catches @aredfan

PR #2972, opened 2025-05-10
camera: maintain TR2 camera glide speed when set

### LINKED ISSUE #2962: OG bug: camera glide values are effectively ignored
Camera triggers with glide values are correctly parsed from floor data, but the TR2 camera resets the speed to `CHASE_SPEED` on every frame, so as a result glide values always default to 1 second. Reported by Lore.

There are many triggers in OG that reference different glide values, such as Great Wall room 0, Diving Area room 57, Barkhang room 26. We'll either need to reset these with injections to 1 second to match OG behaviour, or see how each looks/behaves with the existing unused values.

Image

lahm86:
Resolves #2962.

#### Checklist

#### Description

This fixes TR2's camera always defaulting to the default chase speed and effectively ignoring glide camera movement timers after the first frame. The config option allows retaining the OG approach of defaulting. Affected areas are listed in the readme.

One additional change (only when the option is enabled) is in the fix for Barkhang Monastery, the camera's Y position is shifted up slightly. This is to prevent a noticeable bump because of the nearby geometry, which is emphasised with the slower speed.

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/47


aredfan (approved):
LGTM. 👍 

The only issue I noticed is The Deck info is missing in the readme.

lahm86:
> LGTM. 👍
> 
> The only issue I noticed is The Deck info is missing in the readme.

Thanks, fixed.

rr- (approved):

PR #2973, opened 2025-05-10
camera: improve NO_HEIGHT fallback logic

### LINKED ISSUE #2971: TR2X bug: large camera shift in Great Wall
- `/play 1`
- `/tp switch`, use it (important the door is open)
- go outside and stop on the first tile after the door
- when the camera returns to Lara, notice how it shifts to the top of the guard tower first

Regression since TBD.

lahm86:
Resolves #2971.

#### Checklist

#### Description

This updates the `NO_HEIGHT` check from 798dc16 to first try clamping to the Y bounds of the last valid sector using the new Y value. If that fails, it reverts fully to the previous Y position. This fixes issues when a fixed camera returns to normal and there's no direct path via room portals.

While I'm not completely happy with this approach, other attempts at a fix quickly re-introduced the problems from #1034. I have merged this function into TRX though so we only have to maintain it in one place; it's identical aside from the mic-at-Lara logic in TR2.

As well as the GW issue, I'd appreciate double checking the classic camera problems in #1034. I'll also share a TR1 level which is a replica of the problematic area in GW.


rr- (approved):
It's not that bad. Thank you.

aredfan (approved):

PR #2992, opened 2025-05-16
lot: merge LOT modules into TRX

lahm86:
#### Checklist

#### Description

This merges the LOT modules into TRX, with the only differing function being the setup for different creature types. I plan to follow this up with an alternative approach where we define the LOT's step, drop, fly and mask consts in the objects themselves. This way it should decouple type ID and pathing behaviour in this module.

This also removes previous temporary assertions added in 4daef7f when Lara's TR2 sink behaviour was altered to no longer use a creature slot. I'd forgotten to tidy this part up.


rr- (approved):
Brilliant!

aredfan (approved):

PR #2993, opened 2025-05-16
lot: use object defined LOT setup

lahm86:
#### Checklist

#### Description

Objects are now responsible for picking their own LOT setup. I've made global const vars for this rather than duplicating the same setup in many places. I'm not convinced about the names though, LMK if you have any suggestions.

We'll need to check the enemy types impacted here. Those not listed use the default LOT, so if one of those is behaving OK, the rest should be too.

TR1: bat, crocodile/alligator, t-rex, flying and grounded Atlanteans, centaur, wolf, lion/lioness, panther, ape 
TR2: shark, barracuda, diver, crow, eagle, stick goon 1 and 2, yeti, small spider, skidoo driver, t-rex

### DIFF
--- a/src/libtrx/include/libtrx/game/pathing/vars.h
+++ b/src/libtrx/include/libtrx/game/pathing/vars.h
@@ -1,0 +1,10 @@
+#pragma once
+
+#include "./types.h"
+
+extern const LOT_SETUP g_LOT_Default;
+extern const LOT_SETUP g_LOT_Beast;
+extern const LOT_SETUP g_LOT_Quadruped;
+extern const LOT_SETUP g_LOT_Jumper;
+extern const LOT_SETUP g_LOT_Climber;
+extern const LOT_SETUP g_LOT_Flyer;
--- a/src/tr1/game/objects/creatures/natla.c
+++ b/src/tr1/game/objects/creatures/natla.c
@@ -86,9 +86,9 @@ static void M_Control(const int16_t item_num)
     if (item->hit_points <= 0 && item->hit_points > DONT_TARGET) {
         item->goal_anim_state = NATLA_STATE_DEATH;
     } else if (item->hit_points <= NATLA_NEAR_DEATH) {
-        natla->lot.step = STEP_L;
-        natla->lot.drop = -STEP_L;
-        natla->lot.fly = 0;
+        natla->lot.setup.step = STEP_L;
+        natla->lot.setup.drop = -STEP_L;
+        natla->lot.setup.fly = 0;
 
         AI_INFO info;
         Creature_AIInfo(item, &info);
@@ -188,9 +188,9 @@ static void M_Control(const int16_t item_num)
             break;
         }
     } else {
-        natla->lot.step = STEP_L;
-        natla->lot.drop = -STEP_L;
-        natla->lot.fly = 0;
+        natla->lot.setup.step = STEP_L;
+        natla->lot.setup.drop = -STEP_L;
+        natla->lot.setup.fly = 0;
 
         AI_INFO info;
         Creature_AIInfo(item, &info);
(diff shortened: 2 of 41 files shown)


src/libtrx/include/libtrx/game/pathing/vars.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+#include "./types.h"
+
+extern const LOT_SETUP g_DefaultLOT;
+extern const LOT_SETUP g_BeastLOT;
+extern const LOT_SETUP g_QuadrupedLOT;
+extern const LOT_SETUP g_JumperLOT;
+extern const LOT_SETUP g_ClimberLOT;
+extern const LOT_SETUP g_FlyerLOT;
```
rr-: TBH I'd rename this to `g_LOT_…`.

src/tr1/game/objects/creatures/natla.c
```diff
@@ -206,9 +206,9 @@ static void M_Control(const int16_t item_num)
             if (!(natla->flags & NATLA_FLY_MODE)) {
                 Creature_Mood(item, &info, true);
             }
-            natla->lot.step = WALL_L * 20;
-            natla->lot.drop = -WALL_L * 20;
-            natla->lot.fly = STEP_L / 8;
+            natla->lot.setup.step = WALL_L * 20;
+            natla->lot.setup.drop = -WALL_L * 20;
+            natla->lot.setup.fly = STEP_L / 8;
```
rr-, line 211: Maybe we could make these assignments into `static LOT_SETUP` variables, and assign the entire `setup` structs?

src/tr2/game/objects/creatures/xian_knight.c
```diff
@@ -136,16 +136,16 @@ static void M_Control(const int16_t item_num)
...
     AI_INFO info;
     Creature_AIInfo(item, &info);
     if (item->current_anim_state == XIAN_KNIGHT_STATE_FLY
         && info.zone_num != info.enemy_zone_num) {
-        creature->lot.step = WALL_L * 20;
-        creature->lot.drop = -WALL_L * 20;
-        creature->lot.fly = STEP_L / 4;
+        creature->lot.setup.step = WALL_L * 20;
+        creature->lot.setup.drop = -WALL_L * 20;
+        creature->lot.setup.fly = STEP_L / 4;
```
rr-, line 148: Same as Natla

rr- (commented):
Tiny nitpick about global names.
I'm not convinced about my other comment to introduce static module variables, especially since Lara looks better without such. I'll trust your judgment.

src/tr1/game/objects/creatures/natla.c
lahm86: Thanks, this would be neater but after thinking through it I think we should leave it as is. My thinking is if you look in `mutant.c`, similar happens there, but the problem is flyers have a different block mask to non-flyers but the same control is used to change the lot values. If we were to assign the entire struct during control, we'd have to be careful with the block mask too. I know that's not relevant to Natla and the Xian guards, but perhaps for consistency it's OK to have this as it is?

Or an alternative approach could be splitting out step, drop and fly into a sub struct, which would be safer to replace.

src/libtrx/include/libtrx/game/pathing/vars.h
lahm86: Thank you, renamed.

src/tr1/game/objects/creatures/natla.c
rr-: I think it's fine to leave it as is :)

rr- (approved):

aredfan (approved):

PR #3002, opened 2025-05-17
box: remove zone skipping on load

lahm86:
#### Checklist

#### Description

This removes TR2's zone skipping when reading level data based on the loaded enemy type. I believe this could have been in place for performance reasons, potentially giving a boost on PSX when these values are not needed. I've checked current develop and this PR and see no difference in loading times (`Level_ReadPathingData` in the log file).

This is my thinking.

| LOT step | Zone | Enemy |
| - | - | - |
| `STEP_L * 1` | 0 | default |
| `STEP_L * 2` | 1 | small spider, armed skidoo |
| `STEP_L * 3` | 2 | _nothing_ |
| `STEP_L * 4` | 3 | stick goons, yeti |

LOT setup assigns the above step values, and this is used in the `BOX_ZONE` macro to retrieve the relevant zone index for an enemy when doing pathing checks. So, because the old skipping checks were linked to the only enemy types that use these zones, having them in place even when the enemies themselves aren't loaded should have no impact. Other enemies will continue to use their own zone values regardless.

We can verify by for example, playing Great Wall, opening the guard tower and letting the spiders come outside. If Lara stands on a two-click block, the spiders will jump up to her, and if she's on a four-click block, they will take the "long" route by going up on the lower block first. So that's with the `STEP_L * 4` zone data loaded. A stick goon in this scenario would climb directly on the four-click block.

image

### DIFF
--- a/src/libtrx/game/inject/editors/floor_data.c
+++ b/src/libtrx/game/inject/editors/floor_data.c
@@ -318,11 +318,6 @@ static void M_FixZones(
     const int16_t box_idx = sector->box;
     for (int32_t flip_status = 0; flip_status < 2; flip_status++) {
         for (int32_t zone_idx = 0; zone_idx < MAX_ZONES; zone_idx++) {
-            if (!Box_IsUsableZoneIndex(zone_idx)) {
-                VFile_Skip(injection->fp, sizeof(int16_t));
-                continue;
-            }
-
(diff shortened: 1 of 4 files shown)


aredfan (approved):

rr- (approved):
Makes sense, thank you.

PR #3034, opened 2025-05-19
music: detect ambient trigger tracks

### LINKED ISSUE #811: Feature request: Ambience restricted to certain rooms/areas
Ability to have ambience restricted to certain rooms/areas. Like in TR3 or TR4 where it’s possible to have the ambience change mid-level. I’d like to have that opportunity in custom levels.

### LINKED ISSUE #3032: TRX bug: looped ambient plays end of track only
If you have the option enabled to restore ambient track timestamps, when the track reaches the end after reloading a save, it will then only play a short span of the very end of the track on repeat. Can test by changing a level's ambient in the GF to something short, like track 30.

lahm86:
Resolves #811.
Resolves #3032.

#### Checklist

#### Description

This adds support to have ambient tracks listed in the game flow, so that the game will know to play these tracks looped if Lara crosses a trigger for any of them. The current ambient track is saved and restored on load regardless of music loading option (but the timestamp loading still respects the player's setting). Also fixed the issue with looping tracks after reloading a timestamp.

Would be good to double check timestamp reloading with existing saves as this uprevs the savegame version.

### DIFF
--- a/src/libtrx/engine/audio_stream.c
+++ b/src/libtrx/engine/audio_stream.c
@@ -707,7 +707,6 @@ bool Audio_Stream_SeekTimestamp(const int32_t sound_id, const double timestamp)
     }
 
     AUDIO_STREAM_SOUND *const stream = &m_Streams[sound_id];
-    stream->start_at = timestamp;
     if (!stream->is_used) {
         return false;
     }
--- a/src/libtrx/game/music/common.c
+++ b/src/libtrx/game/music/common.c
@@ -2,6 +2,7 @@
 
 #include "config.h"
 #include "engine/audio.h"
+#include "game/level.h"
 #include "game/music.h"
 #include "game/music/backend_cdaudio.h"
 #include "game/music/backend_files.h"
@@ -10,9 +11,12 @@
 
 static uint16_t m_MusicTrackFlags[MAX_MUSIC_TRACKS] = {};
 static MUSIC_TRACK_ID m_TrackCurrent = MX_INACTIVE;
-static MUSIC_TRACK_ID m_TrackLastPlayed = MX_INACTIVE;
 static MUSIC_TRACK_ID m_TrackDelayed = MX_INACTIVE;
 static MUSIC_TRACK_ID m_TrackLooped = MX_INACTIVE;
+// Remember the last played track, whether normal or looped, to prevent
+// immediately restarting it if Lara remains on the same trigger.
+static MUSIC_TRACK_ID m_TrackLastPlayed = MX_INACTIVE;
+static MUSIC_TRACK_ID m_TrackLastLooped = MX_INACTIVE;
 
 static bool m_Muted = false;
 static int16_t m_MusicVolume = 0;
@@ -24,6 +28,7 @@ static void M_StopActiveStream(void);
 static void M_StreamFinished(int32_t stream_id, void *user_data);
 static bool M_IsBrokenTrack(MUSIC_TRACK_ID track);
 static int32_t M_GetRealTrack(int32_t track_id);
+static bool M_IsAmbientTrack(MUSIC_TRACK_ID track_id);
 static void M_SyncVolume(int32_t audio_stream_id);
 
 static const MUSIC_BACKEND *M_FindBackend(void)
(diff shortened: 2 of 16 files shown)


src/libtrx/engine/audio_stream.c
```diff
@@ -741,6 +742,7 @@ bool Audio_Stream_SeekTimestamp(const int32_t sound_id, const double timestamp)
     }
 
     stream->timestamp = timestamp;
+    stream->start_at = original_start;
```
rr-: Hold on… Is there a point of updating this member at all in this function?
(AFAIR `start_at` is used to support the backend where all music files are concatenated inside a single file).

src/libtrx/game/music/common.c
```diff
@@ -13,6 +14,7 @@ static MUSIC_TRACK_ID m_TrackCurrent = MX_INACTIVE;
 static MUSIC_TRACK_ID m_TrackLastPlayed = MX_INACTIVE;
 static MUSIC_TRACK_ID m_TrackDelayed = MX_INACTIVE;
 static MUSIC_TRACK_ID m_TrackLooped = MX_INACTIVE;
+static MUSIC_TRACK_ID m_TrackLastLooped = MX_INACTIVE;
```
rr-, line 19: Do I understand it right that this is used to prevent restarting the ambience from scratch every time Lara crosses a repeatable music trigger? (If so, can we provide a comment in the code?)

src/libtrx/game/music/common.c
```diff
@@ -205,13 +231,15 @@ bool Music_Play(const MUSIC_TRACK_ID track_id, const MUSIC_PLAY_MODE mode)
...
     M_SyncVolume(m_AudioStreamID);
-    Audio_Stream_SetIsLooped(m_AudioStreamID, mode == MPM_LOOPED);
+    Audio_Stream_SetIsLooped(m_AudioStreamID, is_looped);
     Audio_Stream_SetFinishCallback(m_AudioStreamID, M_StreamFinished, nullptr);
 
 finish:
     m_TrackDelayed = MX_INACTIVE;
-    if (mode == MPM_LOOPED) {
+    if (is_looped) {
+        m_TrackCurrent = MX_INACTIVE;
```
rr-, line 245: Wait, why do we set this to `MX_INACTIVE` if we actually play a track `track_id`?

src/tr1/game/savegame/savegame_bson.c
```diff
@@ -918,37 +918,52 @@ static bool M_LoadLara(
...
     }
+
+    const bool load_timestamp =
+        !is_ambient || g_Config.audio.music_load_condition == MUSIC_LOAD_ALWAYS;
+    if (load_timestamp && !Music_SeekTimestamp(timestamp)) {
+        LOG_WARNING(
+            "Could not load current track %d at timestamp %" PRId64 ".",
+            current_track, timestamp);
+    }
+
```
rr-, line 966: Complex stuff but still coherent more or less. Wish we had tests for this

src/libtrx/game/music/common.c
lahm86: This is needed because `Music_GetCurrentPlayingTrack` determines if the current track is `MX_INACTIVE` then it should return the looped track. Otherwise you could cross a normal trigger, then an ambient trigger, save and load, and the regular music would load in rather than the ambient.

src/libtrx/game/music/common.c
lahm86: Yes, exactly that. I've added a comment for this. Thanks.

src/libtrx/engine/audio_stream.c
lahm86: Good point, it seems to be redundant. Removed, and cdaudio seems to be working fine in TR2X. @aredfan could you confirm it's OK too?

src/tr1/game/savegame/savegame_bson.c
lahm86: Tests would be so good 🙏 One day.

src/libtrx/game/music/common.c
rr-: Thanks, the latter comment is pivotal. Let's include this explanation in the code :relieved:

rr- (approved):

src/libtrx/game/music/common.c
lahm86: Added 😌

src/libtrx/engine/audio_stream.c
aredfan: @lahm86 at a quick glance, cdaudio is working as expected. 👍

aredfan (approved):

PR #3035, opened 2025-05-20
wip: rewrite movable block collision

walkawayy:
#### Checklist

#### Description
WIP code. Ready for testing in TR1 test level: pushblocks-bridges-trapdoors.

Tackling in parts:
- [X] Add floor height and ceiling height functions for movable block
- [X] Blocks drop when a trapdoor below opens
- [X] Rework the `item->priv` bits into `item->data`'s `M_Priv` struct
- [x] Save `item->data`
- [X] Fix saving / loading while blocks are falling
- [X] Camera enters movable block
- [x] Investigate better performance solutions to avoid iterating all walkable items so much
- [x] Improve traversal/activation code and commonize code for multi-sector items
- [x] Drop stack of blocks when drawbridges open
- [x] Enemies don't respect the new movable blocks because of the removal of the alter floor height calls
- [x] Further improve camera behavior
- [ ] Port movable block to TRX
- [ ] TR2 support (lifts?)

### DIFF
--- a/src/libtrx/game/collision.c
+++ b/src/libtrx/game/collision.c
@@ -7,8 +7,58 @@
 #include "game/rooms.h"
 #include "utils.h"
 
+static void M_FillSide(
+    const COLL_INFO *coll, COLL_SIDE *side, int32_t x_pos, int32_t z_pos,
+    int32_t y_pos, int32_t obj_height, int16_t room_num);
 static bool M_IsOnWalkable(
     const SECTOR *sector, int32_t x, int32_t y, int32_t z, int32_t room_height);
+
+static void M_FillSide(
+    const COLL_INFO *const coll, COLL_SIDE *const side, const int32_t x_pos,
+    const int32_t z_pos, const int32_t y_pos, const int32_t obj_height,
+    int16_t room_num)
+{
+    const int32_t y = y_pos - obj_height;
+    const int32_t y_top = y - 160;
+
+    const SECTOR *const sector = Room_GetSector(x_pos, y_top, z_pos, &room_num);
+    int32_t height = Room_GetHeight(sector, x_pos, y_top, z_pos);
+    int32_t ceiling = Room_GetCeiling(sector, x_pos, y_top, z_pos);
+    const int32_t room_height = height;
+    const int32_t room_ceiling = ceiling;
+    const bool sim_wall = room_height == ceiling && room_height != NO_HEIGHT
+        && sector->ceiling.tilt == 0 && sector->floor.tilt == 0;
+    if (height != NO_HEIGHT) {
+        height -= y_pos;
+    }
+    if (ceiling != NO_HEIGHT) {
+        ceiling -= y;
+    }
+
+    side->floor = height;
+    side->ceiling = ceiling;
+    side->type = Room_GetHeightType();
+
+    const bool is_on_walkable =
+        M_IsOnWalkable(sector, x_pos, y_top, z_pos, room_height);
+    if (!is_on_walkable) {
+        if (coll->slopes_are_walls && side->type == HT_BIG_SLOPE
+            && side->floor < 0) {
+            side->floor = -32767;
+        } else if (
+            coll->slopes_are_pits && side->type == HT_BIG_SLOPE
+            && side->floor > 0) {
+            side->floor = STEP_L * 2;
+        } else if (
+            coll->lava_is_pit && side->floor > 0
+            && Room_GetPitSector(sector, x_pos, z_pos)->is_death_sector) {
+            side->floor = STEP_L * 2;
+        }
+    } else if (sim_wall) {
+        side->floor = NO_HEIGHT;
+        side->ceiling = NO_HEIGHT;
+    }
+}
 
 static bool M_IsOnWalkable(
     const SECTOR *const sector, const int32_t x, const int32_t y,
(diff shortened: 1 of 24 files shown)


lahm86:
I think this restores pathfinding blocking, but will need heavy testing. I'm not having much luck with the camera so far.

```diff
diff --git a/src/libtrx/game/objects/traps/movable_block.c b/src/libtrx/game/objects/traps/movable_block.c
index 1e9281510..1709cfaed 100644
--- a/src/libtrx/game/objects/traps/movable_block.c
+++ b/src/libtrx/game/objects/traps/movable_block.c
@@ -4,6 +4,7 @@
 #include "game/game_buf.h"
 #include "game/items.h"
 #include "game/objects/vars.h"
+#include "game/pathing.h"
 #include "vector.h"
 
 #include <stdlib.h>
@@ -98,6 +99,7 @@ void MovableBlock_Initialise(const int16_t item_num)
     data->original_rot =
         (((item->rot.y + DEG_180) / DEG_90) * DEG_90) - DEG_180;
     MovableBlock_UpdateRotation(item, data->original_rot);
+    MovableBlock_UpdateBox(item, true);
 }
 
 // TODO: make private
@@ -108,6 +110,29 @@ void MovableBlock_UpdateRotation(ITEM *const item, const int16_t rot_y)
     data->counter_rot[0] = data->original_rot - rot_y;
 }
 
+// TODO: make private
+void MovableBlock_UpdateBox(const ITEM *const item, const bool blocked)
+{
+    // Might be other cases...
+    if (blocked
+        && (item->status == IS_ACTIVE || item->status == IS_INVISIBLE
+            || (item->flags & IF_KILLED) != 0)) {
+        return;
+    }
+
+    int16_t room_num = item->room_num;
+    const SECTOR *const sector =
+        Room_GetSector(item->pos.x, item->pos.y, item->pos.z, &room_num);
+    if (sector->floor.height == item->pos.y && sector->box != NO_BOX) {
+        BOX_INFO *const box = Box_GetBox(sector->box);
+        if (blocked) {
+            box->overlap_index |= BOX_BLOCKED;
+        } else {
+            box->overlap_index &= ~BOX_BLOCKED;
+        }
+    }
+}
+
 void MovableBlock_SetupFloor(void)
 {
     if (m_UnsortedBlocks == nullptr || m_UnsortedBlocks->count == 0) {
diff --git a/src/libtrx/include/libtrx/game/objects/traps/movable_block.h b/src/libtrx/include/libtrx/game/objects/traps/movable_block.h
index d9eecdc81..10799986e 100644
--- a/src/libtrx/include/libtrx/game/objects/traps/movable_block.h
+++ b/src/libtrx/include/libtrx/game/objects/traps/movable_block.h
@@ -4,6 +4,7 @@
 
 void MovableBlock_Initialise(int16_t item_num);
 void MovableBlock_UpdateRotation(ITEM *item, int16_t rot_y);
+void MovableBlock_UpdateBox(const ITEM *item, bool blocked);
 void MovableBlock_SetupFloor(void);
 void MovableBlock_HandleFlipMap(ROOM_FLIP_STATUS flip_status);
 void MovableBlock_SetPushPull(ITEM *item, bool enable);
diff --git a/src/tr1/game/objects/traps/movable_block.c b/src/tr1/game/objects/traps/movable_block.c
index 256d65fb8..2fbfeb4ac 100644
--- a/src/tr1/game/objects/traps/movable_block.c
+++ b/src/tr1/game/objects/traps/movable_block.c
@@ -407,7 +407,9 @@ static void M_Setup(OBJECT *const obj)
 
 static void M_HandleSave(ITEM *const item, const SAVEGAME_STAGE stage)
 {
-    if (stage == SAVEGAME_STAGE_AFTER_LOAD) {
+    if (stage == SAVEGAME_STAGE_BEFORE_LOAD) {
+        MovableBlock_UpdateBox(item, false);
+    } else if (stage == SAVEGAME_STAGE_AFTER_LOAD) {
         if (item->flags & IF_KILLED) {
             const int16_t item_num = Item_GetIndex(item);
             Item_RemoveWalkable(item_num);
@@ -420,6 +422,7 @@ static void M_HandleSave(ITEM *const item, const SAVEGAME_STAGE stage)
         const bool is_push_pull = item->status == IS_ACTIVE ? true : false;
         MovableBlock_SetPushPull(item, is_push_pull);
         MovableBlock_UpdateRotation(item, item->rot.y);
+        MovableBlock_UpdateBox(item, true);
     }
 }
 
@@ -443,6 +446,7 @@ static void M_Control(const int16_t item_num)
         Item_SortWalkables();
         // TODO Fix enemy boxes.
         // Room_AlterFloorHeight(item, WALL_L);
+        MovableBlock_UpdateBox(item, false);
         return;
     }
 
@@ -512,6 +516,7 @@ static void M_Control(const int16_t item_num)
         // Room_AlterFloorHeight(item, -WALL_L);
         Item_SortWalkables();
         Room_TestTriggers(item);
+        MovableBlock_UpdateBox(item, true);
     }
 }
 
@@ -630,6 +635,7 @@ static void M_Collision(
         Item_AddActive(item_num);
         // TODO Fix enemy boxes.
         // Room_AlterFloorHeight(item, WALL_L);
+        MovableBlock_UpdateBox(item, false);
         Item_Animate(item);
         Lara_Animate(lara_item);
         MovableBlock_SetPushPull(item, true);
```

lahm86:
Some general thoughts I've had going through this:

- `item->priv` isn't saved — this might lead to issues where blocks get stuck if the game is saved while they're falling (e.g., right after a trapdoor opens but before they land). I'm wondering too if this can be simplified, by extending `M_PRIV` to hold values so we wouldn't need to use bit shifting.
- Do you think there might be a more efficient way to manage the walkable list? Right now, `Room_GetHeight` iterates over every walkable item instance, which could potentially become a performance concern on larger levels since it's such a frequently called function.
- You may have already been thinking along these lines, but I wonder if it would be cleaner to merge the current (develop) movable blocks into TRX, then look to integrate the new logic into that.

There were a few smaller code details I noticed, but it might make more sense to agree on the structural points first before going into those.

walkawayy:
> `item->priv` isn't saved — this might lead to issues where blocks get stuck if the game is saved while they're falling (e.g., right after a trapdoor opens but before they land). I'm wondering too if this can be simplified, by extending `M_PRIV` to hold values so we wouldn't need to use bit shifting.

Sorry forgot to mention I also put saving / loading on hold as a future task. I'll add that to the PR task list and also mention the issue about certain multi-sector trapdoor rotation issues.

> Do you think there might be a more efficient way to manage the walkable list? Right now, `Room_GetHeight` iterates over every walkable item instance, which could potentially become a performance concern on larger levels since it's such a frequently called function.

There most certainly is. I just wanted to get some proper playtesting and a look at the code now since people were requesting it. It's not optimized in its current form.

>  You may have already been thinking along these lines, but I wonder if it would be cleaner to merge the current (develop) movable blocks into TRX, then look to integrate the new logic into that.

Ya I can try to tackle that. I branched off a few weeks ago and didn't do any rebasing of develop yet.

> There were a few smaller code details I noticed, but it might make more sense to agree on the structural points first before going into those.

Ya exactly why I wanted to get this open before going into any more rabbit holes than I already have.

lahm86:
This one might be covered already, but just in case:
- raise trapdoor 69
- don't release either of the blocks above it
- go next door and push block 73 to the end of the now raised trapdoor
- release the two blocks above
- lower the trapdoor

The highest block remains floating.

lahm86:
The following gives better camera behaviour, as much as I dislike having to rely on `Camera_IsChunky`. Are the Lara directional tests needed? I'm guessing that would be for potentially supporting > 1 tile sized blocks.

<details><summary>Code</summary>

```c
static bool M_IsItemOnTop(
    const ITEM *const item, const int32_t x, const int32_t z)
{
    const BOUNDS_16 *const bounds = &Item_GetBestFrame(item)->bounds;
    if (bounds == nullptr) {
        return false;
    }

    int32_t dx = x - item->pos.x;
    int32_t dz = z - item->pos.z;

    const ITEM *const lara_item = Lara_GetItem();
    // Creature setups run before Lara is initialized.
    if (lara_item != nullptr && !Camera_IsChunky()) {
        DIRECTION quadrant = ((uint16_t)lara_item->rot.y + DEG_45) / DEG_90;
        switch (quadrant) {
        case DIR_NORTH:
            dx = -dx;
            dz = -dz;
            break;
        case DIR_EAST:
            int32_t t1 = dx;
            dx = dz;
            dz = -t1;
            break;
        case DIR_SOUTH:
            break;
        case DIR_WEST:
            int32_t t2 = dx;
            dx = -dz;
            dz = t2;
            break;
        default:
            break;
        }
    }

    return dx >= bounds->min.x && dx <= bounds->max.x && dz >= bounds->min.z
        && dz <= bounds->max.z;
}
```
</details>

lahm86:
Something else that comes to mind is when a block, or stack of blocks, spans from floor to ceiling. `Room_AlterFloorHeight` would do the following in that scenario, so it'd make a proper wall rather than leaving `floor == ceiling`. Currently Lara can grab as if to pull up, and it can mess with the camera.

```c
if (sector->floor.height == sky_sector->ceiling.height) {
    sector->floor.height = NO_HEIGHT;
}
```

I'm not sure where would be best placed to detect this i.e. in the movable block floor/ceiling functions (creating a dependency between them effectively) or do it once when blocks move somehow (but that's maybe getting back into the awkwardness of altering the floor height in stages).

walkawayy:
> This one might be covered already, but just in case:
> 
>     * raise trapdoor 69
> 
>     * don't release either of the blocks above it
> 
>     * go next door and push block 73 to the end of the now raised trapdoor
> 
>     * release the two blocks above
> 
>     * lower the trapdoor
> 
> 
> The highest block remains floating.

Wow this was a tricky one to solve. I finally realized the issue was the block was retaining its `fall_speed` from its previous fall. So when the trapdoor opened, the top block had a faster fall speed despite the gravity frame delay. The top block was then "landing" on the slower falling block below it. I now set `fall_speed` to 0 in `M_Control` when `gravity` is turned off.

walkawayy:
> Something else that comes to mind is when a block, or stack of blocks, spans from floor to ceiling. `Room_AlterFloorHeight` would do the following in that scenario, so it'd make a proper wall rather than leaving `floor == ceiling`. Currently Lara can grab as if to pull up, and it can mess with the camera.
> 
> ```c
> if (sector->floor.height == sky_sector->ceiling.height) {
>     sector->floor.height = NO_HEIGHT;
> }
> ```
> 
> I'm not sure where would be best placed to detect this i.e. in the movable block floor/ceiling functions (creating a dependency between them effectively) or do it once when blocks move somehow (but that's maybe getting back into the awkwardness of altering the floor height in stages).

Fixed I believe. I added this check at the end of the GetHeight functions to return `NO_HEIGHT`.

lahm86:
A couple of things I've noticed.
- If you use the fly cheat and fly upwards "into" a floating block, Lara will void.
- The camera gets thrown to `NO_HEIGHT` if you `/tp 29 -4 30` and put Lara's back against the block

These weren't happening before, perhaps it's related to e7f1fd08cc5f85d319dd7cb8d5837e3fd25482cd

The only other problem was that I noticed the blocks were silent when they fell, but I hadn't included that SFX in the level, so my bad. Zip updated again.

It's looking great overall though!

walkawayy:
Closing in favor of #3671 which is an entirely new approach of placing walkables into `SECTOR` which should have better performance.

PR #3053, opened 2025-05-24
tr1/savegame_bson: reset no room value in early saves 

### LINKED ISSUE #3052: TR1X Crash in Tomb of Tihocan after Pierre DuPont-Fight
When you finally have won against Pierre DuPont and he sinks to the ground, the game crashes.
Here are the Crashdump and Log: 

[Crashdump.zip](https://github.com/user-attachments/files/20416833/Crashdump.zip)

lahm86:
Resolves #3052.

#### Checklist

#### Description

0351429 merged the `NO_ROOM` definitions; any TR1 savegames made prior to 4.11 would have stored carried items with room numbers of 255, and no check was added with the definition change for this scenario - so this as a result was causing the crash on load as it was seeing 255 as a valid room number. This ensures the numbers get reset to -1 on load to allow the carrier module to proceed safely when the items actually drop. I've had to uprev the savegame version as this could impact saves made both in 4.10 and below and those re-saved from older versions with 4.11.

I've also tidied up some issues in the carrier module (thank you @rr- for finding these). So a general check of drops in both games would be good too here.

### DIFF
--- a/docs/tr1/CHANGELOG.md
+++ b/docs/tr1/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## [Unreleased](https://github.com/LostArtefacts/TRX/compare/tr1-4.11.1...develop) - ××××-××-××
 - changed examine item descriptions to remove extra blank lines
 - fixed examine item overlapping with other UI elements at large text scales
+- fixed a crash related to carried items if using saves made prior to 4.11 (#3052, regression from 4.11)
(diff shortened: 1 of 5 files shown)


aredfan (approved):

rr- (approved):
I think we'll need a hotfix right?

lahm86:
> I think we'll need a hotfix right?

Yeah, that sounds good.

PR #3063, opened 2025-05-25
tr2/objects/final_level_counter: fix boss proximity logic

### LINKED ISSUE #3062: OG bug: HSH always picks the same boss
The final level counter is supposed to pick the nearest boss (type 20) to Lara once all the other enemies are defeated. The logic is flawed however and as a result, item 4 is always selected.

lahm86:
Resolves #3062.

#### Checklist

#### Description

This ensures that the best boss in terms of proximity is selected in the final level.

### DIFF
--- a/docs/tr2/CHANGELOG.md
+++ b/docs/tr2/CHANGELOG.md
@@ -3,6 +3,7 @@
 - fixed a missing collapsible tile trigger in The Cold War room 82 (#3058)
 - fixed missing sound effects for collapsible tiles in Opera House and Catacombs of the Talion (#2262, #2872)
 - fixed texture and visibility issues with the skyboxes in The Cold War and Kingdom (#3056)
+- fixed the same boss item always being selected in Home Sweet Home, regardless of Lara's proximity (#3062)
 - improved word wrapping algorithm in the dev console
 
 ## [1.1](https://github.com/LostArtefacts/TRX/compare/tr2-1.0.2...tr2-1.1) - 2025-05-23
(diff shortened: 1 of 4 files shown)


rr- (approved):
WDYT of the speedrunning? Maybe let's leave a `// NOTE` if we don't want a setting for it (and I don't really want it seeing how nobody even considers TRX for speedrunning of any sort)

lahm86:
> WDYT of the speedrunning? Maybe let's leave a `// NOTE` if we don't want a setting for it (and I don't really want it seeing how nobody even considers TRX for speedrunning of any sort)

Yeah, I think it's fine as you can still manipulate the spawn. If you do the "normal" speedrun route of this level, and finish off the regular enemies in the front courtyard, item 4 still spawns. You could probably get the closer one to spawn too, so if anything it'd be an advantage. I've added a comment.

lahm86:
Oh wait, this brings another bug to light - if you save after the first boss is activated, move somewhere else and reload, another boss gets activated too. In OG, because it always picked the same one, it didn't matter. Will fix.

lahm86:
> Oh wait, this brings another bug to light - if you save after the first boss is activated, move somewhere else and reload, another boss gets activated too. In OG, because it always picked the same one, it didn't matter. Will fix.

Fixed. There was some stale logic related to this in `decomp.c` as it was assuming to be run post load. But the logic didn't make much sense as the timer before the shower cutscene (`g_FinalBossActive`) doesn't start until the boss is defeated. In OG, if you load a save after killing the boss, the timer starts again at 0 anyway.

Scenarios to check as well as proximity:
- Activate a boss, then move elsewhere near to where another boss is. Save and reload.
- Same as above, but save after killing the active boss.

aredfan (approved):

PR #3064, opened 2025-05-25
tr2/vars: remove global object related vars

lahm86:
#### Checklist

#### Description

This eliminates some global object related variables in favour of keeping scope tighter to their linked modules. The final level counter object has also been renamed, and its handling is now managed fully in the same module. The shower shotgun needs a reference to one of the variables still - I considered moving this object's control into the same module as it's so tied together but I've left that for now; I think it's OK to have an API for this bit.

Testing:
- Detonating all mines in Venice e.g. if one blows up, the rest should follow suit
- Home Sweet Home should behave as normal, so starting the level from fresh, from a save before the boss, a save with the boss alive, and another with him defeated


aredfan (approved):

src/tr2/game/objects/general/combat_end.c
```diff
@@ -10,12 +10,26 @@
...
+#define M_BOSS_TYPE O_CULT_3
+
+static const GAME_OBJECT_ID m_EnemyTypes[] = {
+    // clang-format off
+    O_DOG,
+    O_CULT_1,
+    O_WORKER_3,
+    NO_OBJECT,
+    // clang-format on
+};
```
rr-: I'm thinking, maybe this could be enhanced to test for any enemy type except `M_BOSS_TYPE`?

rr- (approved):

src/tr2/game/objects/general/combat_end.c
lahm86: Thanks, that's much neater. I've shared a quick test level for verifying (`hsh-any-enemies`) - @aredfan, if you wouldn't mind giving it a quick check over.

I think the boss type could be moved to either scripting, or potentially when we're adding killable count stats, we can do something here (i.e. those without triggers).

rr- (commented):
The proposed patch will count allies – I think we'll need g_EnemeyObjects.

lahm86:
Good catch, thanks.

rr- (approved):

aredfan (approved):

PR #3070, opened 2025-05-26
objects: rename similar object IDs

lahm86:
#### Checklist

#### Description

This renames similar objects IDs across both games. So for example, TR1 used to have only `O_FALLING_BLOCK`, and TR2 had three variants, so TR1's object is now `O_FALLING_BLOCK_1`. The others are more related to Lara's animation types.

Logically there should be no change, but worth double checking the mesh swaps in TR1's cutscenes.


rr- (approved):
Thanks. Just to make sure, did we check these extra objects that use 2 and 3 numbering that they're adequately named? I originally had taken the names out of OG nomenclature and haven't checked that.

lahm86:
> Thanks. Just to make sure, did we check these extra objects that use 2 and 3 numbering that they're adequately named? I originally had taken the names out of OG nomenclature and haven't checked that.

Yes, I feel it's all good. `O_FALLING_BLOCK_2` is not used in the base game, but `O_FALLING_BLOCK_3` is used for the wooden loose boards in Opera House, so different mesh and animations, but the same control as `O_FALLING_BLOCK_1`. The others like this are similar e.g. barrels and snowballs as aliases for boulders.

aredfan (approved):

PR #3076, opened 2025-05-28
lara: reorganise Lara structs

lahm86:
#### Checklist

#### Description

This reorganises `LARA_ARM`, `AMMO_INFO` and `LARA_INFO` to mark common properties for each, and renames some other properties and changes types to better match their purpose.

Only Lara is affected here so it'd be good to test save/load in various states, such as climbing, holding a flare, a different back gun being holstered and equipped, etc.

### DIFF
--- a/src/tr2/game/savegame/savegame_bson.c
+++ b/src/tr2/game/savegame/savegame_bson.c
@@ -1146,9 +1146,9 @@ static JSON_OBJECT *M_DumpLara(void)
         lara_obj, "grenade", M_DumpAmmo(&lara->grenade_ammo));
     JSON_ObjectAppendObject(lara_obj, "m16", M_DumpAmmo(&lara->m16_ammo));
 
-    if (lara->weapon_item != NO_ITEM) {
+    if (lara->gun_item_num != NO_ITEM) {
         JSON_OBJECT *const weapon_obj = JSON_ObjectNew();
-        const ITEM *const weapon_item = Item_Get(lara->weapon_item);
+        const ITEM *const weapon_item = Item_Get(lara->gun_item_num);
         JSON_ObjectAppendInt(
             weapon_obj, "obj_id", Object_MakeGameID(weapon_item->object_id));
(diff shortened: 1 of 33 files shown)


src/tr2/game/savegame/savegame_bson.c
```diff
@@ -1193,33 +1191,31 @@ static bool M_LoadLara(JSON_OBJECT *const lara_obj)
...
         JSON_ObjectGetInt(lara_obj, "mesh_effects", lara->mesh_effects);
     lara->water_surface_dist = JSON_ObjectGetInt(
         lara_obj, "water_surface_dist", lara->water_surface_dist);
 
-    lara->flare_control_left = JSON_ObjectGetBool(
-        lara_obj, "flare_control_left", lara->flare_control_left);
-    lara->flare_control_right = JSON_ObjectGetBool(
-        lara_obj, "flare_control_right", lara->flare_control_right);
+    lara->flare.control =
+        JSON_ObjectGetBool(lara_obj, "flare_control", lara->flare.control);
```
rr-: I noticed that we don't refer to the previous `flare_control_left` value. I don't mind removing it altogether if it has no consequences on the current saves, but if it creates problems, I'd add a fallback reading of the previous property with a note at which version we stopped supporting this.

lahm86: Good catch, thank you. That was a mistake, I didn't want to change any of the existing JSON names but this one slipped. Reverted.

rr- (approved):
Thanks. As far as I understand, the miss/hits counter still work via RESUME_INFO right? And the properties in AMMO_INFO were redundant.

lahm86:
> Thanks. As far as I understand, the miss/hits counter still work via RESUME_INFO right? And the properties in AMMO_INFO were redundant.

Yes, although they were incremented in the code, their values were never referenced. The `Stats_AddAmmoUsed` and `Stats_AddAmmoHits` take care of it all with `RESUME_INFO` and it's identical in both games 😌

aredfan (approved):

PR #3079, opened 2025-05-28
const: remove game-specific global const headers

lahm86:
#### Checklist

#### Description

The global `const.h` files are now removed from each game and all remaining relevant values moved to more appropriate modules. There are a few I wasn't quite sure where was best to place them, so LMK if you think these (or any others) have a better home. These are the ones in:

- `src/tr2/game/output.h` (some of these are unused but I think they may be linked to semi-transparency so I've left them all in place)
- `src/tr2/game/render/common.h`

An additional point is TR1's const file had a fix for using `strdup` with a VS compiler. I'm pretty sure that's no longer relevant as we don't use that function anywhere, nor do we compile in VS.

The only main point to test would be the boat's sound effects in TR2, as I've adjusted this to pitch shift in a similar way to how the skidoo does it, but it should be an identical result.


src/tr1/game/output/meshes/common.h
```diff
@@ -36,6 +36,7 @@
 #define VERT_NO_LIGHTING 0b0000'1000 // = 0x08
 #define VERT_SPRITE      0b0001'0000 // = 0x10
 // clang-format on
+#define NO_VERT_MOVE 0x2000
```
rr-: IMO this belongs to level/const.h more than it does here.

src/libtrx/include/libtrx/game/const.h
```diff
@@ -3,9 +3,13 @@
 #include "./math/const.h"
 
 #define LOGIC_FPS 30
+#define TICKS_PER_FRAME 2
 
 #define DONT_TARGET (-16384)
 
+#define PHD_ONE 0x10000
+#define PHD_HALF 0x100
```
rr-: Let's move these to `output.h` and leave a TODO to get them removed.

src/tr1/game/inventory_ring/control.c
```diff
@@ -31,6 +31,10 @@
 #define INV_RING_FADE_TIME_FAST                                                \
     (INV_RING_CLOSE_FRAMES / INV_RING_FRAMES / (double)LOGIC_FPS)
 #define INV_RING_FADE_TIME_TITLE_FINISH 0.25
+#define M_RINGSWITCH_FRAMES (96 / 2)
```
rr-: Nitpick: `M_RING_SWITCH`.
Also while we're touching this module let's consider converting other defines here to follow `M_`.

src/tr1/game/level.c
```diff
@@ -8,14 +8,14 @@
 #include "game/objects/creatures/mutant.h"
 #include "game/objects/creatures/pierre.h"
 #include "game/output.h"
+#include "game/output/meshes/common.h"
```
rr-: If possible we should avoid this include – it's bad enough that we're making level.c depend on output.c…

src/tr2/game/output.h
```diff
@@ -6,6 +6,18 @@
...
+#define SPRITE_REL 0x00000000
+#define SPRITE_ABS 0x01000000
+#define SPRITE_SEMI_TRANS 0x02000000
+#define SPRITE_SCALE 0x04000000
+#define SPRITE_SHADE 0x08000000
+#define SPRITE_TINT 0x10000000
+#define SPRITE_TRANS_HALF 0x00000000
+#define SPRITE_TRANS_ADD 0x20000000
+#define SPRITE_TRANS_SUB 0x40000000
+#define SPRITE_TRANS_QUARTER 0x60000000
```
rr-: These were borrowed from TR2Main – it's safe to delete all defines that aren't used.

src/tr2/game/render/common.h
```diff
@@ -7,6 +7,11 @@
 
 #include <stdint.h>
 
+#define SW_DETAIL_LOW (0 * WALL_L << W2V_SHIFT) // = 0
+#define SW_DETAIL_MEDIUM (3 * WALL_L << W2V_SHIFT) // = 50331648
+#define SW_DETAIL_HIGH (6 * WALL_L << W2V_SHIFT) // = 100663296
+#define SW_DETAIL_ULTRA (20 * WALL_L << W2V_SHIFT) // = 335544320
```
rr-, line 14: Let's put a note to get rid of this.

rr-:
BTW I think that lara/const.h is a bit too heavy now but we already have TODOs in place there.

lahm86:
Thanks, @rr- , should all be addressed.
I agree about Lara's consts. At least we can proceed from a single source now and filter them into better homes 😌

rr- (approved):

aredfan (approved):

PR #3082, opened 2025-05-29
inject: support embedded level injections

lahm86:
#### Checklist

#### Description

This adds support to the injection framework to accept a virtual file, which is put to use in level reading to detect when injections have been embedded at the end of regular level data.

I'm not particularly happy with `M_SkimLevel` in TR2 as it now has two purposes (initial secret loading count and to guess the layout), but equally I didn't want to duplicate the code here. WDYT? An alternative option is not to define a TRX version and just have the detection look for the additional data during `M_LoadFromFile`. Perhaps this is better as the "OG version" is technically the same.

I've uploaded a pair of test levels (`te-embedded-injections.zip`) generated from the linked TE PR. Lara should be able to traverse the vertical rooms there.

Ref: https://github.com/MontyTRC89/Tomb-Editor/pull/1038

### DIFF
--- a/src/libtrx/game/inject/common.c
+++ b/src/libtrx/game/inject/common.c
@@ -7,13 +7,12 @@
 #include "game/level.h"
 #include "game/rooms.h"
 #include "memory.h"
-#include "utils.h"
 
 #include <string.h>
 #include <zlib.h>
 
-#define INJECTION_MAGIC MKTAG('T', 'R', 'X', 'J')
-#define INJECTION_CURRENT_VERSION 2
+#define M_INJECTION_CURRENT_VERSION 2
+#define M_VIRTUAL_NAME "virtual_injection"
 
 static bool (*m_Testers[ITT_NUMBER_OF])(const INJECTION *injection) = {};
 static void (*m_Handlers[ICT_NUMBER_OF])(INJECTION_CHUNK chunk) = {};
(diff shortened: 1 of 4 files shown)


aredfan (approved):

rr- (commented):
Wdyt about a third approach where we keep the guess function as atomic and stateless, and have the secret loader be its own function? Like: guess, skim, load fully.

lahm86:
> Wdyt about a third approach where we keep the guess function as atomic and stateless, and have the secret loader be its own function? Like: guess, skim, load fully.

Sounds good, I'll try that out.

lahm86:
I've separated the guessing and skimming, and there is a bit of duplication now, although it's better I think that there is the separation. The skimming does more direct skips than trying/failing, so it assumes that layout guessing has done the work already. WDYT?

rr- (approved):
LGTM. I think this code is pretty maintainable as .phd/.tr2 will not change. (It's a different matter if we talk about shaping it to support saturn / ps1 level files, but we have too little knowledge to tackle this just yet.)

WDYT about creating a strategy pattern for various level files, similar for bson/dat savegames, that would be able to read both .phd and .tr2 files at runtime without `#ifdef`?

lahm86:
Strategies sound like a great way forward 😌 I'll make a note of this.

PR #3096, opened 2025-05-31
decomp/flares: split and refactor logic

lahm86:
#### Checklist

#### Description

This splits the original main logic for flares into the `flare_item.c` module and a new module in `lara/flare.c`. Lara's module deals only with flares in her hand and her animations/state around that; the item module deals only with thrown/dropped flares. There are a few shared functions in the item module that both use, I felt this was the best place for them.

Some minor logic/control changes:
- `gun.c` now has no dealings with flare frame numbers, instead passing control to `lara/flare.c` when appropriate
- `Flare_Create` is now `Lara_Flare_Dispose` and it will handle resetting Lara's hand mesh and control flag (the latter only if not thrown). So this allows a few less additional calls elsewhere.

For testing, general flare checks would be good to try out, but in particular checking behaviour in these cases would be great:
- Mounting a skidoo/boat with a flare lit
- Lighting one on the boat
- Throwing flares (check Lara's hand)
- Having a flare lit and drawing a weapon
- For good measure, it'd be ideal to check save/load, including legacy saves

Thanks!


rr- (approved):
Thanks!

aredfan (approved):

PR #3105, opened 2025-06-01
music: raise slot limit and remove track shifting

### LINKED ISSUE #3100: Feature request: remove music ID shifting
Remove the hard-coded music track ID shifting logic to avoid having to manually work out track numbers when building levels. Provide suitable injections for OG to fix music triggers.

### LINKED ISSUE #3101: Feature request: raise music slot limit
Currently both games are limited to 64 tracks (give or take a few that are skipped). Investigate if the limit can be raised (e.g. check savegame support re flags).

lahm86:
Resolves #3100.
Resolves #3101.

#### Checklist

#### Description

This raises the music track slot limit to 1024 for both TR1 and TR2 (the maximum possible given floor data layout).

In addition, TR2 no longer shifts music track numbers, so whichever track ID is given to it will be played directly just like TR1. OG levels have injections to change the corresponding triggers; TRLE builders can manually change their levels as necessary. There is backward compatibility in place for legacy saves (including TR2X saves made before this version) to restore music track flags (i.e.  one-shot music shouldn't play again if crossing a trigger after load) and to restore the current playing music track and position if applicable.

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/60


src/tr2/game/savegame/savegame_bson.c
```diff
@@ -366,6 +368,9 @@ static bool M_LoadMusic(
         }
     }
 
+    current_track = M_ConvertMusicTrack(current_track, header_version);
+    ambient_track = M_ConvertMusicTrack(ambient_track, header_version);
```
rr-, line 374: Let's include a comment that explains which game versions this fix is applicable for so that we can remove this once sufficient time has passed.

src/libtrx/game/music/backend_cdaudio.c
```diff
@@ -139,7 +141,7 @@ static int32_t M_Play(
 
     const int32_t track_idx = track_id - 1;
     const M_CDAUDIO_TRACK *track = &data->tracks[track_idx];
-    if (track_idx < 0 || track_idx >= MAX_MUSIC_TRACKS) {
+    if (track_idx < 0 || track_idx >= LEGACY_MAX_MUSIC_TRACKS) {
```
rr-: I think we can infer the number of track from the `cdaudio.dat` structure.

src/tr2/game/savegame/savegame_legacy.c
```diff
@@ -849,8 +849,9 @@ static bool M_LoadFromFile(MYFILE *const fp)
         Room_SetFlipSlotFlags(i, M_ReadS8() << 8);
     }
 
-    for (int32_t i = 0; i < MAX_MUSIC_TRACKS; i++) {
-        Music_SetTrackFlags(i, M_ReadU16());
+    for (int32_t i = 0; i < LEGACY_MAX_MUSIC_TRACKS; i++) {
```
rr-: After we remove the dependency on `LEGACY_MAX_MUSIC_TRACKS` in the CDAudio backend, we can make this a module local.

rr- (commented):
Good work. Thank you. Just some proposals that will let us eclipse the mapping quicker.

lahm86:
Thanks, @rr-. Updated.

src/libtrx/game/music/backend_cdaudio.c
```diff
@@ -40,7 +41,8 @@ static bool M_Parse(M_BACKEND_DATA *const data)
         return false;
     }
 
-    data->tracks = Memory_Alloc(sizeof(M_CDAUDIO_TRACK) * MAX_MUSIC_TRACKS);
+    data->num_tracks = track_content_size / sizeof(M_CDAUDIO_TRACK);
```
rr-: I don't think this is right – `cdaudio.dat` is a text file, and while it may be a good enough heuristic I'd rather not combine these units. We'll likely need to iterate through the text once, extract max track ID, allocate and re-parse.
EDIT: or we can just use a vector and grow it dynamically!

lahm86: Oops, so it is. I don't know why I thought it was flat binary 😄 Will adjust.

aredfan (approved):

src/libtrx/game/music/backend_cdaudio.c
lahm86: Updated - we use a vector while parsing, then we allocate to the backend data before continuing as before with the logic to reindex as necessary.

rr- (approved):

PR #3115, opened 2025-06-06
lara/col: begin collision routine merger

lahm86:
#### Checklist

#### Description

This gets the groundwork in place for beginning to merge Lara's collision routines in TRX.  Calling the routine is handled now in one place in TRX, allowing us to define the function pointer array in the same place. I've eliminated the "empty" routine and have merged the default one as a start.

In other commits I've removed several duplicates, such as many different states having their own function, which then just call some other function. These commits are separate for clarity.


rr- (approved):
LGTM. Eventually it'd be nice to see `Lara_Col_*` become module-level functions, though I'm not sure how feasible that strategy is when it comes to supporting more games.

lahm86:
> LGTM. Eventually it'd be nice to see `Lara_Col_*` become module-level functions, though I'm not sure how feasible that strategy is when it comes to supporting more games.

I think that should be doable with this, they're only referenced from the array as-is anyway so provided that this is configured correctly wrt the enum, we should be in a good place 😌

aredfan (approved):

PR #2128, opened 2024-12-27
tr2: fix various gameplay glitches

lahm86:
#### Checklist

- [x] I have read the [coding conventions](https://github.com/LostArtefacts/TR1X/blob/master/CONTRIBUTING.md#coding-conventions)
- [x] I have added a changelog entry about what my pull request accomplishes, or it is an internal change

#### Description

This adds options to fix QWOP, the step bug, free flares and underwater pickup drifting. I've kept the commits separated.

Attached are saves from Rig in the QWOP state, and Bartoli's on a step primed to perform the step bug (holding up should have no effect with the fix on).

[saves.zip](https://github.com/user-attachments/files/18260521/saves.zip)

I've also reorganised the config tool to put all the glitch fixes together and at the bottom of the `Fixes` tab.

PR #3125, opened 2025-06-08
lara/col: port final col routines to TRX

lahm86:
#### Checklist

#### Description

Another big one, but this leaves us with all of Lara's collision routines in TRX (aside from some stragglers in `misc.c` in both games). As the col module is now very big, I am planning another iteration to split that - rough plan as follows for the next stages.

<details><summary>Next stages</summary>

- Determine new config options to reduce the `#if/#else` necessity; for others, make them runtime checks
- Tidy up `misc.c` - collision related functions to move to TRX
- Adopt registration approach to allow splitting `col.c` into sub-modules (not one module per function, but rather a grouping of similar things I think will be best)
- Then start the same process for `state.c`
</details>

Testing for this one:
- Running
- Walking forward and backward, including off steps
- Hopping backward
- Sidestepping
- TR1 descending teleport glitch (sidestepping or walking back when on a breakable tile)
- Hanging on a ledge
- Doing a handstand
- Laddering in TR2, all directions
- TR2 step glitch (ref #2128)
- Running or walking into a wall, and then continuing to hold up and a direction; in TR1 Lara turns extremely slowly, whereas in TR2 she will act more responsively and move away from the wall with ease

There was also a regression in #3121 fixed here, which can be tested in the TR1 wading test level. It's where Lara can exit water in a non-standard way - so horizontally or to the bottom when there is a dry room below a water room.


rr- (approved):
Thanks, also good call on splitting the module into smaller files.

aredfan (approved):

PR #3128, opened 2025-06-08
lara/col: split col module

lahm86:
#### Checklist

#### Description

Splits Lara's collision module into four smaller ones, loosely grouping the routines based on environment. I think this is better, although I had to expose a couple of shared utility functions, but I think that's OK.


aredfan (approved):

src/libtrx/include/libtrx/game/lara/enum.h
```diff
@@ -111,12 +111,14 @@ typedef enum {
...
     LS_KICK          = 69,
     LS_ZIPLINE       = 70,
+    LS_NUMBER_OF     = 71,
 #elif TR_VERSION    == 1
     LS_CONTROLLED    = 56,
     LS_TWIST         = 57,
     LS_WATER_ROLL    = 58,
     LS_WADE          = 59,
     LS_RESPONSIVE    = 60,
+    LS_NUMBER_OF     = 61,
```
rr-: Maybe we could declare this without the value outside of `#endif`? Though that technique is known to be problematic if there are enum aliases present just before the `*_NUMBER_OF` declaration.

rr- (approved):
LGTM. Thanks. It looks so good now.

src/libtrx/include/libtrx/game/lara/enum.h
lahm86: Thanks, done. I think we should be OK with this enum as they're all unique and I _think_ should continue to be as we move forward (famous last words). But if we do need to introduce aliases it's an easy one to revert.

PR #3131, opened 2025-06-09
Respect enforced options in the settings UIs

rr-:
#### Checklist

#### Description

Settings flagged as enforced at the configuration layer can no longer be modified via the UI. On top of it, we append `*` to labels of frozen settings in the settings dialog to give users a clear visual cue that these options are read‑only, and elaborate on it in the option's description dialog.

Normal option:
20250609_120259_Level_0

Enforced option – no arrows and an asterisk marker:
20250609_120300_Level_0

Enforced option – description dialog:
20250609_120302_Level_0

#### Testing

This should work with every config setting in both games. As a reminder here's how we freeze settings:

```
--- a/data/tr1/ship/cfg/TR1X_gameflow.json5
+++ b/data/tr1/ship/cfg/TR1X_gameflow.json5
@@ -526,4 +526,8 @@
         {"path": "fmv/prison.avi"},
         {"path": "fmv/end.avi"},
     ],
+
+    "enforced_config": {
+        "enable_legal": false,
+    },
 }
```

#### Technical considerations

This PR introduces `Config_IsOptionEnforced`, which exposes the enforcement state for config fields. It also upgrades the string-formatter API with a ring buffer of 8 slots for `String_FormatStatic` and similar, which makes it now possible to obtain up to 8 short-lived strings at the same time rather than just 1.


aredfan (approved):

lahm86 (approved):
Great 😌

PR #3133, opened 2025-06-09
Add ability to restore defaults in the settings dialogs

rr-:
#### Checklist

#### Description

Adds a button that lets us immediately restore default settings. Only available if a row is selected and its value is different from the defaults. For compound setting (e.g. water color) resetting affects all components at once.

20250609_134229_Level_0
20250609_134155_Level_0

#### Technical considerations

I've extracted the ProgressButton element from the controls editor dialog intending to reuse it for this function. However, it performed poorly in practice, and the delay for reset was frustrating, so I opted for instant feedback instead. I did leave the ProgressButton isolated, though, in case we need it in the future (though it costs some LOC).

### DIFF
--- a/data/tr1/ship/cfg/TR1X_strings.json5
+++ b/data/tr1/ship/cfg/TR1X_strings.json5
@@ -345,6 +345,7 @@
         "ACTION_UNBIND": "Unbind",
         "ACTION_USE_ITEM": "Use",
         "COMMON_SETTINGS_FROZEN_OPTION_DISCLAIMER": "This setting is enforced by the level builder and cannot be changed.",
+        "COMMON_SETTINGS_RESTORE_DEFAULT": "Restore default",
         "COMMON_SETTINGS_TOGGLE_HELP": "Toggle help",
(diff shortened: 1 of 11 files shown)


rr-:
Not sure but it might be cool to put the backspace icon on the right side for symmetry. WDYT?

aredfan:
Very nice! 😌 

> For compound setting (e.g. water color) resetting affects all components at once.

Regarding the water colours, the `Restore default` prompt doesn't appear when I change the default water colour values to a different value.

> Not sure but it might be cool to put the backspace icon on the right side for symmetry. WDYT?

I think that having both icons to one side of the text feels tidier, it's similar to controls where `Reset` and `Unbind` both have the icon to the right.

lahm86:
Looks great. I agree with aredfan re the text/icon.

rr-:
I fixed the conditions on which the reset function becomes available and haven't touched the icon placement :relieved:

lahm86 (approved):

aredfan (approved):

PR #3158, opened 2025-06-10
lara/col: standardize collision config options

### LINKED ISSUE #3138: TR2X bug: Lara does the bump into wall animation when wading
https://github.com/LostArtefacts/TRX/commit/a643243a33637814deba7e7fd43843500d81bfd3

______________

This bug happens when Lara is a short distance from the wall before wading into it.

**Edit:** Just did another check, it appears Lara does the animation regardless of distance from the wall.

<details><summary>Dev snapshot video</summary>

https://github.com/user-attachments/assets/7b35b062-cea2-4162-97a9-be33553ecdee

</details>

<details><summary>1.1 video (matches OG behaviour)</summary>

https://github.com/user-attachments/assets/1234f1d3-4c81-45a5-b088-23ad74af56d4

</details>

### LINKED ISSUE #3148: Feature: add smooth wall deflection on collide stop
In TR2, if Lara runs into a wall and the player continues to hold forward and either left or right, she will recover and move away from the wall smoothly. In TR1, she will _eventually_ leave the wall. Make the feature optional in both games. Note that TR2's step bug is also linked to this so the fix for that will also need to become an option in TR1.

### LINKED ISSUE #3149: Feature: add one-click roll boost to TR2
In TR1, if Lara rolls off a one-click step near the edge, she gets a boost forward; in TR2 she just rolls. Make the behaviour optional in both games.

### LINKED ISSUE #3150: Feature: make swing cancel optional in TR2
Similar to TR1. One benefit is that Lara takes less damage from high drops when it's off, so it could help with challenge runs.

### LINKED ISSUE #3151: Feature: make lean jumping optional in TR2
When Lara performs a neutral jump and the player holds back or forward, she edges in that direction. Add an option for this like in TR1.

### LINKED ISSUE #3152: Feature: add breakable tile glitch option to TR2
In TR1, Lara can step sideways or backwards off a breaking tile or opening trapdoor to descend to the floor instantly. Add the same option to TR2. One benefit I can think of would be in no-damage runs, to allow getting past the start of Wreck.

### LINKED ISSUE #3153: Feature: add wall bug fix option to TR2
Change TR1's wall bug fix to become an enum of: fixed, TR1 mode or TR2 mode.

### LINKED ISSUE #3154: Feature: allow non-standard water exit
In TR1 Lara can exit water horizontally or into a dry room below; in TR2, she can't.  Additionally, in TR1 she can exit onto a non-standable slope, but can't in TR2.
Make this optional in both games under one setting.

lahm86:
#### Checklist

#### Description

All collision-related config options (excluding wading) are now common between both games. I've spent quite a bit of time testing OG glitch behaviour and these continue to behave as expected. The only difference is that with the wall bug disabled in TR2, QWOP setups tend to fail (but the QWOP itself still works fine). It's just related to how Lara hits the wall if that's part of the setup. This is akin with QWOPs being harder to achieve in TR1, so it balances out IMO. The options are there to pick and mix, which I think is a nice feature.

A few other comments:
- With the wall bug fix on in TR1, Lara deflects the wall in the same way as she would in TR2 (a softer deflection rather than a splat)
- With smooth wall deflection on in TR1, TR2 step bugs become possible, hence the inclusion of that option in TR1 now
- The water exit fix touches on the ability to leave rooms such as 40 in Folly, or 69 in Rig (flip on) as well as the ability to climb out onto a non-standable slope

Other things to test are quite straight-forward per linked issues. Any suggestions about config text/layout/defaults are welcome of course. Thanks!


rr- (approved):
We need to notify Eycore!

aredfan (approved):

PR #3159, opened 2025-06-11
Support translated UI text

rr-:
#### Checklist

* [x] I have read the [[coding conventions](https://github.com/LostArtefacts/TRX/blob/master/CONTRIBUTING.md#coding-conventions)](https://github.com/LostArtefacts/TRX/blob/master/CONTRIBUTING.md#coding-conventions)
* [x] I have added a changelog entry summarizing the changes or marked as internal.
* [x] I have updated the README entry if adding a feature or fixing an OG bug.

#### Description

This PR introduces a language selection setting accessible from the UI tab in the graphical settings dialog, allowing users to switch the game's interface language dynamically.

#### Testing

We're only testing the mechanisms and not the actual translations here – these will be delivered in separate follow-up PRs.

**Recommended testing scenarios include**:

* Switching between languages and verifying that no hardcoded English strings remain.
* Robustness checks, including attempts to induce crashes by:

  * Removing translation files.
  * Injecting invalid or malformed JSON data.
  * Setting nonsensical or unsupported values in the user configuration file, testing both initial launch and subsequent restarts.

Currently, only TR1 has example translated content implemented, although the underlying infrastructure also supports TR2.

#### Technical considerations

The translation system is structured around two main string files:

* **Base strings file** (original content).
* **Mod strings file** (level names and optional overrides).

This PR enhances string management by introducing localized variations, using suffixes in filenames (e.g., `TR1X_strings_ub-de.json5`). Selecting a language (e.g., German `de`) for mod strings automatically loads corresponding base strings but currently does **not** revert to English language files if certain strings are absent. While it should load them from the `GS_DEFINE` hardcoded English strings, it prevents region-specific translations such as `en-us` or `en-gb`. This is something I consider to be a bug.

**Additional aspects**:

* **Translation tooling**:

  * Added a linter to detect missing translation entries.
  * Introduced the `fill_translations` script to generate empty translation placeholders or leverage OpenAI for preliminary translation drafts.

* **Unicode support**:

  * Enhanced `String_ToUpper` function to handle Unicode correctly. It works by using a character-case conversion map .def file, generated by leveraging Python's Unicode facilities and the `mapping.txt` glyph tables to limit the search space to consider only supported characters.

* **Savegame handling**:

  * Savegame files now reference current game flow tables instead of embedded level titles (optional feature, gated behind a compile-time `#if` directive).

* **Review system**:

  * Every time a string is changed, it will be emptied by `update_gameflow`, and can be re-translated by `fill_translations`.
[...]

PR #3163, opened 2025-06-11
savegame: use translated level titles

rr-:
#### Checklist

#### Description

Isolated from #3159, this change is somewhat controversial, and I'm really conflicted about whether we should proceed with it as it might restrict our future options.

The primary objective is to allow the language selected by the user to dictate the level titles in the load dialog. This means that titles would appear in the chosen language, instead of the language used when the save was created. While this works well in the linked PR, it presents a few issues:

1) If a user attempts to load a save from a custom level, such as "Midas Legacy", the current develop will display "Midas Legacy" as expected. However, with this PR, the level would be renamed to a default equivalent, like "City of Vilcabamba", as it only considers the level number regardless of user intent.

2) Should we decide to support save renaming in the future (e.g., using a command like `/save 1 [description]` or perhaps through the UI), this approach would no longer be feasible.

I'm uncertain about the best course of action.


lahm86 (commented):
Code LGTM.

I think I'd be more inclined to leaving the savegames as truth though, certainly if it would mean limiting ourselves for future extensions as you say. Perhaps we just put a note somewhere that save titles will be in the language chosen at the time.

I think saves showing with "wrong" names if you have moved savegame files around can act as a reminder that these won't load normally. If the name looked normal, it would be confusing why they wouldn't load.

rr-:
Thanks. We'll close this and I'll explain it in the readme.

PR #3168, opened 2025-06-11
creature: add option to toggle enemy rotation

### LINKED ISSUE #3129: Feature request: don't rotate enemies on initialisation
Enemies are rotated randomly when they are initialised. Builders may wish to prevent this and have the angle set in the level data maintained e.g. when an enemy is visible, perhaps as a warning, before spawning.

lahm86:
Resolves #3129.

#### Checklist

#### Description

This adds an option to allow toggling whether or not enemies are randomly rotated on initialise. This affects every enemy type in TR1, excluding statues; in TR2 it only affects crows, eagles, the HSH/Opera boss, and wolves and bears in GM.

There is a bit of noise here unfortunately to ensure demo consistency. We can't perform the usual caching of the setting's value as by that time `Creature_Initialise` has already run. LMK if you think there's any potential improvements to this approach.


rr- (commented):
WDYT about exposing `M_PrepareConfig` and `M_RestoreConfig` from `src/tr1/game/demo.c` to `M_HandlePlayLevel` from `src/tr1/game/game_flow/sequencer_events.c`?

Also with the new settings, I think we might need to introduce this concept to TR2.

aredfan (approved):

lahm86:
> WDYT about exposing `M_PrepareConfig` and `M_RestoreConfig` from `src/tr1/game/demo.c` to `M_HandlePlayLevel` from `src/tr1/game/game_flow/sequencer_events.c`?
> 
> Also with the new settings, I think we might need to introduce this concept to TR2.

I think this sounds good. I think we could also merge `demo.c` at this point, it would just need some additional dependent functions marked as `extern` and a few other tweaks. For now, I have reduced the noise by removing the need to call `Creature_Reset` so the rotation check tests the level type directly now as a stopgap. This is similar to how enemy HP is handled in NG+ demo mode.

I will follow this up with a demo merger, which can include the exposed config functions. Probably wiser to land that after fixing #3172 as well.

rr- (approved):

PR #3176, opened 2025-06-12
objects/names: add fallback to built-in names

rr-:
#### Checklist

#### Description

Isolated from #3159. 
The idea is that if a user modifies the strings JSON to contain changed values, we can still use `/give medi` or `/tp wolf`. Won't work with `/play caves`, though, as that is not supplied anywhere other than the strings JSON.

I'm unsure if this change warrants a changelog entry, as it's rather esoteric.

### DIFF
--- a/src/libtrx/game/objects/names.c
+++ b/src/libtrx/game/objects/names.c
@@ -11,19 +11,26 @@
 #include <string.h>
 
 typedef struct {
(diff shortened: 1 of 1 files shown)


aredfan (approved):

lahm86 (approved):
LGTM, I think no changelog is fine for this.

PR #3191, opened 2025-06-15
Improved handling of broken language files

rr-:
#### Checklist

#### Description

Setting the language to something other than English, then removing the relevant file would cause the player to be unable to change the language at all. Now the game will let the player exit of such state by switching back to English.

Also in the event the strings file was inaccessible, the game would exit with a fatal error. This error is not fatal, as we'll just show the builtin English strings in such events, so I changed this to be just a log message.

### DIFF
--- a/src/libtrx/game/game_string_table/common.c
+++ b/src/libtrx/game/game_string_table/common.c
@@ -124,14 +124,16 @@ void GameStringTable_Shutdown(void)
     }
 }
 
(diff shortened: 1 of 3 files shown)


lahm86 (approved):
LGTM.

Btw, I've just noticed TR2 defaults to displaying en (on develop too):

image

aredfan (approved):

PR #3199, opened 2025-06-15
TR1 French translation

JoeyQuint:
#### Checklist

#### Description

This takes the translation attempt from #3159 for TR1, and overhauls it:
- Level and object names match the official translation (with the exception of Lara's Home, which matches TRX)
- A few names were gathered from 2 French official fansites: captain-alban.com & tombraidercie.com
- Many additional strings now have more natural translation - even entirely reworded when necessary (but keeping the original idea)

I started from #3159 instead of develop, and didn't notice at the time there was supposed to be a TR2 translation as well, so it is not in this PR.


rr-:
@Joey79100 Lots of strings are duplicated between TR1 and TR2 and I realized how big of a pain is to check which are the same and which are not. Currently working on extracting common strings into a base file that's repurposed between TR1 and TR2.
In other words: prior to taking on TR2 I advise to wait just a bit.

rr-:
@Joey79100 I'll work on getting your translation merged today :relieved: then we can work on TR2, which should be a lot less work.

rr-:
@Joey79100 I updated the PR so that it uses the new "common strings" file to deal with deduplication, and also provided preliminary translations for TR2 (marked with `\{review}`). I'd appreciate if you could take a look. Thanks :relieved:

rr- (approved):

PR #3209, opened 2025-06-16
Include timestamps in the logs

rr-:
#### Checklist

#### Description

It's high time we did this.

### DIFF
--- a/docs/tr1/CHANGELOG.md
+++ b/docs/tr1/CHANGELOG.md
@@ -45,6 +45,7 @@
 - fixed broken playback of mono music tracks (regression from 2.0)
 - fixed hot-plugging certain audio devices causing glitchy playback (partial fix; regression from 2.0)
 - fixed being unable to toggle fullscreen mode during FMV sequences (#3188, regression from 4.6)
+- fixed logging not outputting anything on Windows terminals
 - improved the teleport cheat if used when Lara is in a special animation, such as grabbing the Scion
 
 ## [4.11.2](https://github.com/LostArtefacts/TRX/compare/tr1-4.11.1...tr1-4.11.2) - 2025-05-24
--- a/docs/tr1/CHANGELOG.md
+++ b/docs/tr1/CHANGELOG.md
@@ -26,6 +26,7 @@
 - changed the music track slot limit from 64 to 1024 (#3101)
 - changed text kerning to a smaller value
 - changed the underwater music volume setting to separate ambient and music volume sliders
+- changed logs format to include timestamps
 - fixed a game crash in custom levels if centaur statues exploded without having centaur objects in the level file (#3155)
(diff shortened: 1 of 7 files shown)


aredfan (approved):

lahm86 (approved):
WDYT about including the logging level as well? E.g. as follows. Could be useful to see errors that may be missed or debugs that should perhaps not be in production, etc.

```
INF | 2025-06-16 08:57:50.890 [foo] bar
DBG | 2025-06-16 08:57:50.928 [foo] bar
ERR | 2025-06-16 08:57:50.928 [foo] bar
```

rr-:
@lahm86 added + added colorization of the terminal output. On Linux, it's working; on Wine, the Windows version is working as well, but I don't trust it really – can we check how the game behaves on actual Windows in an actual terminal? (Eg we open cmd.exe or powershell, drag'n'drop the game .exe there and press enter)

lahm86:
Looks great, thanks!

Launching from cmd exits immediately so nothing is showing there unfortunately. I think it's likely a can of worms with it being a GUI application rather than a console type.

image

rr-:
> Looks great, thanks!
> 
> Launching from cmd exits immediately so nothing is showing there unfortunately. I think it's likely a can of worms with it being a GUI application rather than a console type.
> 
> image

If it works the same on develop, then I think we can leave it as is. Here's how it looks on Linux:

20250616_124207_zbb

lahm86:
Yeah, develop is the same, so it's fine IMO.

This seems to do the trick btw, but I'm fine with not including such Windows-specific calls where we can avoid it.

```diff
diff --git a/src/libtrx/game/shell/main.c b/src/libtrx/game/shell/main.c
index 54a5a0653..b6c5fc1d6 100644
--- a/src/libtrx/game/shell/main.c
+++ b/src/libtrx/game/shell/main.c
@@ -5,6 +5,10 @@
 #include "utils.h"
 
 #include <string.h>
+#if defined(_WIN32)
+    #include <windows.h>
+    #include <stdio.h>
+#endif
 
 int main(int argc, char *argv[])
 {
@@ -12,6 +16,15 @@ int main(int argc, char *argv[])
         return 0;
     }
 
+#if defined(_WIN32)
+    if (AttachConsole(ATTACH_PARENT_PROCESS)) {
+        FILE *fp;
+        freopen_s(&fp, "CONOUT$", "w", stdout);
+        freopen_s(&fp, "CONOUT$", "w", stderr);
+        freopen_s(&fp, "CONIN$", "r", stdin);
+    }
+#endif
+
     char *log_path = File_GetFullPath(PROJECT_NAME ".log");
     Log_Init(log_path);
     Memory_FreePointer(&log_path);
```

image

rr-:
We have a good place for it in Log_Init_Extra.

rr-:
Added!

lahm86 (approved):
Nice! Thank you.

PR #3217, opened 2025-06-16
Fix compass description and obsolete translation

### LINKED ISSUE #3216: Target change option has an outdated description
The description for the target change option (both in-game and in the config tool) still says the Look key is used to change targets, which is no longer correct since a specific Change Target key was added in #1519.

Image

rr-:
#### Checklist

#### Description

Fixes #3216 and the wrong compass name in TR2.


Kappa971:
When I tried to correct the string displayed in the compass, the same string was then displayed in the Gameplay menu.

In TR2X this was missing:
```json
"gameplay": {
            "name": "Gameplay"
 },
```

but adding it doesn't solve the problem so it probably needs to be added in the code?

aredfan:
In TR2X, the PDA object is labelled `Statistics` instead of `Gameplay`.

20250616_115500_Level_0

</details>

rr-:
Thanks. Fixed. We had wrongfully aliased the stopwatch to the PDA.

aredfan (approved):

lahm86 (approved):

PR #3229, opened 2025-06-16
Fix objects live reload

rr-:
#### Checklist

#### Description

Fixes issuing `/strings` clearing the object names from the current level.
Also output in the logs which layers we apply and in what order. This might be adding pointless noise – LMK what you think.


lahm86 (approved):
LGTM. I think the logging is fine, certainly useful for debugging.

PR #3247, opened 2025-06-17
cmd/kill: improve kill all behavior in HSH

rr-:
#### Checklist

#### Description

Running the command `/kill all` will now kill only non-boss enemies. Running it again after the boss gets activated will kill the boss enemies. It will log that it has 5 enemies - one for each possible boss spawn - which could be mitigated but I didn't want to dive in too much into the item flags necessary for this to work.

Furthermore, it will no longer cause a soft lock. To accomplish this, I had to refactor the combat end object to no longer depend on the kill count statistics, which are not updated when the player kills enemies with cheats.


aredfan (approved):
Excellent to see this added, thank you. 👍

lahm86 (approved):

PR #3258, opened 2025-06-19
photo-mode: allow advancing single frames

rr-:
#### Checklist

#### Description

https://github.com/user-attachments/assets/f7536adf-4167-450f-a538-0bb1eee095c6

It's likely possible to do weird stuff with this, but as far as I know, one has to _really_ try.


rr-:
One behavior that's easy to trigger is to have Lara die in this mode – it asks the player to load the game twice.

EDIT: fixed

aredfan (approved):

aredfan:
I found a new issue - the lightning in Folly/Atlantis remains static during frame advancing.

rr-:
Thanks. Fortunately the fix was simple.

aredfan (approved):
> Thanks. Fortunately the fix was simple.

Thank you. 😌

src/libtrx/game/phase/executor.c
```diff
@@ -15,6 +15,7 @@
 #include "game/shell.h"
 #include "game/ui.h"
 #include "gfx/gl/track.h"
+#include "vector.h"
```
lahm86: Looks like this isn't needed.

lahm86 (commented):
Looks great 😌 
One tiny thing is that you can get audio drift in cutscenes in TR1 if you play around with it for a noticeable amount of time; TR2 has `M_FixAudioDrift` already so it resolves itself. But I don't think it's a big issue, when we look at more merging of cutscene logic we could address it then.

src/libtrx/game/phase/executor.c
rr-: Thanks. Let's work on the audio drift in a separate ticket.

PR #3264, opened 2025-06-19
stats: fix non-0 indexed secrets in TR1

### LINKED ISSUE #3262: TR1X bug: non-zero-based secret indexing no longer works
The stats total count calculator caters for non-zero-based indexing, but registering secrets doesn't. Perhaps we need to store the mask in `RESUME_INFO` that's calculated in `Stats_CalculateStats`, and then refer to this rather than the count when registering a secret?
Regression since 4.12.

rr-:
#### Checklist

#### Description

Resolves #3262.

#### Testing

I found it easier to merge TR1 and TR2 secrets management code for this fix for the most part – so to be on the safe side, I recommend testing these routes:

- counting secrets in both tr1 and tr2, and the test level included by Lahm (non-zero-secrets.zip)
- counting total secrets from level to level in all 3 cases
- final stats screen in all ~~3~~ 2 cases 
- saving the game, loading it, then checking both level stats and the final stats
- `/secret` (in the non-zero-secrets.zip you'll have to supply relevant numbers, not 1-4. I'm not sure if this is a feature or an antifeature. LMK your thoughts.)


aredfan:
I found a few issues:

- [x] TR1 test level: after I collect all 4 secrets and then use `/secret`, the console indicate that only secrets 2 and 3 are collected
- [x] TR2: `/secret give x` seem to have no effect
- [x] TR2: picking up a secret doesn't register in the stopwatch, end-of-level stats or final stats. If I save after picking up a secret, then load the SG, the secret will register in the final stats only.

rr-:
Thanks @aredfan fixed the problems. I also updated the secret command to produce more useful information.

aredfan (approved):

lahm86 (approved):

PR #3278, opened 2025-06-21
Add new \{input} dynamic glyphs

rr-:
#### Checklist

#### Description

This PR introduces a new class of icons for input keys, `\\{input …}`, where the argument is consistent with the `g_Input.` property names. It also updates the photo mode help dialog to use these new icons. As a result, it should always show the right input icons, even if the player changes their layout to something else.

image

Controller is not supported at this time, but I plan to tackle this shortly after this PR.


Richard-L:
Looks very nice.

Are the : and [ ] needed in some way? Was thinking if you could even lose them and just keep the /.

lahm86 (approved):
Nice 😌 
@Richard-L The `[]` indicates the different FOV directions, but I agree with dropping the `:` - I think that would look clean.

rr-:
Thanks. Removed the colons.

lahm86 (approved):

aredfan (approved):

PR #3285, opened 2025-06-21
Retain setting dialogs positions

### LINKED ISSUE #3271: Retain settings menu position on re-entry
When a user re-enters a settings menu, the focus should automatically return to the last edited option.

For example, if anisotropic filtering was changed, the focus should be on that option when the menu is reopened.

rr-:
#### Checklist

#### Description

Resolves #3271.

The positions are reset when exiting to the title screen which is intentional.


src/tr2/game/shell/common.c
```diff
@@ -615,6 +615,7 @@ void Shell_Shutdown(void)
     Render_Shutdown();
     Console_Shutdown();
     Overlay_Shutdown();
+    Option_Shutdown();
```
lahm86, line 617: I get a compiler warning here
```
../../../src/tr2/game/shell/common.c:618:5: warning: implicit declaration of function ‘Option_Shutdown’; did you mean ‘Music_Shutdown’? [-Wimplicit-function-declaration]
```

lahm86 (approved):
Very nice too 😌

src/tr2/game/shell/common.c
rr-: Thanks – fixed.

lahm86 (approved):

Richard-L:
Absolutely lovely.

Two thoughts:
- I can't get it to work in the key mapping settings. I wondered if it's needed there too. I thought maybe if someone is actively testing out mappings as they're playing, it could be nice to have.
- Potentially a little overkill: Using the new Step Left/Right tab navigation, I notice this resets the focus per tab. I wondered if to retain the position per-tab would be interesting.

rr-:
> * I can't get it to work in the key mapping settings. I wondered if it's needed there too. I thought maybe if someone is actively testing out mappings as they're playing, it could be nice to have.

Oh yeah definitely, let's do it there too.

> * Potentially a little overkill: Using the new Step Left/Right tab navigation, I notice this resets the focus per tab. I wondered if to retain the position per-tab would be interesting.

Right – the controls editor already keeps the focused row, so it'll only increase consistency.

walkawayy (approved):

rr-:
> Absolutely lovely.
> 
> Two thoughts:
> 
> * I can't get it to work in the key mapping settings. I wondered if it's needed there too. I thought maybe if someone is actively testing out mappings as they're playing, it could be nice to have.
> * Potentially a little overkill: Using the new Step Left/Right tab navigation, I notice this resets the focus per tab. I wondered if to retain the position per-tab would be interesting.

I applied the sidesteps fix so that it's consistent with the controls dialog and keeps the selected row between tabs.

I also updated the controls dialog to retain its state as well (barring the backend dialog, which keeps popping up to not mess with muscle memory).

I then made an additional change to reset the UI state when starting a new level, as it felt a bit weird to begin a new level and see the selection from the title screen. Might be just me.

Please LMK how this plays and feels. @lahm86 @walkawayy @aredfan

Richard-L:
Very very nice 👍 

Just to clarify: I had meant earlier to have it remember the focus per-tab, not across them, since they're not always the same length. Sometimes you may side-step to the next tab which you haven't edited yet, but then won't start editing from the top as you might expect.

But this is a little pedantic.

rr-:
> Very very nice 👍
> 
> Just to clarify: I had meant earlier to have it remember the focus per-tab, not across them, since they're not always the same length. Sometimes you may side-step to the next tab which you haven't edited yet, but then won't start editing from the top as you might expect.
> 
> But this is a little pedantic.

That makes sense; let's see what the rest of the team thinks. (Personally I like the current behavior)

lahm86:
All good as-is IMO. Agree with resetting once you get into a level👍

aredfan (approved):
> Please LMK how this plays and feels.

I think the PR build works well.  The decision to go with focused rows makes sense to me.

aredfan:
I noticed a `dmp` file appearing in the directory for both game.

1. start up the game
2. select the PDA and highlight cutscenes
4. `/exit`

______________

1. start up the game
2. start Lara's Home
3. open the inventory
4. select the PDA and highlight cutscenes
6. `/exit`

rr-:
Thanks – good catch. Fixed this as well.

aredfan (approved):

PR #3292, opened 2025-06-21
Dim unavailable settings and bindings

### LINKED ISSUE #3166: Greying out UI options if they are invalidated
With the config tool being migrated in-game, the menus have naturally become very big. They would benefit greatly from simplification and some UX improvements.

For this issue I'd like to suggest making options greyed out if they are invalidated by another option. For example:

Image

If the green option is set to "On", the two red options have no effect no matter what you set. It would help the user understand the relation between these settings more quickly if they were greyed out, and skipped when browsing through them in this state.

rr-:
#### Checklist

#### Description

This PR adjusts the settings that are currently inactive due to other configurations so that they appear dimmed. I integrated the features suggested by Posix without directly navigating the dialogs myself, except for maintaining the persistent damage and Lara's starting health options as always enabled. I'd appreciate a review if there are any additional settings that might benefit from a similar approach.

Please note that the unavailable settings are still adjustable – it's an intentional design.

Resolves #3166


Richard-L:
It's excellent imo ❤️ 

One quick observation: Is it possible the Lara's Braid + Breeze relation is in reverse? To my thinking, turning the braid off would also dim the Breeze option below.

rr-:
> It's excellent imo ❤️
> 
> One quick observation: Is it possible the Lara's Braid + Breeze relation is in reverse? To my thinking, turning the braid off would also dim the Breeze option below.

Good catch, thank you! Should be fixed now.
(I initially coded the logic in reverse, which was really confusing to read, and forgot to adjust this condition.)

lahm86 (commented):
Looks great. There are just a couple of other combinations that comes to mind for including.
- `Fix step glitch` is only relevant if `Smooth wall deflection` is switched on.
- `Fix wading wall hit` is only relevant if `Wading` is switched on (TR1 only)

rr-:
Thanks @lahm86 – updated.

lahm86 (approved):

aredfan (approved):

PR #3293, opened 2025-06-21
Use controller icons in the UIs

rr-:
#### Checklist

#### Description

This PR updates the configuration to remember the player's preferred input method, whether it is a keyboard or a controller. Additionally, it modifies the user interface to display icons relevant to the player's choice, rather than assuming the default use of the keyboard.

image


lahm86 (approved):

aredfan:
LGTM.

Not sure if it's an issue but I noticed this in the photo mode UI - both `Toggle help` and `Exit` have <kbd>?</kbd> input displayed if I use the controller icons.

rr-:
This is because we don't have inputs assigned for these particular keys. I can choose another character and remove the ?, part from the exit row. I'm not sure what character would be the best, though.

aredfan:
> This is because we don't have inputs assigned for these particular keys. I can choose another character and remove the ?, part from the exit row. I'm not sure what character would be the best, though.

Thank you for explaining. Perhaps if there's no input present, it can show an empty space where the icon is, it would be similar to the controls section where there's an empty space if you unbind an input.

aredfan (approved):

rr-:
I've included a minimal fix to not show `?,OPTIONS` and instead go with `OPTIONS`. However removing `?` would require us to hardcode a lot more formatting logic in case some buttons are missing (for example: `[+]Space: Change FOV` if walk is unbound), so for now I've opted to leave it as-is.

PR #3294, opened 2025-06-21
screenshot: fix screenshot stripping accents

### LINKED ISSUE #3238: Screenshot file names strip away accented level characters

rr-:
#### Checklist

#### Description

Resolves #3238. This PR restores proper support for accented (and other non‑ASCII) characters in screenshot titles.

Previously, we've been removing any characters that aren't alphanumeric, but with the recently added official support for accented characters in level titles, this approach has become problematic.

We were doing this for two reasons - first, to remove sensitive characters such as `/` which can be interpreted as directory separators. The other reason was the challenge of dealing with Unicode support on Windows systems. Unlike other platforms, Windows' "narrow" C `fopen()` does not recognize UTF-8 as it relies on the current ANSI code-page, leading to mishandling or rejection of non-ASCII characters. To accurately open a path containing arbitrary Unicode elements, we need to convert the UTF-8 path to UTF-16LE using `MultiByteToWideChar(CP_UTF8…)` and utilize the wide-char variant `_wfopen()`, which directly engages with the Win32 UTF-16 APIs.

As such I would be grateful to find out if there's any effect on running the game from directories with accented characters, CJK symbols, or paths with spaces.

### DIFF
--- a/docs/tr1/CHANGELOG.md
+++ b/docs/tr1/CHANGELOG.md
@@ -9,6 +9,7 @@
 - changed some settings to be hidden when they're only applicable to specific games or custom levels (#3242)
 - changed photo mode help dialog to show icons for inputs
 - changed settings to retain their active position until exiting to title or starting a new level (#3271)
+- fixed screenshots stripping accented characters (#3238)
 - fixed turbo cheat causing audio desync in cutscenes (#3263)
 - fixed depth buffer problems when closing the inventory ring with fade effects disabled (#3267, regression from 4.8)
 - fixed Lara's braid not being reflective (on Midas' hand) (#3257, regression from 4.9)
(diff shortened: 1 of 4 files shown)


lahm86 (commented):
Paths with spaces on Windows seem to be OK.
If I use an accented character anywhere however, the game crashes after the legal screen (no log file is produced at all).

rr-:
Oh, that's interesting, it seems to be working for me on Wine – not entirely but at least it launches fine. How about the screenshots, are they generating the expected file names? For instance, levels 5 and 7 in the Polish language use accented characters.

In terms of the larger issue with Unicode characters in the game path, I'm considering if we could use libuv to outsource encoding conversions. But let's keep that separate for now.

lahm86:
Yes, screenshots themselves are looking good with accents. Happy to approve

image

lahm86 (approved):

aredfan (approved):

PR #3303, opened 2025-06-23
Fix TR2 underwater hum

### LINKED ISSUE #3305: TR2 underwater hum not playing properly
Steps to reproduce
1) `/play 7`
2) `/fps 60`
3) play on a device with 30 FPS to force frame drops
Regression most likely since 0.10.

rr-:
#### Checklist

#### Description

Resolves #3305.

The underwater hum would play for at most 2 frames before getting reset on certain devices. The reason for this is because of how `Sound_EndScene()` works.

The idea is that looped sound sources need to `Sound_Effect()` their samples on every logical frame to achieve continuous playback, otherwise the samples will stop. Which is actually quite convenient. To achieve this behavior, `Sound_EndScene()` checks if the looped samples are silent. If they aren't, it syncs the parameters to the relevant output stream and updates the samples volume to be silent – but keeps the other parameters. In the following frame, the repeated call to `Sound_Effect()` will either reset the volume back to some non-silent value, or if there are no such calls, the second call to `Sound_EndScene()` will deem the sample as no longer in use, stop the playback, and fully close the slot.

At the same time the call to `Sound_Effect(SFX_UNDERWATER)` would happen inside M_EnsureEnvironment() which is part of the draw phase. Under normal circumstances this is fine, even if we have 60 FPS. But if we experience frame drops, this means the calls to `Sound_Effect()` and the calls to `Sound_EndScene()` can go out of order , and the samples will randomly restart. This is especially visible if the FPS is set to 60 but the display is only capable of rendering 30 FPS.

The fix ended up being simple and was to just move the camera environment sync to the control phase.

#### Testing

The only play is around the camera going into and out of water bodies.
I'd appreciate a version in which the regression occurred.

### DIFF
--- a/docs/tr2/CHANGELOG.md
+++ b/docs/tr2/CHANGELOG.md
@@ -15,6 +15,7 @@
 - changed photo mode help dialog to show icons for inputs
 - changed settings to retain their active position until exiting to title or starting a new level (#3271)
 - removed config tool (we have ingame setting dialogs now)
+- fixed underwater hum not playing properly (#3305, regression from 0.10)
 
 ## [1.2.1](https://github.com/LostArtefacts/TRX/compare/tr2-1.2...tr2-1.2.1) - 2025-06-22
 - fixed some secrets in some levels incorrectly registering by standing on specific tiles (#3280, regression from 1.2)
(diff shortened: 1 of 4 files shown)


aredfan (approved):
> I'd appreciate a version in which the regression occurred.

I'm not 100% sure but I think it's a regression from [0.10](https://github.com/LostArtefacts/TRX/releases/tag/tr2-0.10).

lahm86 (approved):

PR #3323, opened 2025-06-23
tr2/shell: react to fullscreen window movements

### LINKED ISSUE #2842: Backdrop aspect can be squished
https://github.com/user-attachments/assets/8c79f084-25be-45d5-a415-14321923fc66

Sorry for what is perhaps a somewhat annoying issue. Doesn't help that it's hard to describe the steps either:

With a multi-monitor setup, when moving a fullscreened TR2X from display to display via hotkeys (when the above video flickers), it's possible to mess with the backdrop in the main menu. I noticed that setting it to "any", and having the game assume different aspect ratios per display, it would sometimes not update once back on the initial display. If then you manually set a 16:9, it'll reset correctly, and show fine too as "any" (since the display at that moment was also a 16:9). This last step is shown at the very end of the video.

rr-:
#### Checklist

#### Description

The game wouldn't reload the background from the file that has the best aspect ratio. To slightly reduce disk thrashing, output now caches the directory scan results (though the image will be reloaded on every update still).

Resolves #2842.


lahm86 (approved):
LGTM but I'm currently unable to test with a second monitor.

aredfan (approved):
No second monitor here either. I just did a quick sanity test instead.

Richard-L:
Too late, but still to say it works perfectly. Thanks a lot for working on this!!

PR #3324, opened 2025-06-23
shell: fix title bar too small on HiDPI screns

### LINKED ISSUE #2837: Window title bar size is rendered too small on HiDPI screens
Display without scaling:

Image

Display with scaling:

Image

rr-:
#### Checklist

#### Description

Please bear in mind this is a blind fix to a Windows-specific problem, made by a Linux-based dev on a Linux platform from reading online docs alone. It has high likelyhood of not working at all.

(Hopefully) Resolves #2837.

### DIFF
--- a/docs/tr1/CHANGELOG.md
+++ b/docs/tr1/CHANGELOG.md
@@ -62,6 +62,7 @@
 - fixed Lara's braid remaining reflective if the fly cheat is used to resurrect her on the Midas hand
 - fixed invulnerability cheat not getting disabled during the demos (regression from 4.13)
 - fixed health bar flicker on medi packs when cycling the inventory ring (#4211, regression from 4.14)
+- fixed title bar size being too small on HiDPI screens on Windows platform (#2837)
 
 ## [4.15.1](https://github.com/LostArtefacts/TRX/compare/tr1-4.15...tr1-4.15.1) - 2025-10-10
 - changed the examine dialog to be usable with non-puzzle items (#4009)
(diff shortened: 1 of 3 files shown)


aredfan (approved):

lahm86 (approved):
Looks OK my end (but tbh it also looks OK on develop with high DPI scale, so perhaps this is monitor dependent).

rr-:
@Richard-L can you weigh in? Thanks.

Richard-L:
I appreciate so much you're looking into this, despite Windows not being your preferred OS🙏 

Sadly the window header size is still unadjusted and thus too small.

But I have the impression that Windows is now trying to keep the window's overall size for both games roughly the same between displays. I'm not quite sure, but it seemed as though before the window size itself shrunk down as a whole, not just the header, whereas now it's just the header. If this should matter to know exactly, I can compare again with an old build if you like.

rr-:
> I appreciate so much you're looking into this, despite Windows not being your preferred OS🙏
> 
> Sadly the window header size is still unadjusted and thus too small.
> 
> But I have the impression that Windows is now trying to keep the window's overall size for both games roughly the same between displays. I'm not quite sure, but it seemed as though before the window size itself shrunk down as a whole, not just the header, whereas now it's just the header. If this should matter to know exactly, I can compare again with an old build if you like.

I'd appreciate a confirmation, as otherwise I don't think it makes sense to include this patch.

Richard-L:
Hmm, I tried against TR2X v1.0, as it was one day after I had posted the original issue. Unfortunately, it looks to be the same behaviour compared to the unrelease build above.

rr-:
Thanks. I'll keep it opened in case I get my hands on a Windows machine soon to play with it a bit more. (If not, I'll close it, and the ticket will go back to backlog for time being :disappointed: )

rr-:
@Richard-L I've pushed a new fix, which I'm hoping settles the original bug properly. Please LMK once you find the time if it's working as expected.

Richard-L:
Thank you dash.

I think it looks great!

LowDPI
low-dpi

HiDPI
image

PR #3344, opened 2025-06-25
Make UI settings more maintainable

### LINKED ISSUE #3326: Adopt x-macros for ui settings

rr-:
#### Checklist

#### Description

1. Reformats the config module .def files to be prettier, adopts the naming convention for the x-macros.
2. Hides the settings dialog state in its own module
3. Simplifies memory and state management in options all around
4. Moves UI settings-related enum entries to its own module. They're no longer static.
5. Moves UI settings-related custom handler code to its own module. They're no longer static.
6. Introduces x-macros to keep track of the UI setting definitions; moves setting definitions to separate .def files.
    ⚠️ I am not sure if I like the very, very long lines in the .def files. LMK your thoughts. I think I don't like them. Maybe we can do better with the remaining macros?
7. Updates linter to accept the new macro
    To avoid repetition, the definitions only repeat the option name once, rather than declaring separately GS_ID(option_name) and GS_ID(option_name_description). This means that we eagerly try to load descriptions regardless of the option. After this change, it is an error to declare an option without description – the only options that we had like this seems to be limited to individual water color components.

Resolves #3326.

#### Testing

For this we need a quick glance at the common functions of the individual settings – changing a value, fast changes (with TR3 sidesteps off), trying to change it to something out of bounds, descriptions, etc. In general only light checks are necessary I think.


aredfan (approved):

lahm86 (approved):
LGTM. The long lines are fine IMO as they're in the def files (fairly similar to how we have long text lines in the GS def files, but I appreciate it's different as it's actual setup and not just strings). I do think it's a great change overall.

PR #3354, opened 2025-06-26
Add level counters to the statistics

### LINKED ISSUE #1087: Add level number to statistics
Doing a playthrough with my nephew, and he keeps asking me how many levels there are in total, and which one we are in. And I struggle to answer because of all the Level3A/B shenanigans by Core.

Got me thinking if it were worth adding a:

"**LEVEL x OF y**"

At the top of the stats page (compass + level completion).

Let me know your thoughts.

rr-:
#### Checklist

#### Description

Resolves #1087 and moves the stats details mode to the UI tab which I believe is a more fitting place.


lahm86 (approved):

aredfan:
Just one issue found - Return to Egypt is listed as level 0 instead of 1 in the level counter.

rr-:
Thanks – fixed that although I don't know if I like the way I did that. @lahm86 would you mind taking another look? My problem is mainly the ambivalence between numbering and indexing. I tried to be a bit more clear through a docstring and `OrdinalNumber` rather than just `Number`, but I don't know.

lahm86:
> Thanks – fixed that although I don't know if I like the way I did that. @lahm86 would you mind taking another look? My problem is mainly the ambivalence between numbering and indexing. I tried to be a bit more clear through a docstring and `OrdinalNumber` rather than just `Number`, but I don't know.

LGTM. I think it's fine as it's encapsulated in those functions.

aredfan:
> Thanks – fixed that although I don't know if I like the way I did that.

No problem, the issue is fixed. 👍 

I found a new issue, TR2X is experiencing a crash when I select the stopwatch, or if I skip a level to the stats. The crash only happen with the option enabled, and it doesn't happen in the gym.

rr-:
> > Thanks – fixed that although I don't know if I like the way I did that.
> 
> No problem, the issue is fixed. 👍
> 
> I found a new issue, TR2X is experiencing a crash when I select the stopwatch, or if I skip a level to the stats. The crash only happen with the option enabled, and it doesn't happen in the gym.

Thank you – fixed this as well

aredfan (approved):

PR #3355, opened 2025-06-26
tr1/objects/save_crystal: increase vertical collision range

### LINKED ISSUE #3343: Make the save crystals collision more lenient
Crystals are hard to target – even a small nudge or misplacement will cause them to not work.

lahm86:
Resolves #3343.

#### Checklist

#### Description

This increases the vertical range for save crystal collision, but adds an additional check where collision should be blocked by the ceiling. So they can now be up to 4 clicks above Lara, but any ceiling in the way will block collision.

<details><summary>Ceiling block e.g.</summary>

image
</details>

Test level available with crystals at various heights.

One thing I've noticed is that they don't work underwater (OG), that could be a nice feature to include as well.

### DIFF
--- a/docs/tr1/CHANGELOG.md
+++ b/docs/tr1/CHANGELOG.md
@@ -16,6 +16,7 @@
 - changed photo mode help dialog to show icons for inputs
 - changed settings to retain their active position until exiting to title or starting a new level (#3271)
 - changed the dev console to accept compound characters (#2938)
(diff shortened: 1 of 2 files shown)


rr- (approved):
>One thing I've noticed is that they don't work underwater (OG), that could be a nice feature to include as well.

I think it's great – let's add a ticket.

aredfan (approved):

PR #3376, opened 2025-06-29
tr2/objects: avoid testing windows as creatures

### LINKED ISSUE #3375: TR2X bug: harpoon/grenade doesn't break windows
https://github.com/LostArtefacts/TRX/commit/588f92289f6f0fa3c368475f6db7d211b0dd2282

_______________________

When standing in the position shown in the screenshot below, Lara's harpoons/grenades doesn't break the windows in front of her.

Image

</details>

There's an OG bug with harpoons/grenades not breaking windows from the other side (room 3), but that's for a future issue.

lahm86:
Resolves #3375.

#### Checklist

#### Description

This ensures windows aren't tested as targetable creatures for flying harpoons and grenades.

@aredfan you mentioned about this being OG if Lara is facing the opposite way, but also I don't think these projectiles work on bells, do they? Is that also OG? Would be good to sort that out eventually too.

### DIFF
--- a/src/tr2/game/objects/general/grenade.c
+++ b/src/tr2/game/objects/general/grenade.c
@@ -82,13 +82,14 @@ static void M_Control(const int16_t item_num)
             continue;
         }
 
-        if (target_item->object_id != O_WINDOW_1
+        const bool is_window = target_item->object_id == O_WINDOW_1;
+        if (!is_window
             && (!target_obj->intelligent || target_item->status == IS_INVISIBLE
                 || target_obj->collision_func == nullptr)) {
             continue;
         }
 
-        if (!Creature_IsTargetable(target_item)) {
+        if (!is_window && !Creature_IsTargetable(target_item)) {
             continue;
         }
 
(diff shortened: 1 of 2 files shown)


rr- (approved):

rr- (approved):
There is also a function like item is smashable IIRC.

lahm86:
> There is also a function like item is smashable IIRC.

Thanks, yeah, I think that's what'll be needed to get them working on bells too, so can implement that as an enhancement. Fairly sure it's OG that they don't work on these given how the code only checks window 1.

aredfan (approved):
I can confirm the bug is fixed, thank you. 👍 

> @aredfan you mentioned about this being OG if Lara is facing the opposite way, but also I don't think these projectiles work on bells, do they? Is that also OG? Would be good to sort that out eventually too.

Yep, but it depends on the breakable window's position.  If a window is positioned in the middle of a tile, it can be broken by projectiles from both directions (e.g. Venice). Whereas if a window is positioned on the edge of a tile, it can only be broken from one side (e.g. Offshore Rig). Perhaps the portals have something to do with it. 

The projectiles have no effect on bells in OG, just checked.

lahm86:
Thanks, good info! I guess we can get a ticket or tickets raised for the remaining OG problems and tackle those separately.

PR #3386, opened 2025-06-30
output: remove dynamic light limit

### LINKED ISSUE #3384: TR2X bug: dropped flares doesn't emit light when fire lighting is enabled
https://github.com/LostArtefacts/TRX/commit/46f23c82eb4f64eeea815f0617043b7204217667

lahm86:
Resolves #3384.

#### Checklist

#### Description

This replaces the fixed size dynamic light array with a vector, so removing the limit of 10 per frame. I've updated the TR1 test level for fire light, and added one for TR2, both with a silly number of flame objects for testing performance. It behaves OK on my end, but LMK if you think we should retain an upper limit.

### DIFF
--- a/src/libtrx/game/output/common.c
+++ b/src/libtrx/game/output/common.c
@@ -2,8 +2,7 @@
 #include "game/matrix.h"
 #include "game/output.h"
 #include "utils.h"
-
-#define MAX_DYNAMIC_LIGHTS 10
+#include "vector.h"
 
 typedef struct {
     XYZ_32 pos;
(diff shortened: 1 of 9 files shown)


src/libtrx/game/output/common.c
```diff
@@ -279,21 +277,30 @@ void Output_LightRoom(ROOM *const room)
...
 }
 
+void Output_InitLight(void)
+{
+    m_DynamicLights = Vector_Create(sizeof(LIGHT));
+}
+
+void Output_ShutdownLight(void)
+{
+    Vector_Free(m_DynamicLights);
```
rr-: Let's add double free protection and maybe a TODO to consolidate this into `Output_Init()` once it becomes common. Same with `InitLight`, only set it if it's null.

rr- (approved):

aredfan (approved):
> It behaves OK on my end, but LMK if you think we should retain an upper limit.

It also works well on my end, there's no hit to performance as far as I can tell.

PR #3394, opened 2025-06-30
lara/cheat: set animation state post fly cheat

### LINKED ISSUE #3392: TRX bug: Lara can get stuck in swim mode and position when exiting the fly cheat
TR2X regression since [0.3](https://github.com/LostArtefacts/TR2X/releases/tag/0.3) (the fly cheat was introduced in this version)
TR1X regression is unreleased

______________________

Enable the fly cheat and have Lara moving along the ground, then use uw roll just a split second before pressing shift. This may take a few attempts.

<details><summary>Video</summary>

https://github.com/user-attachments/assets/8792b5f4-5adf-400f-a1e5-a82fb1a9c148

</details>

lahm86:
Resolves #3392.

#### Checklist

#### Description

I'll admit I wasn't able to reproduce this, but logically this should be the fix. I think what was happening was there was a one frame window where Lara could continue in her twist state and with the roll key detected there, she would become stuck yet not have a cheat water status. TR2 doesn't have state changes for underwater roll, whereas TR1 does because we inject it; so the merging of the state routine introduced it to TR1.

### DIFF
--- a/docs/tr2/CHANGELOG.md
+++ b/docs/tr2/CHANGELOG.md
@@ -45,6 +45,7 @@
 - fixed the shotgun sound at the end of the shower cutscene in Home Sweet Home being cut off when the credits start (#1579)
 - fixed the camera being partially inside the wall at the end of the Home Sweet Home shower cutscene (#3370)
 - fixed being able to deselect the passport in the game over screen (#3381, regression from 1.0)
+- fixed Lara getting stuck in the fly cheat in rare circumstances (#3392, regression from 0.3)
 - improved the `/tp` command to orient Lara towards keyholes and doors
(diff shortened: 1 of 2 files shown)


rr- (approved):
Haven't tested but LGTM

aredfan (approved):
> I'll admit I wasn't able to reproduce this, but logically this should be the fix.

It's difficult for me to reproduce it too. Sometimes it takes 5 tries, other times it takes more than 30 attempts. It's a pretty elusive bug.

I'm fairly confident the issue is fixed now, thank you. 👍

lahm86:
Thanks, @aredfan. Good find as always 😌

PR #3401, opened 2025-07-01
Rework keyholes

### LINKED ISSUE #3398: Bug:  Lara is able to permanently discard a key item if she gets pushed on the exact frame she interact with a receptacle
I'm not sure which version this introduced or if it's an OG bug, or if it applies to TR1.
Still, logging it

### LINKED ISSUE #3399: Enhancement: have Lara consume the keys on the frame the item is input to a receptacle
Currently the key item gets removed at the start of interaction, and entering the fly cheat will cause the puzzle hole / keyhole to immediately trigger. This is TRX-specific behavior – in the OG I'm pretty sure the item would get consumed at the time of interacting at least for puzzle holes. This can be different for TR1 and TR2.

Haven't checked the gong or the detonator.

Fixing this influences the dupe glitch.

rr-:
#### Checklist

#### Description

- Resolves #3398 
- Resolves #3399 (with the exception of the gong and the detonator, which consume the item immediately)
- Moves keyhole and puzzle hole code to libtrx
- Splits gong and detonator box code
- Renames `detonator 1` to `gong`
- Renames `detonator 2` to `detonator box`

#### Testing

- Normal integration with receptacles
    - Action key / Manual inventory open 
    - Puzzle holes / Keyholes / Detonator / Gong
- Canceling the animation with the flight cheat
- Saving/loading
- Item duplication glitch on and off, whether it's still working as it should, and if it's fixed as it should, and if other interactions with receptacles work
- Trying to use the mystic plaque when Lara is getting hit by endless enemies (the animation should be canceled, but the item should _not_ be consumed unlike in develop – I'm not sure if it's an OG bug or a TRX bug and if it is then what version introduced this)


lahm86 (commented):
Looking great. Item dupe is behaving as expected, along with save/load, the fly cheat interrupt and enemy nudging scenarios.

There is just one issue I've come across, where you can accidentally consume 2 key items.
- `/play 9` (TR1)
- `/tp hole 3`
- `/give 2 rusty`
- Use a key
- Before Lara finishes the animation, go back into the inventory and select the key again

Lara will immediately repeat the key animation, and both keys get consumed. It seems to be dependent on when you go into the inventory during the animation. If you do it from the point below onwards, it seems to be consistent.

20250701_145121_Tomb_of_Tihocan
</details>

rr-:
> Looking great. Item dupe is behaving as expected, along with save/load, the fly cheat interrupt and enemy nudging scenarios.
> 
> There is just one issue I've come across, where you can accidentally consume 2 key items.
> 
> * `/play 9` (TR1)
> * `/tp hole 3`
> * `/give 2 rusty`
> * Use a key
> * Before Lara finishes the animation, go back into the inventory and select the key again
> 
> Lara will immediately repeat the key animation, and both keys get consumed. It seems to be dependent on when you go into the inventory during the animation. If you do it from the point below onwards, it seems to be consistent.

Thanks – added a fix.

lahm86 (approved):

aredfan:
I noticed an issue, it's possible to have a delay in Lara's movements after interacting with a keyhole.

Repeat Lahm's steps, then in the specific moment in the screenshot below, open the inventory and use the second rusty key.  Then hold down.  Lara doesn't back away from the keyhole immediately after using the key.

20250701_155519_Tomb_of_Tihocan

</details>

> Trying to use the mystic plaque when Lara is getting hit by endless enemies (the animation should be canceled, but the item should not be consumed unlike in develop – I'm not sure if it's an OG bug or a TRX bug and if it is then what version introduced this)

I'm not sure how to replicate this bug in OG as it's quite tricky to pull off, but I think it's more likely to occur in TR2X. This bug is only present in develop and not in the latest stable release for TR2X.

rr-:
Thanks @aredfan – great catch – fixed this. Lara should now refuse to use while in the middle of an interaction.

We'll need to test this with savegames in TR1 and TR2 made on develop, and on this branch, so that Lara doesn't refuse a valid interaction due to badly initialized variables. I tested it and seemed OK but I'd appreciate a double check.

aredfan (approved):
> Thanks @aredfan – great catch – fixed this. Lara should now refuse to use while in the middle of an interaction.

No problem, and LGTM. 👍 

> We'll need to test this with savegames in TR1 and TR2 made on develop, and on this branch, so that Lara doesn't refuse a valid interaction due to badly initialized variables. I tested it and seemed OK but I'd appreciate a double check.

I tested with swapping savegames between this branch and develop, it seems to be working as expected.

PR #3412, opened 2025-07-02
data: and end of level stat screens to tr2

### LINKED ISSUE #2682: Add end of level stats for Home Sweet Home and Kingdom
TR2's Home Sweet Home, TR2 The Golden Mask's Kingdom, TR3's Meteorite Cavern, and TR3 The Lost Artifact's Reunion all don't have end of level stats and go straight to the final cutscenes / credits. So TR1 is unique in that it displays the end of level stats for The Great Pyramid. IMO I think it would be nice to have them all show end of level stats. None of them have secrets that affect bonus level unlocks which is maybe why they didn't include it? Or maybe Core got rid of it for style reasons to flow better into a final cutscene and end credits more like a movie? But now that our default stats option shows max kills and pickups in a level, I think it might be nice to have the final level stats screen.

rr-:
#### Checklist

#### Description

Resolves #2682.

### DIFF
--- a/data/tr2/ship/cfg/TR2X_gameflow.json5
+++ b/data/tr2/ship/cfg/TR2X_gameflow.json5
@@ -439,6 +439,7 @@
                 {"type": "remove_medipacks"},
                 {"type": "loop_game"},
                 {"type": "play_music", "music_track": 48},
+                {"type": "level_stats"},
                 {"type": "level_complete"},
                 {"type": "display_picture", "credit": true, "path": "data/images/credit01.webp", "display_time": 15, "fade_in_time": 0.5, "fade_out_time": 0.5},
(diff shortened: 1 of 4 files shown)


aredfan (approved):

lahm86 (commented):
I'm not sure if I'm just too used to the skidoo music starting with the credits, but IMO it sounds a bit odd having this in the level stats screen. Maybe we should use the regular end level music with the stats, and keep the skidoo music starting with the credits?

rr-:
> I'm not sure if I'm just too used to the skidoo music starting with the credits, but IMO it sounds a bit odd having this in the level stats screen. Maybe we should use the regular end level music with the stats, and keep the skidoo music starting with the credits?

I was thinking that, but it sounded even more odd TBH. I think we shouldn't postpone the hype track till credits – imagine a streamer sitting on this screen after finishing a challenge run.

lahm86 (approved):

PR #3422, opened 2025-07-02
Viewport rewrite

### LINKED ISSUE #3332: "Any" aspect ratio in TR1X
It would be lovely if TR2X' "Any" aspect ratio feature could be brought over to TR1X as well. Thanks in advance.

Original message which lead to this:

> I hope it's alright to bring this up and doesn't just come off as totally lazy 🙏
> 
> I have a laptop that I carry around the house to two desks where I like to test TRX on: one has a 16:9 monitor w 1920x1080px native resolution and 240hz, the other desk has a 16:10 monitor at 1920x1200px with 59.94hz. I connect the laptop to both of these, either via a thunderbolt hub, or directly via HDMI, and then extend the display into one wider workspace. I have Windows configured so that the external monitor is set to "Use as my main display". Hence the games will start on this monitor when launched, and assume their fullscreen resolution when toggled.
> 
> Both games will launch in the previously selected resolution, which makes sense, but it would be even nicer if there could be an "auto" (ie max native auto) resolution option that applies dynamically upon launch before the first legal screen or FMV. Currently what happens is that letterboxing occurs until I refresh the resolution in settings.
> 
> And I had an aside thought that could perhaps become a spin-off issue if you're interested: would it make sense to grey out the "Resolution" option while TRX is in windowed mode?

rr-:
#### Checklist

#### Description

Resolves #3332.

- Removes the screen resolution option from TR1
- Brings over the "aspect mode" option from TR2 to TR1
- Brings over the "scaler" option from TR2 to TR1 and renames it to "upscaling factor"
- Brings over the "sizer" option from TR2 to TR1, renames it to "borders" and repurposes it as a percentage field
- Brings over the "fbo filter" option from TR1 to TR2, renames it to "upscaling filter"
- Fixes the translation tools to not complain about the `en-gb` dialect file missing a base file
- Fixes the translation tools showing diffs related to formatting with `--dry-run`
- Rewrites caching of the background images, improving performance during window resizes

#### Testing

- A sanity check whether the mentioned features work as expected
- Updating the relevant settings at runtime in both games, via the UI and via `/set`.
- Toggling fullscreen, moving window around, etc.
- Screenshots (via photo mode and via the print screen key)
- 3D pickups UI
- Docs provided in the PR – I have a feeling I've missed something…
- Choosing aspect-mode specific backgrounds on aspect mode changes etc.


aredfan:
Great overhaul, well done!

- [x] I noticed an issue - the console command to adjust `Borders` doesn't use %, so you'd need to use values like 0.10

rr-:
> Great overhaul, well done!
> 
> * [ ]  I noticed an issue - the console command to adjust `Borders` doesn't use %, so you'd need to use values like 0.10

Thanks – fixed.

aredfan (approved):

lahm86 (approved):

PR #3429, opened 2025-07-03
input: allow skipping credits faster

rr-:
#### Checklist

#### Description

Allows the player to hold the Action or Escape keys to skip through the credits faster, including fade effects. The skip will stop at the final stats screen. A side effect of the way I implemented this is that you can hold the Action key on the final level statistics, and it'll skip the images altogether, but I think it's actually desirable.


lahm86 (approved):
Great!

aredfan (approved):

PR #3431, opened 2025-07-04
photo-mode: add poses

rr-:
#### Checklist

#### Description

Add photo mode poses. Triggered with the fly cheat key in photo mode. Adds a couple fixes to the `/tp` command.
The braid acts funny – it's recommended to cycle a few frames with the pause key for it to settle.

20250704_142614_Lost_Valley
20250704_141930_Colosseum
20250704_141518_Lost_Valley
20250704_145157_Tibetan_Foothills
20250704_145025_Tibetan_Foothills
20250704_142759_Floating_Islands


aredfan:
Those screenshots are lovely. 😌 

I noticed an issue that affects TR1X only.  If Lara has dual guns drawn while switching poses, her left arm will remain in upright position for all poses.

rr-:
> Those screenshots are lovely. 😌
> 
> I noticed an issue that affects TR1X only. If Lara has dual guns drawn while switching poses, her left arm will remain in upright position for all poses.

Thanks – fixed!

aredfan (approved):

lahm86 (approved):
Looking great!
The only problem is cutscenes in TR1, as there is no regular Lara item there. Only the braid takes on the pose. We have the `lara_type` property in the GF, but it might be a can of worms to link to this. Maybe easier to disable posing if regular Lara isn't present?

rr-:
Thanks – you're right, we'd have to contaminate `Object_DrawInterpolatedObject` with Lara stuff and I don't think this is the proper way to go around it. TR2 cutscenes also worked badly, as Lara would reset to her initial position.
I've disabled it for now; we can always revisit it later.

PR #3444, opened 2025-07-05
Improve UI rendering with the upscaling option

### LINKED ISSUE #1944: Text looking terrible in Framebuffer rendering with bilinear filter turned off
Window size (eg drawing everything into a buffer with the same dimensions as the host window, using 800×600 as the resolution for the software 3D vertex calculations):
out2

FBO (eg rendering both game and UI into an offscreen buffer of 800×600 size with the same resolution being used for the vertex calculations):
out1

And after upscaling to match the resolution from the first example:
out1

rr-:
#### Checklist

#### Description

Resolves #1944 by drawing UI at native scaling separately from the game, and making screenshots always capture the images at desktop size.

#### Screenshots

Before:

20250705_154743_Lara's_Home

After:

20250705_160823_Lara's_Home

#### Technical considerations

Best viewed on a commit by commit basis.

- Draws UI into a new separate framebuffer that always uses the target viewport size, regardless of the chosen upscaling factor for the main game.
- Introduces `VIEWPORT_UI` space separately from `VIEWPORT_TARGET` to improve clarity in the code.
- Changes screenshots to always use `VIEWPORT_TARGET` space instead of the small game window as on develop.
- Changes the scaler module to work with floats, improving the calculations fidelity.
- Extracts the framebuffer code from `gfx/renderer.c` into `gfx/gl/fbo.c`.
- Changes `gfx/renderer.c` to keep track of two framebuffers, one for geometry and the other for the UI.
- Changes `gfx/renderer.c` to automatically keep track of the framebuffer sizes.
    - The resizes are now guaranteed to happen right after a swap, which should reduce visual artifacts.
    - The framebuffers are also immediately cleared on resize.
- Removes `GFX_GetDisplayWidth/Height` and `GFX_GetWindowWidth/Height`, instead always using `VIEWPORT_SPACE` enum wherever needed.
- Removes `GFX_Context_SetWindowSize()` and `GFX_Context_SetDisplaySize()`.
- Fixes debug messages with stray `\n` causing pointless blank lines in logs.
- Removes unused `wait()` callback function from the phasers.

Since the UI is now drawn onto an isolated texture, we needed to introduce a new blending mode with premultiplied alpha. This boils down to using `glBlendFunc(GL_ONE, GL_ONE_MINUS_SRC_ALPHA)` rather than `glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA)`. Otherwise, we'd be causing the UI to be more transparent.

With the current setup, with the UI having its own `VIEWPORT_SPACE` and framebuffer, if we ever want it, we can use separate upscaling filter for just the UI elements.

#### Testing

The new blending mode is used to mitigate a scenario where an active UI fade effect will not only fade to black, but also cause the UI elements themselves to become more transparent. We need to look out for that.

Other than that:
- window resizes
- changing render modes in TR2
- changing aspect modes
- changing upscaling factor
- changing upscaling filter
- wireframe mode on/off
- debug overlays esp. trigger tiles transparency
- all UI elements (not necessarily all dialogs, just making sure we check each type of elements):
    - health and air bars, unbind key progress bar
    - dialog frames
    - dialog backgrounds (since these use transparency, we need to pay extra attention to that)
    - text glyphs, spacing between letters on big upscaling factors
    - sprite-based pickups (they should use high-fidelity rendering even with an upscaling factor… at least I think?)
    - 3D pickups just for good measure
    - Lara's Home timer digits
    - debug position digits, ammo counter


aredfan:
I noticed a small issue affecting TR1 only:

1. launch the game
2. on the main menu, execute `/set upscaling-factor 8`
3. `/demo`
4. press action to skip the demo
5. when back on the main menu, execute `/set upscaling-factor 1`
6. notice the backdrop looks blurred, and less pixelated too

**Edit:** It appears to be present in develop too, just checked.

__________________

- [x] TR1: In controls, the `Reset All` and `Unbind` texts get hidden when the bar is drawn.

lahm86 (approved):

rr-:
Thank you @aredfan  – fixed both problems. Great catch with the background, ~~though it's surprising TR2 failed to cache the image~~ (actually I think I know why – playing the game causes the background to be unloaded and load the inventory carpet pattern. Then it gets unloaded again, and this is why it's not cached).

aredfan (approved):
> Thank you @aredfan – fixed both problems. Great catch with the background, ~though it's surprising TR2 failed to cache the image~ (actually I think I know why – playing the game causes the background to be unloaded and load the inventory carpet pattern. Then it gets unloaded again, and this is why it's not cached).

No problem, and thank you.

Both of the issues are fixed. I also rechecked TR2X for good measure, LGTM. 👍

PR #3449, opened 2025-07-05
Further Lara mesh merging

### LINKED ISSUE #3451: TR2X bug: Lara's meshes don't reset properly in fly cheat
For example, equip Uzis then enter and exit the fly cheat - Lara's holsters revert to pistols.
Regression since 0.3 (fly cheat introduction).

lahm86:
Resolves #3451.

#### Checklist

#### Description

This updates TR2 to store Lara's holster and back gun types so that we can correctly restore them when leaving the fly cheat. It should as a result fix that issue - @aredfan, has this been an issue since it was introduced?

We may be able to merge Lara's back gun object ID with the back gun type at a later stage, it will need some work doing in TR1.

Some general save/load with guns in different states would be good to check here too.


src/libtrx/game/savegame/common.c
```diff
@@ -425,8 +415,15 @@ void Savegame_PersistGameToCurrentInfo(const GF_LEVEL *const level)
     } else {
         resume->equipped_gun_type = lara->gun_type;
     }
-    resume->gun_status = LGS_ARMLESS;
 #endif
+
+    resume->holsters_gun_type = lara->holsters_gun_type;
+    resume->back_gun_type = lara->back_gun_type;
+    if (TR_VERSION == 1 && lara->gun_status == LGS_READY) { // TODO: TR2
```
rr-, line 422: Can you please elaborate on this particular condition? I'm not fully following

src/libtrx/include/libtrx/game/lara/types.h
```diff
@@ -84,10 +84,9 @@ typedef struct {
         XYZ_32 initial_pos;
     } interact_target;
 
-#if TR_VERSION == 1
     LARA_GUN_TYPE holsters_gun_type;
     LARA_GUN_TYPE back_gun_type;
-#else
+#if TR_VERSION >= 2
     LARA_GUN_TYPE last_gun_type;
     GAME_OBJECT_ID back_gun_obj_id;
```
rr-, line 91: Can we remove `back_gun_obj_id` with some relatively low effort, or will that be a part of a future PR?

rr- (approved):
Just two cents from me on testing – I recommend testing the legacy saves as well when Lara has different combinations of holsters, back, and equipped gun types in both games.

aredfan:
> @aredfan, has this been an issue since it was introduced?

Yep, it's been the case since [0.3](https://github.com/LostArtefacts/TR2X/releases/tag/0.3) (the good old days when we had to type `/fly` instead of pressing <kbd>o</kbd>).

Before this PR, Lara's weapon meshes wasn't always consistent (when entering/exiting fly cheat) and can depend on factors such as starting a level or creating/loading a SG.

src/libtrx/game/savegame/common.c
lahm86: In TR1, if Lara ends a level with guns out, she starts the next level with guns out. In TR2 she always starts with guns away. The to do is to come back and see about getting it to work in TR2 and perhaps offering an option (I think it may interfere with offshore rig start, will need to investigate).

src/libtrx/include/libtrx/game/lara/types.h
lahm86: I think it should be OK, but yes I think I'll tackle it separately for safety given how fragile these meshes are 😄

aredfan (approved):
I tested with various combinations of pistols, shotgun, auto pistols and harpoon gun in TR2, all works as expected.

TR1 looks good as well. 👍

lahm86:
Thanks! Changelog updated for the fly cheat problem.

PR #3456, opened 2025-07-06
lara/common: offer revert to pistols in TR2

### LINKED ISSUE #3455: Feature: make revert to pistols optional
Similar to TR1, allow Lara's current guns to be remembered when going between levels.

lahm86:
Resolves #3455.

#### Checklist

#### Description

This adds TR1's revert to pistols option to TR2. I looked into the similar feature of Lara keeping her guns out if she finishes a level in that state, but it causes a lot of problems because of cutscenes. I think we can look at this again when refactoring gun control. For now the option should remember her back and holster gun types, and when it's off, she'll revert to pistols/shotgun.

I've also ported `Savegame_ApplyLogicToCurrentInfo` to TRX, and added to the game flow to re-add Lara's guns in Diving Area, so to bring things into line with TR1. TR2 used to enforce having the pistols on every level in this function, but I think this is cleaner.

So other than the gun status checks, a quick test around jumping between levels would be good along with save and load.


lahm86:
Above push just for docs update.

aredfan (approved):

rr- (approved):

PR #3472, opened 2025-07-07
game-strings: fix Italian translation

Kappa971:
#### Checklist

- [ ] I have read the [coding conventions](https://github.com/LostArtefacts/TRX/blob/master/CONTRIBUTING.md#coding-conventions)

#### Description

Small fixes from Remastered versions.

I think that the translated strings should be moved to a separate folder, for example you could create a TEXT folder in the main directory with other folders inside (IT, FR, PL, etc.) that contain the files of each language, otherwise with the addition of other languages ​​and, possibly in the future with the subtitles, it would become too confusing.


rr- (approved):
Thanks for the suggestion re: filesystem, I'll tackle that later.

PR #3493, opened 2025-07-10
creature/common: retain only monks as allies

lahm86:
Resolves #3492.

#### Checklist

- [x] I have read the [coding conventions](https://github.com/LostArtefacts/TRX/blob/master/CONTRIBUTING.md#coding-conventions)
- [x] I have added a changelog entry about what my pull request accomplishes, or it is an internal change
- [x] I have added a readme entry about my new feature or OG bug fix, or it is a different change

#### Description

This removes Lara, Winston and the dying monk in Diving Area from the list of ally types so that only allies who will fight other enemies are retained. Those specific enemies now have their own array as well to add an additional abstraction layer, and Winston is defined as a special type in order to continue allowing him to be killed by a direct console command only.

Testing
- Monks fighting bandits
- Check that bandits still target Lara at times during melee with monks
- Console `/kill` tests for monks and bandits
- Easter egg with angry monks and above
- Killing Winston via console

PR #3494, opened 2025-07-10
Don't stretch backgrounds in TR2 beyond small margin

### LINKED ISSUE #2001: OG bug: pictures get stretched when using widescreen
The following pictures are affected; copyright, main menu, and credits.

rr-:
#### Checklist

#### Description

Resolves #2001. Also adds a 16:10 aspect ratio to test it – while we don't provide 16:10 images, this aspect ratio is close to 3:2 enough that a 3:2 image should be used and stretched to cover the viewport.
In more extreme cases the game should use the current TR1 algorithm which is to match the height of the viewport and on wider resolutions use letterboxing, and on more squarish resolutions trim the sides.


src/libtrx/game/console/cmd/kill.c
```diff
@@ -42,7 +42,7 @@ static bool M_KillSingleEnemyInRange(const int32_t max_dist)
     int16_t best_item_num = NO_ITEM;
     for (int16_t item_num = 0; item_num < Item_GetTotalCount(); item_num++) {
         const ITEM *const item = Item_Get(item_num);
-        if (Creature_IsHostile(item)) {
+        if (Creature_IsHostile(item) && item->object_id != O_LARA) {
```
lahm86: Would you like me to include this in #3493 as an extra guard?

lahm86 (approved):

Richard-L:
I'm sorry I cannot test as I should, I am still just too preoccupied with life. But if I can say: would you consider to not stretch a 16:9 image to 16:10, but instead add small necessary black borders in this case?
We love the image assets so much, so that even smaller deformations like a mild stretch become noticeable in a negative way.
Thanks for considering.

src/libtrx/game/console/cmd/kill.c
rr-: Oh shit, it's a leftover of my own testing. I'll revert that

rr-:
> I'm sorry I cannot test as I should, I am still just too preoccupied with life. But if I can say: would you consider to not stretch a 16:9 image to 16:10, but instead add small necessary black borders in this case? We love the image assets so much, so that even smaller deformations like a mild stretch become noticeable in a negative way. Thanks for considering.

Thanks for the feedback. The current implementation applies letterboxing only for significant aspect ratio mismatches. For 16:10, the stretch from 3:2 that we ship with is minimal (1.6 vs. 1.5), and in practice avoids visible seams or awkward text framing.

I’ve attached comparisons:

### Stretching 3×2 onto 16×10
20250710_114847_Level_0

### Using the 3×2 proportion with letterboxing
20250710_114945_Level_0

### Using the 16×9 proportion with letterboxing
20250710_115525_Level_0

In both letterboxed variants, UI elements float outside the background, which I think looks worse overall.

If you prefer a fully letterboxed look for 16:10, you can override the title screen by placing a custom image at `data/images/16x10/title_eu.png`.

aredfan (approved):
Thank you for working on this. I tested with 4:3 images in widescreen mode, then widescreen images in 4:3 mode.  Both LGTM. 👍

Richard-L:
Thanks for responding and explaining Dash. I see what you mean.

PR #3498, opened 2025-07-10
misc: enhance Lara's immunity

### LINKED ISSUE #3496: TR2X bug: Lara can catch fire when impervious to damage
https://github.com/LostArtefacts/TRX/commit/8b7e9446e5e5b40cc7123438e3c9c0803d08c1e2

__________________

1. launch golden mask
2. `/play 4`
3. `/tp 67 2 62`
4. `/immune`
5. cross the bridge

Image

</details>

lahm86:
#### Checklist

#### Description

While immune, Lara can now survive jumping into spikes; boulders/barrels; Thor's hammer; killer pushblocks and swan-diving onto land from any height. This also resolves catching fire when walking across a bridge.

I think the only one-shot deaths left are: T-rex, Torso and Midas hand. Should we just exclude these too?

### DIFF
--- a/src/tr2/game/objects/traps/rolling_ball.c
+++ b/src/tr2/game/objects/traps/rolling_ball.c
@@ -4,6 +4,7 @@
 #include "game/spawn.h"
 #include "global/vars.h"
 
+#include <libtrx/config.h>
 #include <libtrx/game/camera.h>
 #include <libtrx/game/collision.h>
 #include <libtrx/game/game_buf.h>
@@ -161,11 +162,13 @@ static void M_Collision(
         return;
     }
 
-    if (lara_item->gravity) {
+    if (lara_item->gravity || g_Config.debug.enable_invulnerability) {
         if (coll->enable_baddie_push) {
             Lara_Push(item, coll, coll->enable_hit, true);
         }
-        lara_item->hit_points -= ROLLING_BALL_DAMAGE_AIR;
+        if (!g_Config.debug.enable_invulnerability) {
+            lara_item->hit_points -= ROLLING_BALL_DAMAGE_AIR;
+        }
 
         // TODO: handle overflows
         const int32_t dx = lara_item->pos.x - item->pos.x;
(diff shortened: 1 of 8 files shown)


src/tr2/game/objects/traps/rolling_ball.c
```diff
@@ -161,11 +162,13 @@ static void M_Collision(
...
 
-    if (lara_item->gravity) {
+    if (lara_item->gravity || g_Config.debug.enable_invulnerability) {
         if (coll->enable_baddie_push) {
             Lara_Push(item, coll, coll->enable_hit, true);
         }
-        lara_item->hit_points -= ROLLING_BALL_DAMAGE_AIR;
+        if (!g_Config.debug.enable_invulnerability) {
+            lara_item->hit_points -= ROLLING_BALL_DAMAGE_AIR;
+        }
```
rr-, line 171: We could use `Lara_TakeDamage` TBH, but this can become a part of a bigger cleanup as there are many usages left.
BTW Sometimes Lara receives damage via Item_TakeDamage too. 
Both of these cases are covered by the invulnerability cheat resetting Lara's HP to max in the control routines.

rr- (approved):
I don't have any strong opinion on the special animation deaths. I think Yetis and Xian spearmen can still kill Lara too? If their anim is not HP-dependent. 
I'm leaning towards an opinion that this can be permitted as we still allow them to push Lara around.

src/tr2/game/objects/traps/rolling_ball.c
lahm86: Agreed. I was tempted to start merging a couple of these modules here but best to wait as you say 😌

lahm86:
> I don't have any strong opinion on the special animation deaths. I think Yetis and Xian spearmen can still kill Lara too? If their anim is not HP-dependent. I'm leaning towards an opinion that this can be permitted as we still allow them to push Lara around.

Yeah, the yetis, Xian spearmen and sharks have an HP check, it's just t-rex and Torso that kill her outright. I'm inclined to leave it, as well as the midas touch as these are scenarios you have to really want yourself to be in, rather than accidentally falling into spikes or similar.

aredfan (approved):

PR #3527, opened 2025-07-17
tr1/pickup: reset interact target on pickup

### LINKED ISSUE #3524: Lara says "No" when trying to use the Silver Key on the first door in Cistern (TR1X 4.13)
I'm having a persistent issue (at least on my end) where Lara refuses to use the first Silver Key in Cistern. She just says "No" every time I try to use it on the first door, even though I clearly have the key. I saw that this was supposedly fixed in 4.13, but it's still happening for me.

At first I thought it was caused by saving/loading near the door, so I tried again without saving at all and it still didn't work. The only way I’ve managed to get it working is with a fresh install of TRX1, not touching the settings and going straight for Cistern.
When I do that, the key works fine. So I figured it might be a settings thing. But when I reset *all* settings to default on my First install (still using 4.13), it still didn’t work.

I’m attaching two log files: 
First Install: [TR1X.log](https://github.com/user-attachments/files/21299972/TR1X.log)
Fresh Install: [TR1X.log](https://github.com/user-attachments/files/21299980/TR1X.log)

I didn't wanna make a new issue, especially for a problem that you guys fixed, but it still persists on me.

Btw I use Linux Mint if that is important.

lahm86:
Resolves #3524.

#### Checklist

#### Description

This ensures to reset the interact target after Lara picks up an item when animated interactions is enabled. For existing saves as in the issue, a resolution would be to go to a switch and use it, then return and try to use the keyhole.

### DIFF
--- a/docs/tr1/CHANGELOG.md
+++ b/docs/tr1/CHANGELOG.md
@@ -4,6 +4,7 @@
 - changed the texture page limit from 128 to unlimited (#3517)
 - fixed Lara's first pose in photo mode at times being skipped (#3522, regression from 4.13)
 - fixed Lara's arms being drawn inaccurately when posing in photo mode with dual weapons equipped (#3520, regression from 4.13)
+- fixed Lara being unable to use key items at times with animated interactions enabled (#3524, regression from 4.13)
 
 ## [4.13](https://github.com/LostArtefacts/TRX/compare/tr1-4.12.3...tr1-4.13) - 2025-07-14
 Showcase: https://youtu.be/YKI7u2QOolU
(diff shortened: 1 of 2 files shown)


docs/tr1/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## [Unreleased](https://github.com/LostArtefacts/TRX/compare/tr1-4.13...develop) - ××××-××-××
 - added German translation
 - changed the skybox option to allow toggling in-game without the need to reload the level
+- fixed Lara beign unable to use key items at times with animated interactions enabled (#3524, regression from 4.13)
```
aredfan: Small typo `beign`.

aredfan (approved):

rr- (approved):
Do we need a hotfix?

lahm86:
> Do we need a hotfix?

Sounds like a good idea. While there is a workaround it's not ideal and not hugely obvious I'd say.

PR #3540, opened 2025-07-19
objects/flame: skip effects when object not present

### LINKED ISSUE #3539: TRX bug: dynamic fire light generated despite flames not being present
Temple of the Cat has death tiles, but the flame object is not present so Lara doesn't catch fire, just dies outright. With dynamic fire light enabled, the flickering effect is generated regardless, which looks odd without actual fire.
This affects TR2 as well, although I can't think of a level with death tiles and no flame object present.
Regression from 4.13.

lahm86:
Resolves #3539.

#### Checklist

#### Description

This avoids dynamic light being generated when the flame object is not present in the level.

### DIFF
--- a/docs/tr2/CHANGELOG.md
+++ b/docs/tr2/CHANGELOG.md
@@ -3,6 +3,7 @@
 - changed the texture page limit from 128 to unlimited (#3517)
 - fixed flat/opaque window shards in Lara's Home and Home Sweet Home (#3512)
 - fixed savegame scanner only seeing all-lowercase file names (#3518, regression from 1.0)
+- fixed dynamic fire light being generated despite the flame object not being present in the level (#3539, regression from 1.3)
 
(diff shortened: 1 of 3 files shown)


rr- (approved):

docs/tr2/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 - changed the texture page limit from 128 to unlimited (#3517)
 - fixed flat/opaque window shards in Lara's Home and Home Sweet Home (#3512)
 - fixed savegame scanner only seeing all-lowercase file names (#3518, regression from 1.0)
+- fixed dynamic fire light being generated despite the flame object not being present in the level (#3539, regression from 4.13)
```
aredfan: Just noticed this, probably since 1.3 because the `Fire lighting` was added in that version.

aredfan (approved):

docs/tr2/CHANGELOG.md
lahm86: Thanks. The dangers of copy-paste.

Issue #2593, opened 2025-03-09
TR2X Bug Mismatched cutscenes audio


Checkm8Croft:
When I play any cutscene on TR2X (Same on Windows an macOS), the audio isn't synced with cutscene, the audio will plays few seconds after the cutscene starts.
I tried to use cdaudio.mp3 and single audio tracks (taken from Tomb Raider Remastered) but the issue persists

macOS Note: if I use single audio tracks the music doesn't play

walkawayy:
I can't reproduce this on any cutscene on Windows 11 on the latest dev snapshot ( commit: https://github.com/LostArtefacts/TRX/commit/9caa484cfee8a2ea21f1322f5ae084ec50ecf184 )

aredfan:
It's been a while since I've tested TR2X with cdaudio but I can confirm this is a regression since [0.7](https://github.com/LostArtefacts/TRX/releases/tag/tr2-0.7).

You'll need to remove the music folder and only use the audio folder, then check each cutscene. All 4 cutscenes play the music for the first cutscene albeit with a delay.

PR #3543, opened 2025-07-19
Fix audio drift in HSH shower cutscene with turbo speed

### LINKED ISSUE #3541: TR2X bug: audio desync in shower cutscene when sped up
Regression since [0.6](https://github.com/LostArtefacts/TRX/releases/tag/tr2-0.6) (this is the version the turbo cheat was introduced in)

Partially fixed in [0.7](https://github.com/LostArtefacts/TRX/releases/tag/tr2-0.7) for cutscenes and such, but not the shower cutscene.

rr-:
#### Checklist

#### Description

Resolves #3541.
Affects cutscenes and has potential to affect pausing in the cutscenes. Affects TR1 too.

### DIFF
--- a/docs/tr2/CHANGELOG.md
+++ b/docs/tr2/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## [Unreleased](https://github.com/LostArtefacts/TRX/compare/tr2-1.3.1...develop) - ××××-××-××
 - added German translation
 - changed the texture page limit from 128 to unlimited (#3517)
+- fixed audio in the shower cutscene in Home Sweet Home not being sync with the turbo cheat (#3541)
 - fixed flat/opaque window shards in Lara's Home and Home Sweet Home (#3512)
 
 ## [1.3.1](https://github.com/LostArtefacts/TRX/compare/tr2-1.3...tr2-1.3.1) - 2025-07-18
--- a/src/libtrx/engine/audio_stream.c
+++ b/src/libtrx/engine/audio_stream.c
@@ -424,6 +424,26 @@ void Audio_Stream_Shutdown(void)
     }
 }
 
+bool Audio_Stream_SyncTimestamp(const int32_t sound_id, const double timestamp)
+{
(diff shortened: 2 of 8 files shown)


aredfan:
The shower cutscene's audio drift fix is working great, thank you. 👍 

I noticed that pausing during cutscenes causes the music track to restart. This affects both games.

lahm86:
This also fixes #2593 from what I can tell, could be worth linking and adding a changelog entry for that (@aredfan - maybe you could confirm in case I'm mistaken).

rr-:
Actually this can't go in its current form. Syncing against master clock is nice but turns out master clock goes out of sync with the cutscene clock, as unpausing the game resumes the music during the fade-out effect. Everything after that goes out of sync.

Fun

rr-:
Alright I've rewritten the fix to forgo use of the master clock and instead sync to the animation frame. It should be much less intrusive now. @lahm86 LMK if you think we should create a new API for getting a relative frame number of the active anim, as I'm now doing manual calculation on the frame numbers.

I've decided not to add the master clock API, even though it looks appealing. I think it could be too tempting and make it easy for us to corner ourselves into cross-phase regressions when not careful.

RE: #2593 I don't think it solves that. That one seems to be related to decoding the big cdaudio mp3 file.

lahm86:
> Alright I've rewritten the fix to forgo use of the master clock and instead sync to the animation frame. It should be much less intrusive now. @lahm86 LMK if you think we should create a new API for getting a relative frame number of the active anim, as I'm now doing manual calculation on the frame numbers.
> 
> I've decided not to add the master clock API, even though it looks appealing. I think it could be too tempting and make it easy for us to corner ourselves into cross-phase regressions when not careful.
> 
> RE: #2593 I don't think it solves that. That one seems to be related to decoding the big cdaudio mp3 file.

We have `Item_GetRelativeFrame` already 😌 

#2593 is back to being broken (so probably a good thing, can be handled on its own). The initial PR was certainly working on my end though.

aredfan (approved):

PR #3590, opened 2025-07-23
New APIs in filesystem.c

rr-:
#### Checklist

#### Description

- `File_WriteString(MYFILE *fp, const char *fmt, ...)` which is a no-brainer and uses `String_FormatStaticV` internally
- `const char *File_ReadLine(MYFILE *fp)` which is a bit more interesting – it uses a MYFILE-local growing string buffer to read a single line. Manages the buffer. I'd appreciate a check for logical bugs (corner cases), though in practice it seems to be working OK.

The APIs remain unused at this moment – I'll soon open a PR that uses them.

### DIFF
--- a/src/libtrx/filesystem.c
+++ b/src/libtrx/filesystem.c
@@ -22,6 +22,9 @@
 struct MYFILE {
     FILE *fp;
     const char *path;
+    // line-reading buffer and its capacity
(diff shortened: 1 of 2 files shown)


aredfan (approved):

lahm86 (approved):
LGTM

PR #3591, opened 2025-07-23
Config API updates

rr-:
#### Checklist

#### Description

First, a couple of new functions. I want to use them to interact with external human-readable files in an upcoming PR.

- Introduces `const CONFIG_OPTION *Config_GetOptionByPath(const char *path)`.
- Introduces `const char *Config_ResolveOptionName(const char *name)` although it's more of a side effect to the refactor than a intended change. (It strips the tree prefix so `"gameplay.fix_bear_ai"` becomes `"fix_bear_ai"`).
- Introduces `const char *Config_GetOptionValueAsString(const CONFIG_OPTION *opt)`. It was mostly taken from the `/set` console command, although it got simplified a lot thanks to `String_FormatStatic`. I had to drop `GS()` conversion for boolean values to be able to use it without relying on the player's chosen language. And in order to avoid duplication, I decided to drop the `/set` string serializer specialization in favor of this new simpler function.

**Now onto the big thing**: We no longer call `Config_Write()` after changing options. Instead, we call `Config_Update()`. Underneath it works pretty much the same, except these key differences:

- `Config_Update()` no longer writes to disk. This job is done in a distilled `Config_Write()` and calling this is a responsibility of subscribers to config change events. Specifically, it's now done by the main shell subscriber in libtrx.
- `Config_Update()` no longer consults the disk for whether the config has changed, and instead operates solely by comparing `g_Config` against `g_SavedConfig`.

Additionally, we no longer write the config prior to game exit – persisting the config to disk is supposed to be handled in reaction to events in all cases. And that's what we need to focus with testing here – if all unique ways for changing the config trigger the save properly:
- `/set` (OK to test just one setting)
- UI setting dialogs (OK to test just one setting in the normal dialogs)
- UI key binding dialogs (both game and controller, resetting to defaults, unbinding)
- Hotkeys (need to test all of them separately)
- ?

The idea behind this change is to be able to easily disable saving the config to disk if the game is ran in a special way.
I'm not sure if removing of `Config_Write()` prior to game exit is the right call – it's a good safe guard and we'd have only two places to mock rather than one, both within shell.c, so maybe it's not that bad. LMK WDYT.

Additionally we should check if saving the config works well with enforced options by the builder – enforced settings shouldn't influence the saved settings, which should be kind of "frozen".

### DIFF
--- a/docs/tr1/CHANGELOG.md
+++ b/docs/tr1/CHANGELOG.md
@@ -8,6 +8,7 @@
 - changed the game flow and game strings file placement
 - changed the skybox option to allow toggling in-game without the need to reload the level
 - changed the texture page limit from 128 to unlimited (#3517)
+- changed the `/set` console command to report boolean values as `0` or `1`, language-agnostic
 - fixed glide camera behaviour and position in room 101 in Temple of the Cat (#3533)
 - fixed French translations containing Italian text in some cases (#3567)
 - fixed the camera remaining locked on moving lava if it touches Lara when she is immune (#3578)
--- a/docs/tr2/CHANGELOG.md
+++ b/docs/tr2/CHANGELOG.md
@@ -9,6 +9,7 @@
 - added an option to animate the algae in 40 Fathoms, Wreck of the Maria Doria and The Deck (Gameplay settings → Fixes → Fix sprite animations) (#3141)
 - changed the game flow and game strings file placement
 - changed the texture page limit from 128 to unlimited (#3517)
+- changed the `/set` console command to report boolean values as `0` or `1`, language-agnostic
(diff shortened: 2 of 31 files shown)


aredfan (approved):

lahm86 (commented):
I only noticed one issue - input binding doesn't save e.g. go into the controls dialog, change any input, then Alt+F4. It does save the selected layout, it's only the bindings.

Re not having a failsafe `Config_Write` on exit, I'm fine with this, it makes sense to me it should be handled by the listeners.

rr-:
Added a fix – thanks.

lahm86 (approved):

PR #3597, opened 2025-07-24
objects/waterfall: merge and fix waterfalls

### LINKED ISSUE #3589: TRX/OG bug: odd activation statuses on Waterfall mist
Waterfall mist object activation works differently depending on the engine version - here's what has been tested:

**TR1X**
- doesn't activate with code bits
- can be triggered and antitriggered with triggers

**TR2X**
- activates with code bits
- can be triggered but not antitriggered with triggers

### LINKED ISSUE #3593: OG bug: inaccurate waterfall triggers in Lost Valley
- After using the cog machine, it's feasible to jump over the antipads for waterfall 26, so it remains on after the flipmap
- Even if you run over the antipads for item 26, you can then return to the start of the level and re-activate it
- Waterfalls 15, 16, 17 have only one trigger which is very easy to skip

lahm86:
Resolves #3589.
Resolves #3593.

#### Checklist

#### Description

Merges the waterfall modules into TRX and fixes how they can be triggered/antitriggered to be consistent in both games. Also fixes the skippable triggers in Lost Valley. Test levels uploaded in `waterfalls.zip` with various setups.

One thing I noticed is the game doesn't draw them when Lara's more than 10 tiles away. Presumably this was for performance reasons, should we remove that check?

There is also an initial commit here to fix some doc issues re slide-to-run.

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/90


rr- (approved):
IMO OK to remove the distance check. Ember emitters seem to spawn lots of particles regardless of distances. This change will need a changelog entry.

lahm86:
> IMO OK to remove the distance check. Ember emitters seem to spawn lots of particles regardless of distances. This change will need a changelog entry.

Agreed, it seems odd that this was left in place, perhaps just an oversight. I'll raise a separate ticket to avoid confusing things here.

aredfan (approved):

PR #3603, opened 2025-07-24
data: improve twist animations

### LINKED ISSUE #1589: OG bug: incorrect Lara rotation during uw twist
Cf: https://github.com/LostArtefacts/TR1X/issues/1242

**Copy of Lahm's comment from the TR1X issue:**

> Documenting this here as it will need to be fixed in TR2X eventually.
> 
> - Remove flip effect 0 from the roll start animation and add it to the first frame of the roll end animation instead.
> - Set yaw, pitch and roll on the first frame of the roll end animation to match the 20th frame of the roll start animation (the 20th is unused). Invert the pitch.
> - Restore the defective anim command on the roll start animation for flip effect 3 - use frame 14.
> - Ensure to link to frame 20 of the underwater idle animation from the roll end animation.

lahm86:
Resolves #1589.

#### Checklist

#### Description

Applies the same fix from TR1 to Lara's underwater twist animation in TR2. I've also slightly improved the regular jump twists by moving the 180° flip effect into the end animation in each case, which gets rid of some odd frames in the middle of each action. 

The (OG) issue of pancake Lara remains in the running jump twist animation, I'm really at a loss there I must admit. LMK if the regular twists look OK, I can revert that easily if we'd rather a different approach (just not sure what that would be, would need some help 😄).

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/91


aredfan:
> The (OG) issue of pancake Lara remains in the running jump twist animation, I'm really at a loss there I must admit. LMK if the regular twists look OK, I can revert that easily if we'd rather a different approach (just not sure what that would be, would need some help 😄).

Subtle improvements are always welcome when a solution remains elusive.  I think this PR is good as is.

aredfan (approved):

lahm86:
Updates pushed as discussed on Discord to improve the running jump twist frames. Would appreciate a re-test, thanks.

aredfan (approved):

rr- (approved):

PR #3642, opened 2025-07-27
tr2/data: fix push button offsets

### LINKED ISSUE #3596: OG bug: noticeable gap between buttons and walls in TR2
This appears to be the case with every button in TR2 and Golden Mask.

Image

lahm86:
Resolves #3596.

#### Checklist

#### Description

This fixes the push button in TR2 by correcting its frame offset and bounds in all animations. A new injection type has been introduced to replace a specific object and animation frame set entirely, without having to replace the entire model.

It'd be good to double check the button animations as part of this.

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/100


aredfan (approved):
Very nice! 😌

rr- (approved):

PR #3712, opened 2025-08-04
lara: add sprinting support

### LINKED ISSUE #3711: Feature request: add sprint to TR1 and TR2
Add an option to allow sprinting. Additional enhancements:
- allow Lara to sprint whenever she has stamina, rather than only when the bar is full
- allow sprinting up one-click steps rather than being interrupted with normal running (she can run down them as standard)
- if Lara is stationary and forward and sprint are held together, in OG TR3 she doesn't move - she should start her normal running animation and move to sprint in this case

The new options will be relevant to TR3 as well.

lahm86:
Resolves #3711.

#### Checklist

#### Description

Adds optional sprinting to TR1 and TR2, including an extra option to allow sprinting whenever Lara has stamina, and to sprint up stairs. The only animations not included are those for sprint-roll-to-crawl, plus a few alternative beta animations for sprint-roll that aren't used.

The default input is `.` on keyboard, which I'm not sure is suitable. Also, I'm not sure if we have an available controller input for it, so it's currently blank. I also picked green for the bar, but this does clash with the ally healthbar. WDYT?

Some testing points:
- Sprint roll
- Lara should slide/skid to a stop if forward is released but sprint is still held down
- Saving while sprinting, reloading and checking the bar is accurate

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/112


rr-:
Re: green, maybe we could introduce green 2?
We'll still have to tweak this when we go with PS1 healthbars.

lahm86:
Pushed an alternative green, and I noticed a bug where the timer didn't replenish if Lara sprinted off a ledge into water. Fixed.

aredfan:
> Pushed an alternative green, and I noticed a bug where the timer didn't replenish if Lara sprinted off a ledge into water. Fixed.

Good catch there.

I've finished testing the PR builds and the sprint feature is working as expected, like it does in TR3.  Awesome work. 👍

aredfan (approved):

rr-:
20250804_224237_Caves

Thanks. Not sure if that's intended, but the bar shades appear upside down?
Otherwise LGTM.

lahm86:
> Thanks. Not sure if that's intended, but the bar shades appear upside down? Otherwise LGTM.

Thanks, not intended 😄 Fixed.

PR #3728, opened 2025-08-05
lara/state/land: restore idle pose animation

### LINKED ISSUE #3727: Feature: restore idle pose animation
Add an option to restore Lara's idle pose animation from early 1996 beta levels.

lahm86:
Resolves #3727.

#### Checklist

#### Description

Restores Lara's idle pose from early beta levels, which will trigger after 10 seconds of inactivity provided she is weapon and flare free, on land, not in a vehicle, and not being nudged by an enemy. Any of the previous factors, or otherwise any regular movement input, will take her out of the state.

LMK if you think anything should be tweaked e.g.
- the timeout itself
- the way the camera spins around to face Lara
- currently Lara can look when posing, and look isn't counted as an idle-interrupt activity (so she can stand idle for 10 seconds continuously looking around, and the pose will still trigger)
- we're not saving/loading the idle timer currently: my feeling was it might look odd Lara posing soon after reloading a (potentially old) save if that save was made with the timer close to the limit

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/113


rr- (approved):
LGTM on all ends.

lahm86:
Forgot the changelog, pushed.

rr-:
Now that I played with it a bit:
One thing I noticed is that A/D camera controls do not seem to work in the idle pose.
And I think I'd increase the timer to 30 s – alt tabbing got annoying; she's guaranteed to be idling. Possibly configurable. Or give it a random chance?
Possibly make the camera optional and default to off.

An idea: randomly exit the idle animation, and randomly apply a mirrored pose.

lahm86:
> Now that I played with it a bit: One thing I noticed is that A/D camera controls do not seem to work in the idle pose. And I think I'd increase the timer to 30 s – alt tabbing got annoying; she's guaranteed to be idling. Possibly configurable. Or give it a random chance? Possibly make the camera optional and default to off.
> 
> An idea: randomly exit the idle animation, and randomly apply a mirrored pose.

Thanks, pushed some changes.

- Config option is now an int, so player can set the timeout manually. 0 will disable it
- Separate option for the camera
- The camera will shift on the last frame of the "start" animation rather than forcing the angle each frame. This does mean look will reset it, but I think that's fine

I like the idea of mirrored poses too. I can work on the animations for that.

aredfan:
I noticed an issue. If Lara exits the idle pose animation while the camera is in front of her, the camera will remain in front of Lara until look is pressed to reset it.

Richard-L:
I love love love the fact you decided to add this Lahm 🙏  I'd decided years ago I wasn't going to post an issue for it because it seemed too remote from being interesting, but I personally love it.

I noticed one thing: if you release Lara from the idle pose via the new neutral jump twist, she will instead perform a roll.

I also wondered if there's any other way to release her from the idle state. Currently the command you want her to perform needs to be entered twice: Once to leave idle state, and again to actually perform it. I understand why this is how it is. Do you think it would look weird to have her cancel out of the idle pose directly into the entered move?
The leave-idle-state animation could perhaps be used in order for her to change direction, as Dash suggested. So as to prolong the idle state, she could move from left to right leg with a random number of frames spend per leg each time. Just thinking out loud.

I would also like to suggest to make `Look` cause idle state to end. I find this would be more in line with how you typically use `Look` to regain control over cameras.

Richard-L:
This is likely a bit too detail-obsessed, but if it should be an easy enough fix: in this old animation, you can see her legs wiggle a bit as she breathes in. You can notice it when looking at her calves.

https://github.com/user-attachments/assets/8f58db05-1d63-4ba9-98f7-984c3955e2d1

lahm86:
Added a mirrored animation set, so Lara will now either lean to the left or right by chance. I didn't tackle switching poses during posing itself, I think that could be a future enhancement as it's a bit more involved.

I will look at "look" tomorrow. So is the consensus that look should cancel it fully? Should it be treated as an idle-interrupt to begin with as well?

The issue of the camera not resetting when she finishes the pose should be resolved @aredfan.
The neutral jump twist problem should be resolved @Richard-L 

Re the animation twitch I'll see if I can improve it, but it looks quite finicky.

Thanks for the feedback 😌

aredfan:
> I will look at "look" tomorrow. So is the consensus that look should cancel it fully? Should it be treated as an idle-interrupt to begin with as well?

Sounds good to me. The `Look` input cancels the idle camera, so it might as well cancel the animation too.

I guess after this change, if the player uses `Look`, it'll reset the idle countdown as well.

lahm86:
Look control updated - it will cancel pose and reset the idle countdown.
I can't improve the animation twitch as-is, I think that's maybe something that can be looked at again if anyone else is willing to amend the frames.
Otherwise I think this is good to go, if testing looks OK.

aredfan (approved):

rr-:
> Look control updated - it will cancel pose and reset the idle countdown. I can't improve the animation twitch as-is, I think that's maybe something that can be looked at again if anyone else is willing to amend the frames. Otherwise I think this is good to go, if testing looks OK.

@lahm86 WDYT about making Lara cancel the idle animation by herself at random occurrences, then resetting the idle timer?

lahm86:
> > Look control updated - it will cancel pose and reset the idle countdown. I can't improve the animation twitch as-is, I think that's maybe something that can be looked at again if anyone else is willing to amend the frames. Otherwise I think this is good to go, if testing looks OK.
> 
> @lahm86 WDYT about making Lara cancel the idle animation by herself at random occurrences, then resetting the idle timer?

Added 😌

rr- (approved):

PR #3760, opened 2025-08-07
ui/dialogs/settings: show disclaimer when all settings hidden

### LINKED ISSUE #3637: Hiding all config items results in blank menu
Understandably expected, but if you hide all the config options, the menus becomes blank. Would it be worth hiding the icon itself similar to Lara's Home? I imagine if the video, gameplay and sound options are all hidden, that might make the ring look weird (Or break it entirely) - a suggestion instead would be some text there that says something like "No configurable options"

Here's a extract you can chuck into your gameflow file to test it (For TR1X). This will empty the Gameplay, Graphics and Sound menus

[Blank config ](https://github.com/user-attachments/files/21453312/hidden_config.txt)

Image

lahm86:
Resolves #3637.

#### Checklist

#### Description

If a level builder has hidden all settings for a particular dialog, a suitable message will be shown to the player rather than an empty window.

Adapted config below from Gizzy's post to include some of the latest settings - this will result in the gameplay and sound dialogs being empty, while graphics will have a few remaining.
[hidden.txt](https://github.com/user-attachments/files/21671635/hidden.txt)

### DIFF
--- a/data/common/ship/cfg/base_strings-de.json5
+++ b/data/common/ship/cfg/base_strings-de.json5
@@ -18,6 +18,7 @@
         "CMD_INVALID_SECRET": "Unzulässiges Geheimnis: #%d (zulässige Geheimnisse: %s)",
         "CMD_LUA_RUNTIME_ERROR": "\\{review}Lua-Laufzeitfehler: %s",
         "CMD_LUA_SYNTAX_ERROR": "\\{review}Lua-Syntaxfehler: %s",
+        "COMMON_SETTINGS_ALL_HIDDEN_DISCLAIMER": "\\{review}Die Einstellungen sind für dieses Level-Set deaktiviert.",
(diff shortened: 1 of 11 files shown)


rr- (approved):

aredfan (approved):

rr-:
Actually I'm thinking of using a more neutral wording as this one might be taken as a bit of finger-pointing. Maybe something like

>"Settings are disabled for this level set."
or
>"This level set uses custom settings. In-game options are disabled."
or
>"In-game settings can’t be changed in this level set."

WDYT?

lahm86:
> Actually I'm thinking of using a more neutral wording as this one might be taken as a bit of finger-pointing. Maybe something like
> 
> > "Settings are disabled for this level set."
> > or
> > "This level set uses custom settings. In-game options are disabled."
> > or
> > "In-game settings can’t be changed in this level set."
> 
> WDYT?

Thanks, yes that wasn't my intention so this is a good catch 😄 
I opted for the first, short and simple.

rr- (approved):

PR #3846, opened 2025-08-17
savegame: fix music resuming

### LINKED ISSUE #3845: TR1X Bug: music track is not stopped when loading a save from the same level
If you trigger a music track and load a save of the same level, the music continues to play uninterrupted. If you load a save from a different level, thee music stops as expected. Only affects TR1 from what I tested.

Regression since 4.13

https://github.com/user-attachments/assets/d51f2269-e2d4-4ef1-bb8a-8bdcfcd81958

rr-:
#### Checklist

#### Description

Resolves #3845.

Need updating changelog with the version at which this is broken.

#### Testing

Testing needs to check all 3 modes of resuming active music track on load, and verify if this bug is present in TR2.
I imagine Venice may be problematic as normally it does not have an ambient track.

### DIFF
--- a/docs/tr1/CHANGELOG.md
+++ b/docs/tr1/CHANGELOG.md
@@ -90,6 +90,7 @@
 - fixed Lara walking backwards off ledges into lava (#3745)
 - fixed room scheduling algorithm sometimes drawing overlapping rooms (#3774, regression from 4.1)
 - fixed exiting photo mode on a controller conflicting with the roll input (#3842, regression from 4.8)
+- fixed resuming non-ambient music tracks when loading a savegame (#3845, regression from 4.13)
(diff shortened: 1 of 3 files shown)


walkawayy:
Regression since 4.13. It's actually slightly better in the unreleased build. In 4.13, the music will persist even when loading a different level's savegame. In unreleased, it only persists if you load a savegame from the same level.

walkawayy (approved):
Fix works for me. Just needs the CHANGELOG. I checked Venice even before this build and the bug doesn't seem to affect TR2 AFAIK.

aredfan (approved):

lahm86 (approved):

PR #3855, opened 2025-08-20
game-strings: fix Italian translation

Kappa971:
#### Description

Small fixes.

I noticed that new animation have been added.
I think it would be useful to add a new option that automatically disables some new TRX engine behaviors that could be problematic with the original game levels (I'm thinking of sprinting in TR1, which could have similar issues to enabling jumping in Doom), or even just to have a more authentic experience.
I know it's already possible to do this manually, but having a dedicated option that disables various new features differently between Tomb Raider 1 and 2 (because the two games were slightly different) would be much more practical and immediate, especially for those like me who want to use a new engine that runs natively on modern hardware but at the same time want the games to not deviate too much from the original.


rr- (approved):
Thanks for the suggestion. If you'd like to see this move forward, the best next step would be to open an issue (or drop by our Discord) with a concrete proposal – for example, which features you think should be disabled in TR1 vs TR2, or how the option might look in the UI. That way it's clear what you have in mind and easier for us to discuss.

PR #3859, opened 2025-08-20
tr2/console/cmd/winston: test object presence

lahm86:
#### Checklist

#### Description

This avoids a crash if trying to spawn Winston when the object isn't loaded. The crash only happens on develop (I think because of the rendering changes), hence no changelog.


aredfan (approved):

rr- (approved):
I understand this problem to only take place if the Winston injection is disabled?

lahm86:
> I understand this problem to only take place if the Winston injection is disabled?

Yes, so e.g. if a custom level doesn't ship the file. Shipped TR2X has the correct injections targeted as-is 😌

Issue #3768, opened 2025-08-08
TR1X bug: SFX corruption


lahm86:
- `/play 8`
- `/tp 22 -7 54`
- Pull the lever

Unreleased, still investigating.

lahm86:
a59b01b20

rr-:
Sorry 😥

PR #3890, opened 2025-09-07
Sound refactor

rr-:
#### Checklist

#### Description

Merges TR1 and TR2 sound modules into trx. I'd appreciate a closer reading of the source code to assess whether we like it in its current form. I dislike that TR1 needs `M_UpdateActiveSoundParams` to update for example flame emitters position, whereas 1) TR2 gets this handled with `Sound_Effect` alone 2) the logic of this function duplicates much of `Sound_Effect` initial distance/pan calculation. Speaking of which, I also don't like how distance calculation is different between TR1 and TR2. These are the biggest sources of `#if TR_VERSION` guards. The final one is about `volume` resetting to 0 for ambient sources which I don't fully understand.

It also fixes a bug whether the master volume would be applied twice, resulting in quadratic volume loss for certain samples (so 10% would be rendered as 0.1%, 30% as 9% etc.).

As a follow-up I would like to introduce the setting to disable random pitch changes to TR2, and then get rid of sample LUT to remove the upper limit on the sound slots, and possibly also handle -1 to auto-allocate new sound slots for future injection work.

#### Testing

- sample panning and distance calculation × tr1/tr2
- mic at Lara setting
- player-selected sound volume × tr1/tr2 × various sources (one shot, continuous)
- skidoo motor sounds and pitch changes
- the setting to disable random pitch in TR1
- regressions for #3768
- normal sound sources × tr1/tr2
- ambient sound sources × tr1/tr2 (flames, waterfalls, others?)
- uzis × tr1/tr2
- flip maps (lightning emitters? floods? 40 fathoms payload drop? sand drop? others?)

Ideally listen for differences between develop and this version.


lahm86 (commented):
TR2 is sounding good to me.

I think there are volume issues in TR1; Lara's footsteps can become very quiet and suddenly very loud. I've had them become completely silent too, but it's difficult to reproduce consistently. I was just running around Lost Valley when it happened, I think it was during the t-rex fight.

I think sound sources may be too loud compared to develop? A good example is Lost Valley if you head upstream from the start towards the cog machine, the water sound is prevalent and you struggle to hear Lara's feet. Another is Atlantis room 50, it feels much louder there comparing with develop.

rr-:
Thank you for testing. 

>I think there are volume issues in TR1; Lara's footsteps can become very quiet and suddenly very loud.

I'll see about this later today.

>I think sound sources may be too loud compared to develop? A good example is Lost Valley if you head upstream from the start towards the cog machine, the water sound is prevalent and you struggle to hear Lara's feet. Another is Atlantis room 50, it feels much louder there comparing with develop.

This might be related to the above point, or to this part:

>It also fixes a bug whether the master volume would be applied twice, resulting in quadratic volume loss for certain samples (so 10% would be rendered as 0.1%, 30% as 9% etc.).

Can you please compare the volumes of these sources at 100% sound volume?

lahm86:
Ah yes, I think at 100% the scion sound volumes are matching, so that must be that issue indeed.

rr-:
> Ah yes, I think at 100% the scion sound volumes are matching, so that must be that issue indeed.

Do you think we should revert this fix? Or make the og tr1 behavior uniform between tr1 and tr2?

lahm86:
I think making it uniform would be a great move. Perhaps we could also change the default shipped volumes to match (I think TR1 is 80%, TR2 100% - I personally have it at 50% in both but that's perhaps too low for most people).

rr-:
I've pushed an update, please LMK how this sounds now.
Some things I discovered when preparing this patch:
- TR1 sound cutoff is interpolated with a linear scale between 0…8 tiles, TR2 uses a quadratic scale between 1…10 tiles.
- Before the bugfix TR1 would use inverse square scale for certain sounds.
- TR1 forces the mic to use a mix of Lara's and camera's position, TR2 always uses camera's mic position
- TR1 doesn't stack continuous roll sounds, TR2 allows that
- TR1 updates effects volume on its own in `Sound_UpdateEffects`, TR2 requires calling `Sound_Effect()` for that. This could be fixed by making some TR1 samples (such as `SFX_LOOP_FOR_SMALL_FIRES`) use `SAMPLE_MODE_LOOPED` instead of `SAMPLE_MODE_WAIT`.

I opted to drop some of the `#if`s by keeping the TR1 linear scale parameters and TR2 quadratic scale parameters, but going with TR2 method of calculating the volume attenuation.

lahm86:
Unfortunately the issues persist on my end. I've uploaded a video below to demonstrate - you can hear the volume of Lara's footsteps and the uzis fluctuate, especially when near to sound sources like the Atlantis power-up FX. Some other sounds don't play at times, like levers and doors, and the clang-clang doors are intermittent.

I think the latest build is also applying random pitch to all sounds, regardless of the config option. E.g. if disabled, darts still have randomized pitch.

In the video I have sound at 60%, music at 0%. I have a longer playthrough to the end of Pyramid, that will take a while to upload.

https://www.youtube.com/watch?v=JlFyemwt6ek

aredfan:
Issues:

- [x] The skidoo motor sounds doesn't stop after Lara dismounts from the skidoo.
- [x] TR2: the Uzis sounds doesn't stop when Lara stops firing with the Uzis.
- [x] TR2: the M16 sounds doesn't stop when Lara stops firing with the M16.
- [x] TR2: fire sound effects can be heard while underwater
> 1. `/play 7`
> 2. `/tp 73 -3 51`

rr-:
I've undoed some of the merging, doing as much as I could before walking into regression land. I'd appreciate a quick check if the branch is good to proceed in its current state, and I'll move on to the trickier parts.

lahm86:
> I've undoed some of the merging, doing as much as I could before walking into regression land. I'd appreciate a quick check if the branch is good to proceed in its current state, and I'll move on to the trickier parts.

Looks good to me currently. Would you like this approved, or do you want to add on to this PR? Happy either way.

aredfan:
> I've undoed some of the merging, doing as much as I could before walking into regression land. I'd appreciate a quick check if the branch is good to proceed in its current state, and I'll move on to the trickier parts.

Sounds good to me. 👍

docs/tr1/CHANGELOG.md
```diff
@@ -1,8 +1,5 @@
 ## [Unreleased](https://github.com/LostArtefacts/TRX/compare/tr1-4.14.2...develop) - ××××-××-××
-
-## [4.14.2](https://github.com/LostArtefacts/TRX/compare/tr1-4.14.1...tr1-4.14.2) - 2025-09-07
-- fixed broken rendering in MacOS releases (#3880, regression from 4.14)
-- fixed images from MacOS releases (#3892, regression from 4.14)
+- fixed certain ambient sounds volume scaling wrong on non-100% volumes
```
walkawayy, line 2: I think you just need to fix the changelog since the 4.14.2 hotfix was released.

walkawayy (changes requested):
Code LGTM. I think you just need to fix the changelog since the 4.14.2 hotfix was released.

rr-:
> > I've undoed some of the merging, doing as much as I could before walking into regression land. I'd appreciate a quick check if the branch is good to proceed in its current state, and I'll move on to the trickier parts.
> 
> Looks good to me currently. Would you like this approved, or do you want to add on to this PR? Happy either way.

Thanks. I'll probably merge it to land in develop a sensible part that introduces no audible regressions.

>I think you just need to fix the changelog since the 4.14.2 hotfix was released.

Thank you for spotting this!

docs/tr1/CHANGELOG.md
rr-: @walkawayy updated

walkawayy (approved):

lahm86 (approved):

aredfan (approved):

PR #3897, opened 2025-09-10
Further sound module refactors before final merge

rr-:
#### Checklist

#### Description

This hammers the TR1 and TR2 sound modules into rough alignment, but stops shy of the final merge. The biggest remaining differences:
- `SAMPLE_MODE_WAIT` in TR1 requires `pos`, as some samples (eg explosions) are not using `SAMPLE_MODE_RESTART`
- the way distances and volume are calculated

Please LMK if there are any regressions – we focus on sound repeating, different sources etc. just like in #3890.

The commits are not spectacular in terms of their descriptions, but if we hit regressions, they'll allow me to find out which part of the refactoring went south. I mean to squash everything into a single commit later.

### DIFF
--- a/src/tr1/game/sound.c
+++ b/src/tr1/game/sound.c
@@ -361,20 +361,15 @@ bool Sound_Effect(
         break;
 
     case SAMPLE_MODE_LOOPED:
-        for (int32_t i = 0; i < M_MAX_ACTIVE_SOUNDS; i++) {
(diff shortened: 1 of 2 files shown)


lahm86 (approved):
Looks and sounds good to me 👍

aredfan (approved):

Issue #3898, opened 2025-09-11
Feature: allow more sample slots in TR1/TR2


rr-:
Both .phd and .tr2 use hardcoded arrays of `SFX_NUMBER_OF` size which no longer has to be the case.
In this ticket, we want to get TombEditor to support extra samples and emit sound injections at the end of the file.

PR #3900, opened 2025-09-11
Rewrite sound slots

### LINKED ISSUE #3284: Rewrite sound slots
The goal of this ticket is to let us cleanly inject models across games.

Some internal notes:

```c
typedef struct {
    UUID uuid;
    int32_t tr1_id;
    int32_t tr2_id;
} SOUND_ID_MAPPING;

typedef enum {
    SFX_INVALID = -1,
    SFX_WINSTON_FART,
    SFX_GORILLA_CHEST_POUND,
    SFX_NUMBER_OF,
} SOUND_ID;

static SOUND_ID_MAPPING m_SoundMap[] = {
    [SFX_WINSTON_FART] = { .uuid = …, .tr1_id = SFX_INVALID, .tr2_id = 350 },
    { 0 } // sentinel
};
// Likely defined  with some macros / .def setup

void Sound_Play(SOUND_ID id) {
    // try to look up the properties / wav data through m_SoundMap[id]
}

void Sound_LoadSample(SOUND_ID id, const char *wav_data)
{
    // associate WAV data with uuid or tr1/tr2_id… somehow…
}

static void M_ExampleUsage(void) {
    Sound_Play(SFX_WINSTON_FART); // for hardcoded calls
    Sound_PlayByGameID(350); // for animations and other arbitrary sounds
}
```

`m_SampleLUT` gets in the way but can be worked out, what's more annoying is that we have to ensure we don't cause `Audio_Sample` to decode the same data many times when they're aliased – so likely yet another indirection layer would be needed with this approach.

The big idea is that the level loader can then `LoadSample((SOUND_ID) { .tr1_id = 5 })`  and the injector can `LoadSample((SOUND_ID) { .uuid = … })`.
And the injector itself can do `LoadSample((SOUND_ID) { .tr1_id = 155959 })` for the samples that would now go beyond `SFX_NUMBER_OF` for animations and other things, too .

This approach requires us to readjust the relative numeric IDs within floor and animation data upon injection/reading the level.

It's possible that this approach is flawed and requires more thinking.

rr-:
#### Checklist

#### Description

Resolves #3284.
This rewrites the sound slots to use hash maps instead, which lets the injections use sound slots with arbitrary IDs. 

Just a refresher:
A single sample (such as `SFX_LARA_FALL`) holds volume and arbitrary number of data entries. This means two indexing schemes, one for samples (`SAMPLE_ID`) and one for data. Both schemes now use hashmaps.

Individual data entries can use -1 for their IDs and will be given new IDs increasing linearly when asked. If not, they're free to use direct indices (as compiled by the OG levels).
Samples still have to be assigned concrete numbers, so that they can be referred in animations etc. I'm open to making a refactor which allows negative numbers, and adjusts the animations from within the same injection file to refer to these, but this is a bit over my head. Also I'm not sure how this would play out with embedding multiple injections at the end of the level.

I removed the sample LUT public facades and scoped its work to level/common.c, as it was obscuring the API clarity _a lot_.

TR2 is left untouched. This PR will likely break injected TR2 sounds – @lahm86 can I ask to update TR2 .bin files to use the TR1 samples approach? This should be a good news since we're no longer bound by main.sfx limitations thanks to this.

Unfortunately .phd / .tr2 limitations will not let people use expanded slots out of the box  – this is tracked in #3898.


aredfan (approved):

rr-:
I've included the patch from Lahm and now TR2 injected sounds should be working again. (Thanks!)

lahm86 (approved):

aredfan (approved):

PR #3959, opened 2025-09-21
Improved volume settings

### LINKED ISSUE #3490: Feature request: volume option for FMVs and cutscenes
Currently the cutscenes and FMVs use the music volume which is a bit controversial; IMO this should become a separate setting.

### LINKED ISSUE #3707: Feature request: add separate slider for general ambience volume
Similar to #3490. Since it's possible to set the ambience volume separately from music for the inventory or underwater, it would be nice to be able to do so generally as well.

rr-:
#### Checklist

#### Description

Resolves #3707.
Resolves #3490.

- Moves all volume settings together
- Adds a setting to control general ambiance settings separately from music, not just inventory and underwater
- Adds a setting to control FMVs volume
- Adds a setting to control cutscenes volume
- Since it's getting busy, adds a setting to control master volume
- Fixes sound and music icon sprites were swapped


rr-:
Do you think we should also let the players set the values to >100%? With a disclaimer that this may sound in clipping

lahm86 (approved):
What do you think of having two tabs in `Sound Options`, one for `Volume` and another labelled perhaps `Misc` for everything else? I know we have other tabs with more settings than this, but it might be nice to group all the volume options together for clarity.

lahm86:
> Do you think we should also let the players set the values to >100%? With a disclaimer that this may sound in clipping

I think sticking with 100% max is OK personally.

rr-:
> What do you think of having two tabs in `Sound Options`, one for `Volume` and another labelled perhaps `Misc` for everything else? I know we have other tabs with more settings than this, but it might be nice to group all the volume options together for clarity.

Good call, added.

lahm86 (approved):

aredfan (approved):

PR #3994, opened 2025-10-04
lua: bootstrap level scripting

rr-:
#### Checklist

#### Description

@lahm86 I'd appreciate some thoughts on the design and suggestions on more functions to include in the initial version :relieved:  I'm thinking of:
- exposing phase control so that the lua can trigger something on every frame.
- exposing the config API but that is imo complex enough to warrant its own commit
- silent command execution that doesn't produce logs
- Console.Clear()
- ???

I've included example scripts that need to be removed at the time of merging but should make playing with lua easier.


aredfan (approved):

lahm86:
Looking fantastic 😄 Already, this is very powerful I'd say, being able to warp Lara opens up great possibilities, and the HP modifications settle a long-requested feature 😌

Re hitpoints, I was wondering initially if it would make sense to have a baked-in function that you could just pass the object ID and HP value to, but on considering it further, I don't think it's necessary. If users wish to modify several objects they can optimise it by having just one loop and branches. We may also need to expose something for Natla's HP in her second state. Or we could just alter the logic to use `obj->hit_points / 2`, which I think is the way it's done currently for the dragon in its second state.

I wonder if we should add at this stage the ability to have item string IDs, and the ability to get an item by its ID. Just as when building a level, item indices will change frequently so tracking script changes would become quite the task for a builder. Setting the IDs could be done via injections, I can take on that part. But equally, this could be slotted in later of course.

Having per-frame functionality would be brilliant. I agree re config, that would be useful but warrants its own work.

A couple of other potential functions to expose that come to mind at this stage:
- `Music_Play`
- `Sound_Effect`

rr-:
>I wonder if we should add at this stage the ability to have item string IDs, and the ability to get an item by its ID. Just as when building a level, item indices will change frequently so tracking script changes would become quite the task for a builder. Setting the IDs could be done via injections, I can take on that part. But equally, this could be slotted in later of course.

Let's do it! But maybe as a follow-up PR.

>Having per-frame functionality would be brilliant.

I'll add it here still.

>I agree re config, that would be useful but warrants its own work.

I'll follow up on that on another PR.

>Music_Play
>Sound_Effect

I'll add these as well, within this PR. 😌

rr-:
OK I updated the branch:

- Added new event type, `CONTROL` that runs on every logical frame
- Added new event type, `LEVEL_START` that runs after level completes loading and a potential savegame data is loaded too
- Added Music module
    - `TRX.Music.GetTrack()`
    - `TRX.Music.PlayTrack(id[, opts])`
    - `TRX.Music.Pause()`
    - `TRX.Music.Unpause()`
    - `TRX.Music.Stop()`
- Added Sound module
    - `TRX.Sound.IsAvailable(id)`
    - `TRX.Sound.Play(id[, opts])`
    - `TRX.Sound.Stop(id)`
    - `TRX.Sound.StopAll()`
- Added Config module
    - `TRX.Config.Get(key)`
    - `TRX.Config.Set(key, value)`
    - `TRX.Config.List()`
- Added `TRX.Console.Clear()`
- Fixed `TRX.Console.Log()` logging very confusing `(null)` for variables of complex types
- Changed `TRX.Console.Eval()` to work silently unless user chooses not to
- Changed Natla's near-death state to be her max HP/2 (**requires testing**)

lahm86 (approved):
Amazing! Tested Natla too and that looks good.

aredfan (approved):

PR #4056, opened 2025-10-16
catalog: simplify

rr-:
#### Checklist

#### Description

This update refactors catalog lookups to operate via enum values rather than UUIDs. UUIDs have been removed and are now bypassed during injections. Consequently, the related UUID handling can be cleaned up from both the Rando library and the Injections.

Also makes the changelog more coherent (thanks aredfan!).

If we completely skip emitting UUIDs, the following diff should be applied:

```diff
diff --git a/src/libtrx/game/inject/utils.c b/src/libtrx/game/inject/utils.c
index bef73e061..d857cea54 100644
--- a/src/libtrx/game/inject/utils.c
+++ b/src/libtrx/game/inject/utils.c
@@ -9,9 +9,5 @@ INJECTION_OBJECT_INFO Inject_ReadObjectPtr(VFILE *const fp)
         .id = VFile_ReadS32(fp),
     };
 
-    if (obj_info.type == OBJ_TYPE_OBJECT) {
-        VFile_Skip(fp, 16);
-    }
-
     return obj_info;
 }
```


rr-:
I think we'll need to update teatime injection to emit the new numeric ID in the non-UUID field.
It might be a good idea to regenerate all .bin while at it?

lahm86:
> I think we'll need to update teatime injection to emit the new numeric ID in the non-UUID field. It might be a good idea to regenerate all .bin while at it?

`INJECTION_OBJECT_INFO` already contains the numeric ID, the current injections output both as-is. We could uprev and have the skip in place for older versions though, it'd save having to regenerate them all, wdyt?

rr-:
> > I think we'll need to update teatime injection to emit the new numeric ID in the non-UUID field. It might be a good idea to regenerate all .bin while at it?
> 
> `INJECTION_OBJECT_INFO` already contains the numeric ID, the current injections output both as-is. We could uprev and have the skip in place for older versions though, it'd save having to regenerate them all, wdyt?

Makes sense. However something's not right as with this change, /teatime no longer works in TR1 ❓

lahm86:
Ohhh, I see - the ID is now the same between 1 and 2. That makes sense, sorry, it didn't quite register with me. So yes you're right, we'll need a new TR1 injection, I will check the others too. Currently `O_LARA_M16` is getting the injection 😄

rr-:
Updated as discussed on Discord

docs/tr1/CHANGELOG.md
```diff
@@ -3,9 +3,9 @@
 - added a splash effect when Lara jumps in wading depth water, similar to TR3+ (#3975)
 - added a new easter egg command
 - added bounding box debugging (`/debug 1` or `/set debug-cuboids 1`)
-- added support for object ID, music ID, sound ID, state ID and animation ID catalog overrides
-    This lets the builders to choose which slots to use to provide the game with assets the game uses.
-    In the future, this feature can be used to get "hard-coded" TR2 sounds to work in TR1, and vice versa.
+- added support for object, music, sound, Lara state, and Lara animation slots overrides through CSV catalogs
+    Lets builders link hardcoded logic to slots of their choice, allowing object sharing between games (for example, use TR1 bats in TR2).
```
lahm86: Nitpick, these don't appear on new lines/as a child-list (TR2 changelog as well)

rr-: Fixed

lahm86 (approved):

aredfan (approved):

lahm86:
@rr- I  just realised, `inject/common.c` will need `M_INJECTION_CURRENT_VERSION` up-revved as well.

The injection tool PR is here for the change:
https://github.com/LostArtefacts/TRXInjectionTool/pull/132

rr-:
@lahm86 updated, LMK if it's good.

lahm86:
> @lahm86 updated, LMK if it's good.

Looks good, thanks. Tried a newly built injection and all is well with it and V4.

PR #4069, opened 2025-10-17
More lua

### LINKED ISSUE #4040: Feature Request - Cold Wind Exposure
Related to the closed #4022. 
Allow the builder to enable a cold air, the counterpart of cold water, where Lara's cold bar would decrease when in rooms with the *Wind* flag toggled on. For the duration, I would say it could be made to last [LARA_MAX_EXPOSURE*2], aka twice as long as the cold water max exposure, or longer. I request this in light of the cold air feature being added. 

If I also may suggest:
- Standing would decrease the bar the fastest, while running would slow down the decrease rate, with sprinting making it go down slower to account for the logistics of body heat... Obviously, feel free to ignore the entire issue. If you feel like this feature would not fit into TR2X feel free to close it.

rr-:
#### Checklist

#### Description

- Fixes corrupt strings logged by `TRX_ConsoleLog` and the family (double use of the same `__VA_ARGS__` is illegal)
- Adds `trx.lara.exposure_bar`, allowing builders to build manual interactions with the cold bar (resolves #4040) – I'd appreciate some testing of this
- Adds a new `trx.game` API
- Adds a new `trx.rooms` API
- Changes LUA logs to not emit `,` between multiple arguments – just a single space (similar to Python and LUA REPL which chooses `\t` instead – but tabulators are irrelevant for us)
- Simplifies lara item access from `trx.lara.get_item()` to `trx.lara.item`
- Fixes `trx.items.Item.room_num` using 0-indexing
- Internally, adds a new `GFLT_TITLE` table to be able to map `GF_CurrentLevel()` back to the title level (all FFI accessors use level table lookups)

### DIFF
--- a/data/common/ship/scripting/trx/game.lua
+++ b/data/common/ship/scripting/trx/game.lua
@@ -26,6 +26,15 @@ trx.game = setmetatable({
       t = make_levels(trxc.game.LevelTable.DEMOS)
     elseif key == "cutscenes" then
       t = make_levels(trxc.game.LevelTable.CUTSCENES)
+    elseif key == "current_level" then
+      local table_type = trxc.game.get_current_level_table()
+      local i = trxc.game.get_current_level_idx()
+      return {
+        num = trxc.game.get_level_num(table_type, i),
+        name = trxc.game.get_level_name(table_type, i),
+        path = trxc.game.get_level_path(table_type, i),
+        type = trxc.game.get_level_type(table_type, i),
+      }
     elseif key == "version" then
       return trxc.game.get_version()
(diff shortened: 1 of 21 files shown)


aredfan (approved):

lahm86 (approved):
Looks great 😌 A couple of things that would be nice to have in `item` would be current_anim_state and (relative) anim idx and frame idx. I tried the exposure example in the request ticket and this works really well, but having those extra properties would let the builder change things based on what Lara is doing.

rr-:
Thanks – I'll expose the mentioned properties in a follow-up MR.

PR #4084, opened 2025-10-20
Merge inventory ring control phase; game flow partial tidy-up

rr-:
#### Checklist

#### Description

- Merges `Inventory_Control()` of both games
- Removes `cmd_*` game flow commands from TR2 game flow, as they were a bit bogus, and I'm hoping for LUA to allow more flexibility eventually
- Ports `lockout_option_ring` option to TR1 game flow
- Ports `load_save_disabled` option to TR1 game flow
- Exposes both options via LUA

#### Testing

- Behavior of both options
- Whether they're read properly via game flow
- Whether they can be set from lua
    - `/lua trx.game.settings.lockout_option_ring = true/false`
    - `/lua trx.game.settings.load_save_disabled = true/false`
- Death from natural causes in gym (`/hp 0`)
- Waiting for demos
- Whether pressing input in demo ends it properly
- Ammo counts in the inventory ring in both TR1 and TR2


aredfan:
If `lockout_option_ring` is enabled in TR1:

- `New Game`/`Load Game` is accessible when pressing F5/F6 in the gym.
- `Save Game`/`Load Game` is accessible when pressing F5/F6 in the story levels.


In the case of TR2, pressing F5/F6 would open the inventory with the stopwatch as the focus.

rr-:
>In the case of TR2, pressing F5/F6 would open the inventory with the stopwatch as the focus.

I can't replicate this
Edit: I had to disable save/load. Now it should open Compass like in TR2.

>If lockout_option_ring is enabled in TR1:

This behaves for me the same way in both games.
However, I'm thinking, if someone wants to disable saves, they have the other option for that. I tried not to duplicate the functionality. But maybe that's misguided and it should block quick save/load too?

lahm86:
If you have crystals enabled and save/load disabled, pressing action at a crystal opens the passport but it stops before opening completely and appears frozen (it's not, you can just escape out). I suppose it's a grey area really though, but maybe the crystal collision could do a check before opening the passport?

rr-:
> If you have crystals enabled and save/load disabled, pressing action at a crystal opens the passport but it stops before opening completely and appears frozen (it's not, you can just escape out). I suppose it's a grey area really though, but maybe the crystal collision could do a check before opening the passport?

Good catch. I made Lara say "no" in such scenarios. Cheeky and evil IMO

rr-:
I also added a change to the save crystals' bounding box to make them easier to activate / interact with.

lahm86 (approved):

aredfan:
> However, I'm thinking, if someone wants to disable saves, they have the other option for that. I tried not to duplicate the functionality. But maybe that's misguided and it should block quick save/load too?

I'd go with your approach. I think it's a good move to have a clear difference between the two options.

The less overlap, the better.

aredfan (approved):

PR #4120, opened 2025-10-22
Add Play Any Level to TR1

### LINKED ISSUE #565: Feature request: Play Any Level

rr-:
#### Checklist

#### Description

This does PR several things:
- Moves the `lockout_option_ring` from the game flow to a hidden user setting
- Moves the `load_save_disabled` from the game flow to a hidden user setting
- Moves the `play_any_level `from the TR2 game flow to a hidden user setting
- Implements the TR2 "play any level" feature in TR1 (it allows the player to select the level to play – in place of the new game dialog)
    - This is different from "Select Level" known from TR1 which are bounds to the savegames, and let you also use "story so far" and only play up to the level you unlocked
- Adds game mode selection to the "play any level" dialog to both games

The benefits of moving the game flow settings to user config is that
1) They're now saved across game relaunches
2) There's less code to maintain them as they use faimiliar APIs
3) They now can be discovered and toggled at runtime with the `/set` command, making testing that more convenient

I plan to move some more game flow settings this way (I'm thinking `demo_delay`, `cheat_keys`, `enable_killer_pushblocks`.
I'd like to remove the `single_level` and `is_demo_version` game flow settings.
LMK your thoughts.


lahm86:
In develop on TR2, if you have play any level enabled and jump to say Opera House, Lara will have pistols, shotgun etc  - similar to when we use `/play 4`, it applies the GF mods up to that point. In the PR, her inventory is stripped. Perhaps we could offer the same, although I understand if this is going to conflict with the TR1 select level approach.

Fully agree with moving the other GF options. And I also agree that `single_level` and `is_demo_version` are redundant.

rr-:
> In develop on TR2, if you have play any level enabled and jump to say Opera House, Lara will have pistols, shotgun etc - similar to when we use `/play 4`, it applies the GF mods up to that point. In the PR, her inventory is stripped. Perhaps we could offer the same, although I understand if this is going to conflict with the TR1 select level approach.

Thanks – fixed. 

> Fully agree with moving the other GF options. And I also agree that `single_level` and `is_demo_version` are redundant.

I'll tackle this separately 🙂

lahm86 (approved):

aredfan (approved):

PR #4126, opened 2025-10-22
savegame: make select level available in tr2

rr-:
#### Checklist

#### Description

This PR enables the Select Level feature in TR2.
It's a less-than-ideal change, though – the logic is in place, but there's no UI and it can only be triggered by pressing the Look button while a save is selected. Which makes this feature effectively undiscoverable to players. TR1 still uses the Right Arrow binding, as it doesn't clash with the passport flip logic in it.

Some failed UI attempts, that took inspiration from #545:

- Just replacing the Right Arrow with the Look binding

  <details>
  Looks OK in TR1…
  20251022_202429_Level_0

  …but TR2 uses a different item alignment. No dice.
  20251022_204300_Level_0
  </details>
- A 3-button interface

  <details>
  Looks very busy in TR1, but it's somewhat passable…
  20251022_201734_Level_0

  …improved order to be pseudo-compatible with PS1. Green on green – contrast king! No dice either:
  20251022_214451_Level_0

  …lower placement – still the same:
  20251022_215655_Level_0

  Also, text scaling says hi
  20251022_220059_Level_0
  </details>
- Going with just a single button

  <details>
  Aligned to center:
  20251022_214623_Level_0
  
  Aligned left:
  20251022_215016_Level_0
  
  This being the only hint available, it puts too much focus on this quite niche feature, and sticks out like a sore thumb.
  </details>

At this point I just gave up. I'm happy to try new ideas, though.

Other (from bad to very bad) ideas that I had:
- Make TR2 dialogs use 2-stage approach, where you have to press Action to invoke the dialog. Then the Right button can navigate to the details just fine.
- Make the bottom space an either-or: either you get the strip with the hints, or the "← Load Game →" CTA.
- Swap the places: make the "← Load Game →" above an optional button strip with the hints.
- Just keep the right arrow, and require the players to use sidestep buttons to navigate between passport pages.


lahm86 (approved):

Richard-L:
Thank you so much for iterating so in-depth here, and for continuously holding everything to a high standard. The community will thank you for it.

Other than what I'd already suggested on Discord, I had one more thought:

image

- Always keep `Select` and `Go Back` consistent as the top two navigational options.
- Add any contextual additional options in a second row below.

I reckon this would work with TR1 as well.

PS1 for quick reference:

image

Richard-L:
I noticed one thing in TR1 that surprised me:


https://github.com/user-attachments/assets/7431298a-edc3-40e1-952a-b3891a05cc60

- Go to load game
- Enter select level
- Hit "inventory" to go back twice
- Open new game
- Hit "inventory" again

The game will unexpectedly offer a select level of all levels.

rr-:
> I noticed one thing in TR1 that surprised me:
> 
>  2025-10-23.00-28-39.mp4 
> * Go to load game
> * Enter select level
> * Hit "inventory" to go back twice
> * Open new game
> * Hit "inventory" again
> 
> The game will unexpectedly offer a select level of all levels.

Fixed, I think

aredfan:
**Issues:**

- [x] While doing partial testing of TR2's Select Level in an older PR build (1.5.1-261-gfc47f12-win), I noticed Lara only has pistols in TGW and Venice, then her inventory is stripped bare in later levels.
- [x] On the current PR build, when I open Select Level, then select a level e.g. TGW, this takes me to the Select Mode screen.  I'm unable to proceed further because pressing Action on New Game has no effect.
- [x] Following the same steps as above, when I deselect the Select Mode, screen,  The Select Level section appears to have asserted itself as a main passport page. Closing the passport seems to fix the issue.

> <details><summary>Screenshot</summary>
> 
> 20251023_072353_Level_0
> 
> </details>

**Edit:** the latter 2 issues appear to be present in develop for TR1X.

rr-:
@aredfan I think everything is fixed in the latest version of this PR.

aredfan:
> @aredfan I think everything is fixed in the latest version of this PR.

The first issue is TBD.
The second issue is fixed, thank you. 👍 
The third issue persists.  To reproduce:

> 1. Go into Load Game → Select Level → Select Mode, but don't select New Game
> 2. Deselect Select Mode and you'll return to Select Level, you can now flip the passport page to the right which is incorrect
> 3. Deselect Select Level and the passport closes instead of returning to the Load Game page

**New issue:**

- [x] In the Level Select menu, if I select any level after Caves or TGW, it will always load the first level of the game.

rr-:
Thanks @aredfan – because the bugs were quite interlocked, rather than going after every one at the risk of introducing even more elusive regressions, I decided to implement the UI we had talked about yesterday. This resulted in a rewrite of the passport code to better support hierarchical navigation. I'm hoping whatever bugs we find, they're easy to fix.

- Select Level function was renamed to Play Previous Levels.
- The right-arrow based navigation was completely scrapped, as well as the Look button  experiments. Instead, the function is now available in the New Game menu as a fifth item.
- Likewise, Story So Far was moved from the Select Level dialog, and instead is now placed as the sixth item in the New Game menu.

There are several massive benefits of this decision:
- These functions are now impossible to miss, unlike the previous subtle arrow hint. I'm pretty sure a lot of players will discover this feature for the first time.
- It's now possible to implement #545 more faithfully, without having to think about a lot of additional actions.
- It's now possible to implement TR2's immediate dialog mode in TR1 and vice versa, without having to think of extra selection keys. @walkawayy 
- The UI is straightforward to use. No extra weird keys, no nothing.

Since not everyone is going to like this feature, it is possible to turn off the extra items in the New Game menu with a new setting: Gameplay → General → Play previous levels. It affects both options.

Gotchas:
- The Play Any Level is at odds with the Play Previous Levels menu. It completely overrides the New Game menu with a level selection, followed with a mode selection.
- If the player has disabled the Game Modes, but enabled Play Previous Levels, the dialog will still show up. To make it possible to actually start a new game, it will offer also a single row, "New Game", which in this special scenario should not touch the NG+ flag – if it was active, it stays active, if it was not, it stays too.
- Likewise, if both Game Modes and Play Previous Levels are disabled, the New Game dialog will not show anything – the game should start immediately, without touching the currently set NG+ flag.
- Continuing on, if there are no saves, neither Play Previous Levels or Story So Far will show up.
- If none of the saves support the Play Previous Levels feature (basically OG saves), this option will not be shown in the menu.
- If none of the saves support the Story So Far feature (OG saves and TombUB, which has no FMVs or cutscenes), this option will not be shown in the menu, either.
- In the unfortunate event that none of the saves support the Play Previous Levels _and_ story so far, and the player has disabled game modes, the New Game dialog will only show one item: "New Game". This is a corner case scenario that I believe becomes moot the moment the player creates a new BSON save.
- If there are saves with either feature, but a player selects an incompatible save slot in the second step, an error message should show saying "This save does not support this feature.". It should be possible to close this error message both with Escape _and_ Action keys. Escape navigates back, Action closes the passport IIRC.
- TR2 does not support Story So Far and needs a separate PR.
- Traversing dialogs in and out shouldn't cause any visible transitions – the passport should show dialogs at any time.

### Testing

It's good to test save crystals and death mode ring.
I'd also appreciate trying to "jailbreak" anything passport-related in both TR1 and TR2 by selecting wrong save slots, hitting Escape or Action at unexpected times, etc.

Richard-L:
Thank you for the continuous significant effort put into the new level selection functionality; it feels like this represents a major change. I've tested it and wanted to offer some observations on the design coherence and user experience.

If you'll allow, I wanted to carefully voice concerns about design rationale and clarity. My main suggestions concern the feature's name and placement to align better with the OG and core logic (at least to my understanding):
1. Menu Naming: Could we revisit "Play previous levels"? The original, concise name, "Select Level," is functionally clearer and helps maintain that vital connection to the OG feel.
2. Menu Placement: Integrating this feature under "New Game" seems counter-intuitive. Since level availability depends on an existing save file and progression, it's more logical for the player flow to be Select Save $\rightarrow$ then Level, as was the previous implementation. A 'New Game' implies a fresh start, not accessing data from a playthrough.

I've noted a few specific things I wanted to mention:
- **TR1** - 'Save Not Compatible' Prompt: When this overlay appears, pressing Action currently closes and immediately reopens the ring menu. It should instead return to the save selection screen to allow the player to pick a different save file.
- **TR1** - 'Story So Far' Content: I had a case where the loading screens were cycled and appeared after one another. In the OG, the loading screens appear when it loads the cutscenes. Is there a chance the cutscenes currently do not play in "Story so far"? I wonder if it would just be okay to drop the loading screens from Story so far, and simply keep them as a nice visual when the player loads a save.
- **TR2** - Passport Controls in Level Select: Inside the 'Select Level' menu, the use of Left/Right incorrectly persists and allows flipping the passport, which exits the view. These controls should be limited to Select and Go Back within this specific screen (This is the case in TR1 in fact, where however the "back" icon is still visible on the left side of the save highlight bar).
- **TR2** - Inventory on Level Load: When loading a previously played level, the inventory currently resets to being empty. It nicely reassumes the inventory in TR1 however.

I hope none of this is triggering dash 🙏

aredfan:
**Issues:**

- [x] **TR1/TR2:** New Game → Play previous levels → Save slot → level list
There's an arrow pointing to the left by the highlighted level but it's no longer effective, thus it might be worth hiding said arrow.
- [x] **TR2:** When using `Play previous levels`, Lara only has pistols when loading TGW or Venice, and her inventory is empty in Bartoli's Hideout and Opera House.
- [x] **TR1:** The Restart level is selectable more than once, it's as if it's trying to open a dialog.
- [x] **TR2:** There's an issue with the passport starting a new game unprompted with the following steps:

> 1.  open TR2X.json5 and set "new_game_plus_unlock": to false, then save and close
> 2. launch TR2X
> 3. open the PDA and disable both `Game mode selection` and `Play previous levels`
> 4. if saves are present, open the passport and flip to the right, and a new game starts unprompted
> 5. if no saves are present, simply opening the passport will start a new game unprompted

aredfan:
All the prior issues are resolved, thank you.

- [x] I've found a new bug. This one is similar to my last issue where New Game would start unprompted, but it affects TR1 instead.

> 1. open TR1X.json5 and set "new_game_plus_unlock": to false, then save and close
> 2. launch TR1X
> 3. open the PDA and disable both `Game mode selection` and `Play previous levels`
> 4. load the gym
> 5. press F5 (or F6 if no saves are present)
> 6. a new game starts unprompted

lahm86 (approved):

rr-:
@Richard-L Thanks for taking the time to test it.

> Menu Naming: Could we revisit "Play previous levels"? The original, concise name, "Select Level," is functionally clearer and helps maintain that vital connection to the OG feel.

We deliberately went with "Play Previous Levels" because the original wording is ambiguous. Select Level doesn't clearly describe what's happening – why a save slot is required or why only previously reached levels appear. It reads more like a debug label than part of the Passport UI.

> Menu Placement: Integrating this feature under "New Game" seems counter-intuitive. Since level availability depends on an existing save file and progression, it's more logical for the player flow to be Select Save.

We placed it under New Game after an extended design review and multiple prototypes (some shared in #general, later refined internally). Placing it under Load Game caused issues with discoverability or visual clutter, or added unnecessary friction for the vast majority of players who don't use this feature.
The earlier proposal with triple button hints actually diverges even further from the original design – which relied on a single memory-card slot, two simple button prompts (Select / Go Back), and no special "hidden" navigation with special keys such as Look. This approach preserves that logic while improving clarity for modern users.

We think this setup hits the sweet spot between OG faithfulness and usability, so we’re sticking with it for now.

On a separate note – and this is important – I'd like to ask for a bit more care with tone in written feedback. Phrases like "hope none of this is triggering" or "I hope you don't mind something quite nitpicky, but it's important we don't lose touch with the OG" come across as condescending and unnecessarily personal. They frame the discussion as if one side is emotionally unstable or careless about legacy, which isn't acceptable in a collaborative project.
Constructive criticism is always welcome, but it needs to stay focused on the work itself – not on people's temperament or presumed philosophy.
﻿
Going forward, please also avoid commenting directly on pull requests. Code reviews are handled by maintainers; broader feedback should go to Issues, Discussions, or Discord.

---

@aredfan I think the last version should fix everything mentioned so far. The rewrite was a good decision IMO, it's now much easier to zoom in on the issues and fix them reliably :relieved:

aredfan:
LGTM, and thank you. 👍 

I have a couple more corner cases left to report, these are reproduceable with the following settings.

> "new_game_plus_unlock" set to false
> `Game mode selection` is disabled
> `Play previous levels` is disabled

- [x] **TR1:** In the gym, open the inventory then open the passport.  The New Game page will need to be selected twice, it's as if it's attempting to open a dialog.
- [x] **TR2:** On the main menu, open the passport to the New Game page will start the game unprompted.

rr-:
It's quite difficult to get this condition right but I think it now satisfies all 4 scenarios:
- TR2 title mode
- TR2 gym
- TR1 title mode
- TR1 gym

(with all the options disabled).

aredfan (approved):

walkawayy:
Looks nice. One small inconsistency I noticed is `New Game` is title case while `Play previous levels` is sentence case with only `Play` capitalized.

I have a suggestion for the New Game menu. What do you think about breaking up the New Game menu into two menus? You flip to the `New Game` page and there are 2 options:
1. New Game
2. Play Previous Levels

Selecting `New Game` opens a sub menu with:
1. New Game
2. New Game+
3. Japanese NG
4. Japenese NG+

Selecting `Play Previous Levels` functions as it does now where you choose a previous savegame and then pick the level.

It would add one extra click to the `New Game` menu (for TR2; for TR1 it's a net zero since the page select got removed to behave more like TR2). But to me it makes sense to group the NG options together. At a glance it feels weird to lump the level select feature below the multiple NG options.

rr-:
I don't think introducing additional levels of navigation is acceptable.

I've made the additional items clearly separated, which should justify the capitalization change (as Play Previous Levels reads silly, at least to me) and made it more clear that these aren't exactly new game modes.

image
image
image
image

lahm86:
I think this looks great! Reckon we should merge as-is, given there isn't any loss in functionality and the features are now much more visible to players - it's pretty cool if people will discover it from here on out.
Also a big step towards our engine-merging goal while retaining and improving features like this, so we're moving in the right direction :relieved:
Thanks for working on this @rr-

PR #4137, opened 2025-10-24
items/actions: fix cutscene mesh swaps

### LINKED ISSUE #4136: OG bug: Lara's holsters don't update in cutscenes
E.g. in the GW cutscene, when Lara draws her pistols, they appear in her hands, but also remain in her holsters.

<details>

Image
<details>

### LINKED ISSUE #4142: OG bug: Lara doesn't have pistols holstered in some parts of the Diving Area cutscene
<details><summary>When we see Lara in a swimsuit for the first time, there's no pistols in her holsters.</summary>

Image

</details>

<details><summary>After Lara's exchange of gunfire with Bartoli, there's no pistols in her holsters.</summary>

Image

</details>

### LINKED ISSUE #4145: TR2X bug: Lara's pistols disappear when holstered at the end of TGW cutscene
Regression since [0.9](https://github.com/LostArtefacts/TRX/releases/tag/tr2-0.9)

lahm86:
Resolves #4136.
Resolves #4142.
Resolves #4145.

#### Checklist

#### Description

This corrects mesh swap behaviour in cutscenes for Lara so that status is more accurately tracked. The flip effects are used in the data as toggles, but the difference when drawing/undrawing in game is that Lara's gun status is used to track mesh status effectively. The mesh pointers are now compared in cutscenes in each toggle call.


lahm86:
@aredfan pointed out this is OG - I wonder if this will cause issues elsewhere then as I don't think there is an "undraw" flip effect.

rr- (approved):

rr-:
I think we can just monitor for new regressions – this looks like a very reasonable change.

lahm86:
The effect is only used in cutscenes, so I think it means where you might normally see guns reholstered, they'll no longer be. But I think we could add a new effect quite easily to make it more realistic. I'm thinking Atlantis, GW and Diving Area - aredfan can keep me right though, that's off the top of my head.

rr-:
Sounds good!

aredfan:
> The effect is only used in cutscenes, so I think it means where you might normally see guns reholstered, they'll no longer be. But I think we could add a new effect quite easily to make it more realistic. I'm thinking Atlantis, GW and Diving Area - aredfan can keep me right though, that's off the top of my head.

Thank you for checking, I think you've listed all the relevant cutscenes. 👍 

There's another case worth looking into, it's the part when Lara changes into her swimsuit and her holsters are empty.

_____________

**Edit:** Things to check:

- [x] Atlantis cutscene
- [x] TGW cutscene: does Lara have her pistols holstered at the end? (TR2X bug since 0.9)
- [x] DA cutscene: does Lara have her pistols when she changes into a swimsuit? (OG bug)
- [x] DA cutscene: does Lara have her pistols holstered at the end? (OG bug)

aredfan (approved):

lahm86:
I've updated the description above and this should now fix all the snags with mesh swaps in cutscenes. I've also fixed the textures in DA cutscene as those were very jarring. Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/142

rr- (approved):

aredfan (approved):
Thank you very much @lahm86! 😌

PR #4154, opened 2025-10-25
misc: convert TR1 cutscenes to TR2 format

lahm86:
#### Checklist

#### Description

This updates TR1 cutscene data to make sure that Lara (as `O_LARA`) is always present, meaning the logic can be aligned to TR2 in terms of setup, control and camera behaviour. This brings in a few changes as a result:
- Lara's shadow is now visible in cutscenes
- Lara's lighting in the Atlantis cutscene is slightly different
- Mesh swap, flip map, and camera pos/angle game flow sequence events are now redundant - everything feeds off the data (TR2 as well, no data changes needed)
- Lara's braid injection is much simplified as the cutscene meshes don't need specific handling

Lara's setup and control remain separate for now to allow for testing, although these are practically identical so that should be an easy next task.

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/141


rr- (commented):
Thank you – this looks fantastic 🩷

However, I'd either enable the shadows for all cutscene players, or disable it in TR1, as the Larson cutscene looks unfortunate. It should be possible to target it with `GF_GetCurrentLevel()->type == GFL_CUTSCENE`. LMK WDYT?

lahm86:
> Thank you – this looks fantastic 🩷
> 
> However, I'd either enable the shadows for all cutscene players, or disable it in TR1, as the Larson cutscene looks unfortunate. It should be possible to target it with `GF_GetCurrentLevel()->type == GFL_CUTSCENE`. LMK WDYT?

How about a new flip effect (item action) to switch it off? That way TR2's shadow for Lara in cutscene remains, but we can get rid in TR1. The problem with the other actors is we'd need to define shadow size and they're of course different across the files.

rr-:
On that end, maybe we could make the flip effect to set the shadow size to make it more versatile? And setting it to 0 disables it completely.

lahm86:
That would be ideal, but flip effects don't have scope for args unfortunately (it'd need an entirely new animation command). Maybe the on/off support would be OK for now and in the future Lua could manage `obj->shadow_size`.

Added that change for now, LMK if OK.

rr- (approved):
LGTM and thank you 😌

aredfan (approved):

PR #4155, opened 2025-10-25
tr2/objects: move vehicles code to trx

rr-:
#### Checklist

#### Description

This moves the TR2 vehicle code to TRX, hopefully enabling the support for the vehicles in TR1.
I struggle with creating a test level, though and would appreciate help 🩷

Test level available in `test-tr1-vehicles.zip`. Required TR1 catalog update:
```
195, O_BOAT
196, O_LARA_BOAT
197, O_SKIDOO_FAST
198, O_LARA_SKIDOO
199, O_SKIDOO_DRIVER
200, O_SKIDOO_ARMED
201, O_SKIDOO_TRACK
```


rr-:
@aredfan the last change has the potential to affect TR2 demos – from my testing it doesn't appear anything is wrong, but I'd appreciate if you could work your magic too :)

lahm86 (commented):
Nice!

I think we just need this to avoid a crash if Lara tries to fire the black skidoo guns.
```diff
diff --git a/src/libtrx/game/gun/common.c b/src/libtrx/game/gun/common.c
index bc6200ce6..8231a1b3f 100644
--- a/src/libtrx/game/gun/common.c
+++ b/src/libtrx/game/gun/common.c
@@ -146,8 +146,8 @@ AMMO_INFO *Gun_GetAmmoInfo(const LARA_GUN_TYPE gun_type)
     case LGT_HARPOON: return &lara_info->harpoon_ammo;
     case LGT_M16:     return &lara_info->m16_ammo;
     case LGT_GRENADE: return &lara_info->grenade_ammo;
-    case LGT_SKIDOO:  return &lara_info->pistol_ammo;
 #endif
+    case LGT_SKIDOO:  return &lara_info->pistol_ammo;
     default:          return nullptr;
     }
     // clang-format on
```
She also can't target the driver, but that's a bigger can of worms I understand because of the differences in how the games target enemies.

I think there is a desync in the Venice demo, Lara ends up too far from the switch at the very end to be able to pull it.

rr-:
> Nice!
> 
> I think we just need this to avoid a crash if Lara tries to fire the black skidoo guns.
> 
> ```diff
> diff --git a/src/libtrx/game/gun/common.c b/src/libtrx/game/gun/common.c
> index bc6200ce6..8231a1b3f 100644
> --- a/src/libtrx/game/gun/common.c
> +++ b/src/libtrx/game/gun/common.c
> @@ -146,8 +146,8 @@ AMMO_INFO *Gun_GetAmmoInfo(const LARA_GUN_TYPE gun_type)
>      case LGT_HARPOON: return &lara_info->harpoon_ammo;
>      case LGT_M16:     return &lara_info->m16_ammo;
>      case LGT_GRENADE: return &lara_info->grenade_ammo;
> -    case LGT_SKIDOO:  return &lara_info->pistol_ammo;
>  #endif
> +    case LGT_SKIDOO:  return &lara_info->pistol_ammo;
>      default:          return nullptr;
>      }
>      // clang-format on
> ```

Thanks – added.

> She also can't target the driver, but that's a bigger can of worms I understand because of the differences in how the games target enemies.

I'll see if I can do something about it :eyes: 

> I think there is a desync in the Venice demo, Lara ends up too far from the switch at the very end to be able to pull it.

Yeah, that's what I was afraid of. I patched with a more secure version.

lahm86 (approved):

aredfan (approved):
> @aredfan the last change has the potential to affect TR2 demos – from my testing it doesn't appear anything is wrong, but I'd appreciate if you could work your magic too :)

No worries, I've taken a look at the 3 demos from TR2 and they work as expected. 👍 

The boat and skidoo both LGTM.

rr-:
@lahm86 @aredfan I've pushed a fix for the issue Lahm found where Lara couldn't correctly target the Black Skidoo Driver. The update removes the active item iteration and instead loops through pathfinding slots to pick the right enemy target.

From what I've seen, the only TR1 objects that can be targeted but don't use pathfinding are:
- Bacon Lara
- Scion
- Mummy in Qualopec

The simplest solution would be to set the `intelligent` flag on each, which works for new saves, but it breaks loading in older ones. To avoid that, I added a check for an active trigger in each item's `M_Control` routine. That seems to activate them properly without causing any other issues.

This change also impacts how TR1 enemies choose their targets.

I'd really appreciate a retest on this, ideally using some original saves close to the affected objects.

aredfan (approved):

lahm86 (approved):

PR #4166, opened 2025-10-26
Fix dome portals

### LINKED ISSUE #2685: TR1X - Matrix Stack Overflow
I'm getting a Matrix Stack Overflow error when dropping from one room to another in a custom level. Happens on 4.5.1 (Ships with Tomb Editor) and the latest dev build (4.8.3-256-g2adaf2a), albeit without the assertion error. When I added the small connecting drop between the corridor and the ceiling of the large room, I added the portal normally by just selecting the surface and pressing P

Interestingly, the crash doesn't happen when I just throw lara down the drop, only when I do a safe drop down as seen in the video

Youtube link since I can't get the video under 10mb to attach to the issue, apologies.

https://youtu.be/Ap9AOy4_Cj4

[TR1X.log when using 4.5.1](https://github.com/user-attachments/files/19476747/TR1X.log)

[TR1X.log when using 4.8.3-256-g2adaf2a](https://github.com/user-attachments/files/19476787/TR1X.log)

rr-:
#### Checklist

#### Description

Resolves #2685. The main culprit was the early exit done by the portal normal facing vector check. Removing it fixes the issue and is just 2 lines, but there were still some issues left around how the bounding boxes get computed, so more changes had to follow.

Test level from Lahm: `vis-portal-issues.zip`

This is a pretty fundamental change in how all rooms get scheduled so it might bite us when we least expect. I would appreciate a check if this affects CPU % usage in big levels with lots of rooms (Floater, The Deck). Still, from early testing, it looks good:

<details>
20251026_123646_Test_Level
20251026_123732_Test_Level
20251026_123751_Test_Level
20251026_123851_City_of_Vilcabamba
20251026_123947_City_of_Vilcabamba
</details>

### DIFF
--- a/docs/tr1/CHANGELOG.md
+++ b/docs/tr1/CHANGELOG.md
@@ -53,6 +53,7 @@
 - fixed "Story So Far..." showing loading screens
 - fixed photo mode camera clipping through overlapping rooms
 - fixed matrix stack overflow crash when moving through overlapping or dome portals (#2685)
+- fixed bogus warnings about resume info in logs when playing cutscenes and in the title level
 
 ## [4.15.1](https://github.com/LostArtefacts/TRX/compare/tr1-4.15...tr1-4.15.1) - 2025-10-10
 - changed the examine dialog to be usable with non-puzzle items (#4009)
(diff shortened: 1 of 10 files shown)


lahm86:
I'm getting infinite loops in some places
- In Lost Valley, fly towards the valley and it'll happen near the medipack beside the skeleton
- In Venice, play the start normally and once Lara heads down towards the first water area, it should happen
- The Deck - head to the right from the start
- Rig, 40F and Floating Islands freeze up as soon as they load in

rr-:
Fixed everything

lahm86 (approved):
Looking great. I see no impact in CPU usage on my end compared with develop.
Thank you 😌

aredfan (approved):

PR #4200, opened 2025-10-29
stats: support TR1 sector-based secrets in TR2

### LINKED ISSUE #2047: Feature request: support normal secret triggers in TR2
Add a gameflow option for custom levels to allow general secret rooms/areas to be used rather than the dragon pickups, similar to TR1/3. This would bypass the TR2 reward system to allow builders to place rewards as items along the way.

rr-:
#### Checklist

#### Description

Resolves #2047.

### DIFF
--- a/docs/tr2/CHANGELOG.md
+++ b/docs/tr2/CHANGELOG.md
@@ -31,6 +31,7 @@
 - added an option to draw Shotgun flashes (Graphic Options → Visuals → Shotgun flash)
 - added `enable_debug_camera` setting that shows camera position in realtime (reachable via `/debug` and `/set`)
 - added the ability to fast-forward through cutscenes with the right button (+5 s) or with slow+right (+1 s)
+- added support for TR1-like secret triggers (#2047)
 - improved bilinear filtering for smoother edge blending when multiple objects overlap in depth
 - improved ricochets placement
     - fixed disc ricochets being placed inside walls (#4063)
--- a/src/libtrx/game/inventory.c
+++ b/src/libtrx/game/inventory.c
@@ -295,7 +295,7 @@ bool Inv_AddPickup(const ITEM *const item)
     case O_SECRET_2:
     case O_SECRET_3:
         Stats_MarkSecretCollected(item);
-        if (Stats_CheckAllLevelSecretsCollected()) {
+        if (Stats_CheckAllLevelSecretsPickedUp()) {
             GF_InventoryModifier_Apply(Game_GetCurrentLevel(), GF_INV_SECRET);
         }
         return true;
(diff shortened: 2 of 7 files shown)


aredfan (approved):

aredfan:
I noticed an issue in retrospect. In TR1, the minimal stats shows secrets to be `x of y`. In TR2, the minimal stats differs as it only shows `x` for the secrets in this sort of scenario.

<details><summary>TR1 screenshot</summary>

20251029_195736_Caves

</details>

<details><summary>TR2 screenshot</summary>

20251029_195827_Test_Level

</details>

Here is a screenshot from TR3 which uses TR1's secrets, for reference.

Screenshot

</details>

lahm86 (commented):
In the test level with triggers and the secret item, if you step on the triggers first then collect the item, you get the rewards. If you pick up the item first then cross the triggers, you don't get the rewards. Are we making this setup unsupported so behaviour is undefined as such? Maybe some support docs for builders would be good to make it clearer.

rr-:
> In the test level with triggers and the secret item, if you step on the triggers first then collect the item, you get the rewards. If you pick up the item first then cross the triggers, you don't get the rewards. Are we making this setup unsupported so behaviour is undefined as such? Maybe some support docs for builders would be good to make it clearer.

No that doesn't make sense. 
I'm not sure when to provide the rewards – on the last dragon or on the last secret? IMO dragons only.

lahm86:
> > In the test level with triggers and the secret item, if you step on the triggers first then collect the item, you get the rewards. If you pick up the item first then cross the triggers, you don't get the rewards. Are we making this setup unsupported so behaviour is undefined as such? Maybe some support docs for builders would be good to make it clearer.
> 
> No that doesn't make sense. I'm not sure when to provide the rewards – on the last dragon or on the last secret? IMO dragons only.

Yeah, I would lean towards making gf rewards a dragon-only feature. Pickups as rewards can be placed on normal floor triggers, the "OG style".

aredfan (approved):

rr-:
> > > In the test level with triggers and the secret item, if you step on the triggers first then collect the item, you get the rewards. If you pick up the item first then cross the triggers, you don't get the rewards. Are we making this setup unsupported so behaviour is undefined as such? Maybe some support docs for builders would be good to make it clearer.
> > 
> > 
> > No that doesn't make sense. I'm not sure when to provide the rewards – on the last dragon or on the last secret? IMO dragons only.
> 
> Yeah, I would lean towards making gf rewards a dragon-only feature. Pickups as rewards can be placed on normal floor triggers, the "OG style".

Fixed this – should now only give rewards after the player collects all dragons, regardless of sector data.

lahm86 (approved):

PR #4201, opened 2025-10-29
music/common: improve ambient track detection

### LINKED ISSUE #4181: Bug/Glitch: ambient tracks issues.
It seems like using multiple ambiences in a level (`"ambient_tracks"`) breaks one-shot music triggers. This issue happens to me only when I have `"ambient_tracks"` in my `gameflow.json5` file. 

Also, triggers that play a different ambience don't work all of the time. I have the following setup:
An inside and outside ambience that gets triggered when you enter and exit the base (2 sectors, as shown in the attached image). But these triggers don't work all of the time.

Image

A small detail that bothers me is when Lara steps on a sector that triggers the outside ambience (the left one), the ambience triggers, even though it was already playing. It basically replays the ambience. I think the game should check if track 30 (an example) is already playing - if it is, it ignores the trigger.

I hope I was understandable 😭

lahm86:
Resolves #4181.

#### Checklist

#### Description

This updates music handling to allow the level's default ambient track to be included in floor triggers, and optionally (or accidentally) omitted from the game flow ambient track data.

The one shot problems in the issue are related to the trigger setup from TE (i.e. more than one item in a trigger, but only the music marked as one shot will not yield a compiled one shot trigger - all items need the flag).

### DIFF
--- a/docs/tr1/CHANGELOG.md
+++ b/docs/tr1/CHANGELOG.md
@@ -38,6 +38,7 @@
 - changed save crystal collision to make them easier to activate
 - changed the unrestricted look mode option to include Lara being able to look freely while shooting an enemy (#4090)
 - changed the `ambient_tracks` property to be only available on the root level
+- changed music triggers that match the level's default ambient track to automatically be treated as ambient if omitted from `ambient_tracks` (#4181)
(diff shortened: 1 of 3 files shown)


rr- (approved):

rr-:
Re – item triggers, can we put a comment about this in the original issue? That's a devilish gotcha to be aware of.

aredfan (approved):

PR #4216, opened 2025-10-31
lara: unify extra animation handling

### LINKED ISSUE #4212: OG bug: Lara hit direction retained in extra anims
Quite rare, but if for example a raptor in Qualopec nudges Lara at the same moment she tries to grab the scion, she will retain the hit pose throughout. Applies to all TR1 extra animations, although in OG I think Qualopec is the only level where this is most likely to happen (possibly Sanctuary too with Larson).

<details>

Image
</details>

lahm86:
Resolves #4212.

#### Checklist

#### Description

This combines all of Lara's extra animation logic in terms of data and handling, such that TR1 now uses the state change approach and so that both games are aware of every extra state. All animations are available to both games, and this also means builders can combine where not possible before (such as the yeti death and guard death in the same level). Mesh swaps are now separated from the extra object, so for example you can have a dragon and Midas hand in the same level, and both relevant mesh swaps for these will be honoured.

The code for switching to extra state has been de-duplicated and this in turn fixes the hit direction issue in TR1.

All the usual OG extra animations should work as expected.

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/149


rr- (approved):
This is absolutely amazing, I'm loving this. And the negative net diff. Holy cow!

aredfan (approved):

aredfan:
If it's relevant, I noticed that the tip of Lara's braid in TR2 doesn't turn to gold.

20251101_081140_Test_Level

</details>

lahm86:
> If it's relevant, I noticed that the tip of Lara's braid in TR2 doesn't turn to gold.
> 
> Screenshot

Thanks. I'd missed retexturing this part in WadTool - it's not currently part of the injection, rather part of the test level itself. I'm planning to follow this up with a proper braid mesh swap implementation so it's easier to manage, along with proper braid toggle for TR1.

<details>

image
</details>

PR #4222, opened 2025-11-01
Restructuring

rr-:
#### Checklist

#### Description

This change merges everything into a single `src/trx/` directory and removes the separate `include/` folder.
It also updates meson to build from a single file and removes `src/tr1` and `src/tr2` directories.

It also updates the include style to use angle brackets. Here's my reasoning for this decision:

We sometimes have headers with the same name, for example:
- /config.h
- /game/console/cmd/config.h
- /game/shell/config.h
- /gfx/config.h

When we use quotation marks for includes, like `#include "config.h"`, GCC looks in the current file's directory first. That can easily lead to the wrong file being included when there are duplicates like these, and proved to be really problematic once I tried merging `include/libtrx/*` into `src/`.

Rather than requiring every file to have a unique name, I'm following the Linux kernel approach of always using angle brackets. That way, headers are searched only within the specified include paths, which for our project means `src/`. To avoid conflicts with glibc headers, everything is placed under the `trx/` prefix.

I believe this approach is both elegant and straightforward, even if a bit austere.


rr-:
For testing, we just need to confirm if both games run at all, and if both .exes have the right metadata:

- Title screen version string
- .exe icon
- .exe properties

aredfan (approved):

lahm86 (approved):

PR #4225, opened 2025-11-01
lara/mesh: react to braid toggle in-game

### LINKED ISSUE #2399: Enhancement: allow toggling Lara's braid at runtime
Currently the required data is only injected if the option for it is enabled at load time. Ideally we would like to offer the braid to be able to be turned on/off at runtime and have Lara's meshes respond accordingly.

lahm86:
Resolves #2399.

#### Checklist

#### Description

Allows Lara's head and torso mesh edits in TR1 to be toggled at the same time as the braid, without having to reload the level. It also splits out the golden braid meshes into their own object in TR1 and TR2.

This unfortunately turned out to a bit more complex than I'd hoped, but LMK if you can see any improvements.

Checks:
- Regular gameplay/T-rex death/Midas touch
- Lara firing pistol type weapons
- Save/load
- Gym

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/150


src/trx/game/lara/mesh.c
```diff
@@ -67,6 +89,45 @@ void Lara_Mesh_Initialise(const GF_LEVEL *const level)
...
+                head_obj = O_LARA_EXTRA_SKIN_MIDAS;
+            }
+        }
+    } else {
+        const GF_LEVEL *const level = GF_GetCurrentLevel();
+        const RESUME_INFO *const resume = Savegame_GetCurrentInfo(level);
+        if (resume != nullptr && resume->flags.costume) {
+            torso_obj = O_LARA_EXTRA;
+        }
+    }
```
rr-: I think this function could get extracted to something like this?
```c
static OBJECT_ID M_GetTargetMesh(const LARA_MESH mesh_idx)
{
    const LARA_INFO *const lara = Lara_GetLaraInfo();
    const ITEM *const item = Lara_GetItem();

    if (lara->extra_anim) {
        switch (item->current_anim_state) {
        case LS_EXTRA_TREX_KILL:
            return O_LARA_EXTRA_SKIN_TREX;
        case LS_EXTRA_MIDAS_KILL:
            if ((lara->mesh_effects & (1 << mesh_idx)) != 0) {
                return O_LARA_EXTRA_SKIN_MIDAS;
            }
        default:
            break;
        }
        return O_LARA;
    }

    const GF_LEVEL *const level = GF_GetCurrentLevel();
    const RESUME_INFO *const resume = Savegame_GetCurrentInfo(level);
    if (resume != nullptr && resume->flags.costume && mesh_idx == LM_TORSO /* && … */) {
        return O_LARA_EXTRA;
    }

    return O_LARA;
}
```

Then:

```c
void Lara_Mesh_UpdateHair(const bool enforce)
{
    if (m_BraidStatus != g_Config.visuals.enable_braid
        || (m_BraidStatus && enforce)) {
        M_SwapHairParts();
        m_BraidStatus = g_Config.visuals.enable_braid;
    }
    Lara_Mesh_SwapSingle(LM_TORSO, M_GetTargetMesh(LM_TORSO));
    Lara_Mesh_SwapSingle(LM_HEAD, M_GetTargetMesh(LM_HEAD));
}

lahm86: Thanks, updated.

rr- (approved):
Thanks. My only remaining nitpick is that the gym outfit in `M_GetTargetMesh` doesn't support all of its intended meshes – maybe this can be queried somehow, or we can leave a `// NOTE`.

aredfan (approved):

lahm86:
> Thanks. My only remaining nitpick is that the gym outfit in `M_GetTargetMesh` doesn't support all of its intended meshes – maybe this can be queried somehow, or we can leave a `// NOTE`.

I've flipped it to `!= LM_HEAD` and added a comment too.

rr- (approved):

PR #4230, opened 2025-11-01
output: consolidate fov settings

rr-:
#### Checklist

#### Description

This change removes the **vertical FOV** option from TR1 (it's now on by default) and the **use PS1 FOV** option from TR2 (also on by default for the passport).

#### What this improves

- FOV values now behave consistently across both games.
  Previously, on a 16:9 display, you'd need to use 80° in TR2 but only 65° in TR1 for the same look, unless you select some checkboxes, in which case it's even more complicated. Now it's simple and consistent: 80° gives you the same appearance in both games, no exceptions.
- The passport always uses the nicer PS1-inspired FOV from TR2.
- Both games now display the passport consistently.
  - The passport no longer appears oddly zoomed out if "use PS1 FOV" was disabled in TR2.
  - Disabling TR1's "vertical FOV" setting can no longer break the passport.
- TR1 cutscenes now use the TombATI-style FOV (PS1 looked too zoomed-in – same issue as the PC build).
- TR2 cutscenes use the PC FOV (PS1 felt similarly zoomed and cropped).
- Removing these redundant options also means players can simply use `/set fov` instead of `/set visuals.fov`, since there are no naming conflicts with the now-removed settings.

**However**, I'm leaning toward adding separate options for **cutscene FOV mode** (PC / PS1) and possibly **passport FOV** (PC / PS1), since the passport edges are nearly touching on a 4:3 display.
On the other hand, this would make the configuration UI busier, and I don't really like this. I think what we have in the current PR achieves a reasonable middle ground. Please let me know what you think; I know @Richard-L would prefer us to keep these configurable.

#### Testing

- All cutscenes end-to-end
- Changing the FOV setting
- Cinematic cameras (e.g. Midas death, Scion pedestal; not sure what are the TR2 equivalents?)
- Different aspect ratios
- × TR1 and TR2
- Handling of old TR1 FOV setting when first launching this PR (existing values aren't compatible, so it should correctly fall back to a default 80°)


data/common/ship/cfg/base_strings-en-gb.json5
```diff
@@ -24,6 +24,8 @@
         "SETTING_FOG_COLOR_G_DESCRIPTION": "Green component of the fog colour.",
         "SETTING_FOG_COLOR_R": "Fog colour (R)",
         "SETTING_FOG_COLOR_R_DESCRIPTION": "Red component of the fog colour.",
+        "SETTING_FOV": "\\{review}Field of view",
+        "SETTING_FOV_DESCRIPTION": "\\{review}Viewing angle in degrees. Larger values widen the field of view, smaller ones narrow it.",
```
aredfan: I think it's okay to remove these.

aredfan (approved):

aredfan:
> * Cinematic cameras (e.g. Midas death, Scion pedestal; not sure what are the TR2 equivalents?)

I also checked these TR2 cinematic cameras moments. Not entirely sure if they're considered equivalent, but all looks good.

- Offshore Rig intro
- Lara takes the dagger from the dragon
- HSH intro
- HSH outro

lahm86 (approved):
I think this is the perfect middle ground and avoids too many options in the UI. Everything LGTM.

data/common/ship/cfg/base_strings-en-gb.json5
rr-: Thanks, good catch.

PR #4235, opened 2025-11-02
Agnostic inventory ring

rr-:
#### Checklist

#### Description

This PR moves the hardcoded inventory ring items into external JSON, bringing down `TR_VERSION` usages from 176 to 149.

I'm not sure I like this approach. Please LMK your thoughts.
If this approach is OK, I would like to do the same for `g_Weapons` next.


aredfan (approved):

lahm86 (approved):
I think this approach is a good move 👍

PR #4256, opened 2025-11-03
Merge config reading and writing facilities; reduce TR_VERSION usages

rr-:
#### Checklist

#### Description

This PR reduces the usages of `TR_VERSION` throughout the project down to just 1 in `lot.h`.

Caveat: I changed the way the config serializes the input layouts to use the TR2 approach, which means trying to go back to an earlier TR1X version in the same folder will reset the input configuration. While inconvenient for bisecting, I don't think this is something we should be too concerned about? TR1X and TR2X old options should be migrated forward just fine – it's the downgrading part that will cease to work.


lahm86 (approved):
My bad re item drop leak, thanks for fixing.

lahm86 (approved):

aredfan (approved):

PR #4257, opened 2025-11-03
Make TR2 guns available in TR1

lahm86:
#### Checklist

#### Description

The initial commit provides a resolution to the original (and extremely restrictive) method of determining whether a sprite is a game object or a static 2D based on the original object count. It unfortunately now needs a list of game IDs in its place, but I think this should be fairly manageable given that there are no sprites in TR3, so it shouldn't need to change often.

This then allows us to define the TR2 gun and ammo sprites, and so the injections have been updated to offer TR1 Lara all of TR2's guns and animations as well. These will only be given using the console, so the OG cheat and the TRX cheat will ignore them. Builders can obviously place normal pickups and/or edit the weapons JSON to remove the cheat exclusion.

The injected M16 SFX isn't great in TR1 - there is an audible stop similar to the flare burning. But I've edited the WAV for both of these so it doesn't happen quite so often, and I think it's acceptable. Note that the flare injection files are now merged with the ones for guns.

I'll provide a test level with pickups and room sprites for checking that aspect.

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/154


src/trx/game/input/common.c
```diff
@@ -437,18 +437,3 @@ bool Input_ParseKeyDesc(
...
-        case INPUT_ROLE_EQUIP_HARPOON:
-        case INPUT_ROLE_EQUIP_M16:
-        case INPUT_ROLE_EQUIP_GRENADE_LAUNCHER:
-            return false;
-        default:
-            break;
-        }
-    }
-    return true;
-}
```
rr-, line 454: IMO by displaying the hot-equip hotkeys at all times, we risk the availability of these keys, which are supposed to be an easter egg, look like sloppiness on our part due to engine merging, or a straight bug.
I think we could maybe use this to dim-out the options, and continue to not participate in conflict resolution.
However do not force-disable an input in `Input_Update()`, and let it through. If it connects – good!
WDYT?

src/trx/game/ui/elements/ammo_label.c
```diff
@@ -12,57 +12,56 @@
...
+    const char *string = nullptr;
+    if (icon_str != nullptr) {
+        string = String_FormatStatic("%6d %s", ammo, icon_str);
+    } else {
+        string = String_FormatStatic("%6d", ammo);
     }
+
+    UI_LabelEx(
+        String_StylizeSmallDigitsStatic(string),
+        (UI_LABEL_SETTINGS) { .scale = 1.5f });
```
rr-, line 66: Nice refactor, thanks.

src/trx/game/input/common.c
lahm86: There's also the custom level aspect though, as they'll want to show the keys to players. So maybe we need a config option to link it potentially?

rr-: We can use `g_Weapons[LGT_…].exclude_from_cheats` for this, for the same reason it was introduced for cheats support :)
Maybe this could be renamed to `is_available` and documented somewhere?

BTW – this logic (both cheats, and input availability) should extend nicely to `LGT_FLARE` as well.

lahm86: Oh I see! Sorry, wasn't following properly. I will update tomorrow. I suppose we could create a dummy flare weapon type for this purpose, or perhaps just infer it from those other guns being available (a bit sketchier).

rr-: IMO it's totally OK to go with a dummy flare type – missing properties should not trigger any warnings or errors 😌

Gizzy-91:
Very nice! Played around with this very briefly, a few notes:

- Harpoon gun not playing the holster sound when Lara is in water
- Manually selecting the harpoon gun when it's out of ammo will cause Lara to play the reload animation after equipping
- The grenade launcher explosion plays the same sound each time (Understandable if this is intended as the "longer" explosion sound doesn't exist in TR1
- Are the automatic pistols intended to be included? Trying to give them via the console just gives the magnums

lahm86:
> Very nice! Played around with this very briefly, a few notes:
> 
> * Harpoon gun not playing the holster sound when Lara is in water
> * Manually selecting the harpoon gun when it's out of ammo will cause Lara to play the reload animation after equipping
> * The grenade launcher explosion plays the same sound each time (Understandable if this is intended as the "longer" explosion sound doesn't exist in TR1
> * Are the automatic pistols intended to be included? Trying to give them via the console just gives the magnums

Thanks! I'll check out these points tomorrow. But re the autos, they are the same weapon under the hood in both games - just different meshes and sfx - so that's intended.

aredfan:
> * Manually selecting the harpoon gun when it's out of ammo will cause Lara to play the reload animation after equipping

This appears to be an OG bug.  It happens the first time Lara takes out an empty Harpoon Gun after running out of ammo.

**OG SG:** [savegame.zip](https://github.com/user-attachments/files/23325823/savegame.zip)

docs/02-COMMANDS.md
```diff
@@ -33,7 +33,7 @@ whichever key you have bound, and not include it as part of the command itself.
 - `/give {item_name}`  
   `/give {num} {item_name}`  
   `/give all`  
-  `/give guns`  
+  `/give guns` / `/give more guns`
```
aredfan: > `/give guns` / `/give more guns`

Would this be correct?

> `/give guns` / `/give moreguns`

aredfan:
Very nice, well done! 👍 

It might be worth unhiding the `Fix M16 accuracy` option in the TR1 gameflows.

A few potential issues:

- [x] ToQ doesn't end if Lara uses the grenade launcher to blow up Larson.
- [x] Harpoons and grenades has no effect on Bacon.
- [x] Harpoons and grenades has no effect on the Scion in The Great Pyramid.

docs/02-COMMANDS.md
rr-: (I was actually thinking of `/guns` `/moreguns` aliases in my original proposal, feels more in line with OG video game cheats.)

rr-:
>It might be worth unhiding the Fix M16 accuracy option in the TR1 gameflows.

Or more specifically tying its availability to `g_Weapons[LGT_M16].is_available` :relieved:

lahm86:
Thanks for the feedback. I'll work on the snags today, but rebased against develop for the time being.

lahm86:
- Weapon property is now `is_available` rather than `exclude_from_cheats`
- Flare added to `weapons.json`, so now using the above flag for decision making in cheats
- Fixed missing harpoon UW holster SFX
- Hotkey controls are greyed out, still editable and react to conflicts (left the latter as it's similar to cheat input conflicts showing when they are greyed out, LMK if ok)
- `Harpoon recoil` and `Fix M16 accuracy` config options are now visible (greyed out, but still editable)
- Bacon Lara and the Pyramid scion should now behave better with grenades/harpoons. I'd appreciate some testing in TR2 around hitting things with these projectiles too as I've removed the "intelligent" check, on the basis that both controls already check `Creature_IsTargetable` 
- `/guns` and `/moreguns` aliases added (and `/keys`, as it's fairly similar)
- Larson will use his normal death anim in Qualopec if killed by grenades (he will explode in Sanctuary)
- Fixed softlock if killing Torso with grenades

There are a few things left to address, but I'd like to do those separately in the linked tickets.
- Improve the "can explode" enemy check re grenades
- Improve "smash action" on objects
- Fix the grenade equip animation
- Fix the missing harpoon UW holster SFX in TR2

The only thing not addressed in the above comments is the grenade explosion SFX. If we were to replace it, it would impact normally exploding atlanteans. The alternative may be to create a new SFX ID for both games solely for grenade explosions, but I'm not sure about that.

aredfan:
Thank you for your hard work, as always.  👍

> * Fixed missing harpoon UW holster SFX

I can confirm this bug is fixed while Lara is UW, but it seems to persist if Lara is floating and she withdraws the harpoon gun.

> * `Harpoon recoil` and `Fix M16 accuracy` config options are now visible (greyed out, but still editable)

It might be worth greying out the `Fix flare throw priority` and `Fix free flare glitch` options too in TR1. (Gameplay Options → Fixes)

lahm86:
Thank you, @aredfan - updated 😌

rr-:
The point about conflicts for cheat inputs is very valid – it's OK to leave it as it is.
I'd create a separate SFX for the grenade explosions :relieved: But this is already plenty to have – thank you for all the fixes.

data/tr2/ship/cfg/weapons.json5
```diff
@@ -3,6 +3,10 @@
         "sample_num": "SFX_LARA_NO",
     },
 
+    "LGT_FLARE": {
+        "is_available": true,
```
rr-: I think it's fine to leave it as an empty dict.

src/trx/game/console/cmd/give_item.c
```diff
@@ -77,4 +87,5 @@ static COMMAND_RESULT M_Entrypoint(const COMMAND_CONTEXT *const ctx)
     return CR_SUCCESS;
 }
 
-REGISTER_CONSOLE_COMMAND("give", M_Entrypoint, GS_ID(CONSOLE_HELP_GIVE))
+REGISTER_CONSOLE_COMMAND(
+    "give|keys|(?:more)?guns", M_Entrypoint, GS_ID(CONSOLE_HELP_GIVE))
```
rr-, line 91: Can't believe we have a regex engine in C, lol.

src/trx/game/ui/dialogs/controls_editor.c
```diff
@@ -39,6 +40,26 @@ static bool M_AreCheatsEnabled(void)
...
+
+static bool M_IsM16Enabled(void)
+{
+    return g_Weapons[LGT_M16].is_available;
+}
+
+static bool M_IsGrenadeEnabled(void)
+{
+    return g_Weapons[LGT_GRENADE].is_available;
+}
```
rr-: I'd really prefer if we went back to a centralized `IsRoleAvailable` and removed `.is_available` :sweat_smile: 
This will let us easily search for usages of this function, whereas currently it's a bit decentralized.
(I misssed `M_AreCheatsEnabled` earlier).

`M_GetInputRow` can be inlined now.

lahm86: Thanks, pushed an update. There are two references to `M_GetInputRow` however, so not sure about inlining (unless I'm missing something 😄)

aredfan (approved):

src/trx/game/ui/dialogs/controls_editor.c
rr-: Oh I just meant that now that we show all rows, there's no reason to skip any and compute the actual index for the requester, which means we can just return `group->rows[index]` in that function. But then, the function becomes a pointless boilerplate.
Apologies, should have been more clear about this.

lahm86: Got it. Do you want to get rid of `M_GetInputRole` then, or keep the boilerplate?

rr-: Let's get rid of it.

lahm86: Thanks, done.

rr- (approved):

Issue #3773, opened 2025-08-08
Stray boulder in Floater


rr-:
<img width="2510" height="1363" alt="Image" src="https://github.com/user-attachments/assets/e13f9338-ee1b-459b-b87c-0a2cecb4524c" />

rr-:
I think we have two options. Either this patch, which is not perfect, as it essentially applies the scissor test to the objects, which costs us some performance…

```diff
diff --git a/src/libtrx/game/output/sources/objects.c b/src/libtrx/game/output/sources/objects.c
index b49d89d59..ce87fc8e9 100644
--- a/src/libtrx/game/output/sources/objects.c
+++ b/src/libtrx/game/output/sources/objects.c
@@ -1,6 +1,7 @@
 #include "game/output/sources/objects.h"
 
 #include "config.h"
+#include "game/output/vars.h"
 #include "game/objects/common.h"
 #include "game/output.h"
 #include "game/output/mesh_batcher/mesh_builder.h"
@@ -319,6 +320,13 @@ static void M_Stage(const OBJECT_MESH *const mesh, const bool skybox)
         .ls_vector_view = Output_GetLightVectorView(),
         .update_light_func = skybox ? M_UpdateShadesSkybox : M_UpdateShades,
         .update_light_func_data = (void *)mesh,
+        .enable_scissor = true,
+        .scissor = {
+            .x = g_PhdLeft,
+            .y = g_PhdBottom,
+            .width = g_PhdRight - g_PhdLeft,
+            .height = g_PhdBottom - g_PhdTop,
+        },
     };
     if (skybox) {
         MeshBatcher_Stage(p->batcher, &inst, SCENE_PASS_BACKGROUND);
diff --git a/src/tr2/game/room_draw.c b/src/tr2/game/room_draw.c
index 0b2686a96..1f9e25b26 100644
--- a/src/tr2/game/room_draw.c
+++ b/src/tr2/game/room_draw.c
@@ -299,11 +299,6 @@ void Room_DrawSingleRoomObjects(const int16_t room_num)
         Matrix_Pop();
     }
 
-    g_PhdLeft = Viewport_GetMinX(VIEWPORT_GAME);
-    g_PhdTop = Viewport_GetMinY(VIEWPORT_GAME);
-    g_PhdRight = Viewport_GetMaxX(VIEWPORT_GAME);
-    g_PhdBottom = Viewport_GetMaxY(VIEWPORT_GAME);
-
[...]

PR #4271, opened 2025-11-07
room/draw: add protruding statics to nearby rooms

### LINKED ISSUE #3761: TR1X bug: boulder visible through overlapped rooms
Regression since [4.10](https://github.com/LostArtefacts/TRX/releases/tag/tr1-4.10)

_________________

Image

</details>

rr-:
#### Checklist

#### Description

This hopefully fixes the disappearing gazebo problem in Bartoli without introducing regressions.

We'll need a test level with tons of overlapping stuff to verify if this is indeed a workable approach. I just wanted to put the PR out there for Lahm to play with as I'll be AFK during the weekend.

If it is we can consider putting items into all rooms they appear in, especially as they go mid-portal.

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -252,6 +252,7 @@ common_sources = [
   'trx/game/level/reader.c',
   'trx/game/level/reader_tr1.c',
   'trx/game/level/reader_tr2.c',
+  'trx/game/level/rooms.c',
   'trx/game/level/settings.c',
   'trx/game/los.c',
   'trx/game/lua/common.c',
--- a/src/trx/game/carrier.c
+++ b/src/trx/game/carrier.c
@@ -155,6 +155,7 @@ static void M_InitialiseDataDrops(void)
         CARRIED_ITEM *drop = carrier->carried_item;
         for (int32_t j = 0; j < pickups->count; j++) {
             drop->spawn_num = *(const int16_t *)Vector_Get(pickups, j);
+            Item_RemoveDrawn(drop->spawn_num);
             drop->room_num = NO_ROOM;
             drop->fall_speed = 0;
             drop->status = DS_CARRIED;
--- a/src/trx/game/items/common.c
+++ b/src/trx/game/items/common.c
@@ -21,6 +21,35 @@ static int16_t m_NextItemActive = NO_ITEM;
 static int16_t m_PrevItemActive = NO_ITEM;
 static int16_t m_NextItemFree = NO_ITEM;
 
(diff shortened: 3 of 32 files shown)


rr-:
I've updated the PR to make it possible for both items and statics to belong to multiple rooms. It seems to be passing all of our cursed test cases:
- TR1 Tihocan boulder in room 78 showing up in room 74 when looking from room 77
- TR2 Catacombs trapdoors in room 30 when looking at them close to the wall
- TR2 Bartoli gazebo in room in room 69 when looking at it from room 130
- TR2 Bartoli goon crossing the portal between rooms 75 and 99

That's all I've tested, so more testing would be appreciated. Even worse, if I messed something up, some items might never become visible at all, making them very difficult to catch.

I think this screenshot best sums up just how ridiculous the test cases are – the animation finishes completely outside of the parent room, and completely underneath its floor. The shadow just gives up.
20251110_012102_Bartoli&#39;s_Hideout

The reason why this works now, and the goon gets rendered despite being culled via scissor tests, is because thanks to the bbox tests on portals, he gets assigned also to the room below. So, even though he ends up not getting drawn in the first room as he's essentially outside of room bounds, the second check lets him through as he fits just fine.

I take care not to draw the same item twice even if it appears in two rooms at once, to avoid making its shadow more opaque.

Implementation wise, the magic happens in `Item_RemoveDrawn` and `Item_UpdateRoom`.
I started out with gamebuf-powered vectors and even created an `ALLOCATOR` module for this, but I ended up not liking the performance hit (even if theoretical) – dozens of linear memory scans, `memmove()` calls, etc. I ended up going with a sparse bitset which loses us 1.25 KB of memory per-room, but has O(1) performance for all operations.

aredfan:
**Issues:**

- [x] Issuing `/kill all` in some TR2 levels such as Venice or Floating Islands causes the game to crash.
- [x] In Bartoli's Hideout, the detonator disappears after Lara uses it.

rr-:
Thanks – fixed the `/kill`, but I'm puzzled at the detonator deciding to disappear :sweat_smile: 

Edit: got it – it's because it doesn't get re-added to the flipped room.
Edit 2: fixed as well :relieved:

lahm86 (approved):

aredfan (approved):
Thank you @rr-, the prior issues are all resolved.  I did more tests and haven't encountered any further issues. 👍 

I also noticed #3773 is partially fixed because the stray boulder doesn't appear in the distance.

PR #4290, opened 2025-11-13
tr3: miscellaneous updates

lahm86:
#### Checklist

#### Description

- Uses `O_LARA_SKIN` if loaded
- Updates Lara'a animation, state change and item action catalogs
- Creates TR3 `lara_animations.bin`
- Creates a TR3 PDA model (the jury is out if I like this design but LMK)
- Fixes floor data parsing too many room portals on a single sector (would result in a softlock if sliding down the initial Jungle slope). This was an editor bug I'd forgotten about; the original `GetDoor` engine function would only ever read the first portal, so it never manifested in the game

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/158


rr- (approved):
Brilliant 🩷
And I like the PDA, too. Though it might be a bit subtle that it's from RX-Tech.

lahm86:
> Brilliant 🩷 And I like the PDA, too. Though it might be a bit subtle that it's from RX-Tech.

Thanks. I've added some text on the inside. It's a little awkward as the space is so tight, but hopefully looks OK.

aredfan:
> > Brilliant 🩷 And I like the PDA, too. Though it might be a bit subtle that it's from RX-Tech.
> 
> Thanks. I've added some text on the inside. It's a little awkward as the space is so tight, but hopefully looks OK.

Perfection. 🏆

aredfan (approved):

PR #4291, opened 2025-11-13
tr3: add initial swamp support

lahm86:
#### Checklist

#### Description

Adds swamp support (thanks @rr- for the base commit). The camera is still misbehaving near triangular portals, but I can't figure out what's going on there. Even changing the `NO_BOX` const to 2047 doesn't help (I've left that out of this change as a result). The camera however should behave suitably enough when wading through the mud as it will clamp to a target above Lara's position.

Worth checking going in and out of water as well in all games, including wading.

### DIFF
--- a/src/trx/game/objects/general/flare_item.c
+++ b/src/trx/game/objects/general/flare_item.c
@@ -20,6 +20,12 @@
 static void M_Control(const int16_t item_num)
 {
     ITEM *const item = Item_Get(item_num);
+    const ROOM *const room = Room_Get(item->room_num);
+    if (room->flags.swamp) {
+        Item_Kill(item_num);
+        return;
+    }
+
     if (item->fall_speed) {
         item->rot.x += DEG_1 * 3;
         item->rot.z += DEG_1 * 5;
@@ -34,7 +40,7 @@ static void M_Control(const int16_t item_num)
     item->pos.z += (item->speed * Math_Cos(item->rot.y)) >> W2V_SHIFT;
     item->pos.x += (item->speed * Math_Sin(item->rot.y)) >> W2V_SHIFT;
 
-    if (Room_Get(item->room_num)->flags.underwater) {
+    if (room->flags.underwater) {
         item->fall_speed += (5 - item->fall_speed) / 2;
         item->speed = item->speed + (5 - item->speed) / 2;
     } else {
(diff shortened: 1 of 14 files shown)


aredfan (approved):

src/trx/game/ui/elements/bar_lara_air.c
```diff
@@ -14,9 +14,7 @@ bool UI_LaraAirBar(const bool blink_state)
     const LARA_INFO *const lara = Lara_GetLaraInfo();
     const bool is_blinking = g_Config.ui.enable_bar_flashing
         && lara->air <= LARA_MAX_AIR * UI_BAR_BLINK_THRESHOLD;
-    const bool show = g_Config.ui.show_bars
-        && (lara->water_status == LWS_UNDERWATER
-            || lara->water_status == LWS_SURFACE);
+    const bool show = g_Config.ui.show_bars && lara->air < LARA_MAX_AIR;
```
rr-: This seems to cause the bar to no longer appear if Lara's floating

src/trx/game/objects/general/flare_item.c
```diff
@@ -20,6 +20,12 @@
 static void M_Control(const int16_t item_num)
 {
     ITEM *const item = Item_Get(item_num);
+    const ROOM *const room = Room_Get(item->room_num);
+    if (room->flags.swamp) {
+        Item_Kill(item);
```
rr-: This should have been `Item_Kill(item_num)` – sorry.

src/trx/game/ui/elements/bar_lara_air.c
lahm86: Fixed, thanks.

src/trx/game/objects/general/flare_item.c
lahm86: And fixed too 😌

rr-:
I can't yet run the OG but in this case Lara seems to be unable to climb the slope,
20251113_233225_Level_0
Going from the other side, she slides down:
20251113_233336_Level_0
And falling from height causes her to get stuck like this:
20251113_233421_Level_0

Likely all of this relates to triangular geometry. Happy to approve if this is the case.

lahm86:
Not going up the slope is normal I think, but I don't think she should be sliding. Will check that and the falling issue tomorrow 👍
The triangle problems seem to only be where there are splits directly on a portal.

lahm86:
The sliding and falling issue should be resolved. I'll work  on the triangle portals as a separate task.

rr- (approved):

PR #4297, opened 2025-11-14
Implement additive blending

rr-:
#### Checklist

#### Description

This PR implements the `2` additive-blending scene pass.
It also simplifies face manipulation underneath.

The gun flashes will remain white until we rewrite the lighting system.

20251114_205400_Level_0
20251114_205150_Level_0
20251114_205436_Level_0


lahm86:
Looks great 😌 

Some statics in Lost Valley seem to jump between shades (I think we had this problem a while ago).
- `/play 3`
- `/tp 38 3 20`
- Move forwards and backwards from this tile

<details>

image
</details>

rr-:
> Looks great 😌
> 
> Some statics in Lost Valley seem to jump between shades (I think we had this problem a while ago).
> 
> * `/play 3`
> * `/tp 38 3 20`
> * Move forwards and backwards from this tile
> 
> image

Thank you – excellent observation. I got burned again by the fact that lighting code updates vertex data on meshes and statics data, which is shared between all meshes and statics, rather than on instance level. All the more reason to move towards UBOs or SSBOs.

aredfan:
I found a small issue with a transparent puzzle item not appearing in the HUD when Lara picks it up.  To reproduce, play shore.tr2 and then issue `/tp puzzle` to reach the item.

<details><summary>Screenshot from develop</summary>

20251114_210234_Level_0

</details>

rr-:
I've fixed this + simplified transparency sorting so that it no longer has its own VAO + VBO, and instead uses the same set as any other vertices, passing only vertex indices instead.

lahm86 (approved):

aredfan (approved):

PR #4321, opened 2025-11-17
Shaders refactor

rr-:
#### Checklist

#### Description

This PR:

1. Fixes broken Mac support on develop by slightly downgrading the UBOs syntax to avoid the `binding` keyword.
2. Improves error reporting in shaders to include the path of the affected shader.
3. Improves shader preprocessing to support `#include` macro. Technically this can be achieved in a better way with a virtual filesystem in a modern OpenGL API… but yes, Macs won't have that. That's fine, I did it as a simple preprocessing step.
4. Cleans up shader code a bit, divides into files.
5. Removes the hardcoded `M_UNIFORM` enum from the shader.c code in favor of dynamic lookups. This lets us avoid warnings when loading the `ui.glsl` shader about missing uniforms that are only specific to `meshes.glsl`.
6. Splits the UI work into a new shader, `ui.glsl`, which is really simple. This lets us remove `LIGHTING_MODE` enum and dumb it down to `uLightingEnabled`.
7. Splits the shader code into `mesh.c` and `ui.c`, backed by `generic.c`. This way, individual shaders can maintain their uniforms state.
8. Extracts the recent UBO work into a separate `uniforms.c` module. Since UBOs' data lives independently from shaders and the shaders only reference their data through numeric bindings, it makes sense to keep UBO code isolated from the shaders module. (As an example, updating the view/projection matrices UBO will update the projection and view matrix in both `ui.glsl` and `meshes.glsl` without us having to do individual per-shader transfers).
9. Removes `uProjectionMode` which saves us one branch in the vertex shader, and makes it easier to reason about current projection matrix (which will be either orthogonal or perspective, nothing inbetween). That it can be switched regardless of the active shader is very convenient for the 3D pickups code.
10. Changes sunset timer to be expressed in terms of `Output_GetTimeInGame()`, which gets reset at the start of a level. The `TimeInGame` value is used for the following effects:
    - water caustics, both displacement and lighting,
    - wave effect in the inventory ring,
    - and now sunsets.
11. Removes the notion of light shades from the UI shader, keeping it only RGBA.

#### Testing

Surprisingly easy – we just need to test:

- the caustics appearance
- 3D pickups
- enabling/disabling lighting,
- greyed-out UI options,
- and debug mode stuff.

I tested the sunsets both with and without saves so we should be good there.


aredfan:
There appears to be an issue with the game not starting up.  I've attached the TR2X log and `.dmp` file below.

rr-:
Oof. It's not even our crash, NVidia drivers are crashing.

rr-:
@aredfan I think it's fixed :)

walkawayy (approved):

aredfan (approved):
> @aredfan I think it's fixed :)

It's working now, thank you.

LGTM. 👍

lahm86 (approved):

PR #4356, opened 2025-11-28
rooms/draw: use skybox room flag in TR1

lahm86:
#### Checklist

#### Description

This updates TR1 to use the same approach as TR2+ by marking specific rooms only to draw the skybox in.

This does away with the injection hack to hide the area above Lara at the start of The Cistern as a result. I've left the face injections in Obelisk room 8 however, as the vertex shadows there are nice compared with OG.

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/162


aredfan (approved):

rr- (approved):

rr-:
Since this seems to cause a regression without new TombEditor that does not expose this flag yet, can we sit on this PR until we get a compatible TE out?

lahm86:
TE 1.11 is now released so I think we are in a good position to merge this. I've re-based - would appreciate a quick check in TR1 that the sky still shows up as expected. Thanks!

aredfan (approved):

rr- (approved):

PR #4378, opened 2025-12-07
objects/traps: add support to reset all traps

### LINKED ISSUE #3993: Feature: allow resetting trap types
Investigate the feasibility of being able to antitrigger all trap types. For ones that move, the behaviour should be similar to boulders where the position is reset.

- [x] Collapsible tile
- [x] Damocles sword
- [x] Ember emitter
- [x] Falling ceiling
- [x] Lava wedge
- [x] Pushblock
- ~~Thor's hammer (works currently before it hits the floor, like in Folly)~~
- [x] Hook
- [x] Icicle
- [x] Pendulum
- [x] Spike ceiling

Already supported:
- Boulder
- Dart/disc emitter
- Flame emitter
- Lightning emitter
- Propeller
- Saw
- Slamming door
- Spike wall (not OG, #2762)
- Spinning/statute/wall blades

lahm86:
Resolves #3993.

#### Checklist

#### Description

This adds support to reset all trap types. The one caveat is Thor's hammer, in that it can only be reset as per OG in Folly. Changing the behaviour would result in it always being reset when Lara steps off the pad in OG - I couldn't find a suitable workaround.

For pushblocks, we can only support timed triggers to perform the reset as the control function only runs when the item is falling or being moved by Lara. The outcome is the same so I think this is acceptable, especially given the complexity of blocks.

There are a couple of FD injections for OG as Caves had a timed trigger under collapsible tile 50, and ember emitter 12 in Pyramid was also timed (side effect of the trapdoor timer). Otherwise, no other items of the affected types in OG have timed triggers or antipads/antitriggers.

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/163


src/trx/game/objects/traps/common.c
```diff
@@ -0,0 +1,40 @@
...
+    item->required_anim_state = TRAP_SET;
+    Item_RemoveActive(item_num);
+}
+
+void Trap_HandleSave(ITEM *const item, const SAVEGAME_STAGE stage)
+{
+    if (stage == SAVEGAME_STAGE_AFTER_LOAD) {
+        item->enable_interpolation = item->status == IS_ACTIVE;
+    }
+}
```
rr-: This function naming suggests that it should be called by all traps, but that is not the case.
Because of this confusion, I'd duplicate the function between pendulum.c and hook.c instead.

lahm86: Thanks, updated.

rr- (approved):

aredfan (approved):

Issue #3873, opened 2025-08-25
Feature Request: Angry monks on level start / AI aggression


MaskedGoon:
Hi, 

I think an option to control monk aggressiveness on beginning a level could be a cool feature. Now it would strictly serve a TRLE purpose, allowing the builder to use one more enemy type without doing awkward workarounds, but I figured I'd just post it here.  I've heard that in regular TR2, this already happens if you play Barkhang Monastery in NG+.

Being able to set any AI slot to be either friendly (as in, it attacks other AI than Lara) or aggressive in the level's script would open even more possibilities, but I think it might also go a little too far.

PR #4385, opened 2025-12-11
Lua/creature adjustments

lahm86:
Part of #3873.

#### Checklist

#### Description

This resolves a problem with statistics if a level is set to have hostile allies from the get-go (via `on_level_load`). The order of events on develop would be:

- Unload level
  - Set allies hostile = false
- Load level
  - Calculate stats
  - Allies are not hostile, so are not included as kills
- Run the level's Lua script
  - Allies are now hostile; the stats are inaccurate

I'm not convinced introducing another event is the best approach but wanted to get your thoughts.

This can be tested in Barkhang if you add the following to a Lua file and link it in the GF.

```lua
trx.events.on_level_init(function(level)
  trx.creatures.hostile_allies = true
end)
```

This also prepares for allowing allies to be defined externally, so some checks around the kill command, shooting monks, and seeing that they still pick up items would be good (e.g. item 79 in Kingdom).


aredfan (approved):

src/trx/game/carrier.c
```diff
@@ -54,7 +54,6 @@ static bool M_IsCarrierType(const OBJECT_ID obj_id)
     // also creates issues with /kill, hence a separate check is required here.
     is_enemy &= obj_id != O_EEL && obj_id != O_BIG_EEL;
     is_enemy &= obj_id != O_SKIDOO_DRIVER;
-    is_enemy |= Object_IsType(obj_id, g_AllyObjects);
```
rr-, line 57: Doesn't this break GM allied monk's drops?

src/trx/game/console/cmd/kill.c
```diff
@@ -21,7 +21,6 @@ extern OBJECT_ID CombatEnd_GetBossType(void);
 static bool M_CanTargetObjectCreature(const OBJECT_ID obj_id)
 {
     return (Object_IsType(obj_id, g_EnemyObjects)
-            || Object_IsType(obj_id, g_AllyObjects)
```
rr-, line 24: Is this intended?

src/trx/game/carrier.c
lahm86: They're part of the enemy array as standard (which is checked first) so this becomes redundant. Although admittedly it'd be harmless to leave it in place.

src/trx/game/console/cmd/kill.c
lahm86: Similar to above. Could leave these in place for clarity perhaps?

src/trx/game/carrier.c
rr-: It's pretty odd that enemies are a superset of allies. Sounds like an either-or. Wdyt?

lahm86: It sits ok with me tbh, I'm thinking the ally list is more of an exclusion set rather than a subset. Perhaps we could rename the enemy list to g_CreatureObjects - so conceptually a creature will by default be hostile unless in the ally list?
This should hopefully mean we only need to offer editing the ally and ally-targeting lists in Lua, rather than these plus the global enemy list.

rr-: Yeah I'm in for a rename. Thanks!

lahm86: Done ☺️

rr- (approved):
Thanks!

PR #4389, opened 2025-12-12
creature: control allies/ally-targets via Lua

### LINKED ISSUE #3873: Feature Request: Angry monks on level start / AI aggression

lahm86:
Resolves #3873.

#### Checklist

#### Description

This replaces the hard-coded list of allies and ally-targeting creatures with the ability to instead feed this information via Lua. Original TR2 level scripts are provided. In addition, TRX object IDs are now available in Lua, meaning the need to locally define IDs is removed.

I was going to add `remove_ally`/`remove_ally_target` functions but this might be better to reserve until we have a savestate system. WDYT? Also open to any other thoughts on the API.

I think we can follow up once this gets some usage to look into improving such things as the missile behaviour we discussed.

### DIFF
--- a/data/common/ship/scripting/trx/catalog.lua
+++ b/data/common/ship/scripting/trx/catalog.lua
@@ -1,0 +1,7 @@
+local raw = trxc.catalog
+
+local catalog = {
+  objects = raw.objects,
+}
+
+trx.catalog = catalog
--- a/data/tr2/ship/data/scripts/monastry.lua
+++ b/data/tr2/ship/data/scripts/monastry.lua
@@ -1,0 +1,6 @@
+trx.events.on_level_init(function(level)
+  trx.creatures.add_ally(trx.catalog.objects.O_MONK_1)
+  trx.creatures.add_ally(trx.catalog.objects.O_MONK_2)
+  trx.creatures.add_ally_target(trx.catalog.objects.O_BANDIT_1)
+  trx.creatures.add_ally_target(trx.catalog.objects.O_BANDIT_2)
+end)
--- a/data/common/ship/scripting/trx/creatures.lua
+++ b/data/common/ship/scripting/trx/creatures.lua
@@ -1,6 +1,9 @@
 local raw = trxc.creatures
 
-local creatures = {}
+local creatures = {
+  add_ally = raw.add_ally,
+  add_ally_target = raw.add_ally_target,
+}
 
 local getters = {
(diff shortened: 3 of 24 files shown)


aredfan (approved):

data/common/ship/scripting/trx/catalog.lua
```diff
@@ -0,0 +1,7 @@
+local raw = trxc.catalog
+
+local catalog = {
+  Objects = raw.Objects,
```
rr-: WDYT about making this lowercase?

data/tr2/ship/data/scripts/monastry.lua
```diff
@@ -0,0 +1,6 @@
+trx.events.on_level_init(function(level)
+  trx.creatures.add_ally(trx.catalog.Objects.O_MONK_1)
+  trx.creatures.add_ally(trx.catalog.Objects.O_MONK_2)
+  trx.creatures.add_ally_target(trx.catalog.Objects.O_BANDIT_1)
+  trx.creatures.add_ally_target(trx.catalog.Objects.O_BANDIT_2)
```
rr-: Ideally it would be cool to create a two-way relationship, so we can make bandits shoot both monks AND crows, for example, even though crows are not allies.

rr- (approved):
Suggested making the targeting more robust.
Agree on postponing all of the other things – small bites are good.
Happy to get merged in the current form although I'd prefer `Objects` to be lowercase.

data/common/ship/scripting/trx/catalog.lua
lahm86: Thanks, fixed. Not sure why I picked this originally 😄

data/tr2/ship/data/scripts/monastry.lua
lahm86: That would be great indeed. Happy to continue expanding on this as it gains more traction 😌

rr- (approved):

Issue #2074, opened 2024-12-18
Stray frame in iconic boat camera


Richard-L:
There is this one odd frame which disturbs the moment a little.

https://github.com/user-attachments/assets/10768f68-3292-4540-bbe5-6d63bd463d26

lahm86:
This is fixable on develop (and will be in 1.1) by choosing the TR1 camera mode, which doesn't perform LOS checks for fixed cameras. The ultimate fix would be removing LOS checks from TR2's camera mode, but this would have negative results in several other places so that's not viable. I think picking TR1 mode is the best solution for now given that this is an isolated case.

PR #4393, opened 2025-12-13
camera: introduce strategies

lahm86:
#### Checklist

#### Description

This separates the main camera logic with that of updating the position and target into a strategy-based approach. This will later allow defining the TR3 camera as a separate module.

I've also replaced most game version checks with mode checks instead - for example, removing the LOS check in TR2's fixed camera can be done by choosing TR1 mode. #2074 can be resolved as a result if we feel this is an acceptable approach? I don't want to remove the LOS check entirely as it will impact other areas badly.

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -103,6 +103,7 @@ common_sources = [
   'trx/game/anims/commands.c',
   'trx/game/anims/common.c',
   'trx/game/anims/frames.c',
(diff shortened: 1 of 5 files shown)


aredfan (approved):

rr- (approved):
LGTM

PR #4416, opened 2025-12-18
tr3: fix cutscene setup and NO_BOX const

lahm86:
#### Checklist

#### Description

Adds injections for a couple of TR3's cutscenes to rotate Lara, which I'd forgotten about previously (replaces the old GF sequence). The Jungle cutscene has some bad portals, which was causing the game to loop infinitely during hair control - data fixed for this.
So each cutscene should now play properly (there are niggles in some, like Lara snapping back to the first frame during the stats, but we can deal with those separately).

TR3's `NO_BOX` const is 2047, so sectors are now converted on level load to use `-1` as standard. This stabilises the camera on slopes and other areas, and avoids the random crashes. A good test is Temple Ruins, `/tp 83 -4 31` then drop back, hang from the ledge and press look - on develop the camera is not happy.

And a final update is for TR1's cutscenes - we used to have to get the room index for the braid by a position lookup, but this is no longer needed since we have a proper Lara in place now. Spotted this when debugging hair control above.

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/174


aredfan (approved):

rr- (approved):
This is absolutely amazing

Issue #4407, opened 2025-12-15
TRX bug: alligator and shark doesn't remain tinted when dead and floating


aredfan:
Regression since [TRX 1.0](https://github.com/LostArtefacts/TRX/releases/tag/trx-1.0)

_____________________

1. launch tr2
2. /play 10
3. 2x `/tp key`
4. pick up the key
5. shoot the sharks

____________________

1. launch tr1
2. /tp 7
3. /flip
4. have Lara go into the next area and shoot the alligator

lahm86:
I think this could be because of items being linked to both the dry and wet rooms, as their bounds overlap the portals when floating. If the dry room happens to be drawn first, they'll be drawn as though above water.

PR #4422, opened 2025-12-19
rooms/draw: update water status for each item draw

lahm86:
#### Checklist

#### Description

This sets the output water status before drawing each item, to ensure that the room that has been determined in control updates is the one used for tinting. Sorting by room status has been reverted as there were a few static mesh issues, such as the poles in Venice and the life raft in The Deck. Injections are provided to further fix z-fighting in Venice and shift the poles into their proper positions, doing the equivalent to the life raft didn't make sense as it should be partially submerged.

Refs #4407, https://github.com/LostArtefacts/TRXInjectionTool/pull/175


aredfan (approved):

rr- (approved):
Thank you :)

PR #4435, opened 2025-12-20
level/loader: fix TR3 sample property reading

lahm86:
#### Checklist

#### Description

TR3 samples have signed pitch values and volume is shifted by 7 rather than 8, so this corrects the level loader to match.

Water splash doesn't play (unless in shallow water) because this is no longer an anim command and hardcoded elsewhere, so that's an integration task later. There may be others like this.


aredfan (approved):

rr- (approved):
Missing changelog entry :)

lahm86:
> Missing changelog entry :)

Thanks, added.

Issue #4432, opened 2025-12-20
Wrong final stats yet again


rr-:
[save_tr1_00.dat.zip](https://github.com/user-attachments/files/24273844/save_tr1_00.dat.zip)

<img width="1919" height="1079" alt="Image" src="https://github.com/user-attachments/assets/c4e4325e-63ca-40c8-b0d6-68a813415d86" />

Regression since [TRX 1.0](https://github.com/LostArtefacts/TRX/releases/tag/trx-1.0)

The issue manifests if you start a new game, skip a few levels, save it, close the game, launch it again, load the save, and continue skipping till the end

aredfan:
OP updated.

PR #4436, opened 2025-12-20
Calculate max stats on game launch

rr-:
#### Checklist

#### Description

IIRC, calculating all the stats on game launch was my intention for TRX 1.0 all along… I just forgot to actually implement it. Anyway, here it is now.

The real bummer is that we already dropped storing per-level max stats in the save files. For saves created with TRX 1.0, that data simply doesn't exist anymore, so there's nothing to recover. That effectively locks us into this approach for all existing saves.

And yeah, this implementation is kinda cursed. For each level:
- We skim-load the level file, only processing objects, rooms, floor data, and items.
- We still read all injections, but ~95% get discarded. To reduce overhead, they're now loaded with a thread pool, which gives us around 40% time save.
- Then we do some duct-tape post-processing so that `carrier.c` can aggregate the pickups.
- And the cherry-on-top: we run the Lua level init, since ally setup got moved to Lua.

This code is fragile, but it sort of works. The worst part is Lua – god knows what builders stuff in there. The game is essentially running in a zombie state at that point – no textures, no anims, items unitialized, etc. And it's slow. On my machine, this adds +0.2 s to TR2 launch, and my machine's pretty damn good.

LMK what you think. You can see the calculated stats and the exact performance hit in the log by searching for `Stats_CalculateMaxStats`.

The only real alternative would be to start storing the max counters in save files again, but that still wouldn't fix existing saves made so far, since that data just isn't there anymore.

#### Testing

We need to:

- start the game, level skip to the end, and check the final stats
- load the savegame from #4432, and check the final stats
- compare the total numbers in the tables and see if I got them right (just counting the items)
- compare the total numbers in the tables against the logs / final stats screens
- × TR1, TR2, TR1UB, TR2GM
- pay extra attention to Nightmare in vegas
    - should be played when you collect all secrets
    - shouldn't be played when you fail to collect some secrets
    - collecting some secrets, save, quit, launch, load, collect remaining secrets – should still let you play it
    - final stats screen should include both Nightmare in vegas and the main game pickups/kills/secrets stats (I think this is changed by this PR, please LMK if that's the case, I'll adjust the changelog)
- allies and allied pickups in TR2 / TR2GM are not to be counted


aredfan:
> LMK what you think. You can see the calculated stats and the exact performance hit in the log by searching for `Stats_CalculateMaxStats`.

The game's performance when starting up seems okay to me. These are the figures from my logs.

- TR1 took 0.15 s
- TR2 took 0.22 s

> * final stats screen should include both Nightmare in vegas and the main game pickups/kills/secrets stats (I think this is changed by this PR, please LMK if that's the case, I'll adjust the changelog)

I can confirm this is the case although the final stats screen is shown before starting Nightmare in Vegas.

20251220_184639_Kingdom

</details>

(Testing is in progress. 😄)

lahm86:
I think performance is fine 👍 

- [x] Secret pickups in TR2 don't seem to be registering in the stats (using the console or an older save seems OK).

rr-:
> > LMK what you think. You can see the calculated stats and the exact performance hit in the log by searching for `Stats_CalculateMaxStats`.
> 
> The game's performance when starting up seems okay to me. These are the figures from my logs.
> 
> * TR1 took 0.15 s
> * TR2 took 0.22 s
> 
> > * final stats screen should include both Nightmare in vegas and the main game pickups/kills/secrets stats (I think this is changed by this PR, please LMK if that's the case, I'll adjust the changelog)
> 
> I can confirm this is the case although the final stats screen is shown before starting Nightmare in Vegas.

Thanks, fixed – final stats screen will include the bonus levels only when it's invoked from within a bonus level.

> I think performance is fine 👍

I'm really happy to hear that :)

> * [ ]  Secret pickups in TR2 don't seem to be registering in the stats (using the console or an older save seems OK).

Fixed this as well, though this was [a bit trickier](https://github.com/LostArtefacts/TRX/compare/d9044c5ebcf25a3705e11e0d294a4ab9cb320fb0..ca7818b299b112764be407fffe7269b5eda6f07b).

aredfan:
- [ ] In TR1, if I execute `/set ps1-enemies 0`, and skip through all the levels from start to finish, the total kills shows `295`. In develop, this would show `294` instead.

lahm86 (approved):
LGTM.
Re the PS mummy, I'd be in favour of just changing the config option comment to say to restart the game rather than just the level.

rr-:
> LGTM. Re the PS mummy, I'd be in favour of just changing the config option comment to say to restart the game rather than just the level.

Agreed, this is the simplest fix – pushed.

aredfan (approved):

PR #4443, opened 2025-12-22
Update italian translation

Kappa971:
Minor fixes, general reordering based on the English file.

I noticed a few things:
1. These strings are repeated in `base_strings.json5` and in the `strings.json5` files of TR1 and TR2:
```json
   "objects": {
        "small_medipack": {
            "name": "Small Medipack",
        },
        "large_medipack": {
            "name": "Large Medipack",
        },
        "magnums": {
            "name": "Automatic Pistols",
        },
        "magnums_ammo": {
            "name": "Automatic Pistol Clips",
        }
    },
```
2. If you start a game, go to the passport and choose to return to the menu, the game freezes.
3. It's not a problem but just a curiosity: why are the Windows builds 32 bit?

Can I share the translation of Tomb Raider 3? I did't test if the game runs yet.

### DIFF
--- a/data/common/ship/cfg/base_strings-it.json5
+++ b/data/common/ship/cfg/base_strings-it.json5
@@ -110,7 +110,7 @@
         "ENUM_BILLBOARD_LOCK_PERSPECTIVE": "Prospettiva",
         "ENUM_BILLBOARD_LOCK_ROLL": "Rotazione",
         "ENUM_BILLBOARD_LOCK_ROLL_PITCH": "Rotaz. e inclinaz.",
-        "ENUM_BK_MONOCHROME": "\\{review}Monocromatico",
+        "ENUM_BK_MONOCHROME": "Monocromatico",
         "ENUM_BK_PATTERN_STATIC": "Statico",
         "ENUM_BK_PATTERN_WAVE": "Ondeggiante",
         "ENUM_BK_TRANSPARENT": "Trasparente",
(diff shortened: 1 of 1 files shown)


rr-:
> If you start a game, go to the passport and choose to return to the menu, the game freezes.

Why are you reporting this here?

rr- (approved):
TR3 is not ready, we don't even have English strings in place.

rr-:
> > If you start a game, go to the passport and choose to return to the menu, the game freezes.
> 
> Why are you reporting this here?

I've pushed a fix for that on develop. For future cases like this, please open a separate ticket – we're a tiny team and already juggling quite a bit.

Kappa971:
Regarding the duplicate strings? In the translation, I added them to all three files, just like in the English ones.

rr-:
> Regarding the duplicate strings? In the translation, I added them to all three files, just like in the English ones.

They differ in spelling in the English version – Medi Pack vs Medipack. It's an OG thing.
If the strings are identical, it's just a bit of boilerplate and it's OK to either keep them or remove them from individual games.

Kappa971:
> > Regarding the duplicate strings? In the translation, I added them to all three files, just like in the English ones.
> 
> They differ in spelling in the English version – Medi Pack vs Medipack. It's an OG thing. If the strings are identical, it's just a bit of boilerplate and it's OK to either keep them or remove them from individual games.

Ok, I reported it because I thought it was an oversight.

Regarding the bug, next time I will open an issue (I didn't think it was that problematic to report it here 😅).

Since the game was freezing, I noticed in Windows Task Manager that these precompiled builds are 32 bit. Is there a technical reason behind this? If not, I think there's no point in compiling for 32 bit.

rr-:
> Since the game was freezing, I noticed in Windows Task Manager that these precompiled builds are 32 bit. Is there a technical reason behind this? If not, I think there's no point in compiling for 32 bit.

Switching the builds to 64‑bit would cost me a couple days of rebuild hell for basically a vanity number in Task Manager. If it's stable and not memory‑bound, I'd rather spend that time on actual features. If you're up to it, you can work on it, starting [here](https://github.com/LostArtefacts/TRX/blob/develop/tools/shared/docker/game-win/Dockerfile).

PR #4453, opened 2025-12-23
sound: handle sound positions similarly

### LINKED ISSUE #4451: TRX bug: Lara's footstep SFX can be too quiet
Noted by Sabatu on Discord. In TR1, if Lara has weapons equipped (and mic-at-Lara option is off) her footsteps become very faint.
Regression from 1.0 (I think).

lahm86:
Resolves #4451.

#### Checklist

#### Description

This normalizes sound positioning between TR1 and TR2, aside from attenuation which remains fixed for TR1. This is slightly experimental because I'm not sure how I feel about it - I think some sounds are a bit louder, and getting things to balance came only with changing each of the pieces of logic in this PR (mic pos, default sample range, pan, attenuation factor). In the same breath, with SFX volume at 60% (which I normally have it at) I think it sounds acceptable. At 100% I think it's too loud, but that's maybe just me - LMK what you think.

I also tried a full combine with TR2's attenuation, but some TR1 fixed sound sources would cut off very abruptly when going out of range e.g. the switch in room 9 of The Cistern.

The alternative to this if we feel the results are unsuitable would be going back to using the camera target for TR1 rather than position, but with this the "mic at Lara" option almost becomes redundant.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -44,6 +44,7 @@
 - fixed several OG texture issues in Caves (rooms 0, 1, 2, 6, 24, 30 and 32)
 - fixed Lara automatically being given TR2 weapons in NG+ when playing the OG levels (#4365, regression from 1.0)
 - fixed Lara's pistol holster meshes appearing in NG+ in place of her Uzi holster meshes (#4368, regression from 1.0)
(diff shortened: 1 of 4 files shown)


rr- (approved):
This sounds fine to me!

aredfan (approved):

PR #4457, opened 2025-12-26
objects/switch: amend switch states

### LINKED ISSUE #4456: TRX bug: camera doesn't revert back to Lara after pulling the cog machine switch
Regression since [TRX 1.0](https://github.com/LostArtefacts/TRX/releases/tag/trx-1.0)

________________

1. launch tr1
2. `/play 3`
3. `/tp cog`
4. insert the 3 machine cogs
5. pull the switch and observe that the camera doesn't revert back to Lara

lahm86:
Resolves  #4456.

#### Checklist

#### Description

Level data defines 0 as switch state on and 1 as off. When testing triggers, the check was originally done against the switch having state `LS_RUN` to mean off, but this was fixed to use the correct switch state name, and in turn inherited the wrong value.  The utility switch state functions have been renamed slightly to reflect what they do more specifically.

It'd be good to double check all switch types (looks OK to me, but would appreciate a second test).

Do we want an FD injection fix to avoid the camera repeating if Lara steps off and back on the tile in Valley? I feel if it always happened it'd be fine, but the fact it's one shot in the flip room feels like an oversight in the unflipped room.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -39,6 +39,7 @@
 - fixed flare box pickups containig only one flare if Lara has none in her inventory at that time (#4423, regression from 1.0)
 - fixed water enemies appearing untinted for a frame after dying and moving to the water surface (#4420, regression from TR2X 0.1)
 - fixed the interactive fly cheat breaking with animated interactions enabled (#4444, regression from TR1X 4.14)
+- fixed switch triggers using an incorrect state check, which could result in fixed camera behavior that deviated from OG (#4456, regression from 1.0)
 
 **TR1**:
 - added the ability to change inventory and statistics background styles (pattern + wave are not implemented in TR1)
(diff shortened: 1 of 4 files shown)


aredfan:
LGTM, and thank you. 👍 

> Do we want an FD injection fix to avoid the camera repeating if Lara steps off and back on the tile in Valley? I feel if it always happened it'd be fine, but the fact it's one shot in the flip room feels like an oversight in the unflipped room.

I'm inclined to agree that this behaviour is okay as-is because it's an OG thing. Though a fix is always welcome.

aredfan (approved):

rr- (approved):
I trust your judgement re: what feels more natural re: the extra fix :relieved:

lahm86:
Will just leave the trigger quirk as-is for now - we can always adjust it another time if we feel it needs it.

PR #4462, opened 2025-12-29
Add support for cdaudio.wad to TR3

rr-:
#### Checklist

#### Description

This PR adds support for TR3 music playback by parsing the cdaudio.wad file for TR3+ and loading the tracks from memory, reconstructing the RIFF header from .wad. It also fixes a couple of small memory leaks and other Valgrind warnings.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -65,6 +65,7 @@
 - fixed demos not using loading screens
 
 **TR3**:
+- added cdaudio.wad music playback support
 - added monochrome inventory backgrounds
 - added high-resolutions 16:9 and 4:3 loading screens
 - added high-resolutions 16:9 and 4:3 title and game end screens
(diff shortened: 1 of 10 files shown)


rr-:
I think we should check if all tracks play properly. (`/music` can help with this.)
What I'm not sure is whether they really should all play, or if there are broken entries like in TR2 and TR1. I'd be good to compile a list.

lahm86 (approved):
LGTM. Track 78 is silent but this is intentional as it's used in e.g Tinnos to set the ambient track to be empty (#4463 raised as an aside as it can cut off other music, for example in the wind maze in that level). Tracks 0, 1, 79-81 and 123-129 are also empty, but these aren't referenced in the OG levels. I don't think any special handling is needed for those.

aredfan (approved):

Richard-L:
Iirc TR3 music has no unused tracks.

Maybe just to mention: TRR added a Willard boss fight music track which was supposedly cut due to the game's deadline. It resurfaced later but is clearly not produced by McCree: https://youtu.be/7VS8deMX-cI
(I personally think it's terrible)

PR #4466, opened 2025-12-29
Implement TR3 water and swamp surfaces displacement effect

rr-:
#### Checklist

#### Description

This PR implements the wave effect on the water surfaces. It doesn't implement caustics (as in brightness changes) yet.

It's not a perfect match, but the core idea's the same as in the original: eligible vertices have their Y component displaced by a sine wave with a random phase shift, a 64-frame frequency, and an amplitude based on the room's water setting. The random phase part is what's computed differently from the OG, but it shouldn't matter.

~~It also disables the "wibble" effect known from TR1 and TR2 – in TR3 the vertices don't move.~~ This is false and reverted.

These changes are hardcoded by `g_TRVersion`. To make these a proper setting, we'd need to compute the vertices attributes. I think `VERT_TR3_CAUSTICS_A` refers to water surfaces, and `VERT_TR3_CAUSTICS_B` to underwater vertices – these can be computed I guess. LMK if we want that, if yes, I'll do a follow-up PR.


rr-:
Updated the effect to also work in dry rooms when applicable.

aredfan (approved):

lahm86 (approved):

PR #4482, opened 2025-12-31
gun/rifle: separate M16 and MP5

lahm86:
#### Checklist

#### Description

This splits the M16 and MP5 into their own objects to allow for both to be used in the same level. The MP5 uses different SFX (a high pitched explosion, TIL) and its gunflare position and rotation is different compared to the M16. Other than that, the control was already working, the only other addition is the fix-accuracy-while-running config option now applies.

I also fixed the menu spin in/out SFX IDs.


rr- (approved):

data/tr3/ship/cfg/weapons.json5
```diff
@@ -65,18 +65,18 @@
...
-        "shot_accuracy": 4,
+        "shot_accuracy": 8,
         "gun_height": 500,
         "damage": 3,
         "target_dist": 12.0,
         "equip_anim_idx": 1,
-        "draw_frame": 10,
-        "undraw_frame": 21,
+        "draw_frame": 13,
+        "undraw_frame": 17,
```
aredfan: I've been doing frame-by-frame analysis of Lara drawing and undrawing the MP5 and comparing it with OG. 😅 

I believe the correct frames here are 16 and 21 respectively.

lahm86: You're absolutely right - fixed! Thanks.

aredfan (approved):

PR #4523, opened 2026-01-04
Add TR3 footprints and step sounds

rr-:
#### Checklist

#### Description

20260104_092910_Nevada_Desert

Turns out footprints are rendered as triangles, lol.
20260104_092854_Nevada_Desert


src/trx/game/anims/commands.c
```diff
@@ -33,7 +34,13 @@ static void M_ParseCommand(ANIM_COMMAND *const command, const int16_t **data)
         data->frame_num = *data_ptr++;
         const int16_t effect_data = *data_ptr++;
         data->effect_num = effect_data & 0x3FFF;
-        data->environment = (effect_data & 0xC000) >> 14;
+        data->fx_type = 0;
+        if (command->type == AC_EFFECT && g_TRVersion == 3) {
+            data->fx_type = effect_data & 0xC000;
+            data->environment = ACE_ALL;
```
rr-, line 40: This is a bit questionable and I'd appreciate confirming if this right at all.

src/trx/game/items/actions/common.c
```diff
@@ -2,6 +2,12 @@
 #include <trx/game/rooms.h>
 
 static void (*m_Routines[ITEM_ACTION_NUMBER_OF])(ITEM *item) = {};
+static int16_t m_FXType = 0;
+
+int16_t ItemAction_GetFXType(void)
+{
+    return m_FXType;
```
rr-, line 9: LMK if this should become a parameter to the `action_func` instead.

rr-:
Found a few bugs:
- The footprints seem to be too large.
- They're too dark.
- They appear angled near 1-click surfaces.
    20260104_101233_Lara&#39;s_Home


I'll investigate.

rr-:
I think I got all 3 issues fixed!

src/trx/game/anims/commands.c
lahm86: This should be fine for now. Anim command flip effects don't use the environment flag, it's only for sound commands (we just use the same structure). I think when we later port this to 1/2 we could potentially just reinterpret `environment` as `fx_type`; WADTool already allows setting left/right foot in TR1/2, it just gets ignored and indeed isn't present in OG levels.

src/trx/game/items/actions/common.c
lahm86: This is fine IMO 😌

lahm86 (approved):

aredfan (approved):

PR #4547, opened 2026-01-06
Add TR3 schools of fish

rr-:
#### Checklist

#### Description

This adds piranhas and tropical fishies.
Blood in water not yet implemented.
Crash Site dino carcass object partially implemented (fishies will attack it, but the object itself will have to wait until Crash Site stuff).

OG pretty badly hardcodes each fish group's per-level parameters; I'm leaving that as-is for now.


src/trx/game/objects/vars.c
```diff
@@ -622,3 +622,10 @@ const OBJECT_ID g_SmashableObjects[] = {
     NO_OBJECT,
     // clang-format on
 };
+
+const OBJECT_ID g_ShoalObjects[] = {
+    // clang-format off
+    O_TROPICAL_FISH,
+    O_PIRAHNAS,
+    // clang-format on
```
lahm86, line 631: Missing `NO_OBJECT` to guard `Object_IsType`

rr-: Thanks – fixed

lahm86 (approved):

rr-:
I fixed the following issues:

1. Triggers that do not set the timer (that indicates which of the hardcoded movement ranges to use), will no longer cause the fishes to swim at hypersonic speeds.
2. The issue with blending reported by aredfan on Discord

Testing:

We need to test the performance of weather effects – they _should_ be fine but I'd like a confirmation.

aredfan:
> 1. Triggers that do not set the timer (that indicates which of the hardcoded movement ranges to use), will no longer cause the fishes to swim at hypersonic speeds.
> 2. The issue with blending reported by aredfan on Discord

Thank you for fixing these issues. 👍 

> We need to test the performance of weather effects – they _should_ be fine but I'd like a confirmation.

The rain in Thames Wharf and the snow in Antarctica both LGTM.

I noticed that the fish doesn't always orientate in the right direction when moving. For example in the test level, the fish can give the impression of swimming backwards.  I've checked the fish behaviour OG and they tend to face in the direction they're swimming.

rr-:
Thanks – good point. I believe I have fixed this as well.

aredfan:
> Thanks – good point. I believe I have fixed this as well.

Thank you @rr-, I can confirm the fishes are always facing the right direction. 👍 

I have one more issue, when Lara swims closely with the fishes, there's a slight blur on the fishes which appears to be caused by interpolation.  I made a recording for this, you can enjoy watching Lara struggle to keep up with the fishes. 😅 

[record.txt](https://github.com/user-attachments/files/24466786/record.txt)

rr-:
Inteprolation added :relieved:

aredfan (approved):
> Inteprolation added 😌

Brilliant.  Thank you. 👍 

The tropical fishes looks absolutely stunning.

PR #4560, opened 2026-01-08
Implement TR3 flare effects

rr-:
#### Checklist

#### Description

This PR implements TR3 flare sparks and RGB lighting.
It also fixes the opacity of non-sprite sparks, which OG TR3 caps at 0x80.

#### Testing

Just flare stuff:
- handheld
- thrown
- going out in both variants
- underwater in both variants
- sanity checks in TR1/TR2 for potential regressions

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -115,6 +115,7 @@
 - added surface-based step sounds
 - added cold breath effects
 - added gun shells
+- added flare lighting and sparks
 - added monochrome inventory backgrounds
 - added high-resolutions 16:9 and 4:3 loading screens
 - added high-resolutions 16:9 and 4:3 title and game end screens
--- a/src/trx/game/lara/draw.c
+++ b/src/trx/game/lara/draw.c
@@ -8,6 +8,7 @@
 #include <trx/game/output.h>
 #include <trx/game/output/vars.h>
 #include <trx/game/random.h>
+#include <trx/version.h>
 
 static bool m_CacheMatrices = false;
 
@@ -180,7 +181,8 @@ static bool M_Draw_I(
         M_DrawBodyPart(LM_LARM_L, bone, mesh_rots_1, mesh_rots_2, clip);
         M_DrawBodyPart(LM_HAND_L, bone, mesh_rots_1, mesh_rots_2, clip);
 
-        if (lara->gun_type == LGT_FLARE && lara->left_arm.flash_gun) {
+        if (g_TRVersion < 3 && lara->gun_type == LGT_FLARE
+            && lara->left_arm.flash_gun) {
             Gun_DrawFlash(LGT_FLARE, clip, true);
         }
         Matrix_Pop();
@@ -475,7 +477,8 @@ bool Lara_Draw(const ITEM *const item)
         M_DrawBodyPart(LM_LARM_L, bone, mesh_rots, nullptr, clip);
         M_DrawBodyPart(LM_HAND_L, bone, mesh_rots, nullptr, clip);
 
-        if (lara->gun_type == LGT_FLARE && lara->left_arm.flash_gun) {
+        if (g_TRVersion < 3 && lara->gun_type == LGT_FLARE
+            && lara->left_arm.flash_gun) {
             Gun_DrawFlash(LGT_FLARE, clip, false);
         }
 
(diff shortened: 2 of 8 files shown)


lahm86:
Looks great.

I can't seem to get the sparks to appear when Lara is underwater - OG (below).

<details>

image
</details>

aredfan:
- [x] When an active flare is on the ground, the sparks and smoke doesn't draw.  OG screenshot below, for reference.

Screenshot

</details>

**Edit:** The spark and smoke stops being drawn when an active flare leaves Lara's hand.

rr-:
> Looks great.
> I can't seem to get the sparks to appear when Lara is underwater - OG (below).

Fixed, though I'm not 100% sure if the flare is tinted. If it is, I'd like to tackle this separately.

rr-:
> * [ ]  When an active flare is on the ground, the sparks and smoke doesn't draw.  OG screenshot below, for reference.
> **Edit:** The spark and smoke stops being drawn when an active flare leaves Lara's hand.

Fixed this as well :)

lahm86:
> > Looks great.
> > I can't seem to get the sparks to appear when Lara is underwater - OG (below).
> 
> Fixed, though I'm not 100% sure if the flare is tinted. If it is, I'd like to tackle this separately.

I'm not 100% sure either, but leaning towards thinking they aren't in OG. Could be a nice feature anyway 😌

lahm86 (approved):

aredfan:
> Fixed this as well :)

Thank you @rr- 👍 

While I was retesting, I found a new issue.  

- [x] In most cases, if the inventory/pause screens/photo mode is open while there's an active flare, the sparks continue to disappear.  I've made a recording to show this in action.

[record.txt](https://github.com/user-attachments/files/24497801/record.txt)

rr-:
> > Fixed this as well :)
> 
> Thank you @rr- 👍
> 
> While I was retesting, I found a new issue.
> 
> * [ ]  In most cases, if the inventory/pause screens/photo mode is open while there's an active flare, the sparks continue to disappear.  I've made a recording to show this in action.
> 
> [record.txt](https://github.com/user-attachments/files/24497801/record.txt)

Thanks. The OG was spawning the sparks in the draw routine rather than the control routine. Doing this in TRX is misguided for two reasons:
1. We can end up drawing twice per logical frame, meaning in 60 FPS the flare sparks would appear twice as bright.
2. In phases that do not call the control phase of the game (eg inventory ring, pause screen, photo mode), we saturate the maximum spark slots by constantly trying to allocate new sparks. To illustrate this better: when the sparks array is full, newly spawned sparks push out the oldest spark out. But we only clear the spark array at the start of the logical frame, in the control routines (which get never called here), so the array gets full, and starts to purge the existing sparks after a while. Finally, the new flare sparks that dominate the buffer are almost transparent, hence the decaying effect we could observe prior to the fix.

I had to deviate from the OG by moving the thrown flare sparks to its control phase. This needs retesting of the thrown flares against different environments (surface/underwater), and some sanity tests in TR1/TR2.
This change was only required for the thrown flares. The reason why the bug appears to have affected the handheld flares too, is that it only takes one item on the screen (in this case the thrown flare) to try and continuously schedule new sparks, to push out _all other visible effects_ which in those frozen frames get never scheduled again.

BTW I didn't think of this but it'd be useful to test them with swamps, too.

aredfan:
Thank you again @rr-, the prior issue is resolved too and the games passed sanity checks.

I found a new issue after retesting. 😅 

- [x] TR3: Active flares that are discarded have a gap between the spark/smoke and the rod. In the screenshots below, I had Lara throw away an active flare in water because it falls more slowly, thus easier to observe.

<details><summary>TRX</summary>

20260108_182213_Lara&#39;s_Home

</details>

<details><summary>OG</summary>

OG

</details>

> BTW I didn't think of this but it'd be useful to test them with swamps, too.

I haven't checked this either but it turns out that swamps gobble up flares. This is in line with OG behaviour.

rr-:
> Thank you again @rr-, the prior issue is resolved too and the games passed sanity checks.
> 
> I found a new issue after retesting. 😅
> 
> * [ ]  TR3: Active flares that are discarded have a gap between the spark/smoke and the rod. In the screenshots below, I had Lara throw away an active flare in water because it falls more slowly, thus easier to observe.

Thank you for noticing this. Fixed that as well.

aredfan (approved):

PR #4564, opened 2026-01-08
sound: fix wrongly scaled decibel LUT

rr-:
#### Checklist

#### Description

TR1/2 volume is computed in a log2 "dB-like" domain:

- `m_DecibelLUT[…]` maps a volume index to a "dB-like" gain value
- `M_DecibelToMultiplier(db_gain)` turns that number back into a linear gain multiplier, with which we can multiply the raw samples that we feed into the SDL mixer

The TR1/TR2 code path was originally ported from the glrage compatibility patch, including its "decibel LUT". That LUT uses log2(), but scales it as if one doubling corresponded to 10 dB, i.e. "+1000 per doubling". That's not how real decibels work: for amplitude, +6.02 dB is one doubling (because dB is defined in base‑10 as gain = 10^(dB/20)).

The audio backend was implemented separately under the "real dB" intuition: if a value behaves like centi‑dB, then the natural base‑2 rewrite is "~600 per doubling" (since 6.02 dB ≈ 602 centi‑dB), leading to gain = 2^(value/600).

Those two assumptions (*1000 in the LUT vs /600 in the backend) are not inverses. The mismatch crushes low volumes and makes quiet samples far quieter than intended.

The backend scaling is adjusted to match the inherited glrage LUT (gain = 2^(value/1000)), restoring internal consistency without redesigning the curve yet.

#### Testing

Volumes of distant objects, quiet objects (bat wings, gun shells); of loud objects (waterfalls, machines).

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -65,6 +65,7 @@
 - fixed developer console text editing (backspace, moving the caret) doing weird things with Unicode characters
 - fixed Lara jumping if player holds the swim button when exiting the fly cheat (#4470)
 - fixed game refusing to load savegames made with the JP mode (#4558)
+- fixed quiet sounds being almost inaudible
 
 **TR1**:
 - added the ability to change inventory and statistics background styles (pattern + wave are not implemented in TR1)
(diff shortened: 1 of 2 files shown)


rr-:
Note: we need to compare against the OG in all cases, and all TR1-3.
I think this is a cursed PR because it changes stuff we already got used to, but what prompted me was the gun shells practically *fading into nothing* with the present scaling, despite being *very audible* in the OG.

rr-:
Closing in favor of the other approach.

PR #4566, opened 2026-01-08
Alternative fix to the sound decibel LUT

rr-:
#### Checklist

#### Description

In https://github.com/LostArtefacts/TRX/pull/4564, the TR1/TR2 path stayed more faithful to what we actually ported from the glrage compatibility patch: the game side emits a "log2 * 1000" dB‑like value, and the backend uses the matching inverse 2^(x/1000). That kept the inherited curve (even if it isn't "real decibels") and fixed the immediate bug: the LUT and the inverse stopped disagreeing.

This changeset is the "nuke from orbit" alternative solution: it stops honoring the inherited glrage curve and instead redefines the meaning of the LUT output to match what DirectSound claims to accept (centi‑dB in [-10000..0]) and what TR2 and TR3's DirectSound path is built around conceptually (handing DirectSound a "dB attenuation number", not a custom log2 unit).

Two coordinated changes implement that:

- In src/trx/game/sound/common.c, `m_DecibelLUT[]` is rebuilt as true centi‑dB: for each gain fraction i/512, it stores `centi_db = 2000 * log10(gain)` clamped to [-10000..0]. This is the standard amplitude-decibel relation expressed in hundredths of a dB.
- In src/trx/engine/audio_sample.c, `M_DecibelToMultiplier()` switches to the DirectSound inverse: `gain = 10^(centi_dB/2000)`. That produces the linear multiplier the SDL mixing code needs from a centi‑dB input.

So compared to the "faithful to glrage" PR, this one intentionally changes TR1 loudness behavior: it replaces the legacy log2-scaled "dB-like" units with real base‑10 decibel attenuation semantics, using the same model DirectSound expects and that TR2 and TR3's audio stack revolves around.

#### Testing

Same as the other PR. Ultimately we must decide which curve we want to go with (eg which PR sounds "better"). Biggest differences should be around quiet sources.

Also it's important to test volume slider behavior.

### DIFF
--- a/src/trx/engine/audio_sample.c
+++ b/src/trx/engine/audio_sample.c
@@ -89,8 +89,13 @@ static AUDIO_SAMPLE_SOUND m_Samples[AUDIO_MAX_ACTIVE_SAMPLES] = {};
 
 static double M_DecibelToMultiplier(double db_gain)
 {
-    // DirectSound-style centi-dB domain: gain = 10^(centi_dB/2000).
-    return pow(10.0, db_gain / 2000.0);
+    if (g_TRVersion < 3) {
+        // Legacy scale
+        return pow(2.0, db_gain / 600.0);
+    } else {
+        // DirectSound-style centi-dB domain: gain = 10^(centi_dB/2000).
+        return pow(10.0, db_gain / 2000.0);
+    }
 }
 
 static bool M_RecalculateChannelVolumes(int32_t sound_id)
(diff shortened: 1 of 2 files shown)


lahm86 (approved):
Sounds good to me. I vote for this approach over #4564.

aredfan (approved):
I'm also happy to go with this approach. 👍

rr-:
I ended up gating the new curve behind a version check, and restoring the previous curve for TR1&2. Let's see how this is received.

rr-:
Something weird happened in here but the commits landed in develop safely.

PR #4568, opened 2026-01-08
level: remove statics limit

rr-:
#### Checklist

#### Description

This changes the statics storage to be dynamically allocated with `GameBuf`. It's a sparse array and will just overallocate "holes", so if the builder goes for absurdly high IDs, that's on them.


aredfan (approved):

src/trx/game/inject/common.c
```diff
@@ -112,6 +141,103 @@ static void M_InitialiseBlock(
...
+            }
+            if (obj_type == OBJ_TYPE_OBJECT && version < INJ_VERSION_5) {
+                VFile_Skip(file, 16);
+            }
+            VFile_Skip(file, sizeof(int16_t) * 2);
+        }
+        return;
+    }
+
+    case IDT_ROOM_EDITS: {
```
lahm86: Is this only in place to skip over the data? If so, we can eliminate it as this function will default to doing so anyway using the overall `data_size`. I don't think there is anything in the room meta that affects any new IDs, it's just the statics and sprites cases you've got covered 😌

rr-: No, it's to keep track of the max ID wanted by the injections, should we want to create new statics:
`M_RoomMetaAddStatics(target_room, 1, 0);`

Ideally, the max indices could be provided by `IDT_ROOM_EDIT_META` so that we could forgo looping over it again.

src/trx/game/inject/common.c
```diff
@@ -33,11 +33,40 @@ static int32_t m_NumInjections = 0;
...
 
 static int32_t m_DataCounts[IDT_NUMBER_OF] = {};
+static int32_t m_MaxStaticObject3DId = -1;
+static int32_t m_MaxStaticObject2DId = -1;
 static VECTOR *m_RoomMeta = nullptr;
 static LEVEL_INFO m_CachedInfo = {};
 static uint16_t *m_PaletteMap = nullptr;
 static size_t m_PaletteMapSize = 0;
 
+static void M_RoomMetaAddStatics(
```
lahm86: `IDT_ROOM_EDIT_META` already contains a summary of 2ds and 3ds to add, I think this duplicates information.

src/trx/game/inject/common.c
lahm86: But if we're adding new statics, that will have been captured by `IDT_STATIC_OBJECTS`. `IDT_ROOM_EDITS` is just room instance stuff, so it's not creating new static objects, just references to them in the rooms. And those numbers are handled already in `IDT_ROOM_EDIT_META` so the room knows how much space to allocate for the instances.

rr-: Got it – removed this case item along with `M_RoomMetaAddStatics` (kept `Inject_GetMaxStaticObject[23]DId` though).

src/trx/game/inject/common.c
rr-: Thanks – last night, I failed to notice this:
```c
            if (version >= INJ_VERSION_3) {
                meta.num_static_3ds = VFile_ReadS16(file);
            }
```

rr-:
I've updated the code and there seem to be no problems with any level in TR1 and TR2, though I'd appreciate a quick check if the injected statics work as intended.

lahm86 (approved):
LGTM, thanks!

PR #4585, opened 2026-01-09
camera: implement TR3 camera

lahm86:
#### Checklist

#### Description

This is a draft for now as we have a couple of snags, and I wanted to open this up for experimenting with. Please feel free to add to the issues below. I'm not especially happy with the final commit for `Camera_Update` - the function has become quite disjointed. It may need us to have a copy of the whole function in both strategies for maintainability, but I'm open to any thoughts.

- [ ] In Jungle room 10, have Lara walk onto the spikes so she looks up at the cave target. If she then stands still, the camera shifts every few frames - edit, this looks like an OG bug
- [x] Have Lara with her back against a wall and press look, adjust the camera so Lara is completely out of sight - the camera will void. I think this might be mesh pos problems - as Lara's clipped away, her matrices become invalid
- [x] Refactor look Lara state check once function stable
- [x] There is some rigid movements in the camera when moving against corners,
> **Recording:** [record.txt](https://github.com/user-attachments/files/24540526/record.txt)
> **OG SG:** [savegame.zip](https://github.com/user-attachments/files/24540527/savegame.zip)


rr- (commented):
I think I'd strategize on `Camera_Update()` mainly, and share really common bits in `camera/common.c` similar to what we have for `Lara_Col_`.

lahm86:
This should be ready for a proper review now. The look-at bug (in Jungle) appears to be an OG problem, so we can tackle that separately. The rigid movement should be resolved too, I'll let you double check before ticking it off.

I'm a little uneasy about enforcing Lara draw to always run even if Lara should be fully clipped - we're effectively relying on draw behaviour for the look camera control. The alternative I think would be having to calculate the matrices separately from the draw routine, but that's a sizeable task.

I'll come back to further refactors with the camera modules later, especially in terms of the update routine. These are separated between the two cameras for now given the differences.

aredfan (approved):
Great work. 😌

rr- (approved):
Thank you and I'm sorry for the wait time.

PR #4598, opened 2026-01-10
lara/state: fix walking / sidesteps near slopes

rr-:
#### Checklist

#### Description

This PR fixes the issue where walking sideways, backwards and forwards near slopes would result in Lara twitching in TR3 which doesn't happen in the OG.

Fix only applied to TR3. LMK if we want to backport it to earlier games too – if so, I'd make it non-conditional, unless it breaks some speedrunning strats.


aredfan (approved):

lahm86 (approved):
I think the TR3 guard is fine for now - once we get through more state/col comparisons we can maybe weigh up the differences and decide on any new config options, or merging with existing options.

PR #4600, opened 2026-01-11
Fix TR3 hang tests

rr-:
#### Checklist

#### Description

This PR fixes the hanging issue on triangulated geometry. Turns out the real culprit was `M_TestHangJump` missing the new `switch` statement in one of its branches. 
I also ported an extra offset I found in `M_TestHangJumpUp` and `Lara_Col_HangTest`. The PR even makes use of the extra sides – though that part ended up being a total red herring. I brought the logic over anyway, but honestly, I'm still not sure what it's supposed to fix :sweat_smile: 

#### Testing

We need to test a variety of hanging positions in TR1, TR2, and TR3 across different types of geometry. I don't remember any speedrunning strats that specifically rely on the hanging state, but if there are any, those should be tested as well.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -166,6 +166,7 @@
 - fixed Quad Bike wheels appearing to spin backwards at high speeds
 - fixed the skybox's blue lid for the Thames Wharf and City cutscenes
 - fixed fish schools to no longer swim at supersonic speeds if their triggers do not have timers set, or reuse the same timer
+- fixed Lara letting go of some ledges
 
 ## [1.0.3](https://github.com/LostArtefacts/TRX/compare/trx-1.0.2...trx-1.0.3) - 2025-11-27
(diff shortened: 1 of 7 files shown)


lahm86:
Great find!
I think we'll need to extend this to TR1/2 as well because we support triangles there (bug is present on 1.0.3, but arguably nobody is able to use triangles yet because of TE). Maybe we could have a flag in the room module if it has detected a floor split while parsing FD, and that could be the trigger to use this approach rather than `g_TRVersion`? Or equally we could use TR3 approach as standard, but obviously test everything else in OG thoroughly.

rr-:
I think we should just make the fix unconditional. Pushed a removal of `g_TRVersion` gate for this fix. The other places still use the gates.

lahm86 (approved):
Thank you 😌 LGTM

aredfan (approved):

PR #4602, opened 2026-01-11
Implement TR3 per-mesh underwater tint

rr-:
#### Checklist

#### Description

Only works for Lara (and her braid).


src/trx/game/lara/draw.c
lahm86: I think we can apply the same technique here - currently if Lara is at wading depth (e.g. Gym `/tp 25 3 15`) with a gun on her back, her torso isn't tinted, but the gun is. Similar further below for non-interp.

rr-: Fixed

aredfan (approved):

lahm86 (approved):

PR #4630, opened 2026-01-15
objects: add bubble emitter

rr-:
#### Checklist

#### Description

This is just a draft – I tied the timer to mean "spawn 1 bubble every X frames (on average)" to let the builder control the strength of the effect. Not sure how canon this is.

I've also checked tomb3, tomb4 and tomb5 for potential OG implementations but haven't found anything.

### DIFF
--- a/src/trx/game/objects/traps/bubble_emitter.c
+++ b/src/trx/game/objects/traps/bubble_emitter.c
@@ -1,0 +1,33 @@
+#include <trx/game/effects.h>
+#include <trx/game/objects.h>
+#include <trx/game/random.h>
+#include <trx/game/sound.h>
+#include <trx/game/spawn.h>
+
+static void M_Control(const int16_t item_num)
+{
+    ITEM *const item = Item_Get(item_num);
+    if (!Item_IsTriggerActive(item)) {
+        item->status = IS_INACTIVE;
+        Item_RemoveActive(item_num);
+        return;
+    }
+
+    if (Random_GetControl() % 24) {
+        return;
+    }
+    const int32_t count = 1 + Random_GetControl() % 2;
+    for (int32_t i = 0; i < count; i++) {
+        Spawn_Bubble(&item->pos, item->room_num);
+    }
+}
+
+static void M_Setup(OBJECT *const obj)
+{
+    obj->control_func = M_Control;
+    obj->collision_func = nullptr;
+    obj->draw_func = nullptr;
+    obj->save_flags = true;
+}
+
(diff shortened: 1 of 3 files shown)


rr-:
I've tested it with
```diff
diff --git a/src/trx/game/level/loader.c b/src/trx/game/level/loader.c
index 61497d483..292bf20e8 100644
--- a/src/trx/game/level/loader.c
+++ b/src/trx/game/level/loader.c
@@ -1402,6 +1402,10 @@ void Level_ReadItems(const LEVEL_LOADER *const loader, VFILE *const file)
         item->rot.y = VFile_ReadS16(file);
         M_ReadShade(loader, &item->shade, file);
         item->flags = VFile_ReadS16(file);
+        if (item->object_id == O_EMBER_EMITTER) {
+            item->object_id = O_BUBBLE_EMITTER;
+            LOG_INFO("bubble emitter: %d %d %d / %d", item->pos.x, item->pos.y, item->pos.z, item->room_num);
+        }
     }
 
 finish:
```

and using this in Diving Area's toxic waste room with `/flood`.

rr-:
Here's how it works with the max rate:

https://github.com/user-attachments/assets/5acd9fee-e15b-48c1-94e1-72cf1a37a071

src/trx/game/objects/traps/bubble_emitter.c
```diff
@@ -0,0 +1,31 @@
+#include <trx/game/effects.h>
+#include <trx/game/objects.h>
+#include <trx/game/random.h>
+#include <trx/game/sound.h>
+#include <trx/game/spawn.h>
+
+static void M_Control(const int16_t item_num)
+{
+    ITEM *const item = Item_Get(item_num);
+    if (!Item_IsTriggerActive(item)) {
```
lahm86, line 10: This will change the timer field, but I agree it's important to be able to switch them off so they're not permanently active.
Maybe we could assign a random low value on initialise to priv, or just alter the default randomness. Builders could control the effect with having more than one item.

lahm86 (commented):
This looks great! Just a potential issue below re the timer, I can make a test level later today if needed.

src/trx/game/objects/traps/bubble_emitter.c
rr-: Yeah I'll change it to do that, instead. I'd appreciate a test level.
Come to think of it – we can always introduce the mythical custom properties via lua to control this… once we implement them.

rr-: Updated~

lahm86 (approved):

aredfan (approved):

PR #4632, opened 2026-01-16
sparks: fix burning muzzle effect

rr-:
#### Checklist

#### Description

This fixes the "burning muzzle" effect in Temple of Puna, Lara's Home and Lud's Gate.

These levels were affected because the smoke sparks inherited their first-frame color from the previous sparks – which happened to be flame sparks. The affected levels were the only ones that start with active flames in the opening area.


rr-:
Previously we had a similar bug with the ricochets being black on the first frame; I applied the same fix for every spark effect (explosions, bubbles, smoke effects etc) that we have.

aredfan (approved):

lahm86 (approved):

PR #4633, opened 2026-01-16
lara/look: define TR3 look angles

lahm86:
#### Checklist

#### Description

This defines the look angle behaviour for Lara in TR3; she has a more lenient tilt and there is no difference between regular look and idling on water surface, unlike TR1/2. I've also tidied up a few magic numbers in the camera look function.

I think it'd be nice to have the angles defined in the camera strategies, so that it's dictated by whichever one is in use rather than the game version. But I can tackle that separately at a later date.


rr- (approved):
Thanks and agreed that strategizing the angles is the right direction.

aredfan (approved):

PR #4642, opened 2026-01-17
output: fix drawing shadow twice

### LINKED ISSUE #4640: TRX bug: darker shadows when items are fully clipped
Notice in the first screenshot below the flyer's wing is drawn top-centre screen. In the second, it is fully clipped and the shadow becomes darker.

<details>

Image
Image
</details>

Potentially relevant: #879.

rr-:
#### Checklist

#### Description

This introduces a new module, `output/bind.c`, to avoid requiring everyone who wants to draw items, to pass non-const pointers so that Output_DrawShadow can mutate the binding state.

LMK if we like this, then I'll refactor rooms as a follow-up request to use this pattern too.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## [Unreleased](https://github.com/LostArtefacts/TRX/compare/trx-1.1...develop) - ××××-××-××
 - added a bubble emitter (#4629)
 - improved inventory ring active item highlight for smoother appearance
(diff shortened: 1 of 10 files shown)


lahm86 (approved):
Thanks. Nice approach, so rooms follow-up sounds good 👍

rr-:
Updated the code to also cover rooms, and also treated `(test|bound)_(left|right|top|bottom)` variables to use this pattern as well.

aredfan (approved):

PR #4644, opened 2026-01-17
Implement TR3 quest items

### LINKED ISSUE #4415: OG bug: two of the meteorite artefacts have incorrect names
I figured it's a good time to revive this issue: https://github.com/LostArtefacts/Tomb3Main/issues/1
______________________

> Thanks to Richard Morton, head level designer for Tomb Raider 3, for confirming that the item at the end of Nevada should be named Element 115. Additional thanks go to to Ian Taylor for initially confirming for us that Lara picks up Element 115 at the end of this level in the Sony PlayStation version of Tomb Raider 3. Players of the North American PC version of Tomb Raider 3 will find that the object that Lara picks up is named the Ora Dagger here instead of Element 115. It is a logical conclusion that Element 115 should be found in Nevada because it is claimed by certain individuals to be the power source for space vehicles of extra-terrestrial origins.
> 
> [Source](https://web.archive.org/web/20160503042908/http://trinity.westhost.com/tomb/meteor/07w.html#notes)

___________________

To summarize:

- The Nevada artefact is called `Element 115`
- The South Pacific artefact is called `Ora Dagger`

rr-:
#### Checklist

#### Description

This implements TR3 meteorite artifacts.

#### Testing

- Picking up in any level should end the level
- Picking up in Meteorite Chamber should just get it to the backpack
- It should render properly in the inventory
- They should be carried to the next level
- Saving, loading, and advancing to the next level should carry them still
- They should be examinable
- So should the Crash Site Swamp Map
- The artifacts should glow!


src/trx/game/option/common.c
```diff
@@ -92,15 +93,26 @@ void Option_Control(INVENTORY_ITEM *const inv_item, const bool is_busy)
...
         }
         break;
 
+    case O_QUEST_OPTION_1:
+    case O_QUEST_OPTION_2:
+    case O_QUEST_OPTION_3:
+    case O_QUEST_OPTION_4:
+    case O_PICKUP_OPTION_1:
+    case O_PICKUP_OPTION_2:
+        // This lets the player examine the item with the Action key.
```
lahm86: Builders can currently achieve the same effect by setting `"description": ""` in the strings -  maybe this would be sufficient for this case as well. The version guard here might also cause a regression for pickup 1/2 examining in TR1/2 customs.

data/tr3/ship/cfg/tr3/strings.json5
```diff
@@ -52,7 +57,19 @@
                 "puzzle_1": {
                     "name": "Serpent Stone",
                 },
-            },
+                "quest_1": {
```
lahm86: Just checking if the duplication of the names per level is intended. If they're in `base_strings.json5` they should inherit here I think?

data/common/ship/cfg/base_strings-en-gb.json5
```diff
@@ -41,5 +41,19 @@
         "SETTING_WATER_COLOR_R": "Water colour (R)",
         "SETTING_WATER_COLOR_R_DESCRIPTION": "Red component of the water colour.",
         "STATS_DISTANCE_TRAVELLED": "Distance Travelled",
+    },
+    "objects": {
```
lahm86: `en-gb` strikes again 😄

src/trx/game/option/common.c
rr-: I'll test this, thanks

data/tr3/ship/cfg/tr3/strings.json5
rr-: Yeah – it's intended, and unfortunate. I purposefully left the quest item names in base strings game-agnostic as "Quest Item 1…4" in order to avoid tying them to TR3. But this means that every level which has these items, has to override them all.
I think while this kind of sucks, it's the more elegant solution.

data/common/ship/cfg/base_strings-en-gb.json5
rr-: Thanks – fixed!

src/trx/game/option/common.c
rr-: Added, but this creates a button "Use" which for quest items is not the best. I'll remove it as a follow-up commit but I'm running out of time ATM.

rr-: Got it out as well – I think it's much better now, because the UI now makes it clear that you can examine the item. Thanks for this suggestion.

data/tr3/ship/cfg/tr3/strings.json5
lahm86: Ah, got it. Maybe we could add them to the main `objects` section only in the TR3 strings files (if I'm remembering the inheritance order correctly) instead of each level.

src/trx/game/option/common.c
lahm86: Forgot about the buttons. Thanks.

data/tr3/ship/cfg/tr3/strings.json5
rr-: I didn't think of it… that's brilliant. I'll check it after!

data/tr3/ship/cfg/tr3/strings.json5
```diff
@@ -52,7 +58,23 @@
...
+                    "name": "Infada Stone",
+                    "description": "",
+                },
+                "quest_2": {
+                    "name": "Element 51",
```
aredfan: Element `115`

aredfan:
- [x] The Ora Dagger is slightly buried in Temple of Puna and Meteorite Cavern.

> <details><summary>Screenshot</summary>
> 
> 20260117_162726_Temple_of_Puna
> 
> </details>

I found 4 issues in Sleeping with the Fishes.

- [ ] ~~The Hand of Rathmore rotates in this level whereas it is stationary in OG.~~ [savegame.zip](https://github.com/user-attachments/files/24690730/savegame.zip)
- [x] This pickup is labelled `Quest Item 1` if the all items cheat is used. It's called `Hand of Rathmore`
- [x] Collecting the quest item ends the level. This doesn't happen in OG.
- [ ] ~~There is flickering light on the pipe above the quest item.~~

> 1. ~~launch LA~~
> 2. ~~`/play 4`~~
> 3. ~~`/flip`~~
> 4. ~~`/tp quest`~~

lahm86:
Re the embedded Ora Dagger, LMK if we'd like an injection - its frame bounds are wrong.

data/tr3/ship/cfg/tr3/strings.json5
rr-: LOL. Thanks.

rr-:
> Re the embedded Ora Dagger, LMK if we'd like an injection - its frame bounds are wrong.

Yes please :happybat:

data/tr3/ship/cfg/tr3/strings.json5
rr-: It's working great. Thanks

lahm86:
> > Re the embedded Ora Dagger, LMK if we'd like an injection - its frame bounds are wrong.
> 
> Yes please :happybat:

Here we go 😌 
[ora_dagger.zip](https://github.com/user-attachments/files/24692153/ora_dagger.zip)
https://github.com/LostArtefacts/TRXInjectionTool/pull/196

src/trx/game/option/common.c
```diff
@@ -92,11 +93,15 @@ void Option_Control(INVENTORY_ITEM *const inv_item, const bool is_busy)
...
+    case O_QUEST_OPTION_1:
+    case O_QUEST_OPTION_2:
+    case O_QUEST_OPTION_3:
+    case O_QUEST_OPTION_4:
     case O_PICKUP_OPTION_1:
     case O_PICKUP_OPTION_2:
     case O_SCION_OPTION:
     case O_LEADBAR_OPTION:
-        if (!is_busy) {
+        if (!is_busy && g_TRVersion < 3) {
```
lahm86: I think this is causing an extra input requirement with all key items in TR3 e.g.
- `/play 2`
- `/give all`
- `/tp keyhole`
You have to press action 3 times to use the key rather than twice.

rr-: Sorry, this was a mistake from a previous cleanup. Thanks for spotting it.

lahm86 (approved):
The artefacts look fantastic 😌

lahm86:
[ora_dagger.zip](https://github.com/user-attachments/files/24692431/ora_dagger.zip)
Updated injection to fix https://github.com/LostArtefacts/TRXInjectionTool/pull/196#issuecomment-3764407939 - thanks @aredfan

aredfan (approved):

PR #4651, opened 2026-01-17
objects/save-crystal: add TR3 heal crystals

rr-:
#### Checklist

#### Description

They glow, they spin, and they bounce, too!


aredfan:
- [x] Some crystals are deactivated when starting a level and depend on triggers to get activated.  For example in Antarctica, room 126, the crystal located there is static until activated by a trigger.

rr-:
Thanks. Fixed this, but the fix is disruptive and may affect TR1 (TR2?) save crystals – we need to test for crashes around saving/loading.

src/trx/game/objects/general/save_crystal.c
```diff
@@ -42,10 +55,12 @@ static const OBJECT_BOUNDS *M_Bounds(void)
 
 static void M_Initialise(const int16_t item_num)
 {
-    if (g_Config.gameplay.enable_save_crystals) {
+    ITEM *const item = Item_Get(item_num);
+    if (g_Config.gameplay.enable_save_crystals || g_TRVersion == 3) {
```
lahm86: In OG, crystals don't have an initialise function that forces all to be active - they'll either be activated by their flags in the level (default item initialise mechanic) or can be triggered at some point from FD. I think this is related to aredfan's findings.

rr-:
Thanks – I removed the forced activation, but this means that some of the crystals the players encounters in a normal way, will appear as not active and weird – example: item 90 in Antarctica, when we look at it _before_ we jump.

I can solve this via injections, please LMK if we want that now or later as we go through the levels.

src/trx/game/objects/general/save_crystal.c
```diff
@@ -73,13 +85,55 @@ static void M_HandleSave(ITEM *const item, const SAVEGAME_STAGE stage)
...
+    int32_t c = ABS(angle >> 9);
+    CLAMPG(c, 31);
+    c <<= 3;
+
+    item->pos.y = p->initial_angle - ABS(angle >> 6) - 64;
+
+    Output_AddDynamicLightRGB(item->pos, 8, (RGB_888) { 0, c, 0 });
+
+    ITEM *const lara_item = Lara_GetItem();
+    const int16_t dx = ABS(item->pos.x - lara_item->pos.x);
```
lahm86: I'm able to get a free crystal in Jungle as below - was a bit puzzled :smile: I'm wondering if this is overflowing as it's `int16_t`?

<details>

https://github.com/user-attachments/assets/209788bd-6ecc-479f-8aee-6adac4b2c9c2
</details>

rr-: Oh yes, that was it. IDK why I chose `int16_t` for this! Updated the timer too just in case.

lahm86 (approved):

aredfan (approved):

PR #4653, opened 2026-01-17
lara: implement new currents

rr-:
#### Checklist

#### Description

@lahm86 this differs slightly in how we get the trigger's current value – please LMK if it's OK to stay this way, or if I should update it to match what we have for TR1/2. Thanks.

### DIFF
--- a/src/trx/game/lara/control.c
+++ b/src/trx/game/lara/control.c
@@ -70,29 +70,30 @@ static void M_Cheat(void)
     }
 }
 
-static void M_WaterCurrent(COLL_INFO *const coll)
+static void M_WaterCurrent_TR12(COLL_INFO *const coll)
 {
-    ITEM *const item = Lara_GetItem();
-    LARA_INFO *const lara_info = Lara_GetLaraInfo();
+    ITEM *const lara_item = Lara_GetItem();
+    LARA_INFO *const lara = Lara_GetLaraInfo();
 
-    int16_t room_num = item->room_num;
-    const ROOM *const room = Room_Get(item->room_num);
-    item->box_num = Room_GetWorldSector(room, item->pos.x, item->pos.z)->box;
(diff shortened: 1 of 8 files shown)


src/trx/game/lara/control.c
```diff
@@ -101,51 +102,165 @@ static void M_WaterCurrent(COLL_INFO *const coll)
...
-    coll->old = item->pos;
+    coll->old = lara_item->pos;
+}
+
+static void M_WaterCurrent_TR3(COLL_INFO *const coll)
+{
+    ITEM *const lara_item = Lara_GetItem();
+    LARA_INFO *const lara = Lara_GetLaraInfo();
+
+    int32_t angle;
```
lahm86: Maybe these can be declared directly inside the condition branches as they're not used outside.

src/trx/game/lara/control.c
```diff
@@ -101,51 +102,165 @@ static void M_WaterCurrent(COLL_INFO *const coll)
...
+    int32_t speed;
+    int32_t shifter;
+    int32_t abs_vel;
+
+    if (lara->current.active) {
```
lahm86: `!= 0` for clarity maybe - to avoid any confusion as a flag.

lahm86 (commented):
LGTM. I think the FD setup is fine for now given the differences - maybe we can find a way to combine things in the future, but I have a feeling the position of the sinks could make this tricky.

src/trx/game/lara/control.c
rr-: Thanks, good point – adjusted.

src/trx/game/lara/control.c
rr-: Adjusted, got burned by the naming.

lahm86 (approved):

aredfan (approved):

PR #4664, opened 2026-01-18
misc: implement enemy bleeding after death

rr-:
#### Checklist

#### Description

This implements the effects where the killed creatures continue to bleed for a few frames after they die.
I haven't added a changelog as this can be included in general blood effects that are already present.

#### Testing

We need to check if this conforms visually to the OG. There are gotchas.

#### Implementation

OG TR3 has a fork of the `DoBloodSplat` function, called `DoBloodSplatD`. This distinction actually applies to other functions, like the blood spills in water, blood baths etc. Both branches do pretty much the same (save for a small y offset difference in blood baths) – except one uses `Random_GetControl`, while the other `Random_GetDraw`. And… yes: the `D` effects actually get spawned… in the draw phase. But this approach is generally speaking a no-go for us, as it makes the logic FPS-dependent, which is something we don't want.

In this PR, I've implemented the `D` variants of the blood functions, but I call them in the control phase, as I iterate through the items so that I don't double-spawn them in 60 FPS.

What I'm not sure of is how close we got to replicating the OG effect with this kind of implementation.

### DIFF
--- a/src/trx/game/items/anim.c
+++ b/src/trx/game/items/anim.c
@@ -7,6 +7,7 @@
 #include <trx/game/objects.h>
 #include <trx/game/rooms.h>
 #include <trx/game/sound.h>
+#include <trx/version.h>
 
 #define M_SFX_SURF_DISTANCE ((STEP_L * 2) + 1)
 
@@ -204,6 +205,8 @@ void Item_Animate(ITEM *const item)
             }
 
             case AC_DEACTIVATE:
+                const OBJECT *const obj = Object_Get(item->object_id);
+                item->after_death = obj->intelligent ? 1 : 64;
                 item->status = IS_DEACTIVATED;
                 break;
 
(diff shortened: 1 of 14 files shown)


lahm86:
I think there are more blood spurts after death for the cobra in OG (sorry for the quality, OBS and OG TR3 are not happy today).

<details>

https://github.com/user-attachments/assets/81cea218-df22-4499-9ef8-3a89627bc642
</details>

rr-:
Thanks – I moved it to the draw phase, with a check to only apply the logic on actual "logical-adjacent" draw frames with `Output_SetControlFrame(bool)`.

I'm now concerned about the blood effect directions which seem to be all over the place, though was unable to find the culprit, as everything now seems to check out.

lahm86 (approved):
LGTM

aredfan (approved):

PR #4666, opened 2026-01-18
objects/switch: support one-shot flag

rr-:
#### Checklist

#### Description

This is one of those "I have no idea  what I'm doing" PRs – when auditing the switch/lever code, I noticed that OG TR3 adds support for one-shot flags.

I'm not sure where it's used, or if it has the opportunity to break TR1 / TR2. I'm happy to test for this in automated way, but I'm not sure how – I suppose I'd need to check for any `TO_OBJECT` trigger that targets a switch, but what other properties would we need to check?


rr-:
Let's park it until London, as this is more cursed than we thought.

PR #4680, opened 2026-01-19
Add TR3 bar colors and bar color preview

rr-:
#### Checklist

#### Description

This PR implements the TR3 bar colors as per https://github.com/LostArtefacts/TRX/discussions/3940 and adds a visual preview of the current style directly in the UI. To support that, I’ve rearranged the settings to group by option type rather than bar type.

20260119_180644_Coastal_Village

The change came from my longer‑term goal to stop hardcoding styles and let players or builders customize swatches through a JSON file. That however raises some localization concerns – if players define new colors, how would we know that color 7 in the TR1 style, called "orange," should be "pomarańczowy" in Polish? Then I figured we could skip translations altogether, and focus on visual identity instead. I tried it out and liked how that approach turned out, so I'm proposing it here.

We could add more swatches later (pink, orange, brown…), but that'll be a separate thing.


aredfan:
Looks great! 👍 

- [x] The `Bar scale` option also affects the bars displayed in the menu. It'll be worth tying these bars to the `Text scale` option instead.  The difference is noticeable if `Text scale` and `Bar scale` use different values.

rr-:
> Looks great! 👍
> 
> * [ ]  The `Bar scale` option also affects the bars displayed in the menu. It'll be worth tying these bars to the `Text scale` option instead.  The difference is noticeable if `Text scale` and `Bar scale` use different values.

Thanks – addressed :)

aredfan (approved):

lahm86 (approved):
Very nice!

PR #4689, opened 2026-01-19
objects/spikes: implement animations and sounds

rr-:
#### Checklist

#### Description

This adds the animation and sound effects to the spikes in Madubu Gorge, and in Coastal Village (trigger in room 41).
Note that the spikes appear to be floating midair – not sure if it's an OG bug?


aredfan (approved):
> Note that the spikes appear to be floating midair – not sure if it's an OG bug?

I can confirm the spikes do float in OG.

src/trx/game/objects/traps/spikes.c
```diff
@@ -53,9 +54,35 @@ static void M_Collision(
...
+    if (!Item_IsTriggerActive(item)) {
+        return;
+    }
+
+    const int32_t level_num = GF_BadGetLevelNum();
+    if (level_num != 5 && level_num != 7) {
+        return;
+    }
+
+    if (item->frame_num == Anim_GetAnim(item->anim_num)->frame_base) {
```
lahm86, line 69: We could use `Item_GetRelativeFrame(item) == 0` - I meant to mention this in the Tony PR too, but of course we can refactor in the future.

lahm86 (approved):

src/trx/game/objects/traps/spikes.c
rr-: I'll do that as a follow-up – thanks.

PR #4700, opened 2026-01-20
lara/skin: offer outfit swaps

### LINKED ISSUE #110: Feature Request: Ability to change Braid Offset
Enabling the Braid works fine, but it's position is hardcoded. Having a customizable height offset would be great.

### LINKED ISSUE #4383: Outfit / Lara Model change?
First of all thanks for your hardwork, this project is a godsend for people who still plays original Tomb Raiders.

I wanna ask if it is possible to change Lara's costume (as in Remastered version)? Can we use Lara's Tomb Raider 2 model in original Tomb Raider?

Sorry if it is a stupid question or answered before.

lahm86:
#### Checklist

#### Description

This is very much WIP - I wanted to open it up for some early testing/experimenting. The logic I will most likely tweak as I want to use similar mechanisms for extra animation mesh swaps (I'm thinking of an "equipment" mesh pointer array for Lara, on top of her regular array). For now, you can cycle through the outfits, which is bound to `T` by default (and `walk` will go backwards, like poses) - not sure if we'll need a cycle key, but this is a stopgap until there is a UI.

For guns to appear, you have to cycle the outfit at least once when the level loads.

Commented out code is temporary, mostly extra mesh swaps as I'm repurposing object IDs.

TODO:
- [x] Define ~~18~~ 20 available outfits as standard
- [x] Remove hands from equipped guns
- [x] Remove holsters from thighs
- [x] Realign TR3 gym/Nevada/Antarctica holsters
- [x] Rework gun version consistency (always TR1 pistols in TR1, always TR3 shotgun in TR3 etc)
- [x] Hide holsters in TR1/2 gym and HSH unless weapon cheats used
- [x] Alter logic to use equipment pointers
- [x] Handle level load
- [x] Handle braid
- [x] Apply braid torso/head changes for TR1 outfits
- [x] Handle combat face
- [x] Handle ~~resume info and~~ saves
- [x] Handle default outfit swaps between levels (player choice rules, unless `Default` is chosen)
- [x] Handle cutscenes ~~(maybe...Diving Area could be tricky)~~
  - [x] Allow swaps
  - [x] Handle reflections
  - [x] Handle gun swaps
  - [x] Handle DA outfit change
  - [x] Handle HSC cutscene mesh swap
- [x] Handle extra animations
  - [x] Mauled by T-rex
  - [x] Midas hand
  - [x] Pulling dagger
  - [x] Inspecting dagger
  - [x] Shower cutscene
  - [x] Kayak (prep)
  - [x] Minecart (prep)
- [x] Handle barefoot SFX
- [x] Handle fly cheat reset
- [x] Fix all mesh spheres
- [x] Finish fixing OG outfit bugs
- [x] UI
- [x] Translations and documentation

Feel free to add to above anything I may have missed.


rr-:
> Handle extra animations

This bullet probably already covers it, but I've found that Lara reverts to her default outfit when using the fly cheat.

---

A small patch that avoids the crash on the game exit :)
```diff
diff --git a/src/trx/game/lara/skin.c b/src/trx/game/lara/skin.c
index 9bbe310f7..84c384597 100644
--- a/src/trx/game/lara/skin.c
+++ b/src/trx/game/lara/skin.c
@@ -294,7 +294,7 @@ void Lara_Skin_Shutdown(void)
     if (m_Outfits != nullptr) {
         for (int32_t i = 0; i < m_Outfits->count; i++) {
             M_OUTFIT *const outfit = Vector_Get(m_Outfits, i);
-            Memory_FreePointer(outfit->name);
+            Memory_FreePointer(&outfit->name);
         }
 
         Vector_Free(m_Outfits);
```

---

I also noticed the outfit names defined in the JSON, which raises a question – how do we make the game recognize a third-party provided string as part of the localization system? I've hit this issue in two previous tickets and ended up solving it differently each time:
- For the UI bar theme definitions, I avoided using the color names directly (aside from config storage) and represented the selection visually instead. Applying that same approach to outfits would mean showing a small Lara model, like the 3D pickups – or at least a tiny textured quad preview. Both are just way beyond scope and a technical nightmare either way.
- For the globe's user-defined zones (Nevada, South Pacific Islands, etc.), I went with hardcoded strings like `GS_DEFINE(GLOBE_SELECT_AREA_1, …)`. It was simple and effective.

I think we could lean on that second method again – even if it's not the most elegant, couples outfit styles tightly to their indexes, and imposes a fixed upper limit on supported outfits. At least we'd still get to leverage the AI-powered translation tools.
LMK what you think about this.

lahm86:
Thanks, @rr-. I've gone with a similar approach for strings, which I think is perfectly fine for now. It in turn fixes the crash on exit - thanks for noticing this too.

src/trx/game/lara/skin/common.c
```diff
@@ -0,0 +1,175 @@
...
+            continue;
+        }
+
+        for (int32_t j = 0; j < hair_segment_count; j++) {
+            Object_SetMeshReflective(
+                O_LARA_SKIN_SWAP_EXTRA, braid->gold_offset + j, true);
+        }
+    }
+
+    // TODO: if config outfit doesn't match level outfit, override?
```
rr-: IMO config should be able to specify skin: default which then just takes the skin from the level, otherwise it always forces player choice.

rr- (commented):
With the new `lara_skin_idx` property, do builders need to learn the new skin workflow to swap Lara’s textures between levels, even if they’re not using player-specific skins?

Also I'm afraid that `idx` in the gameflow might be a bit too technical given it's builder-facing – WDYT?

lahm86:
> With the new `lara_skin_idx` property, do builders need to learn the new skin workflow to swap Lara’s textures between levels, even if they’re not using player-specific skins?
> 
> Also I'm afraid that `idx` in the gameflow might be a bit too technical given it's builder-facing – WDYT?

I am in two minds about forcing this to be the way for editing Lara. I'll see if falling back to OG approach is still doable when skins aren't present. I think forcing for the base game is fine as this task will also fix the wealth of texture bugs on Lara, I don't want to do that in two places.

Agreed re idx, thanks. I'll tweak and amend outfit json to show the numbers in comments as an extra help.

aredfan:
- [x] If Desert Eagle or Uzis are holstered, there's a mesh issue with their placement on Lara for the following outfits:

> - TR3 Gym
> - Nevada
> - Antarctica

- [ ] ~~Lara can have the incorrect grenade launcher mesh on her back depending on which game is open.~~

> - ~~Lara's TR1/TR2 model in TR3 has the TR3 grenade on her back.~~
> - ~~Lara's TR3 model in TR1/TR2 has the TR2 grenade on her back.~~

lahm86:
Thanks @aredfan. Deagle/Uzi should be fixed for now, although there is a minor holster alignment issue on some outfits I'll tackle later.

I was going to ask our thoughts on "gun consistency". At the moment the outfit dictates the gun flavour - so you'll see TR2 pistols in TR1, TR3 shotgun in TR2 etc. Do we think it should be consistent with what's in Lara's inventory instead? I'm leaning towards that, especially as things like gun flash offsets are dictated by the game version (well the version's JSON) and getting all of that to line up and be interchangeable would be too complex I think.

So TLDR - regardless of outfit, whatever gun Lara has on her back, her holsters or in her hands should match the style in the inventory.

aredfan:
> Thanks @aredfan. Deagle/Uzi should be fixed for now, although there is a minor holster alignment issue on some outfits I'll tackle later.
> 
> I was going to ask our thoughts on "gun consistency". At the moment the outfit dictates the gun flavour - so you'll see TR2 pistols in TR1, TR3 shotgun in TR2 etc. Do we think it should be consistent with what's in Lara's inventory instead? I'm leaning towards that, especially as things like gun flash offsets are dictated by the game version (well the version's JSON) and getting all of that to line up and be interchangeable would be too complex I think.
> 
> So TLDR - regardless of outfit, whatever gun Lara has on her back, her holsters or in her hands should match the style in the inventory.

Thank you for fixing the issue. 👍 

I think your proposal for the inventory to determine which guns appears on Lara's model makes the most sense.

rr-:
> Thanks @aredfan. Deagle/Uzi should be fixed for now, although there is a minor holster alignment issue on some outfits I'll tackle later.
> 
> I was going to ask our thoughts on "gun consistency". At the moment the outfit dictates the gun flavour - so you'll see TR2 pistols in TR1, TR3 shotgun in TR2 etc. Do we think it should be consistent with what's in Lara's inventory instead? I'm leaning towards that, especially as things like gun flash offsets are dictated by the game version (well the version's JSON) and getting all of that to line up and be interchangeable would be too complex I think.
> 
> So TLDR - regardless of outfit, whatever gun Lara has on her back, her holsters or in her hands should match the style in the inventory.

I'm thinking the same. Guns are guns, outfits are outfits. Pairing guns appearance to outfits isn't most logical. Unless I'm misunderstanding

src/trx/game/lara/skin.h
```diff
@@ -0,0 +1,31 @@
...
+void Lara_Skin_SwapSingleExtra(LARA_MESH mesh, LARA_EXTRA_STATE state);
+const ANIM_BONE *Lara_Skin_GetBoneBase(void);
+
+bool Lara_Skin_IsBraidSupported(void);
+XYZ_32 Lara_Skin_GetBraidOffset(void);
+int32_t Lara_Skin_GetBraidMeshIdx(void);
+
+void Lara_Skin_SetEquipment(LARA_MESH mesh, LARA_SKIN_EXTRA_MESH extra_mesh);
+void Lara_Skin_ClearEquipment(LARA_MESH mesh);
+const OBJECT_MESH *Lara_Skin_GetEquipment(LARA_MESH mesh);
```
rr-: Let's move these headers into `lara/skin/` and do an `#include` here, like with other compound modules.

aredfan:
These issues are probably already known, I figured I should list them just in case.

- [x] The holstered guns doesn't remain visible when a rifle-type weapon is drawn
- [x] The holstered guns doesn't remain visible when a flare is held
- [x] TR1: Lara starts the gym with holsters
- [x] TR2: Lara starts HSH with holsters
- [x] TR2: If the fly cheat is used during the HSH intro cutscene, Lara ends up with a dagger in her hand
- [x] TR2: Lara has the dagger attached to her belt during the shower cutscene
- [x] Lara doesn't take out guns during cutscenes (e.g. 1st cutscene in TR1/TR2 and 2nd cutscene in TR3)

lahm86:
> These issues are probably already known, I figured I should list them just in case.
> 
> * [ ]  The holstered guns doesn't remain visible when a rifle-type weapon is drawn
> * [ ]  The holstered guns doesn't remain visible when a flare is held
> * [ ]  TR1: Lara starts the gym with holsters
> * [ ]  TR2: Lara starts HSH with holsters
> * [ ]  TR2: If the fly cheat is used during the HSH intro cutscene, Lara ends up with a dagger in her hand
> * [ ]  TR2: Lara has the dagger attached to her belt during the shower cutscene
> * [ ]  Lara doesn't take out guns during cutscenes (e.g. 1st cutscene in TR1/TR2 and 2nd cutscene in TR3)

Thanks. 1, 2, 6, and 7 should be resolved now. Will tackle the rest later, along with the holster/gun flavour mesh madness.

data/tr2/ship/data/scripts/cut3.lua
```diff
@@ -0,0 +1,8 @@
+local lara_diving_suit = 10 --skin_type_10
+
+trx.events.after_control(function()
+  local lara_item = trx.lara.item
+  if lara_item.anim == 7 and lara_item.frame < 5 and trx.lara.skin ~= lara_diving_suit then
```
rr-: I may be mistaken, but I think we now can use `catalog` for the anim? (Need to confirm the game-to-trx id translations)

data/tr2/ship/data/scripts/cut3.lua
```diff
@@ -0,0 +1,8 @@
+local lara_diving_suit = 10 --skin_type_10
```
rr-: Let's put a space after `--`

data/common/ship/cfg/outfits.json5
```diff
@@ -0,0 +1,502 @@
+{
+    "outfits": {
+        // TR1 Gym
+        "skin_type_01" : {
```
rr-, line 4: I'd suggest either to allow the skins to be renamed, and look them up by a dynamic name, or to just go with integers. I'm leaning on letting the players rename them – keep the dictionary keys, and for each key, gets its index to mean skin index, and set the skin's name to the key content.
The benefit is that it gives more meaningful names here in the JSON, and in the game flows, which would make things more readable IMO.

EDIT: I think I understand why these are enums – this way it's much easier to localize it, right?
But we could keep the GS constants just for localization.

LMK if you would like me to propose a patch for this.

src/trx/game/lara/skin/common.c
```diff
@@ -0,0 +1,368 @@
...
+        extra_id = mesh_idx == LM_TORSO ? EXTRA_MESH_TR1_BRAID_DEFAULT_TORSO
+                                        : EXTRA_MESH_TR1_BRAID_DEFAULT_HEAD;
+        break;
+    case BRAID_MODE_TR1_MAULED:
+        extra_id = mesh_idx == LM_TORSO ? EXTRA_MESH_TR1_BRAID_MAULED_TORSO
+                                        : EXTRA_MESH_TR1_BRAID_DEFAULT_HEAD;
+        break;
+    case BRAID_MODE_TR1_GOLD:
+        extra_id = mesh_idx == LM_TORSO ? EXTRA_MESH_TR1_BRAID_GOLD_TORSO
+                                        : EXTRA_MESH_TR1_BRAID_GOLD_HEAD;
```
rr-, line 50: WDYT of removing a braid map, which currently hardcodes some outfits to behave special, and instead storing all of this directly in the outfits, allowing the data to be duplicated a bit?

This would be ideal with YAML anchors, because this way, we could benefit from having to declare the offsets only once, but they'd get copied to each outfit. Sadly, not even JSON6 supports these.

Furthermore, maybe we could (somehow) get the extra stuff – Xian dagger etc – to become completely data-driven too…? (Highly doubt it but just putting this idea out there)

src/trx/game/lara/hair.c
```diff
@@ -96,7 +92,8 @@ static void M_CalculateSpheres(const ANIM_FRAME *const frame)
     m_HairSpheres[2].r = mesh->radius;
     Matrix_Pop();
 
-    Matrix_TranslateRel(M_HAIR_OFFSET_X, M_HAIR_OFFSET_Y, M_HAIR_OFFSET_Z);
+    const XYZ_32 offset = Lara_Skin_GetBraidOffset();
+    Matrix_TranslateRel32(offset);
```
rr-: Nitpick: I think we can just inline the variable

src/trx/game/lara/hair.c
```diff
@@ -183,7 +180,8 @@ static void M_CalculateSpheres_I(
     m_HairSpheres[2].r = mesh->radius;
     Matrix_Pop_I();
 
-    Matrix_TranslateRel_I(M_HAIR_OFFSET_X, M_HAIR_OFFSET_Y, M_HAIR_OFFSET_Z);
+    const XYZ_32 offset = Lara_Skin_GetBraidOffset();
```
rr-: Nitpick: I think we can just inline the variable

src/trx/game/shell/common.c
```diff
@@ -152,5 +153,9 @@ void Shell_HandleConfigChange(const CONFIG *const old, const CONFIG *const new)
         Savegame_Init();
         Savegame_ScanSavedGames();
     }
+
+    if (L_CHANGED(visuals.lara_skin_type)) {
+        Lara_Skin_ApplyOutfitFromConfig();
+    }
```
rr-: Ouch, found a leftover from the engine merging era – `HandleConfigChange` should be merged to `HandleCommonConfigChange`, and then we should drop `Common` :D

data/tr2/ship/data/scripts/cut3.lua
lahm86: I think we could yeah. It's a cutscene though so Lara's anim names don't really match up (`LA_RUN_TO_WALK_RIGHT` here).

data/common/ship/cfg/outfits.json5
lahm86: Yes, the named approach is to handle translation keys, but also my thinking was if a builder only wants to offer say 2 outfits from the 20 we ship, they can delete the ones they don't need from the JSON and still refer to the same strings. Otherwise they'd need to re-evaluate the indices, and it felt a bit clumsy. As outfits also point to other outfits (e.g. TR1 Classic knows it can turn to TR1 Golden Lara) that's done by name too instead of by index.

A nicer naming system would be welcome indeed - I agree, it'd be nice if they were more descriptive at this level.

src/trx/game/shell/common.c
lahm86: Oops, and I've even updated both of these functions in this PR without realising 😄

src/trx/game/lara/skin/common.c
lahm86: I'm not sure about this. So we'd have a copy of TR1 Lara with and without her braid? It's a lot of duplication - I'm not sure how it'd work with the braid option itself and TR2/3. 
I don't mind the current approach too much, although yes, it'd be ideal if the engine didn't have to worry about these things and just apply the meshes it's fed. It's sort of a hybrid situation at the moment, which maybe is OK for the first pass.

rr-: My suggestion was a bit multi-leveled and not well explained, sorry.

As a first step I'd suggest removing the additional indirection layer between outfits and braids, by pulling the braid definition directly into the outfit that currently references it with a special field:
```
        "skin_type_02": {
            "mesh_object": "O_LARA_SKIN_SWAP_1",
            "mesh_offset": 16,
            "gun_map": 0,
            "braid": {
                "mesh_offset": 10,
                "gold_offset": 16,
                "hair_pos": {
                    "x": 0,
                    "y": 20,
                    "z": -45,
                },
            },
            "braid_mode": "BRAID_MODE_TR1_FULL",
            "combat_face_offset": 1,
            "extra_outfits": {
                "LS_EXTRA_TREX_KILL": "skin_type_03",
                "LS_EXTRA_MIDAS_KILL": "skin_type_05",
            },
        },
```
My understanding is that this doesn't change anything on the logic level, apart from reducing some of the code complexity, at the cost of making the data more duplicated. The benefit of this is to let the outfits be eventually more self-contained with fewer cross-references.

As the second suggestion, pull torso and braid meshes directly into outfit:
```
        "skin_type_02": {
            "mesh_object": "O_LARA_SKIN_SWAP_1",
            "mesh_offset": 16,
            "gun_map": 0,
            "braid": {
                "mesh_offset": 10,
                "gold_offset": 16,
                "hair_pos": {
                    "x": 0,
                    "y": 20,
                    "z": -45,
                },
                "head_mesh": 7,
                "torso_mesh": 4,
            },
            "combat_face_offset": 1,
            "extra_outfits": {
                "LS_EXTRA_TREX_KILL": "skin_type_03",
                "LS_EXTRA_MIDAS_KILL": "skin_type_05",
            },
        },
```
My understanding is that disabling the braid would continue to not render `braid.mesh_offset`, and would not replace `braid.head_mesh` or `braid.torso_mesh`. 
The challenge with this is handling combat faces effectively but this can get pulled into another `braid.combat_face_offset` and `combat_face_offset`. I think. And if braid.combat_face_offset is -1 (not set) then we can use `outfit.combat_face_offset`?

Another proposal is to simplify Golden Lara handling, by just swapping the meshes one by one to `skin_type_05`, which would allow us to remove the need for `braid.gold_offset` hopefully?
OR, redo golden Lara handling altogether, by doing this purely in the render stage with a flag to disable texturing and override the color to yellow-ish, which would work then on any outfit. This is more difficult and I'm not sure I like this.

LMK what you think of this.
My thinking is that with these edits, we'd have self-contained outfits, and self-contained equipment definitions, without need to think much of the braid or combat face mappings (at least not on the JSON level).

data/common/ship/cfg/outfits.json5
rr-: Hmm maybe let's do it like this: land the current version as-is into develop, and then I play with it for a bit, and make a PR with a proposed change, assuming I manage to get it right? Happy to also try this out with my braid handling suggestion, because these are pretty heavy changes that feel too heavy for this initial version.

src/trx/game/lara/skin/common.c
rr-: Edit: suggested to postpone this for a potential outfits v2 in https://github.com/LostArtefacts/TRX/pull/4700#discussion_r2746599035

lahm86: Oh I see - sorry, I was thinking full outfit duplication at the data level.
This makes a lot of sense, thanks. My initial thinking with the separation was to try to be efficient with JSON size, but yeah, it's extra overhead for the lookups and it's not easy to follow in the JSON. Self-contained is a winner.
I'll look at incorporating this, thanks 😌

data/common/ship/cfg/outfits.json5
lahm86: Sure, happy to delay it for now.
I'll focus on finishing the holsters/guns/saves over the weekend and then we should be in a good position to tidy up.

rr- (commented):
Just so that we don't forget – we'll need to document the new LUA stuff.

lahm86:
Saves should now work, with the added benefit of no longer needing to store absolute mesh offsets for Lara anywhere. Skins are applied automatically, and any extra equipment is saved in terms of its type rather than a specific mesh. For pre 1.2 saves, holsters and guns should be restored (by comparing against the old mesh offsets) and in HSH, the dagger should appear on Lara's hips. Note that old saves during the start of HSH (via the console) won't attempt to load the dagger into her hand, but I think this is a reasonably small enough edge case - we could do it though, LMK.

Some NBs as we discussed:
- If the config option is set to default and has subsequently been changed since making a save, the player's new choice will apply to the old save
- If the skin is changed by code (e.g. `/lua trx.lara.skin = 13`) and you then save and change the skin manually, loading the save will re-apply the saved skin

rr-:
> I think this is a reasonably small enough edge case

It is :)

> we could do it though, LMK.

Nah. So long as re-saving now works, it's fine.

Thank you.

data/tr2/ship/data/scripts/assault.lua
```diff
@@ -1,4 +1,5 @@
 trx.events.on_game_start(function(level, is_save)
+  trx.lara.holsters_visible = trx.lara.has_pistol_weapon -- for pre-skin saves
```
rr-: Let's take a note when to remove this with the wording "remove after 1.x" (I vote supporting up to -0.3 versions inclusively)

rr- (approved):

aredfan (approved):

lahm86:
Docs and translations are now updated, and the commits have been tidied-up. I'd appreciate a quick sanity check, thanks!

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/203

docs/03-MIGRATING.md
```diff
@@ -30,6 +30,10 @@ title: Migrating levels
 4. **Legacy savegame pattern removed**:
     Remove the `savegame_fmt_legacy` key from game flow files.
 
+5. **Lara skin system**:
```
rr-: Let's make it number one.

rr- (approved):

aredfan:
- [x] During sanity checks, I found a small corner case bug affecting TR1 and TR2, where Lara can start a regular level without holsters.

> 1. `/launch TR1`
> 2. `/play 0`
> 3. `/play 13` and observe Lara is without holsters

Alternative steps:

> 1. `/play 13` then `/save 1`
> 2. `/play 0`
> 3. Exit to Title
> 4. open the passport to New Game and select Play previous levels
> 5. select Natla's Mines and observe Lara is without holsters

lahm86:
Excellent catch @aredfan, thanks. Should be fixed.

aredfan (approved):
Thank you for fixing the issue @lahm86. 👍 

Amazing PR, well done, and thank you. 🎉

PR #4708, opened 2026-01-21
rooms: remove max drawn rooms limit

rr-:
#### Checklist

#### Description

There's still a hidden limit of binding 128 rooms at maximum, but this doesn't seem to be a problem: when I cap this list to just 50 entries, it still renders the big area just fine with no clipping.

### DIFF
--- a/src/trx/game/rooms/draw.c
+++ b/src/trx/game/rooms/draw.c
@@ -11,6 +11,7 @@
 #include <trx/game/water_fx.h>
 #include <trx/game/weather_fx.h>
 #include <trx/utils.h>
+#include <trx/vector.h>
 #include <trx/version.h>
 
 #include <string.h>
@@ -78,8 +79,7 @@ static inline void M_DrawSet_ForEach(
     }
 }
 
-static int32_t m_DrawCount = 0;
-static int16_t m_RoomsToDraw[MAX_ROOMS_TO_DRAW] = {};
+static VECTOR *m_RoomsToDraw = nullptr;
 
 static int32_t m_Outside;
 static int32_t m_OutsideRight;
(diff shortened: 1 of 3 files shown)


src/trx/game/rooms/draw.c
```diff
@@ -91,6 +91,14 @@ static int32_t m_BoundStart;
 static int32_t m_BoundEnd;
 static int32_t m_BoundRooms[M_MAX_BOUND_ROOMS] = {};
 
+static void M_EnsureRoomsToDraw(void)
+{
+    if (m_RoomsToDraw != nullptr) {
+        return;
+    }
+    m_RoomsToDraw = Vector_CreateAtCapacity(sizeof(int16_t), 100);
```
lahm86, line 99: Should we free this on shutdown? Just thinking valgrind may flag it if we're debugging something in the future.

src/trx/game/rooms/draw.c
```diff
@@ -414,32 +422,30 @@ static void M_DrawSingleRoom(const ROOM *const room)
...
 
 void Room_MarkToBeDrawn(const int16_t room_num)
 {
-    if (m_DrawCount + 1 == MAX_ROOMS_TO_DRAW) {
-        return;
-    }
-
-    for (int32_t i = 0; i < m_DrawCount; i++) {
-        if (m_RoomsToDraw[i] == room_num) {
+    for (int32_t i = 0; i < m_RoomsToDraw->count; i++) {
```
lahm86: Could we make use of `Vector_Contains`?

src/trx/game/rooms/draw.c
rr-: Oh definitely. Added a static destructor – should be fine.

src/trx/game/rooms/draw.c
rr-: Thanks, I forgot this function exists. Adjusted.

lahm86 (approved):
Thanks!

aredfan (approved):

PR #4709, opened 2026-01-22
objects/kill_all_triggered: implement kill all control

lahm86:
#### Checklist

#### Description

This implements the control for `O_KILL_ALL_TRIGGERS`, which kills all active items and effects, bar some exceptions such as Lara, doors and pickups. Items that have also been activated by standard in the level (like some save crystals) are also excluded.

I don't really like the name as it implies we're killing FD triggers. `O_KILL_ALL_TRIGGERED` would make more sense IMO - WDYT?

Temple Ruins is a good test - have Lara activate everything at the start, then make it to room 77 and cross the kill all trigger there. Then with the fly cheat, head back to the start to see everything has gone, bar the exclusions of course.


rr- (approved):
I agree on the naming.

aredfan (approved):

lahm86:
Thanks. Updated the naming, and defined the GS names too, which I'd forgotten to do.

rr- (approved):

PR #4710, opened 2026-01-22
Implement TR3 monkeys

rr-:
#### Checklist

- [x] I have read the [coding conventions](https://github.com/LostArtefacts/TRX/blob/master/CONTRIBUTING.md#coding-conventions)
- [x] I have added a changelog entry about what my pull request accomplishes, or it is an internal change
- [x] I have added a readme entry about my new feature or OG bug fix, or it is a different change

#### Description

This is WIP as in all likelihood, it requires new pathfinding  for them to fully work.

PR #4713, opened 2026-01-23
objects/bird: register Vulture

lahm86:
#### Checklist

#### Description

This registers the vulture object, which behaves in the same way as the TR2 eagle, bar having different HP.

The difference with TR3 flyers in general is that they will fly over water/swamp rooms, while in TR1/2 they remain over land only (bar occasionally in Great Wall where the crows can drown, but that's an OG bug). The change for this in TR3 was part of #4710; the end of the River Ganges and the start of Nevada are good places to test. They'll dip their ~~toes~~ claws if Lara is on the surface, but will never go fully underwater.


rr- (approved):
Thanks. Quick question – the anims are in line with those of the eagles right?

lahm86:
> Thanks. Quick question – the anims are in line with those of the eagles right?

They are 😌 Pretty much just a re-skin.

aredfan (approved):

PR #4715, opened 2026-01-23
objects/rolling_ball: integrate TR3 boulders

lahm86:
#### Checklist

#### Description

This updates boulder control to take ceiling height into account when deciding whether or not to come to a stop, and registers the giant boulder from Temple of Puna to be used. I kept the name in line with the other types, but LMK if we'd prefer the distinction with this one. OG TR3 levels only use the regular boulder (the barrels in Tinnos are just these as well) and the giant one in Puna, although it still has the TR2 drums and avalanche registered.


aredfan (approved):

rr- (approved):
I'm ok with the naming. Thank you!

PR #4721, opened 2026-01-24
monkey: do not hardcode hostility

rr-:
#### Checklist

#### Description

This changes the Monkey behavior to no longer hardcode hostility based on presence of Tigers in the level.
As a side effect, when you shoot one monkey, you'll now need to shoot them all.

#### Testing

1) Jungle monkeys do not attack Lara, unless she attacks them
2) Attacking a single monkey in Jungle gets others to turn hostile
3) Hostile monkeys prioritize playing with medi packs (seems to be OG behavior…)
4) Jungle monkeys are not included in the level max kill count (you may need to remove `cfg/tr3/max_stats.cache.json`)
5) Temple Ruins monkeys do attack Lara
6) The River Ganges monkeys do attack Lara
7) In both levels they _are_ included in the level kill count

### DIFF
--- a/data/tr3/ship/cfg/tr3-la/gameflow.json5
+++ b/data/tr3/ship/cfg/tr3-la/gameflow.json5
@@ -124,6 +124,7 @@
         // 5. It's a Madhouse!
         {
             "path": "data/zoo.tr2",
+            "script": "data/scripts/zoo.lua",
             "music_track": 34,
             "sequence": [
                 {"type": "loading_screen", "path": "data/images/zoo.webp", "fade_in_time": 1.0, "fade_out_time": 1.0},
--- a/data/tr3/ship/cfg/tr3/gameflow.json5
+++ b/data/tr3/ship/cfg/tr3/gameflow.json5
@@ -63,6 +63,7 @@
         // 1. Jungle
         {
             "path": "data/jungle.tr2",
+            "script": "data/scripts/jungle.lua",
             "music_track": 34,
             "weather_type": "rain",
             "sequence": [
--- a/data/tr3/ship/data/scripts/jungle.lua
+++ b/data/tr3/ship/data/scripts/jungle.lua
@@ -1,0 +1,3 @@
+trx.events.before_level_file(function(level)
+  trx.creatures.add_ally(trx.catalog.objects.monkey)
+end)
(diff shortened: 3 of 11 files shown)


lahm86 (approved):

rr-:
Rebased to include the recent catalog property rename.

aredfan:
> 7\. In both levels they _are_ included in the level kill count

- [x] In Temple Ruins, the total kill count is 8 which only counts the Cobras. Likewise in The River Ganges, the total kill count is 11 which only counts Cobras and Vultures.
- [x] There's a couple of corner cases where attacking one monkey doesn't turn the other monkeys hostile.

> 1. `/play 1`
> 2. `/tp key` and shoot the monkey before it gets to the key
> 3. `/tp 29 26 63`
> 4. the other monkey doesn't attack Lara (after collecting and dropping off the small medi pack)

> 1. `/play 1`
> 2. `/tp 29 26 63`
> 3. shoot the monkey before it gets to the medi pack, or when it has the pickup in its mouth
> 4. `/tp key`
> 5. the other monkey doesn't attack Lara (after collecting the key)

rr-:
The first issue requires this fix: https://github.com/LostArtefacts/TRX/pull/4726

rr-:
Rebased on top of develop with #4726 to fix the first problem – the other two problems will be tackled this evening.

aredfan:
- [ ] In It's a Madhouse, the monkeys are also allies like they're in Jungle. In the PR build, they'll attack Lara.

rr-:
Re: hostility corner cases, it really was one. The friendly-fire threshold that triggers hostility is 10, but monkeys only have 8 HP. So if you one-shot a monkey while it's busy (such as picking up an item or actively in combat), the damage never exceeds the threshold, meaning the hostility trigger doesn't fire – even though you end up killing the poor thing.
Weapons like rocket launchers easily exceed the 10 HP threshold, which is why the Barkhang monks will still turn hostile if you blow them up.

To fix this, I updated the rules so that all allies turn hostile if you:
- One-shot or kill any individual ally (new rule)
- Attack any idle ally
- Injure any ally engaged in combat (or otherwise busy) by 10 HP or more

I also moved the friendly-fire accumulator into its own field instead of using creature flags. Flags felt unreliable, since many objects repurpose them for other data, which could easily cause conflicts. This new field is only recognized by BSON savegames, so loading older save files technically introduces a minor regression – you can injure monks in those cases up to 18 HP. However, downright killing them will still trigger hostility as per the new rules, and once the game is re-saved in BSON format, the issue disappears. Just a small quirk worth noting.

Lastly, I adjusted the monkey behavior so that hostile monkeys no longer get distracted by pickups. It felt awkward seeing them prioritize retrieving and carrying a medpack before engaging in combat, instead of attacking the player immediately.

aredfan (approved):

PR #4729, opened 2026-01-25
Reflect player's stance on killing allies in the stats dialog

### LINKED ISSUE #4731: TR3 Monkey Behaviour Change
A recently merged PR (https://github.com/LostArtefacts/TRX/pull/4721) introduced code to make the TR3 monkey passivity subject to the Lua creature ally API, rather than the presence of the TIGER object being loaded in the level. I think this is perfectly acceptable feature since the ally behaviour can easily be replicated via the included Lua scripts, and this subsequently opens up lots of possibilities for custom level builders. However, the PR explicitly states that how the monkey's ally status determined during play is now intentionally different, mimicking the TR2 Barkhang monks who all become aggressive if one monk is aggro'd, amongst other things. This is explicitly different from the TR3 monkeys who do not immediately all become aggressive if one monkey is aggro'd which leads to very distinctly different behaviour which I feel goes too far in changing the gameplay for a source port which has otherwise been good at preserving gameplay quirks of the original games, even including explicit glitches used by speedrunners.

This change means that in the first level, if a player is too trigger-happy and accidentally attacks the first monkey, the second monkey never interacts with the medipack, and the player never gets another opportunity to see the monkey's more interesting passive behaviour. It also arguably doesn't make much sense from a realism sense for all the monkeys in the level to immediately aggro. It feels believable in the context of the monks who are densely packed enough that you can imagine them signalling the player's hostility to their friends, but it doesn't really make much sense for why killing a monkey suddenly makes all the other monkeys in completely unrelated areas of the level immediately aggressive.

While I would not recommend reverting to the original hardcoded logic, I would recommend extending the APIs with additional flags which would allow the quirks of the original monkey behaviour to be preserved, since I feel this subtly tilts the project away from an enhanced refactored source port with optional bug fixes and fancy new features to something more resembling a personal interpretation of the game.

rr-:
#### Checklist

#### Description

This PR updates the stats dialog to count ally kills once the player turns against them. If allies are attacked in any level, they'll be included in the kill count for that level and all that follow. At the end of the game, if the player ever fought their allies, the final stats will include all ally kills across every level.

This is for 100% runs. It was awkward to see kills: 60/40, and made max stats flawed for players that take the path of evil.

#### Testing

- Load Barkhang, confirm the max kill count hasn't changed
- Load Barkhang, finish the game, confirm the max kill count in the final stats hasn't changed
- Attack monks in Barkhang, confirm the max kill count has increased
- Attack monks in Barkhang, save, load, check the max kill count stays increased
- Attack monks in Barkhang, finish the game, confirm that the max kill count in the final stats has increased
- Similar tests with Monkeys – useful to add a Lua script in Temple Ruins or The River Ganges to make them friendly by default, similar to Jungle, then we can check the multi-level stats!
- Similar tests with the Play Previous Levels feature – loading a prior level should remember when the player has hurt any ally and reflect that in the loaded level and in the final stats, and likewise – loading states prior to hurting anyone should not include allies in that level or the final stats


lahm86:
> Load Barkhang, confirm the max kill count hasn't changed

The kill count is showing 52 in Barkhang on level load, on develop it's 30.

rr-:
> > Load Barkhang, confirm the max kill count hasn't changed
> 
> The kill count is showing 52 in Barkhang on level load, on develop it's 30.

Thanks fortunately it was a silly mistake in the stats dialog code, not the main logic.

aredfan:
- [x] Barking Monastery is showing 31 kills.  It looks like the "unobtainable_kills": has no effect.

rr-:
> * [ ]  Barking Monastery is showing 31 kills.  It looks like the "unobtainable_kills": has no effect.

Thanks, fixed – another silly mistake.

Also, fixed "Play Previous Levels" hostility status propagation with non-linear order of levels.

Also introduced a (hidden) setting called "Ally hostility policy" to close #4731 which for TR3 defaults to "individual", meaning monkeys will continue to not share their hostility state across the level, and will continue to absurdly prioritize pickups even as you actively are shooting at them.
In this policy, killing a monkey, or exceeding a threshold, or attacking an idle monkey (so basically all TR2 hostility conditions) continue to change the stats dialog to include allies in the kill count, even if this doesn't cause others to attack you.
This change will require any saves created in the previous versions of this PR to be re-taken.

src/trx/game/objects/creatures/monkey.c
```diff
@@ -195,7 +195,7 @@ static void M_Control(const int16_t item_num)
     } else {
         Creature_GetAITarget(creature);
 
-        if (creature->hurt_by_lara) {
+        if (creature->hurt_by_lara && creature->enemy == lara_item) {
```
lahm86: Isn't this redundant?

docs/CHANGELOG.md
lahm86: We can replace this with the config option notes maybe.

lahm86 (commented):
Stats and monkey behaviour are looking good to me.
My only preference would be making the config option default to the shared mode for TR3 as well, the OG behaviour is non-sensical IMO. But of course happy to go with whatever we feel is best.

src/trx/game/objects/creatures/monkey.c
rr-: This is necessary for monkey to prioritize medpacks when attacked in the individual hostility policy.

lahm86: Got it. I just read it as "if ... and enemy is Lara, set enemy to Lara"

docs/CHANGELOG.md
rr-: Changed + exposed this new setting via UI

rr-:
> Stats and monkey behaviour are looking good to me. My only preference would be making the config option default to the shared mode for TR3 as well, the OG behaviour is non-sensical IMO. But of course happy to go with whatever we feel is best.

Agreed. Changed to default to shared, though it's important to test both.

I've included three more fixes:

- Targeting allies was not possible if "Enable ally targeting" was off, and you previously turned them against you
- `/kill` will now target actively hostile allies in the level (eg it'll only kill truly all under the shared policy)
- `/kill` will now no longer report already dead enemies  as killed (needs testing with Winston)

src/trx/game/objects/creatures/monkey.c
rr-: You're right, it's complete nonsense. I'm getting lost in these `if`s. 
I'll sort it out.

rr-: Ended up creating a separate setting for it, inspired by `fix bear ai` which is yet another setting which does not make imo sense to disable.

rr-:
To recap the full list of changes of this PR encompasses:

- Adds a configurable “Ally hostility policy” (Individual vs Shared), exposed in gameplay settings. Individual means allies don't mind watching their peers get slaughtered (TR3), Shared means TR2 monks – all or none. Defaults to Shared.
- Adds a configurable "Fix monkey pickup priority" that makes attacked monkeys retaliate before picking up items. Defaults to on.
- Updates the max kill count in the stats dialog to reflect whether the player has hurt allies. This uses the "Shared" policy logic, so we bump the max kill count to include all allies, even if we have the "Individual" policy set and they won't go after Lara.
- Fixes one-shotting allies with max HP < 10 not upgrading ally hostility status.
- Fixes "Enable AI targeting" to let you target ex-allies that are now hostile.
- Improves `/kill` to avoid counting already-dead enemies.
- Improves `/kill` to handle hostile allies correctly. In the "Individual" policy, it'll continue to only target the creatures that are already aggro'd.

Implementation details:

- Splits max kills into ally vs non‑ally when calculating the level stats.
- Introduces undocumented `unobtainable_ally_kills` that gives us control over both max kill counts.
- Persists if the player has harmed allies in save/resume data so it carries across levels, and gets the stats dialog max kill counts to work well with the “Play Previous Levels” feature.

#### Testing

- As in the OP
- Visibility of the new settings – hostility policy should be TR2+TR3, monkey pickups should be TR3 only
- Monks with both policies
- TR3 enemies with both policies
- Attacking the monkeys on their way to pick up a medpack/key with both policies and with the pickup priority fix on/off (4 scenarios)
- `/kill winston` in TR2/TR3
- `/teatime` before and after killing Winston
- Targeting Winston + "Enable ally targeting" on/off
- Targeting allies + "Enable ally targeting" on/off + both hostility policies
- Targeting ex-allies + "Enable ally targeting" on/off + both hostility policies
- `/kill` + both policies + before/after hurting an ally
- "Play Previous Levels" max kills with non-linear gameplay ideally, though this requires tinkering with LUA

lahm86:
- [x] `/kill winston` seems to always yield `object not found`
- ~~[ ] In TR2, set policy to individual and start Barkhang and trigger the monks/goons at the start. Issue `/kill goon` - all monks are aggressive towards Lara~~ Ignore, not sure what I did
- [x] With individual policy set, if you shoot a monk, he doesn't attack Lara

rr-:
Fixed both issues

lahm86 (approved):

aredfan (approved):

PR #4736, opened 2026-01-26
Add Shiva + animating Puzzle Holes, and fix TR3 projectiles

rr-:
#### Checklist

#### Description

This PR adds Shiva control along with the blade shielding and statue metamorphosis, topped with animating puzzle holes support (check out Serpent Stone receptacles in Coastal Village!).


aredfan:
**Issues (testing in progress):**

- [x] The shiva statue's head doesn't turn towards Lara.

rr-:
> **Issues (testing in progress):**
> 
> * [ ]  The shiva statue's head doesn't turn towards Lara.

I had a silly mistake in the bone setup – can you please LMK if it's better now?
(I'm struggling to see differences myself and am afraid I have a mistake somewhere deeper.)

EDIT: it's not working for sure :(

src/trx/game/objects/creatures/shiva.c
```diff
@@ -0,0 +1,544 @@
...
+                creature->flags = 4;
+                item->goal_anim_state = M_STATE_WAIT;
+            }
+            break;
+        }
+    }
+
+    if (lara_alive && lara_item->hit_points <= 0) {
+        Creature_SpecialKill(
+            item, M_ANIM_KILL_ANIM, M_STATE_KILL, LS_EXTRA_YETI_KILL);
```
lahm86: We'll need to use the new TRX state for Lara's death:  `LS_EXTRA_SHIVA_KILL` (#4649)

src/trx/game/objects/creatures/shiva.c
```diff
@@ -0,0 +1,544 @@
...
+    return true;
+}
+
+static void M_Setup(OBJECT *const obj)
+{
+    if (!obj->loaded) {
+        return;
+    }
+
+    if (!Object_Get(O_MESH_SWAP_2)->loaded) {
```
lahm86: `O_MESH_SWAP_1` I think

src/trx/game/objects/creatures/shiva.c
rr-: Fixed

src/trx/game/objects/creatures/shiva.c
rr-: Oops – fixed

rr-:
I fixed the joint rotations, and the fix was a bit wild.

20260127_093748_Temple_Ruins

Many output routines use `item->data` as the `extra_rotation` array. This never made sense to me, because for intelligent objects, `item->data` holds a `CREATURE*`, not extra rotations, but I shrugged it away thinking that for sure no bones use the extra rotation flags, eg it never tries to resolve it for those cases.

But actually the bones **do** use those flags and it **is** the extra rotations list, because in TR1 and TR2, `CREATURE` begins with `head_rotation` and `neck_rotation`, while in TR3 it starts with `joint_rotation[4]`. So when code treats `item->data` as the start of an `extra_rotation` array, it's effectively indexing into those early rotation fields of `CREATURE`. And in the OG, everything works!

TR3 does no longer uses `head_rotation`  / `neck_rotation` fields, and instead, introduces `joint_rotation[4]`. Some time ago, I copied them over, placing them… at the end of the `CREATURE` array – and that's the mistake. This only hit us with Shiva because it's the first TR3 object to use `joint_rotation[4]` in such an evident manner. So, to fix this, I modified the `CREATURE` definition so its first members intentionally match the expected layout of `extra_rotation`. That way, accessing `item->data` as rotations is now correct again.

It's still not a great design, as it relies on the field layout, but fixing that is out of scope for this PR.

lahm86 (approved):
Good find on the rots. Agreed that changing it to not rely on layout can be done another time.

rr-:
@aredfan Let's move to the ticket re: guns. I fixed incorrect damage values for the Grenade Launcher and the Harpoon Gun, and ensured projectiles always damage Shivas even while she's in her defensive state. While this fixes the Harpoon counts, Rocket Launcher and Grenade Launcher counts remain incorrect. Investigating further.

aredfan:
> @aredfan Let's move to the ticket re: guns. I fixed incorrect damage values for the Grenade Launcher and the Harpoon Gun, and ensured projectiles always damage Shivas even while she's in her defensive state. While this fixes the Harpoon counts, Rocket Launcher and Grenade Launcher counts remain incorrect. Investigating further.

Thank you, both the Shiva's head rotation and the Harpoon Gun's damage LGTM. 👍 

- [x] I've come across a new issue while checking out the serpent stone receptacle.  After Lara inserts the stone, save and reload, and the stone has stopped rotating in the receptacle.

rr-:
> > @aredfan Let's move to the ticket re: guns. I fixed incorrect damage values for the Grenade Launcher and the Harpoon Gun, and ensured projectiles always damage Shivas even while she's in her defensive state. While this fixes the Harpoon counts, Rocket Launcher and Grenade Launcher counts remain incorrect. Investigating further.
> 
> Thank you, both the Shiva's head rotation and the Harpoon Gun's damage LGTM. 👍

I am not seeing from the code how you can take down a Shiva with just 2 rockets / 3 grenades. Do you have any idea Lahm?

> * [ ]  I've come across a new issue while checking out the serpent stone receptacle.  After Lara inserts the stone, save and reload, and the stone has stopped rotating in the receptacle.

Thanks – good catch, fixed. Will need testing OG saves near complete and incomplete puzzle holes in TR1/2, though.

aredfan:
> Thanks – good catch, fixed. Will need testing OG saves near complete and incomplete puzzle holes in TR1/2, though.

No problem. The OG saves looks good re empty and filled puzzle slots.

lahm86:
> I am not seeing from the code how you can take down a Shiva with just 2 rockets / 3 grenades. Do you have any idea Lahm?

I've stepped through and it seems that in OG, `HitTarget` is called twice per frame for rockets/grenades 🫨 I'm not sure why, it just stems from `Item_Control` and it's the same item number each time. Maybe this quirk is an OG bug?

rr-:
Thanks, that's really revealing. I'll focus on finding out why this happens.

@aredfan It would be good to see if it's a TOMB3 regression, or an OG thing. Do we by any chance have the same kinds of differences against the OG TR2 – can we confirm the hit counts for the goons in OG Barkhang? (IIRC everyone takes 2 grenade shots in that level).

rr-:
> > I am not seeing from the code how you can take down a Shiva with just 2 rockets / 3 grenades. Do you have any idea Lahm?
> 
> I've stepped through and it seems that in OG, `HitTarget` is called twice per frame for rockets/grenades 🫨 I'm not sure why, it just stems from `Item_Control` and it's the same item number each time. Maybe this quirk is an OG bug?

Excellent find and I can explain why this happens in TOMB3 – the question is if it also happens in the OG TR2 and TR3. 

Basically, whenever we detonate the projectile, we restart the outer nearby rooms traversal from scratch (with different AoE radius). The outer loop is guaranteed to run at most twice thanks to the `exploded` check. Target-wise, it'll reliably hit the same item twice if there are no nearby targets and it's not a special type (like bosses or smashables). If other items are close, the damage spreads out, and in crowded spots, multiple targets can get hit more than twice.

To achieve the exact accuracy, I'll need to port the TR3 ways of iterating. The big question that remains is whether going with it for grenades isn't going to regress TR2.

rr-:
I've unconditionally applied the fix to rockets, which are TR3-exclusive. It seems to double the damage to Shivas as expected :relieved: So now only Grenade Launcher needs decision.

Ideally, we should test the hit counts with an isolated enemy, to simplify the risk of spilling the damage to nearby enemies.

aredfan:
Thanks for fixing the OG bug with the smoke effect. 👍 

- [x] If the shiva statue kills Lara in a special way which involves picking her up, the camera ends up behind the shiva statue.  In OG, the camera remains in front of the shiva statue.

lahm86:
Camera issue is my bad - this should fix it.

```diff
diff --git a/src/trx/game/lara/state/extra.c b/src/trx/game/lara/state/extra.c
index 1bfcc3dc1..7dcd33ea8 100644
--- a/src/trx/game/lara/state/extra.c
+++ b/src/trx/game/lara/state/extra.c
@@ -277,4 +277,5 @@ REGISTER_LARA_EXTRA(LS_EXTRA_PULL_DAGGER,    M_PullDagger)
 REGISTER_LARA_EXTRA(LS_EXTRA_START_ANIM,     M_StartAnim)
 REGISTER_LARA_EXTRA(LS_EXTRA_START_HOUSE,    M_StartHouse)
 REGISTER_LARA_EXTRA(LS_EXTRA_END_HOUSE,      M_EndHouse)
+REGISTER_LARA_EXTRA(LS_EXTRA_SHIVA_KILL,     M_BeastKill)
 // clang-format on
```

rr-:
Thanks – fixed the camera issue.

aredfan (approved):

lahm86 (approved):

rr-:
@aredfan I'd like to address the grenades mismatch here if possible – can you please confirm how many shots the bandits take in OG TR2? Thanks.

aredfan:
> @aredfan I'd like to address the grenades mismatch here if possible – can you please confirm how many shots the bandits take in OG TR2? Thanks.

@rr- yes no problem, I also included the xian statues for good measure.

- The monastery goons each takes 2 grenades to kill
- The xian swordsmen each takes 3 grenades to kill
- The xian spearmen each take 4 grenades to kill

rr-:
Thank you so much – especially for attaching info on the Xian spearmen, which proved really useful for testing. Unfortunately, the original TR2 function *was* ported correctly, so I ended up having to add a version gate, after all.

Currently, Rocket Launcher has the OP double damage as per OG TR3, whereas the Grenade Launcher is OP only in TR3, remaining weakened in earlier games. I would like to give this one more iteration. What’s our take on these three possible routes?

1) We consider this double damage to be a bug, and we nerf both Rocket and Grenade Launchers. This breaks OG parity with TR3. Personally, this option would leave me salty.
2) Nerf Rocket Launcher only in TR1/TR2, by doing a similar version gate as in the Grenade Launcher. This would make its output consistent with the Grenade Launcher damage values across all three games.
3) My recommendation: turn the target iteration logic into a setting, `Projectile area damage`, with options `Single-sweep` and `Multi-sweep`. That way we can tune balance per game – nerf for TR3, buff for TR2 – while keeping OG behavior as default. (Personally, I'd love to see the grenades hit harder in TR1/2 by default, but I understand this is not ideal to everyone.)

lahm86:
Option 3 sounds ideal to me ☺️

aredfan:
> 3\. My recommendation: turn the target iteration logic into a setting, `Projectile area damage`, with options `Single-sweep` and `Multi-sweep`. That way we can tune balance per game – nerf for TR3, buff for TR2 – while keeping OG behavior as default. (Personally, I'd love to see the grenades hit harder in TR1/2 by default, but I understand this is not ideal to everyone.)

I'd go with your recommendation too.  I think having an option for the rocket/grenade sweeps behaviour offers more flexibility across the 3 games.

Besides, the bug itself has been around for long enough so we might as well consider it a feature.

rr-:
Added the toggle. I must say, TR2's Grenade Launcher finally feels *fun* to use. And while this option is pretty much a completely arbitrary mod, it somehow feels far more legitimate than manually fiddling with `weapons.json5`.

data/common/ship/cfg/base_strings.json5
```diff
@@ -665,6 +667,8 @@
         "SETTING_PAUSE_FADE_EFFECTS_DESCRIPTION": "Fine-tunes the fade effects to be enabled or disabled in the pause screen. Needs the Fade Effects option to be enabled to work.",
         "SETTING_PAUSE_MUSIC_IN_INVENTORY": "Pause music in inventory",
         "SETTING_PAUSE_MUSIC_IN_INVENTORY_DESCRIPTION": "Pauses game sounds, ambient and music in the inventory screen.",
+        "SETTING_PROJECTILE_AREA_DAMAGE": "Projectile area damage",
+        "SETTING_PROJECTILE_AREA_DAMAGE_DESCRIPTION": "Controls how the area-of-effect for Rocket Launcher and Grenade Launcher propagates.\n\n- Single-sweep: TR1 & TR2 behavior.\n- Multi-sweep: TR3 behavior.\n\nThe multi-sweep option often ends up doing double damage to individual enemies.",
```
aredfan, line 671: We'll need a British translation for the 2 instances of `behaviour` with a `u`.

lahm86 (approved):
LGTM. A great option 😌

data/common/ship/cfg/base_strings.json5
rr-: Thank you – supplied!

aredfan: Thank you very much 😌

aredfan (approved):

PR #4737, opened 2026-01-27
rooms/floor-data: fix tr3 antitriggers

rr-:
#### Checklist

#### Description

This implements the split one-shot flags for antipads/antitriggers and for switches. @lahm86 can I ask for assessment whether this is something that's safe to land engine-agnostic, or if we should keep the version gate? 

It should fix the problems with antitriggers not working in certain cases, such as the Madhouse door.
The blades in Coastal Village never stop for a different reason, and it's an OG bug: https://www.youtube.com/watch?v=hdrrTGAB39c&t=9135s . To fix that, I implemented Lahm's pattern of resetting other traps.

### DIFF
--- a/src/trx/game/rooms/floor_data.c
+++ b/src/trx/game/rooms/floor_data.c
@@ -454,7 +454,25 @@ void Room_TestSectorTrigger(const ITEM *const item, const SECTOR *const sector)
         case TO_OBJECT: {
             const int16_t item_num = (int16_t)(intptr_t)cmd->parameter;
             ITEM *const trig_item = Item_Get(item_num);
-            if (trig_item->flags & IF_ONE_SHOT) {
+
+            bool one_shot = false;
+            if (g_TRVersion == 3) {
+                switch (trigger->type) {
+                case TT_SWITCH:
+                    one_shot = trig_item->flags & IF_ONE_SHOT_SWITCH;
+                    break;
+                case TT_ANTIPAD:
+                case TT_ANTITRIGGER:
+                    one_shot = trig_item->flags & IF_ONE_SHOT_ANTITRIGGER;
+                    break;
+                default:
+                    one_shot = trig_item->flags & IF_ONE_SHOT;
+                    break;
+                }
+            } else {
+                one_shot = trig_item->flags & IF_ONE_SHOT;
+            }
+            if (one_shot) {
                 break;
             }
 
@@ -483,6 +501,9 @@ void Room_TestSectorTrigger(const ITEM *const item, const SECTOR *const sector)
 
             if (trigger->type == TT_SWITCH) {
                 trig_item->flags ^= trigger->mask;
+                if (trigger->one_shot && g_TRVersion == 3) {
+                    trig_item->flags |= IF_ONE_SHOT_SWITCH;
+                }
             } else if (
                 trigger->type == TT_ANTIPAD
                 || trigger->type == TT_ANTITRIGGER) {
@@ -491,7 +512,11 @@ void Room_TestSectorTrigger(const ITEM *const item, const SECTOR *const sector)
                 }
                 trig_item->flags &= ~trigger->mask;
                 if (trigger->one_shot) {
-                    trig_item->flags |= IF_ONE_SHOT;
+                    if (g_TRVersion == 3) {
+                        trig_item->flags |= IF_ONE_SHOT_ANTITRIGGER;
+                    } else {
+                        trig_item->flags |= IF_ONE_SHOT;
+                    }
                 }
             } else {
                 trig_item->flags |= trigger->mask;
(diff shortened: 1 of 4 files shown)


aredfan (approved):

lahm86:
I'm pretty sure this will be safe to extend into TR1/2, but I will scan through OG FD later today to double check level state.

src/trx/game/rooms/floor_data.c
```diff
@@ -466,23 +485,26 @@ void Room_TestSectorTrigger(const ITEM *const item, const SECTOR *const sector)
             if (is_shoal_object && trigger->type != TT_ANTIPAD
                 && trigger->type != TT_ANTITRIGGER) {
                 Shoal_TriggerActivate(trig_item, trigger->timer);
-            } else {
```
lahm86, line 469: This seems to cause fish to freeze. I think it's guarded for things like the fish that use the timer field for their own context.

src/trx/game/rooms/floor_data.c
```diff
@@ -361,6 +361,7 @@ void Room_TestTriggers(const ITEM *const item)
     }
 }
 
+#include <trx/log.h>
```
lahm86: Leftover include

lahm86 (commented):
I think we'll have to leave it guarded for TR3 for now. One example for testing it enabled in TR2:
In Offshore Rig room 42, run into room 43 to trigger the door ahead. Keep going and cross the (one-shot) antitrigger. Go back to the first trigger to open it again - crossing the antitrigger now has no effect, so you can reach the unreachable area.

We can likely fix TR1/2 cases with injections if we want to backport it in the future.

src/trx/game/rooms/floor_data.c
rr-: Fixed

src/trx/game/rooms/floor_data.c
rr-: Oh yeah it's a leftover mistake – reverted.

rr-:
> I think we'll have to leave it guarded for TR3 for now. One example for testing it enabled in TR2: In Offshore Rig room 42, run into room 43 to trigger the door ahead. Keep going and cross the (one-shot) antitrigger. Go back to the first trigger to open it again - crossing the antitrigger now has no effect, so you can reach the unreachable area.
> 
> We can likely fix TR1/2 cases with injections if we want to backport it in the future.

Thanks – removed the version gate.

lahm86:
> > I think we'll have to leave it guarded for TR3 for now. One example for testing it enabled in TR2: In Offshore Rig room 42, run into room 43 to trigger the door ahead. Keep going and cross the (one-shot) antitrigger. Go back to the first trigger to open it again - crossing the antitrigger now has no effect, so you can reach the unreachable area.
> > We can likely fix TR1/2 cases with injections if we want to backport it in the future.
> 
> Thanks – removed the version gate.

Sorry, I meant I had tried a local build with the version gate removed and ended up with the scenario above in Rig. So I think it'll need to remain - or we can look at injections for TR1/2, which I could try to fit in today.

rr-:
> Sorry, I meant I had tried a local build with the version gate removed and ended up with the scenario above in Rig. So I think it'll need to remain - or we can look at injections for TR1/2, which I could try to fit in today.

Oh, I skimmed over the first sentence, and understood you considered this an optional fix  – sorry! In this case, let's go with an injection so that we can fully close this. I think it'll also let us mark Switches as done in the TR3 roadmap, as the only reason they were marked in yellow were the Londons antitriggers.

lahm86:
> > Sorry, I meant I had tried a local build with the version gate removed and ended up with the scenario above in Rig. So I think it'll need to remain - or we can look at injections for TR1/2, which I could try to fit in today.
> 
> Oh, I skimmed over the first sentence, and understood you considered this an optional fix  – sorry! In this case, let's go with an injection so that we can fully close this. I think it'll also let us mark Switches as done in the TR3 roadmap, as the only reason they were marked in yellow were the Londons antitriggers.

I found a snag with changing TR1/2 unfortunately. This area in Wreck I think would be impossible to rearrange seamlessly as the door closes behind Lara when she enters, then the antipad by the switch reopens it. It too marks it as oneshot, so that means the pad on the way back doesn't re-close it. 

<details>

image
</details>

TR3 has changed the oneshot antitrigger concept to mean the antitrigger will only run once, but re-triggering the item normally  can work. Like below in Crash Site, the door is antitriggered when you enter, you use the switches to re-open it, and then the antitrigger is ignored when you leave.

<details>

image
</details>

In TR1/2 it means antitrigger the item, and then never allow it to be re-triggered again, like the statues in Bartoli's.

<details>

image
</details>

So I think we'll need to guard this behaviour still 😢

rr-:
Thank you for the detailed investigation. I have restored the gate check, and made sure TR1/2 does not use the new flags at all.

Maybe one day we can address this in TE to make both behaviors possible with new masks…

lahm86 (approved):
Thanks. Yes, hopefully we can find a way to merge things later.

PR #4742, opened 2026-01-28
Fix counting kills from quads and manually aimed projectiles

rr-:
#### Checklist

#### Description

This PR fixes kills from quads and manually aimed projectiles not getting counted in the level stats.

Honestly I don't like `DONT_TARGET`. How do we feel about refactoring it away in favor of introducing `item->targetable`? This way we can get destructible objects to have proper HP (sleeping Cobras now are one-shot as a semi-hardcoded logic, and it's not really possible to change that).
It's a big, big task with tons of objects to go over, though. I think we could tackle it iteratively, however, by declaring `item->targetable` as true by default, getting the targeting logic right, and only going object-by-object to sunset `DONT_TARGET`.

The reason why I'm bringing this up here, is because I find it really upsetting that you can kill Army Winston with the Rocket Launcher, and it's related because we can't use `DONT_TARGET` for him. And instead of complicating the matters further by mixing the existing special HP values with flags, I'd like to tidy things up for good.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -129,6 +129,7 @@
 - fixed ally Lua API not working with most of the TR3 enemies supported so far
 - fixed one-shot antitriggers / antipads behavior
 - fixed Blades in Coastal Village not respecting antitrigger
+- fixed running down an enemy with a Quad not counting as a kill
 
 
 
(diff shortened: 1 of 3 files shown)


aredfan (approved):

lahm86 (approved):
I agree, replacing `DONT_TARGET` with a proper flag sounds much more logical.

PR #4744, opened 2026-01-28
tr1/data: prevent overwriting pushblock animations

### LINKED ISSUE #4735: TR1 1.1 - Natlas Mines - Moveable blocks in the golden rooms just before end
Hi,

there's a strange behaviour of the moveable blocks near the end (before entering the place with the shotgun guy):
If I push a block it just rotates about 45 degrees and back but stands at it's place - if you pull it, it just don't move. Afterwards you can just walk through the blocks.

I've added 4 screenshoss (pulling a block):
1-Start pulling
2-During pull
3-End of pull (Lara moved, the block seems to stay at it's initial place)
4-You can walk inside and through the block to the next room

Image
Image
Image
Image

lahm86:
Resolves #4735.

#### Checklist

#### Description

This replaces the fix for the pushblocks in Natla's Mines to add the missing SFX commands rather than replacing the entire animation set of each block, so to prevent the original level items having different animation numbers to the original.

The code addition here I am uneasy about, but it ultimately makes the save in the issue workable, and any saves made in the current 1.1 will work in 1.2 etc. WDYT?

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/201


rr- (commented):
WDYT of making this into a generic fix in the BSON savegame loader with a `LOG_ERROR`?
This approach would be more attractive to keep around for longer, which I'm not sure how I feel about. It's also be similar to how I started to relax some of the essential fields by using `M_SHOULD` rather than `M_MUST`.

The fix proposed in this PR might be better, because it's more likely that we'd eventually get it removed.

rr- (approved):

lahm86:
> WDYT of making this into a generic fix in the BSON savegame loader with a `LOG_ERROR`? This approach would be more attractive to keep around for longer, which I'm not sure how I feel about. It's also be similar to how I started to relax some of the essential fields by using `M_SHOULD` rather than `M_MUST`.
> 
> The fix proposed in this PR might be better, because it's more likely that we'd eventually get it removed.

Yeah, I think we'd have to guard it by object in the BSON because items can validly use another object's animation e.g. `Item_SwitchToObjAnim` cases. So as much as I dislike the PR approach, maybe it's the lesser of two evils.

Perhaps something for the future would be to save `anim_object` and the relative anim and frame number, but the new `anim_object` field would also need to be stored and tracked on the item, and integrating this with older saves could be difficult too.

aredfan (approved):

PR #4752, opened 2026-01-29
objects/boat: use new priv data pattern

rr-:
#### Checklist

#### Description

This moves Boat private data to be fully owned by `boat.c` module, making use of the private data storage pattern.

#### Testing

Just the Boat behavior with OG and BSON saves. Only two objects left after this PR – Sliding Pillars and Skidoos.


aredfan:
- [x] The only issue I found is the gondola boats disappear as soon as the motor boat is within touching distance. [record.txt](https://github.com/user-attachments/files/24944345/record.txt)

src/trx/game/objects/vehicles/boat.c
```diff
@@ -197,23 +246,21 @@ static void M_DoWakeEffect(const ITEM *const boat_item)
 
 static void M_DoShift(const int32_t boat_num)
 {
-    ITEM *const boat_item = Item_Get(boat_num);
-    int16_t item_num = Room_Get(boat_item->room_num)->item_num;
+    ITEM *const item = Item_Get(boat_num);
```
lahm86: I think this rename could be causing the gondola problem as `item` is re-used in the loop further below.
Was wondering if we should use `-Wshadow` but it looks like it'll open a can of worms 😅

rr-:
Apologies for this, I went a bit overzealous with my refactor. I reverted the `boat_item` rename. I've added a TODO for myself to look at `-Wshadow`, as it's a very useful warning.

lahm86 (approved):

aredfan (approved):

PR #4761, opened 2026-01-30
Use priv data pattern for the flame emitters

rr-:
#### Checklist

#### Description

This is a bit weird because I found the save routines' presence, or lack thereof, to have no seeming effect on the game.
I checked this by fiddling with the enhanced save games on/off, and trying to get Lara to catch fire twice by saving/reloading and walking into fire emitters. All good all around.
And TR3 flames still restart, because they're controlled by sparks, which we do not save (yet?).

So…
What I did here was to remove the save/load special handlers for the flame emitters, and replace `item->data` with `item->priv` usage.

### DIFF
--- a/src/trx/game/effects.c
+++ b/src/trx/game/effects.c
@@ -84,6 +84,19 @@ int16_t Effect_GetNum(const EFFECT *effect)
     return effect - m_Effects;
 }
 
+int16_t Effect_GetInOrderNum(const int16_t effect_num)
+{
+    int16_t order_num = 0;
+    for (int16_t link_num = Effect_GetActiveNum(); link_num != NO_EFFECT;
+         link_num = Effect_Get(link_num)->next_active) {
+        if (link_num == effect_num) {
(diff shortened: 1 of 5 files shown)


lahm86:
I think flames are doubling up on load with the `save effects` option on:
- `/play 7` (TR1)
- `/fly`
- `/tp emit`
- `/save 1`
- `/load 1`

rr-:
> I think flames are doubling up on load with the `save effects` option on:
> 
> * `/play 7` (TR1)
> * `/fly`
> * `/tp emit`
> * `/save 1`
> * `/load 1`

Thank you for spotting this – I was looking for the wrong thing (double flames on Lara), and it didn't occur to me to listen to volume differences or look for z-fighting on duplicated flames on the emitters themselves. Updated!

lahm86 (approved):
Thanks! LGTM

aredfan (approved):

PR #4764, opened 2026-01-31
Drop support for <TRX 1.0 BSON saves

rr-:
#### Checklist

- [x] I have read the [coding conventions](https://github.com/LostArtefacts/TRX/blob/master/CONTRIBUTING.md#coding-conventions)
- [x] I have added a changelog entry about what my pull request accomplishes, or it is an internal change
- [x] I have added a readme entry about my new feature or OG bug fix, or it is a different change

#### Description

This PR tidies up a big batch of old code used for pre‑merge savegames (TR1X/TR2X era), making the savegames module a whole lot easier to maintain. A follow‑up PR will fully retire support for the original save formats (TombATI, TR2 Steam/GOG).

The legacy save formats were vital during early development, but maintaining them now imposes a disproportionate cost on testing and stability. Deprecating this path lets us focus our time on actively improving the engine for everyone.

Nonetheless, players who want to hold onto their older saves can re‑save them with version 1.0 or 1.1 before upgrading, and it should all continue to work – just with an extra step.

#### Testing

1. Create new saves in TR1–3 using versions 1.0 and 1.1 with various gameplay elements. Confirm they still load correctly and don't kick the player out to the main menu.
    Areas of interest:
    - Active enemies
    - Pickups
    - Effects
    - Holster/back/equipped guns

    Note: 1.0 did not support TR3 savegames and it is expected to crash when trying to save the game.
3. Create new saves in the current build for all three games, and confirm they load properly as well.

PR #4766, opened 2026-02-01
Drop support for legacy saves

rr-:
#### Checklist

#### Description

This is part 3 of the savegames refactor, that drops the support for the OG savegame files.

Part 1: https://github.com/LostArtefacts/TRX/pull/4764 
Part 2: https://github.com/LostArtefacts/TRX/pull/4765

Since the strategy pattern no longer fit, I have flattened all the related code into a single module. With BSON now being the only supported format, there was no need to emphasize it in the module name, so I went with a more neutral name, `SG_File`.

I've also updated the gameflow files to match this change. They support the old field for now for backwards compatibility, but it's planned for removal in TRX 1.5.

#### Testing

- Lightweight testing if the game continues to load any of the supported saves across all 3 games – if one works, all will work (though it'd be good to check the Golds and PC demo to verify gameflow edit mistakes)
- Have TombATI / TR2 PC / TR1X / TR2X saves present. The game should complain in the logs about them, but not crash.


lahm86 (approved):
Great 😌

aredfan (approved):

PR #4775, opened 2026-02-01
camera/los: reduce bobbing in tr3 look cameras

rr-:
#### Checklist

#### Description

This PR stabilizes the look-camera ray by "straightening" the head-derived sample points before they're used.

Right now, we pick three points around Lara's head (`pos_1/2/3`) and use them to build the look ray; those points come from animated head/torso matrices, so when Lara does idle breathing, the head bone gets a tiny roll/tilt wobble and the points wobble, too. The camera then follows that wobble, which feels like bobbing/rolling even when the player is not moving.

The fix keeps the same 3 points, but adjusts them in a controlled way:

1. Take `head_pos` as the anchor point.
2. Compute a stable forward direction axis from just yaw + pitch (`XYZ_32_FromYawPitch(yaw, pitch)`), which ignores the roll/tilt coming from breathing.
3. For each `pos_i`, snap it onto the straight line `head_pos + t*axis` by replacing it with the closest point on that line (`XYZ_32_ProjectPointOntoAxis(…)`).

After this, the ray is still head-locked and faithful to the existing setup, but it no longer inherits breathing roll/tilt noise.

The projection math and small 64-bit helpers were added to `math/` to avoid clouding the camera code.
I also renamed `XYZ_32_DirShift` to `XYZ_32_OffsetYaw` to better match intent.

#### Testing

I'm afraid to test the behavior, we'll either need a test level that has some unique scenarios: cameras next to walls, cameras in very open spaces (think Lara standing on a pillar), pointing at targets down and above Lara…, etc. I only tested this with Jungle room 10 as this is the only look camera I know of :sweat_smile:  but it seemed to work as expected.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -114,6 +114,7 @@
     - added Beacon Light control
     - added On/Off Light control
 - added Lara's backwards-hop stumble if there is a slope behind her
+- improved look camera stability to reduce idle-breathing camera bobbing/roll
 - improved Monkeys to no longer hardcode hostility status based on Tiger presence
 - changed hostile Monkeys to share hostility status, like TR2 Barkhang monks (the original TR3 behavior can be restored in Gameplay → General → Ally hostility policy)
 - changed enemy drops to appear at the tile center, to conform with the OG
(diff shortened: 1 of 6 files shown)


lahm86 (approved):
This is a great improvement! I'll share a test level on Discord.

aredfan (approved):

PR #4796, opened 2026-02-06
ui: expand bar theme support

rr-:
#### Checklist

#### Description

This PR adds TR3 PS1 bar color theme and streamlines TR1-3 PC bar color themes (thanks Aredfan!).
It also adds theoretical support for more themes and bar colors.

#### Implementation details

To achieve this, we de-hardcode UI bar themes, and migrate bar look selection from enum-backed values to dynamic theme keys loaded from `cfg/ui.json5`. The bar settings backend now loads all themes dynamically by key, dispatches PC/PS1 parsing from each theme’s `style`, and uses per-theme `name_gs` labels for the settings UI. The bars settings tab now uses `X_UI_CFG_STRING(ui.bar_look, ...)` with custom handlers instead of enum entries, and the old hardcoded `UI_Settings_BarLookEnumEntries` path is removed.

Config defaults and mappings were updated to the new theme/color naming. Legacy config migration was extended for old bar look IDs and renamed color names, then gated behind a schema version (`config_version`) so that migration is one-shot and cannot be re-applied on already-migrated configs. The concept of `config_version` is new, but is similar to what we have been using for savegames.

It's worth highlighting that the theme names aren't hardcoded either. `name_gs` is transient, and the keys exist only in `base_strings*.json5`, not in `game_string.def`. `tools/update_game_strings` now treats dynamic theme labels from `cfg/ui.json5` as used, preventing `BAR_LOOK_…` entries in `base_strings*.json5` from being removed even though they’re never defined in `game_string.def`. The other change that was needed for this was to get `game_string_table` loader to accept unknown game string keys from tables without warnings, and define them directly.


aredfan:
- [x] I found just one issue. If the game is started without TR*X.json5 being present in the cfg folder, the PC bar colour for sprint defaults to teal instead of green.

lahm86 (approved):

rr-:
> * [ ]  I found just one issue. If the game is started without TR*X.json5 being present in the cfg folder, the PC bar colour for sprint defaults to teal instead of green.

Thank you – applied

aredfan (approved):

PR #4801, opened 2026-02-06
Replace LARA_SKIN_TYPE with dynamic keys

rr-:
#### Checklist

#### Description

Following the pattern introduced in daee8e6f634072e0b54e98c80b7798d9f9cd5107, this PR migrates Lara outfit selection to transient game strings and JSON-key-based identifiers, removing enum-driven outfit selection from the player-facing flow.

`cfg/outfits.json5` is now the source of truth for selectable outfits via stable string keys (for example `tr2_diving_suit`) plus `name_gs` entries (`LARA_OUTFIT_*`) for localized labels. Base game strings were updated to the new `LARA_OUTFIT_*` names, and `tools/update_game_strings` was updated so these outfit `name_gs` references are treated as in-use. The hardcoded "Default" UI label remains declared with `GS_DEFINE(LARA_OUTFIT_DEFAULT, "Default")`.

Player config now stores `visuals.lara_outfit` (string or `nullptr` for default behavior) instead of `visuals.lara_skin_type`. Gameflow level config also uses `lara_outfit` string values. All shipped TR1/TR2/TR3 gameflow files and relevant schema/docs were updated accordingly.

Runtime outfit resolution follows the intended fallback flow: `nullptr` means default/level-driven outfit behavior, invalid configured outfit names log a warning and fall back to default outfit selection.

Lua-facing outfit control now uses outfit strings and the public property is trx.lara.outfit (with docs/scripts updated), matching the new string-key model.

Compatibility notes: this is an unreleased feature, so no config migration logic was added despite the rename.

---

I opted to use `outfit` for builder-facing names throughout, as "skin" sounded too… fleshy to me :sweat_smile: LMK if you would like to have that reverted.

#### Testing

- [x] Config default behavior. Set `lara_outfit` in `TRX.json5` to `nullptr` and load multiple levels with different `lara_outfit` in gameflow. Expected: level-defined outfit changes apply normally.
- [x] Config override behavior. Set `lara_outfit` in `TRX.json5` to a valid key like `"tr2_diving_suit"`. Expected: Lara uses that outfit persistently until changed.
- [x] Invalid config fallback. Set `lara_outfit` in `TRX.json5` to an unknown key. Expected: warning is logged and default outfit is used instead of crashing.
- [x] Diving Area cutscene. Play the Diving Area cutscene until Lara changes into her diving suit. Expected: Lara switches her outfit (unless the player has overridden it).
- [x] UI outfit cycling. Change outfit in settings UI and relaunch level. Expected: selection labels come from `name_gs` and are subject to the language selection, and selected outfit persists via `visuals.lara_outfit`.
- [x] UI Default outfit. Change outfit to Default in settings UI and relaunch level. Expected: outfit persists via `lara_outfit` as NULL and does not get written to `TRX.json5`.
- [x] Enforcing the outfit by the builder. Edit any game's game flow to include: `"enforced_config": { "lara_outfit": null }` on the top level, and try playing the game as Bacon Lara. Expected: the outfit should be set to one appropriate for the played level.
- [x] Double check the docs if they reflect the changes well.


lahm86:
Thanks! Do we need to define the remaining 12 entries in the strings as placeholders? Just thinking if a builder wants to offer the full 32 options, they'd run out of `name_gs` options I think atm.

- [x] There's one issue not on develop, that's a bit weird to reproduce - if repeatedly toggling the outfit (using the input key) it sometimes gets "stuck" and you need to keep pressing it a couple of times for it to shift forward. Looking into it.

rr-:
>  Just thinking if a builder wants to offer the full 32 options, they'd run out of name_gs options I think atm.

Nah, they can just add any missing strings themselves in `base_strings*.json5`. The goal's for those keys to actually mean something – so if they’re doing `outfit_batman`, I'd rather see `LARA_OUTFIT_BATMAN` than `LARA_OUTFIT_21` :relieved:

lahm86:
> > Just thinking if a builder wants to offer the full 32 options, they'd run out of name_gs options I think atm.
> 
> Nah, they can just add any missing strings themselves in `base_strings*.json5`. The goal's for those keys to actually mean something – so if they’re doing `outfit_batman`, I'd rather see `LARA_OUTFIT_BATMAN` than `LARA_OUTFIT_21` 😌

Got it, thanks. I was misunderstanding initially. This is really nice ☺️

rr-:
> * [ ]  There's one issue not on develop, that's a bit weird to reproduce - if repeatedly toggling the outfit (using the input key) it sometimes gets "stuck" and you need to keep pressing it a couple of times for it to shift forward. Looking into it.

I think I got this:

- `Config_Update()` detects `COT_STRING` changes by pointer identity (`src/trx/config/common.c:372`).
- In `Lara_Skin_CycleOutfit`, the new code was doing free then dup for `g_Config.visuals.lara_outfit`, which can occasionally reuse the same address, so config change events are dropped intermittently.
- That matches the symptom of the outfit toggle sometimes appearing stuck and needing extra presses.

I changed this to dup first, and then drop the old memory, forcing the change to get detected. I haven't tested it, though.

lahm86:
Thanks, looking good. That didn't cross my mind to try.
Will check through the other scenarios.

lahm86:
>  Invalid config fallback. Set lara_outfit in TRX.json5 to an unknown key. Expected: warning is logged and default outfit is used instead of crashing.

Tried this in TR2, then loaded Venice. Lara's TR1 gym outfit is used, but the UI says "Default". I feel Default infers the GF level default as opposed to the JSON default. 
Similarly, if you use the console with an invalid value e.g. `/set outfit banana` it defaults to TR1 gym. I think in this case it should just fail and leave the existing value (granted on develop, the console approach is most unfriendly 😄).

Everything else LGTM.

rr-:
> > Invalid config fallback. Set lara_outfit in TRX.json5 to an unknown key. Expected: warning is logged and default outfit is used instead of crashing.
> 
> Tried this in TR2, then loaded Venice. Lara's TR1 gym outfit is used, but the UI says "Default". I feel Default infers the GF level default as opposed to the JSON default. Similarly, if you use the console with an invalid value e.g. `/set outfit banana` it defaults to TR1 gym. I think in this case it should just fail and leave the existing value (granted on develop, the console approach is most unfriendly 😄).
> 
> Everything else LGTM.

Thanks – I think I got that fixed, and also have fixed `banana` to fail and display available values instead.
I don't like the tight coupling in `cmd/set.c` with skins, but… Okay, I guess. I should do the same for the bar themes and colors later on.

lahm86 (approved):
Looking great, thanks again for doing this 😌

rr-:
(Squashed fixes into the main commit)

aredfan (approved):

PR #4802, opened 2026-02-06
game/cutscene: allow seeking back; fix slow seek forward

rr-:
#### Checklist

#### Description

This PR adds backward seeking to cutscenes and fixes the sluggish fast-forward behavior.

Cutscenes can now seek in both directions. During seeking, actors are replayed frame-by-frame through their control functions with `Camera_UpdateCutscene()`, keeping transforms and camera state perfectly aligned. When rewinding, actors reset to the start of their animation and then play forward up to the target frame.

The slow seek issue came from using debounced input instead of regular input, which made slow-fast-forward basically unusable.

#### Testing

- [x] **Backward seek:** Play a cutscene, move forward, then seek left. Actors should maintain correct orientation and smooth animation while rewinding.
- [x] **Forward seek:** Jump right from various points in the cutscene. Actor and camera states should stay consistent after each seek.
- [x] **Slow seek modifier:** Try seeking with and without the slow key held. The speed should reflect the intended modifier cleanly, with no big jumps. (I recommend using `/debug anim` to see exact frame step values.)

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -38,6 +38,8 @@
 - added an option to control whether or not moving boulders should shake the camera (Gameplay options → General → Enable boulder shake)
 - added an option to make Lara stumble if she hops backwards and there is a slope behind her (Gameplay options → Controls → Backwards slope stumble)
 - added `/trigger` and `/untrigger` console commands, with support for targeting by item ID, item name, or object name
+- added the ability to seek backwards through cutscenes with left button
+- improved the ability to seek through cutscenes to support even faster seeks (Slow = ±1 s, default = ±5 s, new: Draw = ±15 s)
 - improved `/tp` to accept `room`/`item` prefixes and `rN`/`iN` shortcuts
 - improved inventory ring active item highlight for smoother appearance
 - improved savegame file size by reducing it about 20–30%.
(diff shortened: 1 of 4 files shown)


lahm86:
Looks good, but I think we might face some difficult to resolve issues with animation commands. E.g. in Great Wall cutscene, go forwards and backwards a few times over the part where Lara draws her guns - her hands become quite inconsistent.

rr-:
> Looks good, but I think we might face some difficult to resolve issues with animation commands. E.g. in Great Wall cutscene, go forwards and backwards a few times over the part where Lara draws her guns - her hands become quite inconsistent.

Thanks. Fixed this, and the DA cutscene, to support both backwards and forwards seeking just fine.
Also, since DA cutscene is absolutely long, I've added an even faster seek shortcut with Draw+Right/Left (Action was my first choice, but it's used to terminate the cutscene immediately).

rr-:
I was wondering if the added code isn't too big of a maintenance burden, but I think it ended up just about acceptable. The ability to do big forward seeks followed by precise backward seeks is IMO just about enough convenient to justify this PR.

lahm86 (approved):

aredfan:
The ability to rewind in cutscenes is working great. 👍 

In TR2's 1st and 4th cutscenes, if left/right is pressed a few times, it can make the poison drink or the dagger disappear.  Not sure if it's fixable.  I also made a recording for these 2 scenarios.

[record.txt](https://github.com/user-attachments/files/25139418/record.txt)

aredfan (approved):

rr-:
Thanks – got that fixed as well :relieved:

aredfan (approved):
> Thanks – got that fixed as well 😌

Thank you.  I can confirm the poison drink and dagger is working well. 👍 

I also checked the Antarctica cutscene because Willard has a spoon and remote control at various points, all works well.

PR #4852, opened 2026-02-11
inv-ring: fix volume not being reset correctly

rr-:
#### Checklist

#### Description

This fixes a title-menu music regression introduced in 91364c8d9c530acce34a677a059f1aeabf9ca921.

The regression happens because title music can inherit the last runtime-adjusted gameplay volume (for example, inventory ring ambient attenuation) when returning to title.
Since the volume reset was removed from unload flow in that commit and title startup does not always pass through a `play_music` sequence event, the title ring can start with effectively muted music until a settings change forces `Music_SetVolume()`.

This change restores correct behavior by explicitly resetting menu/title music volume to `g_Config.audio.music_volume` when entering title inventory flow, so returning to title no longer depends on previous in-level attenuation state.

#### Testing

Compare the following TR2 recording on develop and on this PR:

```
config gameplay.enable_fmv 0
config gameplay.enable_legal 0

@+0:    noop
@+15:   cmd "set audio.inventory-ambient-volume 0.5"
@+15:   cmd "play 7"
@+2:    ● "escape"
@+2:    ○ "escape"
@+18:   ● "down"
@+6:    ○ "down"
@+20:   ● "left ctrl"
@+11:   ○ "left ctrl"
@+15:   ● "right"
@+6:    ○ "right"
@+7:    ● "right"
@+6:    ○ "right"
@+6:    ● "left ctrl"
@+10:   ○ "left ctrl"
@+80:   quit
```

Expected outcome: develop – the inventory ring gets muted upon return; PR – it gets reset back to 100%.

### DIFF
--- a/src/trx/game/inventory_ring/priv.c
+++ b/src/trx/game/inventory_ring/priv.c
@@ -74,6 +74,7 @@ static void M_HandleRequestedObject(INV_RING *const ring)
 void InvRing_AdjustMusicVolume(const INV_RING *const ring)
 {
     if (ring->mode == INV_TITLE_MODE) {
+        Music_SetVolume(g_Config.audio.music_volume);
         return;
     }
     const bool is_ambient =
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## [Unreleased](https://github.com/LostArtefacts/TRX/compare/trx-1.2...develop) - ××××-××-××
-- fixed settings dialog changing size when cycling through non-scrollable tabs
+- fixed title ring music inheriting the wrong audio volume (regression from 1.2)
+- fixed settings dialog changing size when cycling through non-scrollable tabs (regression from 1.2)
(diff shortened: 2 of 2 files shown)


src/trx/game/inventory_ring/priv.c
```diff
@@ -74,7 +74,7 @@ static void M_HandleRequestedObject(INV_RING *const ring)
 void InvRing_AdjustMusicVolume(const INV_RING *const ring)
 {
     if (ring->mode == INV_TITLE_MODE) {
-        return;
+        Music_SetVolume(g_Config.audio.music_volume);
```
lahm86, line 77: I think we still need the `return` - example is if you have ambient at 0%, start a level then go back to title via passport, there is no music.

rr-: Of course – fixed.

lahm86 (approved):

aredfan (approved):

PR #4862, opened 2026-02-12
config: allow showing loading screens for saves

### LINKED ISSUE #1290: Show loading screens on load
As we'd discussed, theoretically still missing as of yet. Had ideally been consistent with later TR2X (and even a TR3X 🤞).

To be decided how long the images should show. On a new level start, it seems desirable to keep the 5 sec length. On a regular load, would these only show briefly and fade into the level?

rr-:
#### Checklist

#### Description

This PR replaces the loading screens boolean with an enum mode:

- `disabled`
- `always`
- `new-games` (skip on save loads)

This now lets players see the loading screens when loading a save, which were previously hardcoded to be skipped.
The PR includes backward-compatible migration from legacy `enable_loading_screens` bool.

#### Testing

- [x] Open Gameplay settings and cycle loading screens mode
  Expected outcome: three options are shown and selectable (`Disabled`, `Always`, `New games`)
- [x] Legacy config migration (`enable_loading_screens: true`)
  Expected outcome: migrated behavior matches `Loading screens: Always`
- [x] Legacy config migration (`enable_loading_screens: false`)
  Expected outcome: migrated behavior matches `Loading screens: Disabled`
- [x] Load from a save with mode set to `new-games`
  Expected outcome: loading screen is skipped for save loads
- [x] Use "Play previous levels" and "Select level" features
  Expected outcome: loading screens shown for `new-games` and `always`, and do not show for `disabled`
- [x] Trigger "Story so far…" flow in all 3 loading screens modes
  Expected outcome: loading screens are skipped unconditionally and only FMVs and cutscenes play
  (I know this might be a small deviation… but I think it's reasonable? If we don't like it I'm happy to revert.)


data/common/ship/cfg/base_strings.json5
```diff
@@ -521,7 +524,7 @@
         "SETTING_ENABLE_LEGAL": "Legal screens",
         "SETTING_ENABLE_LEGAL_DESCRIPTION": "Enables legal screen and Core Design FMV at the game start.",
         "SETTING_ENABLE_LOADING_SCREENS": "Loading screens",
-        "SETTING_ENABLE_LOADING_SCREENS_DESCRIPTION": "Enables PS1 style loading screens before each level.",
+        "SETTING_ENABLE_LOADING_SCREENS_DESCRIPTION": "Controls PS1-style loading screens before level loads.\n\n- Disabled: never show loading screens.\n- Always: show loading screens.\n- New games: skip showing loading screens when loading a save.",
```
aredfan: I think we can remove the `PS1-style` here because loading screens are also in PC TR3.

rr-: Good catch – fixed all locales.

aredfan:
> Trigger "Story so far…" flow in all 3 loading screens modes
> Expected outcome: loading screens are skipped unconditionally and only FMVs and cutscenes play
> (I know this might be a small deviation… but I think it's reasonable? If we don't like it I'm happy to revert.)

- [ ] If the loading screens are set to `Always`,  they will continue displaying in `Story so far...`

rr-:
Thank you – fixed.

aredfan (approved):

lahm86 (approved):

PR #4868, opened 2026-02-13
Fix GL memory leaks

rr-:
#### Checklist

- [x] I have read the coding conventions (https://github.com/LostArtefacts/TRX/blob/develop/docs/CONTRIBUTING.md#coding-conventions)

#### Description

This PR fixes several OpenGL resource leaks in the output pipeline that could accumulate over repeated level reloads and eventually trigger out-of-memory errors.

As the main fix, it corrects mesh batcher layout lifetime handling: add/remove now only marks layout dirty, and packed vertex/index offsets are rebuilt at `MeshBatcher_Seal()`, so EBO/VBO allocations no longer grow across unload/load cycles when persistent meshes remain bound. It didn't work because the old cleanup code assumed unloading a level would completely clear the mesh batcher, while in reality, some output sources (like cuboids and spheres) only upload once per game launch by design.

Less importantly, it also adds missing cleanup for UI and debug VAO/VBO objects, releases output UBOs in `Output_Uniforms_Free()`, and closes renderer sampler/buffer objects during renderer shutdown. Finally, `OutputSource_UI_Shutdown()` is now called from `Output_Shutdown()` so that cleanup path is actually executed.

No gameplay behavior is changed; this is lifecycle/resource management hardening.

#### Testing

[oom.txt](https://github.com/user-attachments/files/25277581/oom.txt)
This recording repeatedly loads Meteorite Cavern and shows the load attempt number in the console. I chose this level, because surprisingly it has the most vertices overall. For me, on develop, the game ceases to render at around 860 mark.

- [x] Run `oom.txt` on develop
  Expected outcome: the counter gets laggier with time, and eventually the game stops rendering logging a lot of GL errors.
- [x] Run `oom.txt` on this branch
  Expected outcome: the levels load with uniform times, and the recording can complete without glitching.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,6 +9,7 @@
 - fixed not restoring Lara's back weapon mesh between levels when "remember guns" is enabled and a rifle-type weapon is equipped at level end
 - fixed a missing footstep sound when Lara starts to sprint
 - fixed Lara's flare undraw animation being skippable on specific late draw frames (#1593)
+- fixed a potential `GL_OUT_OF_MEMORY` error that could occur after reloading levels many times (regression from <1.0)
 
 **TR3**:
 - added a slide-to-sprint animation state change for Lara, similar to TR1 and TR2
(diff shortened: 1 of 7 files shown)


aredfan (approved):
> [x] Run `oom.txt` on this branch
> Expected outcome: the levels load with uniform times, and the recording can complete without glitching.

I F4-ed at 5000 as it showed no sign of slowing down. 😅 

LGTM.

lahm86 (approved):

PR #4872, opened 2026-02-13
Improve poly fx z fighting scenarios

rr-:
#### Checklist

#### Description

This PR:
- changes footprints to no longer get elevated by 16 (!) world units, and instead adds the z discriminator we use for shape-based shadows among other things.
- changes water ripples to also use the z discriminator, which avoids z-fighting with ground tiles of the same height.

#### Testing

- [x] Run around on various terrain
  Expected outcome: footprints continue to appear visible at slopes of all angles, and behave better in the photo mode.
- [x] `/play crash`, `/tp 23 2 59`, run around
  Expected outcome: water ripples do not trigger z-fighting
- [x] `/play area`, `/tp 16 -4 44`, run around near the tile edge
  Expected outcome: water ripples draw over the ground tiles

I was thinking of whether to discriminate the depth to favor draw over or under the neighboring dirt tiles, but Crash Site looked really weird when we drew under, and in Area 51, it's a bit extreme corner case. We could do subdivisions but I don't think it's worth it.
LMK what you think?

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,6 +9,7 @@
 - fixed not restoring Lara's back weapon mesh between levels when "remember guns" is enabled and a rifle-type weapon is equipped at level end
 - fixed a missing footstep sound when Lara starts to sprint
 - fixed Lara's flare undraw animation being skippable on specific late draw frames (#1593)
+- fixed footprints rendering with an excessive Y offset
 - fixed water ripples triggering z-fighting with 0-click ground surfaces
 
 **TR3**:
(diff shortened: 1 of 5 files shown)


aredfan:
Thank you @rr-, the small water ripples LGTM.

The median and large water ripples can still have z-fighting with the ground.  I made a recording to show this occurring. 

[record.txt](https://github.com/user-attachments/files/25297139/record.txt)

lahm86 (approved):
> We could do subdivisions but I don't think it's worth it.

Agreed, it looks great as-is with this update imo.

rr-:
> Thank you @rr-, the small water ripples LGTM.
> 
> The median and large water ripples can still have z-fighting with the ground. I made a recording to show this occurring.
> 
> [record.txt](https://github.com/user-attachments/files/25297139/record.txt)

Fixed :)

aredfan (approved):
> Fixed :)

The water ripples by 0-click surfaces is looking great. Thank you. 👍

PR #4875, opened 2026-02-13
docs: update game-flow schema

lahm86:
#### Checklist

#### Description

This aligns the game-flow schema for each game into a single file, and in turn targets TR3 for game-flow validation. It also updates the level object properties to mark Lara's outfit as a required field. The title level already has its own distinct setup in the schema so is unaffected, plus the added conditions ensure levels of type current and dummy are not affected.


aredfan (approved):

tools/shared/linting.py
```diff
@@ -134,11 +134,11 @@ def get_project_game_strings_path(
...
+    if not schema_path.exists():
+        return
+    schema = load_json5(schema_path)
     for project, paths in PROJECT_PATHS.items():
-        schema_path = paths.docs_dir / "gameflow.schema.json"
-        if not schema_path.exists():
-            continue
-        schema = load_json5(schema_path)
         validator_cls = jsonschema.validators.validator_for(schema)
         validator = validator_cls(schema=schema)
```
rr-: We can move this near `schema` creation

tools/shared/paths.py
```diff
@@ -16,6 +16,7 @@ class BasePaths:
     data_dir = Path
     shipped_data_dir = Path
     src_dir = Path
+    docs_dir = Path
```
rr-: This is my mistake, but all ` = ` should have been `:` in this class.

rr- (approved):

tools/shared/linting.py
lahm86: Thanks, updated

tools/shared/paths.py
lahm86: Updated 😌

rr- (approved):

PR #4903, opened 2026-02-16
objects/carcass: add carcass control

lahm86:
#### Checklist

#### Description

This adds the hanging raptor carcass control used in Crash Site. The piranhas in the area where it's used will divert their attention to it when it falls - all seems to be in order there, so thanks @rr- for previous implementation of that. This object is only used in Crash Site so the original level sequence check was redundant, and I also dropped the OG global single carcass item number in favour of allowing piranhas to find the target on their own; this is similar to the small dinos.

The one remaining hard-coded aspect is the item's roll when it hits the riverbed, that's currently tied to the specific slope in Crash Site. Left a `TODO`.


src/trx/game/objects/general/shoal.c
```diff
@@ -274,8 +279,10 @@ static void M_Control(const int16_t item_num)
 
     int32_t piranha_attack = 0;
     if (item->object_id == O_PIRAHNAS && lara_item != nullptr) {
-        if (g_TRVersion == 3 && GF_BadGetLevelNum() == 6
-            && m_CarcassItemNum != NO_ITEM) {
+        if (m_CarcassItemNum == NO_ITEM) {
+            M_FindCarcass(item);
+        }
+        if (m_CarcassItemNum != NO_ITEM) {
```
rr-: For this to become usable by multiple schools of fish independently, I'd suggest moving it to `M_PRIV`.

lahm86: Fixed

rr- (approved):

aredfan:
- [ ] In OG, when the carcass falls into the water, it makes a really tall splash. In TRX, the splash is quite small instead.

> <details><summary>OG screenshot</summary>
> 
> OG
> 
> </details>
> 
> [OG SG.zip](https://github.com/user-attachments/files/25350153/OG.SG.zip)
> 
> [record.txt](https://github.com/user-attachments/files/25350155/record.txt)

lahm86:
> * [ ]  In OG, when the carcass falls into the water, it makes a really tall splash. In TRX, the splash is quite small instead.

Thanks, should be fixed. I was throttling the fall speed too early effectively, so the splash speed was being determined by the  item's underwater speed rather than just before hitting the water.

aredfan (approved):

PR #4906, opened 2026-02-16
camera: fix interrupts from heavy triggers

### LINKED ISSUE #4902: Boulder kills cameras in TR3
Is it possible to change the behaviour of boulders? In tr1/2 boulder on trigger sector does not break the camera so you can user it with special settings with boulders and time more cameras, unless it's stops on the trigger sector. 
In in tr3 and TR3X if the camera touch a sector with trigger the camera stops and you see Lara again.

rr-:
#### Checklist

#### Description

This PR fixes camera getting interrupted by heavy triggers, which can happen in scenarios where the builder uses Boulders to achieve precisely timed triggers.

#### Implementation details

Heavy triggers with no `TO_CAMERA` / `TO_TARGET` were coming through with `status == TARGET_UNKNOWN` and were incorrectly forcing `g_Camera.timer = -1`.
The fix is in `Camera_RefreshFromTrigger` and narrows the timer reset guard so it only applies when the trigger actually produced camera-related status (`status != TARGET_UNKNOWN`). This prevents non-camera heavy refreshes from canceling an active camera timer while preserving existing invalidation behavior for true camera trigger paths.

#### Testing

- [x] Start the test level shared on Discord (same as with the other boulder issue), run `/tp push` three times, press the button a few times.
  Expected outcome: the cameras do not get randomly interrupted and last mostly the same duration.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -14,6 +14,7 @@
 - fixed UI bar scale option not updating the padding and borders (regression from 1.2)
 - fixed Blade stopping in the wrong position when anti-triggered (#4894)
 - fixed very distant Boulders causing camera shake (similar to the Tihocan crocodile targeting bug)
+- fixed heavy triggers with no `TO_TARGET` / `TO_CAMERA` resetting cameras
 
 **TR2**:
 - fixed flickering switches and spike ceilings in Temple of Xian and Floating Islands (#4874)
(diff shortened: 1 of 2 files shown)


lahm86 (approved):
Thanks, LGTM 😌

aredfan (approved):

PR #4908, opened 2026-02-16
objects/smashable: add smashable wall

lahm86:
#### Checklist

#### Description

This is a draft for now to get some thoughts on the approach.

- The smashable wall in Crash Site can only be broken by the mounted gun
- Firing a regular rocket at it will explode the rocket on the wall instead of passing through
- Firing regular guns has no effect

Rather than passing flags to the rocket item produced by the mounted gun, I've created a clone object to determine that it's a "heavier" rocket. Not sure I like this approach. We could inject a proper object as an alternative, but LMK if you can think of other alternatives.

The guarding in `Gun_SmashItems` has grown quite complex; I've tried to streamline it but if anything, I think the naming of the object vars is a bit confusing.

On the plus, this does allow us to allocate the fourth smashable as the regular shutters in TR3 (e.g. Aldwych), which use a different SFX from TR2 - so the slot count has matched up nicely.


docs/CHANGELOG.md
```diff
@@ -2,7 +2,10 @@
 - added the ability to freely rotate examinable items
 - added a color editor dialog for fog and water colors in Graphic Options → Visuals
 - added `O_SWITCH_TYPE_WHEEL`, which is similar to `O_SWITCH_TYPE_AIRLOCK` but can be used more than once
+- added `O_SMASH_OBJECT_3`, which can only be broken with triggers or the Crash Site gun
+- added `O_SMASH_OBJECT_4`, which behaves like `O_SMASH_OBJECT_1` but uses `SFX_SHUTTERS_BREAK`
 - added support for using more sound slots than originally possible in custom levels (#3898)
+- changed `O_WINDOW_1` and `O_WINDOW_2` to `O_SMASH_OBJECT_1` and `O_SMASH_OBJECT_2` respectively
```
rr-, line 8: This needs a migration note

src/trx/game/items/utils.c
```diff
@@ -79,7 +79,7 @@ int32_t Item_Explode(
     Matrix_Rot16(best_frame->mesh_rots[0]);
 
     const int32_t speed_shift = item->object_id == O_TORSO ? 7 : 8;
-    const bool is_tr3 = g_TRVersion == 3;
+    const bool is_tr3 = g_TRVersion == 3 && obj->intelligent; // TODO: improve
```
rr-: Can we add a more descriptive comment?

rr- (approved):
I'll think about it but it looks good to me and I find this solution elegant.

src/trx/game/items/utils.c
lahm86: Thanks for the reminder. This was temporary to avoid the windows burning when they break. I'll add a proper smash object detection here.

rr- (approved):

rr-:
Yeah I couldn't think of anything better short of creating yet another callback on all objects, which is crowding the design. Maybe it'll come to us later.

aredfan (approved):

PR #4913, opened 2026-02-17
lara/common: prevent interact flag on underwater switches

### LINKED ISSUE #4912: TR2 (V1.2.2) - Problem: Lara won't move after climbing out of water
Hi,

now the problem occurred the second time I send you a report:

After finishing some underwater lever actions in the 'Temple of Xian' Lara climbs out on of the water a ledge to collect a medpack before continuing - but she's stuck and refuses to move in any direction.

The save file and the screenshot are attached below. FYI: The first occurence of this problem was at 'The Deck' after climbing out of the water in the stern room (opening stern key door - jump in the water - pull lever - climbing out - stucked)

Thank you

Greetings 
kara

Image

[save_tr2_00.zip](https://github.com/user-attachments/files/25371643/save_tr2_00.zip)

lahm86:
Resolves #4912.

#### Checklist

#### Description

This prevents Lara's item interact flag being set on underwater switches, which do not use controlled collision but share the same move position function as land switches.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -23,6 +23,7 @@
 - fixed flickering switches and spike ceilings in Temple of Xian and Floating Islands (#4874)
 - fixed Airlock door handles not getting drawn from certain angles (#4886, regression from 1.0)
 - fixed loading screens showing before playing FMVs on most levels
+- fixed Lara not being able to move after exiting water, having used an underwater lever with the animated interactions setting enabled (#4912, regression from 1.0)
 - removed the requirement to use `main.sfx` in custom levels (#3898)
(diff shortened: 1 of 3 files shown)


rr- (approved):
I think the tomb5 safety check is good to include too, as we've hit this flag several times since its introduction with similar reported bugs. (With comments of course)

lahm86:
> I think the tomb5 safety check is good to include too, as we've hit this flag several times since its introduction with similar reported bugs. (With comments of course)

It turns out we had the check already, but it was version gated to TR1 only. Fixed.

aredfan (approved):

PR #4921, opened 2026-02-18
objects/crocodile: add TR3 support

lahm86:
#### Checklist

#### Description

Most of this is a refactor, the final commit contains the TR3 differences, which are minimal. In theory, we could support switching to land crocs, similar to TR1, but I've left it gated for now. The animations, state changes etc are identical for the water type, so I've made the "fix alligator AI" option visible in TR3. Their bites should draw blood on the correct frame.

Some testing in The Cistern would also be good, flooding and draining with the crocs both alive and dead.


rr- (approved):
Thank you!

aredfan (approved):

PR #4935, opened 2026-02-19
objects/mercenary: add South Pacific mercenary

lahm86:
#### Checklist

#### Description

This adds the mercenary in Crash Site, who is friendly towards Lara and will target any creature. A few remaining visual TODOs below, but this opens the PR for testing the main control logic/behaviour for now.

TODOs:
- [ ] Check gun visuals - follow-up task
- [x] Blood when shooting other creatures


aredfan:
- [x] There's a couple of spots by the plane's tail where the soldiers can get stuck when pathing.  This doesn't happen in OG.

> [record.txt](https://github.com/user-attachments/files/25426084/record.txt)

- [x] If Lara has 2 mercenaries nearby and she attacks one of them, when the other shoots at Lara, he appears to have no gun fire or gun lighting.

> [record.txt](https://github.com/user-attachments/files/25426088/record.txt)

lahm86:
> If Lara has 2 mercenaries nearby and she attacks one of them, when the other shoots at Lara, he appears to have no gun fire or gun lighting.

This should be resolved now. I've also added blood effects when they hit raptors (or any other creature) per OG. This will also apply to bandits in TR2 shooting monks, WDYT? We could version gate it, but it seems like a sensible change.

Still looking at the pathing issue around the plane.

aredfan:
> This should be resolved now.

Thank you. 👍 

> I've also added blood effects when they hit raptors (or any other creature) per OG. This will also apply to bandits in TR2 shooting monks, WDYT? We could version gate it, but it seems like a sensible change.
> 
> Still looking at the pathing issue around the plane.

Letting the monks bleed shows that they're taking hits. I agree that it'd be a sensible change.

lahm86:
Struggling a bit with the pathing issue. All I can see is that on the data side, the pathfinding box is marked as `BLOCKABLE`, which means if there were a door or pushblock there it would get blocked off from pathfinding. The soldier's block mask is `BLOCKED` meaning he will happily use boxes that are `BLOCKABLE` but not ones that are `BLOCKED`. If I change the data to make this box `BLOCKED`, he ignores it.
I'm struggling to see any logic that could prevent this scenario happening in OG. Super odd.

lahm86:
Found it. TR3 marks all blockable boxes as being blocked when the level loads. Added this, but version-gated as it will change behaviour significantly.

rr- (approved):

aredfan (approved):

PR #4994, opened 2026-02-27
walkable: fix triangle portal sector mismatches

lahm86:
#### Checklist

#### Description

This resolves collision issues when trapdoors, bridges etc are placed over a sector that contains a triangle portal. The original implementation allocated a walkable to a single sector at a time; this update will test four points around the item's mid-point and allocate the walkable to each unique pit sector it finds.

The first commit contains the basics for this, but it leads to an arena memory leak when moving pushblocks. In the original approach a moving walkable was removed from its old sector and added to the new, and the previously allocated `WALKABLE` was singularly re-used. Because now a source and destination may not necessarily have the same number of required pit sectors, we need to allocate up-front the total number of potential sectors, cache these, then use them when the items move. That's the gist of the second commit.

Not an ideal solution. The alternative I think would be changing the `WALKABLE` linked list in `SECTOR` to a direct `VECTOR`, but that moves into memory management in level data rather than allowing game buf to handle it. LMK what you think.

I also fixed Lara's footprints sometimes showing up when walking on bridges e.g. RX-Tech room 1.

Will provide test levels.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -114,6 +114,7 @@
 - fixed Tony briefly appearing for a single frame when loading a save after his death
 - fixed Lara sometimes getting stuck when crawling backwards off a tilted ledge (#4956)
 - fixed the Tribe Pipeman sometimes not being able to aim darts at Lara correctly (Gameplay → Fixes → Fix Pipeman aim)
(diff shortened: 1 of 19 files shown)


lahm86:
Build updated, thanks to dash for a leaner solution 😌

rr- (approved):

aredfan (approved):

PR #5008, opened 2026-03-01
mercenary: add gun effects

rr-:
#### Checklist

#### Description

This is a proposal on how we could do mercenary gun flash, smoke, and shell case effects. I had to redo the OG code into something very different to avoid introducing bad patterns. Please LMK if this approach looks authentic enough.

Note: the gun flashes don't move with joints, they're spawned and stay this way until they disappear, which happens after 2-3 frames.

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -235,6 +235,7 @@ common_sources = [
   'trx/game/fmv.c',
   'trx/game/fx/explosion_ring.c',
   'trx/game/fx/footprint.c',
+  'trx/game/fx/gun_flash.c',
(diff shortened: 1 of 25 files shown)


rr-:
I noticed the rotations are completely screwed up. Investigating.

rr-:
I rewrote this again to not spawn an `EFFECT` and instead use a dedicated `fx/` module.

aredfan (approved):
It's working great. 😌

lahm86 (approved):
This looks brilliant. Thank you 😌

PR #5013, opened 2026-03-02
savegame: add real quick saves

### LINKED ISSUE #1897: Feature request: option to disable passport animation for quick save/load like TR4+
This feature was requested by a member on the Lost Artefacts server earlier this year. Logging it so it doesn't get lost.

**Update:** This is a video of how saving/loading looks in TR4.  The quick save/load menu is instant without the delay which accompanies the passport animations.

https://github.com/user-attachments/assets/ae21b830-1bf4-4b58-ad95-12ac40fe2271

rr-:
#### Checklist

- [x] I have read the coding conventions (https://github.com/LostArtefacts/TRX/blob/develop/docs/CONTRIBUTING.md#coding-conventions)

#### Description

This PR introduces real quick save/quick load support with dedicated slot handling, keybinds, UI integration, and console command aliases.
This feature is completely opt-in and pretty much invisible unless the player goes out of their way to use it (requires new bindings; quick save slots are only visible if there's an actual quick save bound to it.)

#### Implementation details

The savegame system now supports two slot pools (normal + quick) through one unified slot reference API (`SAVEGAME_SLOT_REF`) used across save/load, binding, story checks, sequencer flow, and passport/inventory actions. Quick saves are round-robin, use standard counters, and update bound-slot behavior the same way as normal saves (including death counter updates). Quick slot metadata is probeable from file headers without BSON parsing.

Input roles now distinguish Save/Load from Quick Save/Quick Load (`INPUT_ROLE_SAVE`, `INPUT_ROLE_LOAD` remain; new `INPUT_ROLE_QUICK_SAVE`, `INPUT_ROLE_QUICK_LOAD` added). Default keyboard bindings for quick roles are unbound, while existing F5/F6 stay on normal save/load. In-game OSD messages and control labels were updated accordingly, and new strings/config options were added (including `gameplay.maximum_quick_save_slots

UI and flow were updated to work with unified `SAVEGAME_SLOT_REF` across the board. Save slot dialogs now render quick slots at the top, style them distinctly (QS [Qn]), and only show quick rows when those quick saves actually exist meaning no visual changes happen until the player starts using this feature. The files are stored in a separate pattern, but that pattern is derived from the existing `savegame_file_fmt` field by prepending the counter with `"q"` rather than requiring the builders to ship a new field.

Console commands were extended so `/save` and `/load` now also accept quick forms:

- `/save quick {slot}` and `/save q{slot}`
- `/load quick {slot}` and `/load q{slot}`

#### Questions to the team

1. ~~Should we show quick saves in the Play Previous Levels/Story So Far… save slot selection? IMO yes, currently not shown.~~ Implemented
2. ~~For 2+ quick save slots, should we sort them by the save counter value visually, so that the newest saves always appear at the top? IMO yes but we need to be careful to sync the sorting to `/save` and `/load` commands, and the game flow parameter emission.~~ Implemented

#### Testing

- [x] Set `gameplay.maximum_quick_save_slots` to 0, trigger quick save/load input.
  Expected outcome: quick actions fail gracefully with OSD errors, no crash, no passport flow.
- [x] Set quick slots to 2, bind quick input roles, press quick save three times in-level.
  Expected outcome: saves rotate Q1 -> Q2 -> Q1 (round-robin), with quick-save OSD logs.
- [x] Bind a quick slot, die/reload/continue, then quick load.
  Expected outcome: bound slot behavior works the same for quick and normal saves, including death-counter updates.
- [x] Open Save Game and Load Game dialogs with some quick saves present.
  Expected outcome: quick slots appear at top, visually marked as quick, only present quick rows are shown.
- [x] Use console commands: `/save 1`, `/load 1`, `/save quick 1`, `/save q1`, `/load quick 1`, `/load q1`.
  Expected outcome: normal and quick command variants both work and target the correct pool. Same slots across different pools do not overwrite each other.
- [x] Enter Play Previous Levels flow from passport with mixed slot states.
  Expected outcome: no regressions; quick saves shown too and don't collide with normal save pool
- [x] Enter Story So Far… flow from passport with mixed slot states.
  Expected outcome: no regressions; quick saves shown too and don't collide with normal save pool
- [x] Enter Play Any Level… flow from passport. (`/set flow.play_any_level 1`)
  Expected outcome: no regressions.
- [x] Hunt for other passport / select level / select save slot / commands / slot binding regressions.
- [x] Hunt for bugs where choosing a quicksave X actually interacts with quicksave Y (in logs, behavior, …)


lahm86 (approved):
This is fantastic and I'm sure will be very well-received by players ☺️ 

> Should we show quick saves in the Play Previous Levels/Story So Far… save slot selection? IMO yes, currently not shown.
For 2+ quick save slots, should we sort them by the save counter value visually, so that the newest saves always appear at the top? IMO yes but we need to be careful to sync the sorting to /save and /load commands, and the game flow parameter emission.

I agree on both counts here, this will be a nice addition.

rr-:
Thanks – I implemented both.

aredfan (approved):
Amazing, absolutely top-tier work, thank you. 😌

PR #5014, opened 2026-03-02
objects/dragon: add independent dragon mode

### LINKED ISSUE #5011: [Possible Feature?] Dragon as a normal boss
The Dragon in TR2 is always tied to the "Bartoli" object [ID 40], where you need to trigger it and get within 3 tiles for the Dragon to appear. Then pulling the dagger is the only way of killing it as you know.

Would it be possible to "detach" the Dragon from this object, so it can be triggered at will and act like a normal boss?
The dragon bubble effect is already a nice way of spawning it into action so it's not sudden. :)

lahm86:
Resolves #5011.

#### Checklist

#### Description

This allows builders to use dragons with no ties to Bartoli by placing a back item and triggering it like any other enemy. It has no dagger concept so has only one phase, and does not turn to bones. It can still activate heavy triggers; for this, the handling is moved out of Lara's extra state and into the dragon, and as a result the dragon will "dissolve" for a few frames more so that the timing matches up. I think the result is fine, and it actually avoids a jump in the skin disappearing (difficult to see any difference anyway with the camera angle in OG).

Most of this is moving some things around and some const refactors; the final commit has the main change.

Will share a test level, but a sanity check of the OG dragon would be good, including save/load while alive and dead.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -6,6 +6,9 @@
 - added `O_SMASH_OBJECT_3`, which can only be broken with triggers or the Crash Site gun
 - added `O_SMASH_OBJECT_4`, which behaves like `O_SMASH_OBJECT_1` but uses `SFX_SHUTTERS_BREAK`
 - added `O_TREX_ALPHA`, which can target raptors and be distracted by flares
+- added the ability to trigger dragons independently of Bartoli in custom levels (#5011)
+  - place an `O_DRAGON_BACK` item in the editor and trigger it normally
+  - the dragon will spawn immediately when triggered and will be one-phase, so no dagger needs to be pulled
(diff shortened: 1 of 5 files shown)


rr- (approved):
Thank you for this :)

aredfan (approved):

aredfan:
**Issues:**

- [x] If Lara kills the independent dragon, it's possible to have an overlap of flesh and bones

1. launch the test level
2. trigger the independent dragon
3. `/kill dragon`
4. after the doors opens, save and reload

- [x] It's possible for the independent dragon to enter the semi death state

1. launch the test level
2. trigger the independent dragon
3. save and reload
4. `/kill dragon`

lahm86:
Thanks, @aredfan. Should be fixed, but it'll need new saves to be made for these scenarios.

aredfan (approved):

PR #5031, opened 2026-03-06
game-strings: rework settings storage

rr-:
#### Checklist

#### Description

This changes the way we store the translations for the setting names and descriptions, by moving it from
```json
{
    "game_strings": {
        "SETTING_UI_PICKUPS_OVERLAY": "Pickups overlay",
        "SETTING_UI_PICKUPS_OVERLAY_DESCRIPTION": "Shows items in the bottom-right corner when Lara picks something up.",
        "SETTING_UI_SCROLL_WRAPAROUND": "Scroll wrap",
        "SETTING_UI_SCROLL_WRAPAROUND_DESCRIPTION": "Lets directional navigation in menus loop around."
    }
}
```
to
```json
{
    "settings": {
        "ui.show_pickups_overlay": {
            "title": "Pickups overlay",
            "description": "Shows items in the bottom-right corner when Lara picks something up."
        },
        "ui.enable_wraparound": {
            "title": "Scroll wrap",
            "description": "Lets directional navigation in menus loop around."
        }
    }
}
```

The reason for this change is that we used to map the game strings by hand in the UI setting dialog tabs definitions. Now they are automatically derived from the setting internal names, which lets us avoid the manual mapping, and easily ask for the option canonical name and description via a much less specialized API. This in turn lets us reuse these labels in other places (config presets, to be landed soon).
The added benefit is that the strings names (barring the additional prefix) are now identical to what we have in `cfg/TRX.json5` and `"enforced_settings"`/`"hidden_settings"` in `gameflow.json5`, reducing cognitive complexity too.

The big increase in the lines of code introduced by this pull request comes from each setting requiring 4 lines instead of 2 due to using nested objects for the title and description fields.

In follow-up PRs I would like to explore:

- simplifying the reader code for game string tables to reuse the `M_LoadNestedGameStrings` for game strings
- maybe try to hook this approach into object names as well, to reduce special handling
- give the same treatment to enums, so that we can easily ask for human-readable name of any enum without hardcoded mappings
- likely lowercase all game string defines
- possibly use nested namespaces for commands

But since this PR on its own is pretty massive, I didn't want to pollute it further by doing all of that at once.

#### Testing

- [x] Skim all setting dialogs and tags
  Expected outcome: no empty labels or values text
- [x] Randomly check several options' descriptions
  Expected outcome: no missing descriptions


rr-:
Semi-important note: in case we do have a manually defined setting JSON key such as:
```c
X_CFG_DYNAMIC_ENUM_EX("ui.enemy_healthbar_color", ui.enemy_health_bar.color, "grey")
```
the game strings reuse that JSON path, eg `"ui.enemy_healthbar_color"`, rather than `"ui.enemy_health_bar.color"`.

aredfan (approved):

lahm86 (approved):

PR #5042, opened 2026-03-06
objects/train: add train control

lahm86:
#### Checklist

#### Description

This adds the Aldwych subway train and support for its different forms in The Lost Artefact. The train's speed is determined by its animation, as regular item animation is not used. This allows removing hard-coded aspects.

LMK if the fixed camera behaviour is reasonable. It avoids any changes in the camera implementations, so we just slot in to existing fixed behaviour. But I'm not sold on this fully.

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/223


rr- (approved):
This looks great to me (only tested Aldwych, though).

aredfan (approved):

PR #5063, opened 2026-03-08
game_flow/inventory: convert reward guns to ammo

lahm86:
#### Checklist

#### Description

This converts secret reward guns to the equivalent ammo if Lara already has the gun. This is in line with TR1/3 secret mode, where if the item were a physical pickup it would be converted by the global replace logic.

- `/play 4`
- use item cheat
- collect all secrets
- Lara should receive 5 clips instead of 4

### DIFF
--- a/src/trx/game/game_flow/inventory.c
+++ b/src/trx/game/game_flow/inventory.c
@@ -167,8 +167,12 @@ static void M_ModifyInventory_GunOrAmmo(
 
     if (Inv_RequestItem(gun_object_id)) {
         if (type == GF_INV_SECRET) {
-            ammo_info->ammo += ammo_qty * m_SecretInvItems[ammo_object_id];
-            for (int32_t i = 0; i < m_SecretInvItems[ammo_object_id]; i++) {
+            // Convert already collected guns into ammo to maintain stats
+            // accuracy.
+            const int32_t count = m_SecretInvItems[ammo_object_id]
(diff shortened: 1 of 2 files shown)


src/trx/game/game_flow/inventory.c
```diff
@@ -167,8 +167,12 @@ static void M_ModifyInventory_GunOrAmmo(
...
     if (Inv_RequestItem(gun_object_id)) {
         if (type == GF_INV_SECRET) {
-            ammo_info->ammo += ammo_qty * m_SecretInvItems[ammo_object_id];
-            for (int32_t i = 0; i < m_SecretInvItems[ammo_object_id]; i++) {
+            // Convert already collected guns into ammo to maintain stats
+            // accuracy.
+            const int32_t count = m_SecretInvItems[ammo_object_id]
+                + m_SecretInvItems[gun_object_id];
+            ammo_info->ammo += ammo_qty * count;
+            for (int32_t i = 0; i < count; i++) {
```
rr-, line 175: Doesn't this cause the bug to happen in the other direction: if I _don't_ have Uzis, I'll now have Uzis _and_ an additional clip?

lahm86: The existing guard on line 168 prevents this - so the other conditional branch will play out in that scenario (line 185 for Opera specifically) 😌

rr- (approved):

aredfan (approved):

PR #5069, opened 2026-03-09
objects/upv: add

rr-:
#### Checklist

#### Description

This is a draft that adds UPV support.
Currently known issues:
- [ ] Wake FX not working
- [x] Harpoons only firing once, then never again


lahm86:
TIL that UPVs need to be triggered, so when we were TP'ing to them, they weren't activated - so `M_Control` wasn't playing, which meant harpoons and the wake effect weren't kicking in. Technically an OG bug as you can reach item 117 in Lud's Gate with a big skip, which means you also skip its triggers. I've added a check in `M_GetOn` to ensure the item is activated when Lara's getting on, which makes sense IMO.

Lara's air bar should be responding too and the SFX playing. I added the save/load control and did a little semantic adjustment, hopefully all OK. Didn't touch the changelog or object support JSON.

aredfan:
- [x] While riding the UPV, Lara is able to breath when brushing against angled ceilings, it can happen in OG too but is mostly limited to the top of the diamond-shaped tunnels. One of these tunnels is shown at the end of the recording attached below.

> [record.txt](https://github.com/user-attachments/files/25849924/record.txt)

rr-:
> TIL that UPVs need to be triggered, so when we were TP'ing to them, they weren't activated - so `M_Control` wasn't playing, which meant harpoons and the wake effect weren't kicking in. Technically an OG bug as you can reach item 117 in Lud's Gate with a big skip, which means you also skip its triggers. I've added a check in `M_GetOn` to ensure the item is activated when Lara's getting on, which makes sense IMO.

Thank you!! This is the part that was driving me insane, it never occurred to me to check for real triggers when the vehicle was perfectly mountable.

docs/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 - fixed guns as secret rewards not being converted to the equivalent ammo if Lara already has the gun
 
 **TR3**:
+- added UPV control
```
lahm86, line 15: We could add a mention for the hair fix too as that's a nice improvement 😌

lahm86 (approved):
LGTM 👍

docs/CHANGELOG.md
rr-: Added + squashed all the commits into one.

lahm86 (approved):

rr-:
I have removed unused includes, removed the dead wake FX code, and simplified the bitfield flags to dedicated bool members in `M_PRIV`.

aredfan:
I noticed 2 bugs (both could be the same thing) when comparing to OG.

- [x] When Lara disengages from the UPV, there is a shift in her position. It can happen in OG as well but not as often. In the OG video below, Lara disengage from the UPV elegantly after the first time which is a shift.

> <details><summary>OG video</summary>
> 
> https://github.com/user-attachments/assets/b5d1a94a-52d4-4b80-948c-4d807933d62a
> 
> </details>
>
> [record.txt](https://github.com/user-attachments/files/25883581/record.txt)

- [x] I also tried having Lara disengage from the UPV in various places and she does appear to get nudged in some places.  I can confirm this doesn't happen in OG.  Recording attached below.

> [record2.txt](https://github.com/user-attachments/files/25883704/record2.txt)

lahm86 (approved):

aredfan (approved):

PR #5071, opened 2026-03-09
Add WebGL/Emscripten build support

goldfish-g:
#### Checklist

#### Description

Full WebGL/Emscripten port of TRX, allowing TR1 and TR2 to be played directly in the browser. Both games compile, run, and are fully playable on desktop and mobile browsers.

### What this adds

- **Meson cross-compilation** with Emscripten via `tools/shared/emscripten/emscripten_cross.ini`
- **GL ES 3.0 compatibility layer** (`src/trx/gl/gl_webgl_compat.h/.c`) — shims for desktop-only GL functions (`glPolygonMode`, `glMapBuffer`, `glDrawElementsBaseVertex`, etc.) so the existing renderer works unchanged on WebGL 2
- **ASYNCIFY integration** — yields to the browser event loop from synchronous C code, enabling the existing game loop to run without restructuring
- **IndexedDB persistence** (IDBFS) — saves and config persist across sessions in the browser
- **FMV streaming** — cutscenes are fetched on demand via HTTP rather than preloaded, keeping initial load times small
- **FFmpeg in WASM** — full audio/video decoding including `.fmv` remastered format support
- **Touch controls overlay** — on-screen D-pad and action buttons for mobile, auto-enabled on first touch device visit
- **PWA support** — installable web app with manifest, service worker for offline play, and app icons
- **Game data upload UI** — users can upload their own game files (including remastered content) at runtime when data isn't bundled
- **Docker build** — reproducible builds via `tools/shared/docker/game-webgl/Dockerfile`

### Desktop impact

Minimal. The port adds small `#ifdef EMSCRIPTEN_BUILD` blocks (typically 1–10 lines) in ~30 source files. These are well-contained platform-specific branches — no behavioral changes on desktop. Key areas:
- `src/trx/gl/` — GL version selection, shader `#version` injection, pixel format (RGBA vs BGRA)
- `src/trx/game/shell/flow.c` — ASYNCIFY yield points, IDBFS init, touch controls setup
- `src/trx/av/video.c`, `src/trx/game/fmv.c` — surface format selection, sleep method
- `src/trx/core/strings/common.c` — POSIX `<regex.h>` fallback (PCRE2 unavailable under Emscripten)

### How to build and test

**WebGL (via Docker):**
```bash
# From the TRX repo root:
./tools/build_webgl.sh tr1 release
./tools/build_webgl.sh tr2 release
```

Or use the Docker image directly:
```bash
docker build . -f tools/shared/docker/game-webgl/Dockerfile -t trx-webgl
docker run --rm --user "$(id -u):$(id -g)" -v "$(pwd):/app" trx-webgl \
    build --target release --game tr1
```

**Serving locally** (COOP/COEP headers required for SharedArrayBuffer/pthreads):

Save this as `serve_coop.py` and run `python3 serve_coop.py 8080 build/webgl-tr1`:
```python
#!/usr/bin/env python3
import sys
from functools import partial
from http.server import HTTPServer, SimpleHTTPRequestHandler

class COOPCOEPHandler(SimpleHTTPRequestHandler):
    def end_headers(self):
        self.send_header("Cross-Origin-Opener-Policy", "same-origin")
        self.send_header("Cross-Origin-Embedder-Policy", "require-corp")
        super().end_headers()

if __name__ == "__main__":
    port = int(sys.argv[1])
    directory = sys.argv[2]
    handler = partial(COOPCOEPHandler, directory=directory)
    HTTPServer(("0.0.0.0", port), handler).serve_forever()
```
Then open `http://localhost:8080/TRX.html`.

<details>
<summary><b>Full deploy script</b> (builds both games, serves on localhost)</summary>

Save as `deploy.sh` in the **parent directory** of the TRX repo, alongside a `serve_coop.py`:

```bash
#!/usr/bin/env bash
set -euo pipefail
SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
TRX_DIR="$SCRIPT_DIR/TRX"
DEPLOY_BASE="$SCRIPT_DIR/deployments"
usage() {
    echo "Usage: $0 <test|prod> [--game-data|--no-game-data]"
    echo "  test  - build & serve on ports 8081 (tr1) / 8082 (tr2)"
    echo "  prod  - build & serve on ports 9081 (tr1) / 9082 (tr2)"
    echo "  --game-data      Bundle game data in the build"
    echo "  --no-game-data   Exclude game data (users upload at runtime)"
    echo "  Defaults: test bundles game data, prod does not."
    exit 1
}
[[ $# -lt 1 ]] && usage
ENV="$1"; shift
case "$ENV" in
    test) PORT_TR1=8081; PORT_TR2=8082; GAME_DATA_DEFAULT="yes" ;;
    prod) PORT_TR1=9081; PORT_TR2=9082; GAME_DATA_DEFAULT="no" ;;
    *) usage ;;
esac
GAME_DATA_OPT="$GAME_DATA_DEFAULT"
while [[ $# -gt 0 ]]; do
    case "$1" in
        --game-data)    GAME_DATA_OPT="yes"; shift ;;
        --no-game-data) GAME_DATA_OPT="no";  shift ;;
        *)              echo "Unknown option: $1"; usage ;;
    esac
done
DEPLOY_DIR="$DEPLOY_BASE/$ENV"
DOCKER_IMAGE="rrdash/trx-webgl"
DOCKER_TAG="20251017.rev1"
DOCKER_CONTEXT_HASH=$(cat \
    "$TRX_DIR/tools/shared/docker/game-webgl/Dockerfile" \
    "$TRX_DIR/tools/ffmpeg_flags_webgl.txt" \
    | sha256sum | cut -c1-12)
DOCKER_FULL_TAG="${DOCKER_TAG}-${DOCKER_CONTEXT_HASH}"
if ! docker images --format '{{.Repository}}:{{.Tag}}' | grep -q "^${DOCKER_IMAGE}:${DOCKER_FULL_TAG}$"; then
    echo "=== Building Docker image (tag: ${DOCKER_FULL_TAG}) ==="
    docker build "$TRX_DIR" \
        -f "$TRX_DIR/tools/shared/docker/game-webgl/Dockerfile" \
        -t "${DOCKER_IMAGE}:${DOCKER_FULL_TAG}"
fi
docker_build() {
    local game="$1"; local extra_args=()
    [[ "$GAME_DATA_OPT" == "no" ]] && extra_args+=(--no-game-data)
    [[ "$ENV" == "test" ]] && extra_args+=(--eruda)
    docker run --rm --user "$(id -u):$(id -g)" -v "$TRX_DIR:/app" \
        "${DOCKER_IMAGE}:${DOCKER_FULL_TAG}" \
        build --target release --game "$game" "${extra_args[@]+"${extra_args[@]}"}"
}
for game in tr1 tr2; do rm -rf "$TRX_DIR/build/webgl-${game}"; done
echo "=== Building TR1 ===" && docker_build tr1
echo "=== Building TR2 ===" && docker_build tr2
echo "=== Deploying to $DEPLOY_DIR ==="
for game in tr1 tr2; do
    dest="$DEPLOY_DIR/$game"; mkdir -p "$dest"
    rsync -a --delete "$TRX_DIR/build/webgl-${game}/" "$dest/"
done
for port in $PORT_TR1 $PORT_TR2; do fuser -k "$port/tcp" 2>/dev/null || true; done
SERVE_SCRIPT="$SCRIPT_DIR/serve_coop.py"
nohup python3 "$SERVE_SCRIPT" "$PORT_TR1" "$DEPLOY_DIR/tr1" >/dev/null 2>&1 &
echo "TR1: http://localhost:$PORT_TR1/TRX.html"
nohup python3 "$SERVE_SCRIPT" "$PORT_TR2" "$DEPLOY_DIR/tr2" >/dev/null 2>&1 &
echo "TR2: http://localhost:$PORT_TR2/TRX.html"
echo "=== Done ($ENV) ==="
```
</details>

### Architecture notes

- **ASYNCIFY**: The game's synchronous C main loop is preserved. Emscripten's ASYNCIFY automatically yields to the browser at sleep/delay points. See `docs/WEBGL_ASYNCIFY.md` for the full list of unwound functions.
- **GL compat**: A single `gl_webgl_compat.h` header remaps desktop-only GL calls to WebGL 2 equivalents (or no-ops where unsupported). This keeps the renderer code unified — no per-function ifdefs in rendering code.
- **Preload mappings**: Game config, shaders, and ship data are preloaded into the Emscripten virtual filesystem at build time via `--preload-file` in `src/meson.build`.
- **Platform ifdefs**: Kept intentionally minimal and inline. Each ifdef is a small (1–10 line) block at the point of use, with comments explaining why the platform-specific behavior is needed.


src/trx/config/common.c
```diff
@@ -135,7 +135,13 @@ bool Config_Write(void)
         .enforced_path = g_Config.enforced_path,
         .action = &Config_DumpToJSON,
     };
-    return ConfigFile_Write(&args);
+    const bool result = ConfigFile_Write(&args);
+#ifdef EMSCRIPTEN_BUILD
+    if (result) {
+        Shell_PersistConfigToIDBFS();
```
rr-: Can we do this from within `ConfigFile_Write`?

rr- (commented):
We still have a lot of `#if` – I'd like the new APIs to be specialized in `*_emscripten.c`, with common decls available in `*.h`, and a backwards-compatible `*_generic.c`.

Example:
Before - `video.c`:
```c
#if EMSCRIPTEN_BUILD
    .pixel_format = RGB
#else
    .pixel_format = BGR
#endif
```

After:
- `video.h`
    ```c
    #if EMSCRIPTEN_BUILD
        #inlcude <.../video_emscripten.h>
    #else
        #inlcude <.../video_generic.h>
    #endif
    ```
- `video_emscripten.h`
    ```c
    #define VIDEO_PIXEL_FORMAT RGB
    ```
- `video_generic.h`
    ```c
    #define VIDEO_PIXEL_FORMAT BGR
    ```
- `video.c`
    ```c
        .pixel_format = VIDEO_PIXEL_FORMAT
    ```

Slightly more boilerplate, but easier on the eyes when reading core modules, as we'll have less places where we do the fencing.
The same can be extended to implementation files (see for example the platform-specific implementations of `Log_Shutdown_Extra` across `log_linux.c` vs `log_windows.c` and how that's getting instrumented in `meson.build`).

src/trx/game/input/backends/keyboard.c
```diff
@@ -428,6 +428,10 @@ static bool M_CustomUpdate(INPUT_STATE *const result, const INPUT_LAYOUT layout)
     result->menu_confirm |= result->action;
     result->toggle_fullscreen =
         KEY_DOWN(SDL_SCANCODE_RETURN) && KEY_DOWN(SDL_SCANCODE_LALT);
+    if (!result->toggle_fullscreen
+        && (KEY_DOWN(SDL_SCANCODE_LALT) || KEY_DOWN(SDL_SCANCODE_RALT))) {
+        result->menu_back = true;
```
lahm86: This seems undesirable to hard-code inputs if users have these keys remapped. Is this needed for a particular scenario? `menu_back` can be freely assigned in the input dialog (System > Inventory).

src/trx/game/output/mesh_batcher/batcher.c
```diff
@@ -730,18 +730,36 @@ void MeshBatcher_Seal(MESH_BATCHER *const batcher)
...
-                &opaque_indices[bind->opaque_index_start],
-                Vector_GetData(bind->mesh->opaque_vertex_indices),
+                &opaque_indices[bind->opaque_index_start], src,
                 bind->opaque_index_count * sizeof(uint32_t));
+#endif
         }
 
         // Copy Blend Indices
         if (bind->blend_add_index_count > 0) {
+            const uint32_t *src =
```
lahm86: Nitpick: `*const  `

src/trx/game/phase/executor.c
```diff
@@ -247,9 +252,16 @@ GF_COMMAND PhaseExecutor_Run(PHASE *const phase)
         }
     }
 
+    WEBGL_LOG("[WEBGL] Entering main frame loop");
+    static int s_FrameLog = 0;
```
lahm86: We tend to opt for `int32_t` and static vars are placed near the top of the module rather than inside functions. The prefix should be `m_`.

src/trx/game/shell/flow.c
```diff
@@ -57,10 +60,13 @@ static void M_CreateGameWindow(void)
     if (m_Window != nullptr) {
         return; // Window persists across mod switches
     }
+
+    uint32_t window_flags = SDL_WINDOW_HIDDEN | SDL_WINDOW_RESIZABLE
```
lahm86: `const`

src/trx/game/fmv.c
```diff
@@ -130,19 +132,51 @@ static bool M_Play(const char *const file_name)
     Video_SetSurfaceUnlockFunc(video, M_UnlockSurface, nullptr);
     Video_SetSurfaceUploadFunc(video, M_UploadSurface, renderer_2d);
 
+    int32_t audio_id = Audio_Stream_CreateFromFile(file_name);
+    if (audio_id == AUDIO_NO_SOUND) {
+        // The video file may lack an audio stream (e.g. remastered .ogv).
+        // Try other FMV extensions to find a file that contains audio.
+        static const char *const fallback_exts[] = {
```
lahm86: Can `TRXPath_GuessExtension` perhaps be utilised here?

rr-:
Nit: we don't need `#include <stdbool.h>` as we use c2x.

src/trx/game/input/backends/keyboard.c
goldfish-g: This is to make touch controls more intuitive, and work like a controller. I want to keep the menu button as a back out button, but also makes sense to have the jump button back out.

https://github.com/user-attachments/assets/96c93e85-1164-407f-a7a4-6328beb424cb

What would be the better approach?

src/trx/game/fmv.c
goldfish-g: The issue is that TRXPath_GuessExtension returns the first existing file. But we need to try each candidate with Audio_Stream_CreateFromFile because a file might exist without an audio stream. So a direct replacement doesn't work cleanly.

Why I do this: The remaster ships the HD FMVs with no audio, and instead uses the audio from the original files, it's bizarre.

src/trx/game/input/backends/keyboard.c
rr-: The _best_ approach would be to add a separate backend for touch controls.
Instead of sending keyboard keys, it could emit custom Emscripten events that get consumed by this new backend?

src/trx/game/fmv.c
rr-: I would say this falls outside of the scope of the WebGL builds and deserves attention separately.

src/trx/platform/yield.h
```diff
@@ -0,0 +1,23 @@
...
+}
+#else
+    #include <SDL2/SDL_timer.h>
+static inline void Platform_Yield(unsigned int ms)
+{
+    if (ms > 0) {
+        SDL_Delay((Uint32)ms);
+    }
+}
+#endif
```
rr-: This module should be a responsibility of clock (`Clock_Delay` as a proxy to `SDL_Delay`), and implementations should go into `clock_emscripten` and `clock_generic`.

src/meson.options
```diff
@@ -1 +1,3 @@
 option('staticdeps', type: 'boolean', value: true, description: 'Try to build against static dependencies. default: true')
+option('game', type: 'combo', choices: ['tr1', 'tr2'], value: 'tr1', description: 'Which game to build for Emscripten (tr1 or tr2)')
```
rr-, line 2: Since you'd prefer merging everything at once, I think it makes more sense to wait for the .zip integration first and then build a single target, instead of doing this interim step.

Example how the hierarchy is supposed to look when using a single directory for all games (last develop):

https://files.bad.pink/serve/pub/single-tree.txt

rr- (commented):
> tools/shared/emscripten/include/uthash.h

This shouldn't be checked in, and if it has to be compiled statically, it should be made available by the Docker build process.

src/trx/game/input/backends/keyboard.c
goldfish-g: That makes a lot of sense, I made a new backend for touch controls, now it no longer "emulates" keyboard presses

src/trx/game/fmv.c
goldfish-g: Fair enough, maybe I'll open a PR for that in the future. Given that, I will change the priority to prefer the .fmv files over the .ogv files, at least for now.

src/meson.options
goldfish-g: Oh, I had no idea. This significantly changes what I had in mind for the "profile selector". How will you handle TRLE's? Well, I can just wait for this then, this is huge, and will change the way I handle things for sure 😇

goldfish-g:
> > tools/shared/emscripten/include/uthash.h
> 
> This shouldn't be checked in, and if it has to be compiled statically, it should be made available by the Docker build process.

That was a huge oversight on my part, thank you, I just removed 1.2k lines of code with that one 😅

Well, I think I went through all the feedback, thank you so much. We can await the zip integration, for sure, no rush. I'll keep rebasing in the meantime 😉

goldfish-g:
Rebased and linted.

Issue #5086, opened 2026-03-11
TRX-3: No guards and prisoners in 'High Security Compound'


kara20004:
Hi,

did I miss something or is it a bug: When starting HSC no guard enters the room and any other prison cell is empty!? 

I checked the game folders (audio, cuts, data, fmv and pix) but everything seems complete.

Attached you find 2 saves (last from 'Nevada desert' and first of 'HSC' - can you help me find the problem?

Thanks in advance
kara

[saves.zip](https://github.com/user-attachments/files/25898128/saves.zip)

ShadowXJaw724:
Nevada is not complete yet. Work is currently in progress to make TRX playable. Lara's Home, India and South Pacific are playable at the moment. Nevada, London and Antarctica and Expansions are in the works.

PR #5091, opened 2026-03-11
docs: add issue templates

rr-:
#### Checklist

#### Description

This builds on top of https://github.com/LostArtefacts/TRX/pull/5090 and replaces the freeform issue forms with structured ones, ensuring users include the key details we keep having to request manually.

Example how this works can be found in my fork:

https://github.com/rr-/TRX/issues
https://github.com/rr-/TRX/issues/11

It also updates the website to just redirect to GH instead of repeating the guidelines in yet another place.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -34,6 +34,12 @@ body:
     validations:
       required: true
 
+  - type: markdown
+    attributes:
+      value: |
+        If your report is about TR3, please note that current support is limited to Lara's Home, India, and South Pacific.
+        Reports for other TR3 areas are still out of scope while that work is in progress, so those issues are not being accepted yet.
+
   - type: input
     id: trx_version
     attributes:
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,0 +1,102 @@
+name: Bug report
+description: Report a reproducible problem in TRX
+title: "[Bug] "
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Please open one issue per bug.
+        Before submitting, read the bug reporting guide:
+        https://github.com/LostArtefacts/TRX/blob/develop/BUG_REPORTING.md
+
+  - type: dropdown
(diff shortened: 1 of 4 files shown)


rr-:
This might work against us, actually – confronted with a large form, many people will probably just close the tab, and we'll lose some reports. On the other hand, having to chase missing info at the current frequency is likely even less favorable long-term.

LMK what you think. We don't have to push this if it feels too heavy. I can also simplify parts of the form, though our field type options are pretty limited.

ShadowXJaw724:
I've given it a look, and as someone who would make an issue, this would prevent people from spamming issues or not putting effort into them, and while it may lead to some people not posting issues as you said, it ensures a certain line of quality when it comes to issues for TRX. Until TR3 is complete, I would suggest adding a note if it's about TR3 that, if reporting on an issue for an incomplete area, it should say that bug reports aren't being taken until that area is complete, to prevent cases such as #5086, but that's an optional suggestion if anything.

rr-:
> TR3 completion progress

That's a great suggestion esp. considering the project does receive issues about unfinished areas every now and then, so this should save times for both sides – added.

aredfan (approved):

lahm86 (approved):
Great stuff 😌

Richard-L:
I think the template is well designed. I also think it'll stop people from reporting. And I'm afraid people will _always_ come to discord to chat about the thing of their momentary interest. It's what discord promotes, probably.

My mentality here is more that of a sponge that soaks up everything regardless of what it is, to then hopefully be able to distil something. But that's very easy for me to say who is not in your shoes.

The only idea I had was to have a team of helpers who close or prep issues for you when they miss information, so you don't have to and can focus on stuff. But I don't know how realistic that is.

rr-:
> The only idea I had was to have a team of helpers who close or prep issues for you when they miss information, so you don't have to and can focus on stuff. But I don't know how realistic that is.

Yeah I'd really appreciate someone who had longer fuse to take care of the community management :) Some projects are blessed with such people, but unfortunately, they emerge rarely, and naturally. I don't think it's a position to advertise for, unless you're an actual commercial company.

rr-:
Let's pilot this. If we end up not liking it, we can always revert/adjust.

PR #5099, opened 2026-03-12
av/fmv: route FMV audio through Audio_Stream

goldfish-g:
## Summary

- Stop opening a second SDL audio device inside the ffplay-derived video
  player. Instead, decode FMV audio through the game's existing mixer via
  `Audio_Stream_CreateFromFile` and feed the resulting timestamp back via
  `Video_SetExternalAudioClock` for A/V sync.
- Add `.ogv` and `.fmv` to the FMV extension search list. `.ogv` is
  preferred over `.fmv` so that remastered HD video files are used when
  available (remastered `.ogv` files contain HD video but no audio).
- When the selected video file lacks an audio stream (e.g. a remastered
  `.ogv`), probe alternative extensions (`.mp4`, `.mpeg`, `.webm`,
  `.avi`, `.fmv`, `.rpl`) to locate a companion file that carries the
  original audio track.

## Why

The Tomb Raider remasters ship `.ogv` files with HD video but strip the
audio track. The original `.fmv`/`.rpl` files still contain audio. By
preferring `.ogv` for video and falling back to `.fmv` (or other
formats) for audio, players with remaster assets get HD cutscenes with
working sound.

Routing all FMV audio through the game's existing mixer also avoids
issues on platforms that only support a single audio device (e.g. WebGL
via Emscripten), and ensures FMV volume respects the in-game master and
FMV volume settings.

## Checklist

## Test plan

- Play an FMV with only `.fmv` files present — audio and video should
  work from the same file
- Play an FMV with both `.ogv` (remaster, no audio) and `.fmv` (original,
  with audio) present — video should come from `.ogv`, audio should fall
  back to `.fmv`
- Verify FMV volume slider and mute toggle still work during playback
- Verify A/V sync is maintained throughout a full cutscene

### DIFF
--- a/src/trx/av/video.c
+++ b/src/trx/av/video.c
@@ -271,6 +271,8 @@ typedef struct {
 
     void (*surface_upload_func)(void *surface, void *user_data);
     void *surface_upload_func_user_data;
+
+    bool audio_enabled;
 } M_STATE;
 
 static int64_t m_AudioCallbackTime;
@@ -1795,7 +1797,7 @@ static int M_ReadThread(void *arg)
         AVRational sar = av_guess_sample_aspect_ratio(ic, st, nullptr);
     }
 
-    if (st_index[AVMEDIA_TYPE_AUDIO] >= 0) {
+    if (is->audio_enabled && st_index[AVMEDIA_TYPE_AUDIO] >= 0) {
         M_StreamComponentOpen(is, st_index[AVMEDIA_TYPE_AUDIO]);
     }
 
@@ -1997,6 +1999,12 @@ void Video_PumpEvents(VIDEO *video)
     video->is_playing = !is->abort_request && !is->playback_finished;
 }
 
+void Video_SetAudioEnabled(VIDEO *const video, const bool enabled)
+{
+    M_STATE *const is = video->priv;
+    is->audio_enabled = enabled;
+}
+
 void Video_SetVolume(VIDEO *const video, const double volume)
 {
     M_STATE *const is = video->priv;
@@ -2137,4 +2145,10 @@ void Video_SetRenderEndFunc(
     M_STATE *const is = video->priv;
     is->render_end_func = func;
     is->render_end_func_user_data = user_data;
+}
+
+void Video_SetExternalAudioClock(VIDEO *const video, const double timestamp)
+{
+    M_STATE *const is = video->priv;
+    M_SetClock(&is->extclk, timestamp, is->extclk.serial);
 }
--- a/src/trx/game/fmv.c
+++ b/src/trx/game/fmv.c
@@ -3,6 +3,7 @@
 #include <trx/av/audio.h>
 #include <trx/av/video.h>
 #include <trx/config.h>
+#include <trx/core/filesystem.h>
 #include <trx/core/log.h>
 #include <trx/core/memory.h>
 #include <trx/core/strings.h>
@@ -21,6 +22,10 @@
 #include <string.h>
 
 static bool m_IsPlaying = false;
+
+static const char *const m_FallbackExts[] = {
+    ".mp4", ".mpeg", ".webm", ".avi", ".fmv", ".rpl", nullptr,
+};
 
 typedef struct {
     OUTPUT_QUAD_SURFACE_DESC desc;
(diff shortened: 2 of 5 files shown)


src/trx/game/fmv.c
```diff
@@ -130,15 +131,47 @@ static bool M_Play(const char *const file_name)
     Video_SetSurfaceUnlockFunc(video, M_UnlockSurface, nullptr);
     Video_SetSurfaceUploadFunc(video, M_UploadSurface, renderer_2d);
 
+    int32_t audio_id = Audio_Stream_CreateFromFile(file_name);
+    if (audio_id == AUDIO_NO_SOUND) {
+        // The video file may lack an audio stream (e.g. remastered .ogv).
+        // Try other FMV extensions to find a file that contains audio.
+        static const char *const fallback_exts[] = {
```
rr-: We avoid `static` variables mid-functions – this should become `static const char *const m_FallbackExts[]` at the top of the module.

src/trx/game/fmv.c
```diff
@@ -130,15 +131,47 @@ static bool M_Play(const char *const file_name)
...
+                    Memory_Free(candidate);
+                    if (audio_id != AUDIO_NO_SOUND) {
+                        break;
+                    }
+                } else {
+                    Memory_Free(candidate);
+                }
+            }
+        }
+    }
```
rr-: I suggest wrapping this entire block in a static function, `M_OpenAudioStream`.

rr- (commented):
Doesn't this open the audio stream twice, once via `Audio_Stream`, and once via `Video`, for normal FMVs?

goldfish-g:
> Doesn't this open the audio stream twice, once via `Audio_Stream`, and once via `Video`, for normal FMVs?

It doesn't, the change to video.c skips opening the audio track completly now, the video decoder only handles video now.

goldfish-g:
Implemented the suggested improvements 👍

src/trx/av/video.c
```diff
@@ -1795,9 +1795,8 @@ static int M_ReadThread(void *arg)
         AVRational sar = av_guess_sample_aspect_ratio(ic, st, nullptr);
     }
 
-    if (st_index[AVMEDIA_TYPE_AUDIO] >= 0) {
-        M_StreamComponentOpen(is, st_index[AVMEDIA_TYPE_AUDIO]);
-    }
+    // Audio is handled externally via Audio_Stream, so skip opening
+    // the audio component here to avoid a second SDL audio device.
```
rr-: Maybe let's keep the original call behind a flag, like `Video_SetAudioEnabled(bool)`.
This way:
- it justifies not deleting a lot of code from video.c related to audio processing right now,
- keeps it clear how to orchestrate audio via video.c if it's ever needed,
- doesn't interfere with the new changes.

rr- (approved):
LGTM. Thank you. Let's wait for @lahm86 for additional code check and @aredfan to do QA

goldfish-g:
> LGTM. Thank you. Let's wait for @lahm86 for additional code check and @aredfan to do QA

Sounds good 😊
Just a warning that when using audio_enabled = true, we don't have the logic of the audio falling back to another file, so if the engine uses the ogv files, the FMVs will have no audio. I figure that's fine, since that format wasn't even considered before

rr-:
> > LGTM. Thank you. Let's wait for @lahm86 for additional code check and @aredfan to do QA
> 
> Sounds good 😊 Just a warning that when using audio_enabled = true, we don't have the logic of the audio falling back to another file, so if the engine uses the ogv files, the FMVs will have no audio. I figure that's fine, since that format wasn't even considered before

Would `Video_SetAudioEnabled(audio_stream == AUDIO_NO_SOUND)` help with that?

goldfish-g:
> > > LGTM. Thank you. Let's wait for @lahm86 for additional code check and @aredfan to do QA
> > 
> > 
> > Sounds good 😊 Just a warning that when using audio_enabled = true, we don't have the logic of the audio falling back to another file, so if the engine uses the ogv files, the FMVs will have no audio. I figure that's fine, since that format wasn't even considered before
> 
> Would `Video_SetAudioEnabled(audio_stream == AUDIO_NO_SOUND)` help with that?

I mean, it would, but it means we would only use the new logic when the video file has no audio. Currently I am using a hardcoded false outright, to always use the new logic. WebGL requires this to be false always.

rr-:
> > > > LGTM. Thank you. Let's wait for @lahm86 for additional code check and @aredfan to do QA
> > > 
> > > 
> > > Sounds good 😊 Just a warning that when using audio_enabled = true, we don't have the logic of the audio falling back to another file, so if the engine uses the ogv files, the FMVs will have no audio. I figure that's fine, since that format wasn't even considered before
> > 
> > 
> > Would `Video_SetAudioEnabled(audio_stream == AUDIO_NO_SOUND)` help with that?
> 
> I mean, it would, but it means we would only use the new logic when the video file has no audio. Currently I am using a hardcoded false outright, to always use the new logic. WebGL requires this to be false always.

I don't understand. I meant muting video audio only if we failed to create a separate-file audio stream, and if we fail to use an external file, allow the video orchestrator to do its work.
(I may be very much speaking nonsense.)

goldfish-g:
> I don't understand. I meant muting video audio only if we failed to create a separate-file audio stream, and if we fail to use an external file, allow the video orchestrator to do its work. (I may be very much speaking nonsense.)

Ah, you meant adding the old logic as a backup in case the new one fails at either creating a seperate audio stream or loading the second file for audio? 
I guess I could do that, but don't know if it's the right approach. It could silently fail on desktop, but still work with the fallback, but then on the web the audio wouldn't work and we'd be wondering why it works on desktop and not on the web 😅

I don't know, feels like an extra source of confusion down the line. Or am I still misunderstaing what you are asking for? I may be confused too

rr-:
> > I don't understand. I meant muting video audio only if we failed to create a separate-file audio stream, and if we fail to use an external file, allow the video orchestrator to do its work. (I may be very much speaking nonsense.)
> 
> Ah, you meant adding the old logic as a backup in case the new one fails at either creating a seperate audio stream or loading the second file for audio? I guess I could do that, but don't know if it's the right approach. It could silently fail on desktop, but still work with the fallback, but then on the web the audio wouldn't work and we'd be wondering why it works on desktop and not on the web 😅
> 
> I don't know, feels like an extra source of confusion down the line. Or am I still misunderstaing what you are asking for? I may be confused too

Yeah not worth it at all. I misunderstood what you meant about playing .ogv with `audio_enabled=true`. Since we never set it to true from `fmv.c`, it's all good :relieved: it's dead code, but it's there in case we ever needed, and the warning is now understood correctly on my part.

lahm86 (approved):

aredfan (approved):

PR #5101, opened 2026-03-12
objects/security_guard: add security guard control

lahm86:
#### Checklist

#### Description

This adds control for the London Security Guard (type ID 56). I've also aligned usage of `AIInfo`'s `bite` and `ahead` properties as bools rather than ints throughout the creature modules as they are only ever set as such.


rr- (approved):

data/common/ship/cfg/base_strings.json5
```diff
@@ -2451,6 +2451,9 @@
         "secret_3": {
             "name": "Secret 3",
         },
+        "security_guard": {
+            "name": "Securty Guard",
```
aredfan: `Security` 😄

lahm86: ~~Fxed~~ fixed 😄

aredfan:
Issue:

- [x] Explode a security guard, save and reload, and Lara will end up targeting where the guard was standing.

> [record.txt](https://github.com/user-attachments/files/25937242/record.txt)

It looks like saving and reloading doesn't retain the enemy's hp if it's at half-full, for example.

lahm86:
> Issue:
> 
> * [x]  Explode a security guard, save and reload, and Lara will end up targeting where the guard was standing.
> 
> > [record.txt](https://github.com/user-attachments/files/25937242/record.txt)
> 
> It looks like saving and reloading doesn't retain the enemy's hp if it's at half-full, for example.

Great catch, thanks. Fixed.

aredfan (approved):

PR #5105, opened 2026-03-13
savegame: restore boulder floor on load

lahm86:
#### Checklist

#### Description

This ensures boulder floor values are restored on load to prevent them from re-activating for a frame due to a mismatch with their current position.

An alternative to this would be having `M_PRIV` for boulders that replicates `item->floor` and have that saved/loaded. Or, we could drop the shadow check in the load function, but it'd need heavier testing as there are dozens of objects that save position but don't have a shadow size.

### DIFF
--- a/src/trx/game/objects/setup.c
+++ b/src/trx/game/objects/setup.c
@@ -54,6 +54,7 @@ void Object_SetupAllObjects(void)
         obj->save_hitpoints = false;
         obj->save_flags = false;
         obj->save_anim = false;
+        obj->load_floor = false;
         obj->intelligent = false;
         obj->smartness = -1;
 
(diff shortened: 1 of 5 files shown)


src/trx/game/objects/setup.c
```diff
@@ -55,6 +55,7 @@ void Object_SetupAllObjects(void)
         obj->save_flags = false;
         obj->save_anim = false;
         obj->intelligent = false;
+        obj->load_floor = false;
```
rr-, line 57: Let's move it near `save_anim`

rr- (approved):

rr-:
Thank you, this looks very reasonable. Agree on keeping the shadow check.

I can't help but wonder if we could make it completely unconditional, and depend on `save_position` alone. Though I'm almost certain there will be undesired side effects to this…

lahm86:
> Thank you, this looks very reasonable. Agree on keeping the shadow check.
> 
> I can't help but wonder if we could make it completely unconditional, and depend on `save_position` alone. Though I'm almost certain there will be undesired side effects to this…

Yeah, there are things like grenades, vehicles and skidoo drivers - thought it best not to rock the boat 😄

aredfan (approved):

PR #5111, opened 2026-03-14
objects/diver: update TR3 diver control

lahm86:
#### Checklist

#### Description

This updates the scuba diver in line with TR3. Everything is the same except TR3 does not use extra bone rotation. Harpoons also don't spawn blood when they hit Lara, but that's OG; it'd be easy enough to add the effect, WDYT?


rr- (approved):
IMO they should spawn blood, and we could add it to the changelog.

lahm86:
> IMO they should spawn blood, and we could add it to the changelog.

Done ☺️

rr- (approved):

aredfan (approved):

PR #5117, opened 2026-03-15
objects/pendulum: update TR3 pendulum control

lahm86:
#### Checklist

#### Description

This updates pendulums for TR3, and adds the ability to use flaming pendulums in TR1 and TR2. The setup mimics that of punks, so a flame emitter is placed at the same position; this avoids the OG hack of using the item timer, and so allows all pendulum types to continue to use timers normally. Injections are provided for Tinnos and Reunion to update the triggers and place these extra emitters.

Will share a TR1 test level, but the other areas to check are the different forms it can be in TR3.

- Lud's Gate: metallic pendulums and stampers
- RX-Tech: drill bits and rolling machines
- Tinnos: flaming pendulums and underwater blades
- All Hallows: same stamper as Lud's Gate
- Reunion: flaming pendulums

A sanity check of TR1 and TR2's regular pendulums would be good too.

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/227


rr- (approved):
Thank you :)

aredfan (approved):

PR #5122, opened 2026-03-15
gun: check smashables behind target

rr-:
#### Checklist

#### Description

This fixes the Fuse box shooting in Reunion.

Doesn't apply to Rockets because come on. For conventional bullets that aren't drawn, it might make sense that some "get past" the target even when it gets hit. But for a rocket that explodes at the hit target, using a raycast to also smash items way behind the hit item, is actually harmful IMO.

LMK what you think.


lahm86 (approved):
Looks like a good approach to me 👍

aredfan (approved):

PR #5171, opened 2026-03-21
lara/col/climb: fix ladder woes

### LINKED ISSUE #5169: [Bug] Lara cannot climb up on a triangle surface?
### Game type

Original game

### Game

TR3

### TRX version

1.4

### Operating system

Windows 10

### GPU

NVIDIA GeForse GTX 1050ti

She cannot climb when the reaching point isn't anything but a flat square? This doesn't occur in the original game, only on the current TR3 dev build (1.4) for London.

To climb up normally

### Attachments

https://github.com/user-attachments/assets/627abdf1-9793-49bc-b4a1-25800baf1637

### Before submitting

lahm86:
Resolves #5169.

#### Checklist

#### Description

TR3 uses a general X/Z shift of 4 for hanging and ladder checks, whereas TR1/2 both use 2. We had this version gate in place in one area but it was missing in two others, which seems to be the cause of the faulty ladder in Thames Wharf.

Also fixed Lara going from ladder to crouch in this area when pulling up, which only became evident with the above fix.

### DIFF
--- a/src/trx/game/lara/col/climb.c
+++ b/src/trx/game/lara/col/climb.c
@@ -21,6 +21,7 @@
 #define M_LEDGE_JUMP_PUSH_HEIGHT (STEP_L - 16)                    // = 240
 #define M_LEDGE_JUMP_HEIGHT_UP   (LARA_HEIGHT + (STEP_L * 3) / 8) // = 858
 #define M_LEDGE_JUMP_HEIGHT_BACK (LARA_HEIGHT - (STEP_L * 5) / 4) // = 442
+#define M_HANG_SHIFT             (g_TRVersion >= 3 ? 4 : 2)
 // clang-format on
 
 typedef enum {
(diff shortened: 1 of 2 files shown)


aredfan (approved):

src/trx/game/lara/col/climb.c
```diff
@@ -808,7 +808,7 @@ static void M_UpLadder(ITEM *const item, COLL_INFO *const coll)
         item->goal_anim_state = LS(LS_CLIMB_STANCE);
         Lara_Animate(item);
         if (ABS(ledge_l - ledge_r) <= 120) {
-            if (result_r == CLIMB_RESULT_NEG && result_l == CLIMB_RESULT_NEG) {
+            if (result_r == CLIMB_RESULT_NEG || result_l == CLIMB_RESULT_NEG) {
```
rr-, line 811: Are we sure about this?  What does tomb3 say?

rr- (approved):

src/trx/game/lara/col/climb.c
lahm86: Yes, should be good now. This isn't in TR3 originally, I took it from TR5. But I misread the condition originally. TR2/3 doesn't compare anything beyond `ABS(ledge_l - ledge_r) <= 120` as the earlier condition guards the climb result states.

PR #5194, opened 2026-03-22
cmd/tp: add enemy special target

### LINKED ISSUE #5187: [Feature] Add /tp enemy

Hey!
I'm totally enjoying your builds! 
But when something is good, people tend to get lazy and greedy. And, after nearly 29 years of playing, I don't feel like doing NLNMPO again... especially TR3... no way lol

So I was thinking about a save files editor - pretty much self-explanatory.
Or...
Similarly to the "secret give" command, "kill all" to also be registered in the statistics. 
And a new command: "pickups" which gives you all the level pickups, and is also registered in the final stats. 
Just an idea!

Thanks for keeping the flame of the originals!
Cheers!
 

_No response_

### Extra context

_No response_

rr-:
#### Checklist

#### Description

This adds `/tp enemy`.

As an additional convenient enhancement, it changes `/tp` to activate the target item if it's a creature. This is actually a can of worms for the following reasons:

- certain debugging scenarios, like teleporting to Xian statues to check if they take damage in their inactive state, will become harder and require workarounds, as the creatures will immediately activate.
- it is not complete because many of the creatures (Marco Bartoli, for example) will not activate due to special handling.
- it creates inconsistency because traps are not activated.

LMK if you like this or if I should remove this change. I'm leaning towards removing.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## [Unreleased](https://github.com/LostArtefacts/TRX/compare/trx-1.4.1...develop) - ××××-××-××
 - added remembering of the last played mod
+- added a new console command, `/tp enemy`, to cycle Lara through hostile creatures in the current level
 - changed `--level` to no longer require `-e/--engine` to work
 
 ## [1.4.1](https://github.com/LostArtefacts/TRX/compare/trx-1.4...trx-1.4.1) - 2026-03-23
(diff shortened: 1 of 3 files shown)


lahm86:
I would definitely lean towards not triggering the item; `/trigger` is available and sufficient IMO.

rr-:
> I would definitely lean towards not triggering the item; `/trigger` is available and sufficient IMO.

Your voice is my command – removed :relieved:

lahm86 (approved):
LGTM, thanks!

aredfan (approved):

PR #5198, opened 2026-03-23
data: restore High Security Compound cutscene

lahm86:
#### Checklist

#### Description

This restores the cutscene at the beginning of High Security Compound. It introduces an animation command to allow rotating an item by 90 degrees, such that Lara can be angled properly during the scene.

LMK if we want this to be optional; if so, I'd be inclined to have the option affect the equivalent TR2 starting animations (Rig and HSH) as well.

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/236


aredfan (approved):

rr- (approved):
I'd leave this to be non-optional.

PR #5209, opened 2026-03-24
Add security lasers

rr-:
#### Checklist

#### Description

This adds the three thin laser types: alarm / damage / kill.

Additionally it changes the line segment drawing function to no longer draw two quads in a + pattern, and instead emit a single camera-facing quad. This gets rid of the opacity increase at the point of quads intersection which looks bad in photo mode and looked bad for the newly added lasers.


src/trx/game/objects/traps/security_laser.c
```diff
@@ -0,0 +1,397 @@
...
+        step = (XZ_32) { 0, WALL_L };
+        break;
+    default:
+        step = (XZ_32) { WALL_L, 0 };
+        break;
+    }
+
+    int32_t x = item->pos.x;
+    int32_t z = item->pos.z;
+    while (sector->box != 2047
```
lahm86: I think this'll need to be `NO_BOX` - we translate 2047 as -1 during level load.

rr-: Thanks – fixed.

lahm86 (approved):

aredfan (approved):

PR #5218, opened 2026-03-24
objects/detonator_box: fix bounds and rotation

### LINKED ISSUE #5215: TRX bug: issue with the detonator when Lara interacts with it
Regression since [TR2X 1.3](https://github.com/LostArtefacts/TRX/releases/tag/tr2-1.3)

_____________

I came across a bug where Lara is prevented from manually selecting the detonator key.
It's an OG bug which we've fixed in #1965 and subsequently released as part of [TR2X 0.7](https://github.com/LostArtefacts/TRX/releases/tag/tr2-0.7).

To reproduce:

> 1. launch tr2
> 2. `/play 3`
> 3. use the all items cheat
> 4. `/tp det` x2
> 5. have Lara slightly turn to the left or right instead of directly facing the detonator
> 6. open the inventory and **manually** select the detonator key
> 7.  Lara says "No"

[record.txt](https://github.com/user-attachments/files/26218850/record.txt)

As a result, manually selecting the key for the detonator is quite challenging because Lara must be facing an exact angle.

_____________________

There is a second possibly related bug.
If Lara interacts with a detonator without using the key, the detonator orientates towards Lara when this happens.
In OG, the detonator would remain unmoved when following the same steps, it only turns towards Lara when she uses the key.

To reproduce:

> 1. launch tr2
> 2. `/play 3`
> 3. use the all items cheat
> 4. `/tp det` x2
> 5. have Lara slightly turn to the left or right instead of directly facing the detonator
> 6. **press action**
> 7. close the inventory **without** using the key

[record.txt](https://github.com/user-attachments/files/26218817/record.txt)

____________________

Both bugs are a regression since [TR2X 1.3](https://github.com/LostArtefacts/TRX/releases/tag/tr2-1.3).
The second issue isn't hugely important and can be dispensed if needed.

lahm86:
Resolves #5215.

#### Checklist

#### Description

This replicates pickup bounds used by the Detonator Box (which is OG) into its own module, so allowing it to use the `ignore_rot` property such that it can be activated manually from the inventory. It also prevents it remaining rotated if Lara begins interacting with it but then cancels out of the inventory.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 **TR2**:
 - changed the Detonator Box to no longer hard-code dynamic light output; refer to the migration guide for custom levels
+- fixed the Detonator Box being difficult to activate when selecting the key manually from the inventory (#5215, regression from TR2X 1.3)
 
 **TR3**:
 - added Security Laser (Alarm) control
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -8,6 +8,7 @@
 **TR2**:
 - changed the Detonator Box to no longer hard-code dynamic light output; refer to the migration guide for custom levels
 - fixed the Detonator Box being difficult to activate when selecting the key manually from the inventory (#5215, regression from TR2X 1.3)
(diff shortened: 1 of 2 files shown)


rr- (approved):

aredfan (approved):
Thank you for fixing the issue. 😌

PR #5236, opened 2026-03-27
creature: fix collision code

rr-:
#### Checklist

#### Description

This fixes a drift from OG TR3 code in `Creature_Collision` but I don't know of the consequences.
It only sets Lara's hit direction and frame.
Maybe worth checking what happens after defeating Tony and we try to pick up Infada Stone too fast?

### DIFF
--- a/src/trx/game/creature/common.c
+++ b/src/trx/game/creature/common.c
@@ -1,5 +1,6 @@
 #include <trx/config.h>
 #include <trx/core/log.h>
+#include <trx/core/math.h>
 #include <trx/core/utils.h>
 #include <trx/game/camera/vars.h>
 #include <trx/game/collision/los.h>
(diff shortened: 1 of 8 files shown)


src/trx/game/creature/common.c
```diff
@@ -800,27 +801,44 @@ void Creature_Collision(
...
-        item, coll,
-        (g_TRVersion >= 2 || item->hit_points > 0) ? coll->enable_hit : false,
-        false);
+    if (coll->enable_baddie_push) {
+        Lara_Col_ItemPush(
+            item, coll,
+            (g_TRVersion >= 2 || item->hit_points > 0) ? coll->enable_hit
+                                                       : false,
+            false);
+    } else if (g_TRVersion == 3) {
```
lahm86: I think we also need `&& coll->enable_hit`, otherwise Lara is in her hit animation through picking up Tony's artefact if she gets there before he's completely gone.

rr-: That's absolutely right – sorry about that. Fixed.

lahm86 (approved):

aredfan (approved):

PR #5240, opened 2026-03-27
stats: update HSC/Area51 kill stats

lahm86:
#### Checklist

#### Description

This marks Orcas to not be included in the kill total, in a similar way to how we exclude TR2 eels. I've also amended the unobtainable kills in HSC, but would appreciate a sanity check in particular of this. Following are those I believe cannot be killed (by Lara) without using cheats as they're all before the point of no return.

<details>

| Index | Type     |
|-------|----------|
| 0     | Prisoner |
| 4     | Prisoner |
| 5     | Prisoner |
| 17    | MP 2     |
| 126   | MP 1     |
| 132   | Prisoner |
| 147   | Prisoner |
| 148   | Prisoner |
| 149   | Prisoner |
| 156   | Prisoner |
| 165   | MP 1     |
| 180   | MP 1     |
| 185   | Sentry   |
| ~~188~~   | ~~MP 2~~     |
| 194   | Sentry   |
| 207   | Sentry   |
| 209   | MP 1     |
| 213   | Sentry   |
</details>

### DIFF
--- a/data/trx/ship/games/tr3/gameflow.json5
+++ b/data/trx/ship/games/tr3/gameflow.json5
@@ -355,6 +355,8 @@
                 "lara_guns.bin",
                 "compound_cine.bin",
             ],
+            "unobtainable_kills": 9,
+            "unobtainable_ally_kills": 8,
         },
 
         // 15. Area 51
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -27,6 +27,7 @@
 - added Hook control
 - restored the cutscene at the beginning of High Security Compound
 - changed Area 51 Rocket to no longer hardcode room 52 as the fire blast room; instead it checks for presence of an upwards portal pointing to the rocket room
+- changed enemies who are killed by lasers to be included in the stats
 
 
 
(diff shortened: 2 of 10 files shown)


rr- (approved):
Thanks. Does killing the guard with the security laser count as a kill?

lahm86:
> Thanks. Does killing the guard with the security laser count as a kill?

It doesn't, but we could add that. It does "box damage" on each frame but we could check when HP hits 0 and add a kill then. I suppose it's an indirect kill, but it's still Lara as the source.

rr-:
> > Thanks. Does killing the guard with the security laser count as a kill?
> 
> It doesn't, but we could add that. It does "box damage" on each frame but we could check when HP hits 0 and add a kill then. I suppose it's an indirect kill, but it's still Lara as the source.

Yeah let's do that 🩷 Thanks.

I wonder if we should also count kills by allies. I think _yes_, because I've seen once a streamer struggled to get a 100% kill count due to monks in a custom level. Perhaps an option. I'd open a separate task for that though.

lahm86:
Sounds good for that kind of case with the monks. I'll log it.
Pushed a commit for blocked box kills, which turned out to be a nice little refactor too as it was identical logic in all the guards.
Reduced unobtainables by 1 in HSC.

rr- (approved):

aredfan (approved):

PR #5254, opened 2026-03-29
Fix combo deferral for non-starter keys and tab navigation after rebinding

### LINKED ISSUE #5249: [Bug] Bugged combo binds
### Game type

Original game

### Game

TR2

### TRX version

1.4.2-45-g8227514

### Operating system

Linux Bazzite

### GPU

Intel UHD Graphics @ 0.75 GHz [Integrated]

I combo bound the harpoon gun and M16 hotkeys to Select + LT and Select + RT respectively and it broke Look and Walk functionality (I had previously bound Look to LT and Walk to RT), leaving me unable to do either.

Combo Bindings should not interfere or conflict with regular ones.

### Attachments

[TRX.log](https://github.com/user-attachments/files/26326587/TRX.log)

### Before submitting

### LINKED ISSUE #5250: [Bug] Rebinding Shoulder Buttons and Triggers slightly breaks UI navigation
### Game type

Original game

### Game

All games

### TRX version

1.4.2-45-g8227514

### Operating system

Linux Bazzite

### GPU

Intel UHD Graphics @ 0.75 GHz [Integrated]

I rebound Look from LB to LT and Walk from RB to RT, and this disabled being able to switch tabs in menus.

Menu navigation should not be affected by custom binds.

### Attachments

[TRX.log](https://github.com/user-attachments/files/26326596/TRX.log)

### Before submitting

goldfish-g:
Fixes #5249, fixes #5250.

#### Checklist

#### Description

Fix three related input binding bugs:

**Combo deferral for non-starter keys (#5249)**

Phase 3 and Phase 6 of combo resolution incorrectly deferred any key
appearing anywhere in a multi-key combo. For example, with
harpoon=SELECT+LT and M16=SELECT+RT, pressing LT or RT alone would get
blocked — even though SELECT is the combo starter, not LT/RT.

`M_IsComboKey`/`M_IsComboInput` now only match `keys[0]` (the combo
starter) instead of all positions. `M_HasLongerCombo` additionally
requires the first key to match between the short binding and the longer
combo. This means a key that only appears as the 2nd or 3rd element of a
combo is never deferred or suppressed on its own.

**New role classifications: sustained and capturing (#5249)**

Two new role classifications replace the previous approach of marking
look and walk as immediate:

*Sustained* roles (look, walk, step, crouch, sprint) are held states
that fire immediately without deferral but, unlike immediate roles, can
be the first key of a combo. However, combos starting with a
non-capturing sustained role cannot contain immediate keys (e.g.
sprint+jump is invalid) because the immediate action would fire before
the combo could form. Invalid combos are flagged as conflicts in the
controls editor and suppressed at runtime.

*Capturing* roles (currently only look) are sustained roles that redirect
other inputs while held, so combos like look+jump are valid since the
immediate key loses its normal meaning under look mode.

**Tab navigation breaks when LT/RT are rebound (#5250)**

Tab switching in menus checked the rebindable `step_left`/`step_right`
roles as a proxy for "triggers pressed". After rebinding LT/RT away from
sidestep (e.g. to look/walk), triggers no longer switched tabs.

New hardcoded roles `MENU_TAB_LEFT`/`MENU_TAB_RIGHT` are now bound to
LT/RT on controller (and DELETE/PAGEDOWN on keyboard, matching the
existing step defaults). These are non-rebindable, so triggers always
switch tabs regardless of in-game bindings.


rr-:
But this means we no longer can bind to Shift+X right?

rr-:
Also since Delete is now hardcoded does this mean the players can no longer bind it to other actions?

goldfish-g:
> But this means we no longer can bind to Shift+X right?

The restrictions are role based, not key based. If shift is walk, then shift + x can no longer be bound, like you said. But x + shift can. There was already a list of roles that had an "immediate" flag, meaning they cannot be the start of combos.

> Also since Delete is now hardcoded does this mean the players can no longer bind it to other actions?

Actually no! I added new roles for tab switching, so that delete will work for those roles regardless of what other roles it has assigned. So delete can still be bound freely. And LT and RT too of course.

goldfish-g:
Actually, now that I think about it, maybe we should have another type of restriction other than "immediate"... For example, roll can't be the start of a combo, and that makes sense, but something like look and walk can, as long as they don't block their primary function 🤔

rr-:
Yeah let's explore alternatives.

goldfish-g:
Okay, I think I found something I am happy with. I added two new role classifications, I amended the PR body with an explanation so I won't repeat here.
TLDR walk + action is a valid shortcut, action and walk now have the sustained role, but walk + jump is not, as jump is still an immediate tole, and takes priority as to not ruin gameplay.

rr- (commented):
Can the sustained inputs policy be merged together?

Right now the keyboard and controller backends each seem to carry their own copy of:

- combo-starter detection
- has longer combo with same starter logic
- sustained/capturing validation policy

IMO the backend-specific parts should really just answer low-level questions like "is this binding pressed?", "are these two key tokens equal?" etc. The combo policy itself could live in one shared layer with small backend adapters only. LMK if this makes sense.

---

Additionally I think there is still possibility to achieve illegal combos, by doing:

- bind walk to G + H (walk now never works)
- bind pistols to G+H+J; keep UI hiding with H (GHJ both equips pistols and toggles UI)

But that is just ridiculous.

goldfish-g:
> Can the sustained inputs policy be merged together?
> 
> Right now the keyboard and controller backends each seem to carry their own copy of:
> 
>     * combo-starter detection
> 
>     * has longer combo with same starter logic
> 
>     * sustained/capturing validation policy
> 
> 
> IMO the backend-specific parts should really just answer low-level questions like "is this binding pressed?", "are these two key tokens equal?" etc. The combo policy itself could live in one shared layer with small backend adapters only. LMK if this makes sense.

That's fair enough, I should avoid duplicate code, I will improve this.

> Additionally I think there is still possibility to achieve illegal combos, by doing:
> 
>     * bind walk to G + H (walk now never works)
> 
>     * bind pistols to G+H+J; keep UI hiding with H (GHJ both equips pistols and toggles UI)
> 
> 
> But that is just ridiculous.

Well, yes, I guess it's not 100% fool proof, but I think at some point we can call it user error 🤣
I think all reasonable checks are accounted for.

docs/CHANGELOG.md
```diff
@@ -11,6 +11,8 @@
 - fixed recordings keeping unbound hotkeys active during playback
 - fixed being unable to change FOV after using photo mode without restarting the level (#5246, regression from 1.0)
 - fixed Lara getting stuck if using crouch-roll near very low ceilings (#5248)
+- fixed combo bindings blocking non-starter keys from firing (e.g. SELECT+LT blocking standalone LT) (#5249)
```
lahm86: I think the bugs are unreleased so we can skip the changelog entries (although best to confirm if that's the case with 5250 - I don't have a controller unfortunately).

rr- (approved):
`backends/combo.[ch]` → `combo.[ch]` and we're good (backends/combo dilutes the folder contents a bit which should focus on strategies, with a small exception made for internal.c).
If we worry about this being private and not part of the public APIs, we could introduce input/priv/combo.[ch] and move backends/internal.[ch] there too (under a more meaningful name)?

rr- (approved):

lahm86 (approved):

aredfan (approved):

PR #5263, opened 2026-03-29
objects/security_laser: add heavy trigger support

### LINKED ISSUE #5225: [Feature] Add the ability for TR3 tripwires to activate heavy triggers

From what I learned from @lahm86 on Discord, the TR3 tripwires can only activate sentry guns and strobe lights. I think it would be a great feature to make tripwires activate heavy triggers placed on squares beneath them - like the bells in TR2 for example. 
If it is an unreasonable request / not possible, feel free to close. Just figured I'd ask. :) 

_No response_

### Extra context

_No response_

lahm86:
Resolves #5225.

#### Checklist

#### Description

This adds the ability for security lasers to activate heavy triggers when tripped by Lara. The trigger must be on the same tile as the null mesh and it's activated if Lara touches any of the beams. None of the lasers in OG sit on heavy triggers, so this is an unobtrusive change. Test level available.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -6,6 +6,7 @@
 - added dynamic mod discovery from the games/ directory using new `extends` and `name` fields in gameflow.json5
 - added expanded statistics screen customization, including per-row toggles and a choice between bare and bordered layouts (Graphic Options → Stats)
 - added optional save/heal crystal counts to level and final statistics (#5180)
+- added the ability for security lasers to activate heavy triggers when tripped by Lara (#5225)
 - changed `--level` to no longer require `-e/--engine` to work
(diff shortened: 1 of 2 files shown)


rr- (approved):
Maybe we could extract `M_……()` for batch activation?

lahm86:
> Maybe we could extract `M_……()` for batch activation?

Sounds good. I just noticed the strobe light/sentry gun loop happens for each beam (same in OG). I _think_ that should be safe to extract and just run it once if tripped, WDYT?

rr-:
It should be safe since this underneath unconditionally sets internal state to specific constants of the affected items, rather than doing stateful updates. (`M_HandleEvent`)

lahm86:
Thanks, updated 😌

rr- (approved):

aredfan (approved):

PR #5268, opened 2026-03-29
savegame: allow quick load when Lara's dead

### LINKED ISSUE #5264: [Feature] Allow quick load when Lara is dead and cmd line loading option

When Lara is dead or dying, you can't use quick load. You need to open the savebook to go back to the last save (not so quick anymore).

Would be nice if quick load (maybe not save) is still working when death animation is happening.

Introducing an argument to access quick saves would solve this "kind of". As command line "/" works everywhere, quick saves could be loaded on startup and when Lara dies with "load q1" or "load q2" for example. Would be the easiest way to deal with it.

### Extra context

Nice feature! I use it all the time.

_No response_

rr-:
#### Checklist

#### Description

Resolves #5264.


lahm86 (approved):

aredfan (approved):
The quick load hotkey is blocked during special events like getting eaten by the T-Rex, or standing on Midas Hand.  Would we like to extend the fix to these events as well?

rr-:
Yes – good catch. Adjusted. 😌

aredfan (approved):

PR #5275, opened 2026-03-31
output: force bilinear filter in overlay images

rr-:
#### Checklist

#### Description

This forces the background images to use bilinear filtering regardless of user setting, which avoid pixellated look.
Only applies to big images like title screen backgrounds and credit images.
Does not affect patterns which continue to respect the setting.


lahm86 (approved):

aredfan:
> Does not affect patterns which continue to respect the setting.

- [ ] Toggling the `Texture filter` option has no effect on the pattern background, it remains on `Point` regardless of the config setting.

rr-:
I had left another erroneous comment coming from testing on the wrong branch, which I had since deleted.

The issue is now fixed

aredfan (approved):

PR #5368, opened 2026-04-09
objects/crawler_mutant: add Dying Mutant control

lahm86:
#### Checklist

#### Description

This adds control for the Dying Mutant in Antarctica, which uses the same gas death control as the Crawler.

I don't think OG included it as a kill, do we want to change that?


rr- (approved):
Since the mutant dies as soon as it's activated I'd say no, and exclude it from max kills. This will make the game a bit friendlier for pacifist runs by reducing forced kills to minimum.

lahm86:
> Since the mutant dies as soon as it's activated I'd say no, and exclude it from max kills. This will make the game a bit friendlier for pacifist runs by reducing forced kills to minimum.

Sounds good. It should be excluded already as standard, as it's not marked up as a creature. Essentially it's just a trap/hazard 😌

aredfan (approved):

PR #5371, opened 2026-04-09
objects/wasp_mutant: add wasp and emitter control

lahm86:
#### Checklist

#### Description

This adds control for the wasps in Tinnos, and their emitters.

There is what I believe to be an OG bug that I'm still trying to track down, which is causing the max kill stats to show one less than what the game produces. The first spawn in Tinnos (item 37) has its triggers set-up to spawn 20 wasps, but in fact 21 show up.


src/trx/game/objects/creatures/wasp_mutant.c
```diff
@@ -0,0 +1,298 @@
...
+{
+    const ITEM *const lara_item = Lara_GetItem();
+    const int32_t dx = lara_item->pos.x - item->pos.x;
+    const int32_t dz = lara_item->pos.z - item->pos.z;
+    if (dx < -0x4000 || dx > 0x4000 || dz < -0x4000 || dz > 0x4000) {
+        return;
+    }
+
+    SPARK *const spark = Sparks_GetFreeSpark();
+    spark->on = 1;
```
rr-: `true`

rr- (approved):

lahm86:
The issue was non one-shot triggers, so the longer the player stands on one of these the more the item's timer gets reset and hence it messes with the spawn numbers. This is now standardized so the item is always one shot when activated, which makes sense to be able to support finite spawn counts.

src/trx/game/objects/creatures/wasp_mutant.c
lahm86: Fixed 😌

aredfan:
Issues (testing in progress):

- [x] In room 44 where Lara first encounters the wasps, 2 spawn as a time in OG, whereas 3 spawn as a time in TRX.
- [x] When a wasp is active, save and reload, and the wasp is gone.  Lara is still able to target where the wasp used to be if she has her weapon out.

lahm86:
Thanks, @aredfan - those two issues should be fixed (will need fresh saves).
I also noticed something else that I think is OG (briefly mentioned on Stella's site). If you have two active and take them out with the rocket launcher, then take the next one that spawns out with a rocket as well, no more seem to spawn again (this is the case in tomb3). I've put in a fix for that, so that shouldn't happen now in TRX. If that's an OG bug you're aware of, I'll mark it up in the changelog.

aredfan:
> Thanks, @aredfan - those two issues should be fixed (will need fresh saves). I also noticed something else that I think is OG (briefly mentioned on Stella's site). If you have two active and take them out with the rocket launcher, then take the next one that spawns out with a rocket as well, no more seem to spawn again (this is the case in tomb3). I've put in a fix for that, so that shouldn't happen now in TRX. If that's an OG bug you're aware of, I'll mark it up in the changelog.

Thank you @lahm86, I'll check out the new build in a short while.
I can confirm this bug doesn't happen in OG. I've attempted it in room 44.

aredfan:
- [x] If Lara crosses a wasp spawner trigger in room 208, then another one in 210, it causes the current wasps to despawn, and Lara is able to target one where the wasps were despawned.

> [record.txt](https://github.com/user-attachments/files/26638725/record.txt)

lahm86:
Should be fixed now. I think a nice future enhancement could be to lift the limit so that if you leave some behind alive and cross a trigger for another nest, it'll spawn new ones and still leave the originals alive. We can make a note of that one.

aredfan (approved):

PR #5375, opened 2026-04-11
objects/fire_head: add Fire Head control

lahm86:
#### Checklist

#### Description

This adds control for the Fire Head traps in Tinnos.


rr- (approved):

lahm86:
^ Just a rebase.

aredfan:
- [x] In Highland Fling room 128, the fire head doesn't seem to set Lara on fire.  I've checked OG LA just in case and Lara does catch fire when standing on the tile at these coordinates `/30 -3 69`.

lahm86:
Thanks - I fear there is another LA hard-coded difference; it must use different Y range to determine the hit, as the item's Y pos is much lower than the mouth. I'll have a think for an elegant solution for better detection to cover base game and LA.

lahm86:
Got something in place @aredfan - she will catch fire at the given coords now, and the test for Tinnos should match up.
Do you know how far away from the mouth (i.e. towards the water) is the safe zone in OG? It seems she can be quite close in TRX and not catch fire, I'm wondering if LA also adjusted the "deadliness".

aredfan:
Thank you @lahm86, LGTM. 👍 

I made a OG video showing roughly where Lara can catch fire and how quickly her health drains. It's the same behaviour in TRX as far as I can tell.

<details><summary>OG video</summary>

https://github.com/user-attachments/assets/27e81ddd-6d13-4cf8-9ef6-0eada6c7358a

</details>

If Lara is standing any further back, there's no damage.

aredfan (approved):

lahm86:
Thank you 😌

PR #5382, opened 2026-04-14
savegame: fix reading item flags

lahm86:
#### Checklist

#### Description

This ensures that TR3's special one shot flags are retained on items when loading a save. I'm wondering if we can remove the masking on load but I'm not sure of the consequences, so this is probably safest for now.


aredfan (approved):

aredfan:
Good catch. 👍

rr- (commented):
Let's remove masking and see what happens. Maybe pass it through an ai to look for problems preemptively. 

(IMO)

lahm86:
AI agrees 😌

rr- (approved):

PR #5384, opened 2026-04-14
savegame: handle item count differences

lahm86:
#### Checklist

#### Description

This improves loading saves when the number of items in the level has changed after making the save. Item names (when set) and indices are now stored in the save, to loosen the load system relying on array order. The injection tool handles ensuring any new items are assigned unique names.

Attached are a couple of test injections for City of Khamoon. One places a pickup in room 0, the other a pickup in room 25. Combinations of these and the existing mummy injection can be used to test the functionality e.g.

- reference the test bins in the game flow
- launch Khamoon with the PS1 enemy option enabled
- trigger the mummy
- save
- disable the mummy option
- load
  - pickups should exist, mummy does not
- re-enable the mummy option
- exit game
- comment out one of the test injections
- load the previous save
  - one pickup exists, and the mummy is present

[khamoon_test.zip](https://github.com/user-attachments/files/26718307/khamoon_test.zip)

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/246


aredfan (approved):

src/trx/game/inject/common.c
```diff
@@ -15,7 +15,7 @@
 #include <string.h>
 #include <zlib.h>
 
-#define M_INJECTION_CURRENT_VERSION 6
```
rr-, line 18: Let's move this define into the enum.

rr- (commented):
This approach ties heavily with lua feature of naming items. Are we sure we want that?

rr- (approved):

lahm86:
My thinking was to be able to support extra items being added in an arbitrary order, we would need a string identifier for them. Rather than introducing another field, using the name I felt made sense as this is essentially an id in practice. If by chance a name does get changed, the save will fall back to the index (which, granted, is then subject to injection order again, but it's a fallback at least).
Open to alternatives of course. Maybe an additional id string after all, which would be invisible or read-only to Lua?

rr- (approved):
Nah it's good. I didn't think of reordering :)

PR #5388, opened 2026-04-15
data: add crystals to UB, TR2 and GM

### LINKED ISSUE #1525: Add UB save crystals
I'm not 100% how desirable this is, but I thought I'd collect what there is to know, for your consideration.

This was the first fan conversion of UB to the PS1, ca 2012: https://www.tombraiderforums.com/showthread.php?t=194134

It changed/added:
- Save crystal positions
- Loading screens for UB (we might look at these separately)
- A demoplay in Atlantean Stronghold
- Replaced some mummies with panthers

Here are its level files: [ub-ps1-levels_ghostblade.zip](https://github.com/user-attachments/files/17007322/ub-ps1-levels_ghostblade.zip)

For reference, there was also a [second UB PS1 conversion by b122251](https://www.tombraiderforums.com/showpost.php?p=8168525&postcount=652), ca 2020. Some aspects of it are different, but it takes over the save crystals from the 2012 release.

The save crystals can easily be extracted from the above level files using https://github.com/chreden/trview/

image

There's then also a Saturn conversion of UB by jonathanrij (2023): https://youtu.be/ImmNJRxv6yQ . The Saturn data can't be accessed easily sadly as there is no tooling for it (other than OpenLara). But I imagine since jonathanrij also worked on the previous conversion with Ghostblade, he likely took the save crystals one by one.

-----

Since the save crystals are now receiving reflections and their OG purple colour in TR1X, I would imagine a player combining the crystal save mode and UB should ideally find crystals in those levels. In fact, I wouldn't know how else you could save in UB otherwise. If the decision is against adding the crystals, then UB levels would need to default to PC style saving always.

Needless to say, we could also just decide on our own TR1X crystals for UB and add some, if we're not comfortable to take from others.

lahm86:
Resolves #1525.

#### Checklist

#### Description

This adds crystals to Unfinished Business, TR2 and The Golden Mask. UB locations were taken from the community mod linked in the OP; TR2 base locations were taken from the beta version, with some minor modifications where geometry differs; some TR2 levels were missing locations - thanks to @BradTRe for helping to fill the gaps here.

The TR2 beta model was the same as TR1 PC, so appearance is identical in the injections. The purple tint option still applies.

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/247


rr- (approved):

aredfan:
Testing (in progress):

- [x] The TR2 demos are missing the injection for the crystals and colour, if we'd like to include those.
- [x] One of the save crystals is invisible in Wreck of the Maria Doria.
- [x] The Dragon's Lair doesn't appear to have a save crystal.

lahm86:
Should be all fixed. Some of the room numbers were wrong on the locations so it would have been making `tp` fail and causing invisibility. Lair has just one crystal; the only level that has none should be Home Sweet Home (which feels like an intentional move imo, but LMK what you think).

aredfan:
- [x] ~~Offshore Rig~~ Diving Area room 35, triggering the flipmap may be removing the crystal.

lahm86:
Thanks, that one was placed in the flipped room indeed. Fixed.

I've also fixed another issue:
- `/set cine 1`
- `/play 17`
- level skip
- save in HSH when prompted
- `/set cine 0`
- load the save
- Lara resumes her cinematic sequence, but the camera does not

Until we support saving during cinematics, I've changed the automatic save prompt to wait for these to finish.

aredfan (approved):

lahm86:
Fixed one more from the beta in Barkhang room 70 - Lara couldn't get below the lantern.

PR #5405, opened 2026-04-19
lara/col/crouch: fix crawling underwater

lahm86:
#### Checklist

#### Description

This prevents Lara from crawling and crouch-rolling too far into water or swamp rooms. Let me know if we think this should be more or less tolerant in terms of how far she can go into the water. Also not sure about making it optional, given it can cause issues with Lara not being able to use her hands if she ventures too far.


aredfan:
> This prevents Lara from crawling and crouch-rolling too far into water or swamp rooms. Let me know if we think this should be more or less tolerant in terms of how far she can go into the water. Also not sure about making it optional, given it can cause issues with Lara not being able to use her hands if she ventures too far.

I think it's a good limit for now.  We can adjust in the future if the need arises.

aredfan (approved):

rr- (approved):

PR #5407, opened 2026-04-20
camera/box_camera: guard no height in shift

lahm86:
#### Checklist

#### Description

This guards `g_Camera.shift` against using `NO_HEIGHT` in some rare cases. This is used only in the TR1/2 camera; TR3's camera bakes the shift into the Y pos as standard, and has its own strict checks in place already. Although we've only noticed it in the Venice demo since 1.2 (81e5d1e), this is ultimately an OG bug, I just think it's quite rare.

### DIFF
--- a/src/trx/game/camera/box_camera.c
+++ b/src/trx/game/camera/box_camera.c
@@ -518,26 +518,33 @@ static void M_Move(const GAME_VECTOR *const target, const int32_t speed)
         }
     }
 
-    height -= STEP_L;
+    if (height != NO_HEIGHT) {
+        height -= STEP_L;
+    }
     if (pos.y >= height && target->y >= height) {
         LOS_Check(&g_Camera.target, &pos, false);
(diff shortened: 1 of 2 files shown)


src/trx/game/camera/box_camera.c
```diff
@@ -518,26 +518,36 @@ static void M_Move(const GAME_VECTOR *const target, const int32_t speed)
         }
     }
 
-    height -= STEP_L;
+#define L_APPLY_CLEARANCE(v, step)                                             \
+    do {                                                                       \
```
rr-: I think it's a bit overengineered; I'd inline it.

rr- (approved):

src/trx/game/camera/box_camera.c
lahm86: Thanks, inlined.

aredfan (approved):

PR #5424, opened 2026-05-01
objects/wasp_emitter: improve spawning

lahm86:
#### Checklist

#### Description

This improves wasp emitter spawning by ensuring that each nest has the default number of items allocated, as opposed to only three being shared throughout a single level.


rr- (approved):
Do we need the same treatment to the raptor emitter?
Will this break existing savegames?

lahm86:
> Do we need the same treatment to the raptor emitter? Will this break existing savegames?

I think raptor emitters are OK as they don't have a shared max active count between the emitters. It would break existing saves there; I think we're OK to introduce this at this point for wasps as Antarctica is unreleased.

aredfan (approved):

PR #5425, opened 2026-05-01
inventory: improve ammo tracking; fix harpoons

rr-:
#### Checklist

#### Description

- Fixes UPV not loading loose Harpoons
- Fixes loose Harpoons displaying 2 instead of 3 for each pack (OG bug for TR2, TRX bug for TR3)


rr-:
Testing will require paying special attention to ammo multipliers in the following places:

- [x] loose ammo pickup inventory counter for all weapons
- [x] ammo + weapon pickups inventory counter for all weapons
- [x] save/reload restores ammo correctly
    - [x] loose ammo
    - [x] guns
- [x] **existing saves loading ammo correctly**
    - [x] loose ammo
    - [x] guns

lahm86 (approved):

victor0596:
Even after this update, I still can't shoot the harpoon in UPV on an existing save file. I can only shoot at newly picked harpoons; the ones I already have in my inventory don't work.

rr-:
> Even after this update, I still can't shoot the harpoon in UPV on an existing save file. I can only shoot at newly picked harpoons; the ones I already have in my inventory don't work.

Thanks. This particular problem for existing saves will not be fixed retroactively as adding a migration path here is not feasible.

victor0596:
> > Even after this update, I still can't shoot the harpoon in UPV on an existing save file. I can only shoot at newly picked harpoons; the ones I already have in my inventory don't work.
> 
> Thanks. This particular problem for existing saves will not be fixed retroactively as adding a migration path here is not feasible.

Thank you for your reply. Is there a way to remove the harpoon ammunition from my inventory and add it back?

rr-:
Not at the moment but you could use Restart Level feature to not lose much progress.
Alternatively I can edit the files manually for you but the soonest I can do that is Sunday.

aredfan (approved):

victor0596:
I understand. So I'll restart the level so it doesn't require any further work for you. Thank you again.

PR #5243, opened 2026-03-28
tr3/data: inject pickup aid sprite

lahm86:
Resolves #5239.

#### Checklist

- [x] I have read the [coding conventions](https://github.com/LostArtefacts/TRX/blob/develop/docs/CONTRIBUTING.md#coding-conventions)
- [x] I have added a changelog entry about what my pull request accomplishes, or it is an internal change
- [x] I have added a readme entry about my new feature or OG bug fix, or it is a different change

#### Description

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/238

PR #5435, opened 2026-05-03
objects/pickup: fix twinkles + quest item status

rr-:
#### Checklist

#### Description

This PR fixes Hand of Rathmore rotating in Sleeping with the Fishes, and quest items not showing pickup aid twinkle effects.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,4 +1,8 @@
 ## [Unreleased](https://github.com/LostArtefacts/TRX/compare/trx-1.6...develop) - ××××-××-××
+**TR3**:
+- fixed Hand of Rathmore rotating in Sleeping with the Fishes
+
+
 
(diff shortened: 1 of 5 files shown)


lahm86:
The quest items were intentionally excluded from having pickup aids - https://github.com/LostArtefacts/TRX/pull/5243#issuecomment-4148473590 - but I realise this causes an inconsistency with the item in Sleeping with the Fishes. I don't mind either way tbh ☺️ 
Maybe an alternative solution would be injecting the Hand as a regular puzzle type (at the cost of impacting saves of course, but maybe that's OK at this stage).

aredfan:
Issues:

- [x] In Meteorite Cavern, Lara starts on the trigger which activates the quest items, but this has no effect as the items remain static.
- [x] When a boss drops a quest item, it remains static. Tested with Tony, Puna and Sophia.

rr-:
Thanks. I've reverted the twinkles and fixed aredfan's reports.

lahm86 (approved):

aredfan (approved):

PR #5446, opened 2026-05-05
tr3/data: fix RX-Tech texture issues

### LINKED ISSUE #5432: [Bug] RX Tech Mines submarine ropes are cut
### Game type

Original game

### Game

TR3

### TRX version

1.6

### Operating system

Windows 10

### GPU

GTX 1050ti

Might not even be a bug, but is the rope meant to be cut when Lara activates the submarine ship near the end of RX Tech Mines?

Rope stays as it is when the submarine drops

### Attachments

Screenshot taken from the latest video, where I noticed the rope bug

Image

### Before submitting

lahm86:
Resolves #5432.

#### Checklist

#### Description

This fixes the cut ropes in RX-Tech Mines where the sub is lowered; static meshes can't be used here as there is one room in the pool that doesn't flip, so the solution is to use animatings that are activated when the puzzle items are used.

I couldn't unsee texture problems in this area so have fixed those too along with the others in this level that aredfan had pointed out as below.

- [x] Cut ropes (#5432)
- [x] Misaligned ropes (https://github.com/LostArtefacts/TRX/discussions/3854#discussioncomment-16815854)
- [x] RX section of https://github.com/LostArtefacts/TRX/discussions/3854#discussioncomment-15796421

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/255


aredfan (approved):
Very diligently done, thank you. 👍

rr- (approved):

PR #5455, opened 2026-05-07
config: refactor enforced config

rr-:
#### Checklist

#### Description

This PR refactors the enforced config handling to use the push/pop stack-based API introduced in the last PR.
Additionally, the `/set` command now disallows overriding enforced settings (including the ones enforced during demos), unless the user adds `-f` / `--force` argument, in which case the game changes the runtime value of the config option without changing underlying value.

#### Testing

- [x] Launch TR1 or TR2 with a gameflow that enforces a gameplay option, then open the settings UI.
  Expected outcome: the enforced option is shown as locked/read-only and cannot be changed from the UI.
- [x] In the console, try changing a gameflow-enforced option with `/set {option} {value}`.
  Expected outcome: the command is rejected with an enforced-option message, and the setting remains unchanged.
- [x] In the console, change the same enforced option with `/set --force {option} {value}`.
  Expected outcome: the command succeeds for the current session.
- [x] Repeat the forced change with `/set {option} {value} --force`, `/set {option} --force {value}`, and `/set {option} {value} -f`.
  Expected outcome: all forms are accepted and behave the same as `/set --force {option} {value}`.
- [x] Force-change an enforced option, exit the game, then launch it again with the same gameflow.
  Expected outcome: the forced value is not saved over the user config while the option remains enforced.
- [x] Start a demo, change a demo-enforced setting (such as wall glitch mode) during playback (via `/set --force`), then end the demo and exit.
  Expected outcome: demo-enforced gameplay options are restored after playback and are not written into the user config.
- [x] Start a demo, change a normal setting (such as FPS counter) during playback, then end the demo and exit.
  Expected outcome: non-enforced settings are saved and restored upon game restart.
- [x] Play with the normal non-enforced settings in the UIs of each type (string, integer, decimal, percentage, color, boolean)
  Expected outcome: no regressions around config option effectiveness and saving the values across game restarts


aredfan (approved):

rr-:
@aredfan one more thing I thought of - if switching between two mods that enforce different options correctly unlocks the old mod and locks the new mod options :)

aredfan:
> @aredfan one more thing I thought of - if switching between two mods that enforce different options correctly unlocks the old mod and locks the new mod options :)

@rr- I've tested this and the enforced options are working correctly when switching mods and between relaunches.  I couldn't find any regressions. 😌

lahm86 (approved):

PR #5489, opened 2026-05-12
Add support for custom properties

rr-:
#### Checklist

#### Description

This PR introduces a system to let objects and individual items define their own configurable data, called "object properties".

It adds a new LUA event, `before_item_setup`, that runs just before `Item_Initialise`. This is to let builders adjust default object- or item-level properties _before_ they can be read the first time (which includes `Item_Initialise` itself).

With this change, `ENEMY_DEFAULTS.md` has been replaced by `OBJECTS.md`, which automatically generates documentation for every available custom property. It handles complex defaults expressions such as `#define M_HITPOINTS (g_TRVersion == 1 ? 6 : 10)`, too.

To streamline the LUA APIs a bit, `max_hit_points` is no longer considered to be a general `Item` property from LUA's perspective, and all defaults have been migrated to this new system.
Additionally, the music played when mounting the Quad Bike in The River Ganges is no longer hardcoded and that setup has been moved to LUA. This was mostly done as a test how this system fares in real world.
Finally, I adjusted the example on how to customize max enemy HP from LUA to cover both item- and object-level customization. Both seem to be working fine.

All in all, only enemy HP and quad bike mount tracks can be customized, but in time, I expect many more properties to be migrated to this system – most importantly **enemy damage values** and **currently hardcoded level-specific behaviors** in TR3 objects.

No TombEditor integration is available at this time; the properties can only be changed from LUA. With time we may integrate with https://github.com/TombEngine/Tomb-Editor/pull/1144 by offering catalog XMLs.


aredfan (approved):

lahm86 (approved):
Looks brilliant, thank you ☺️

PR #5490, opened 2026-05-12
items/col: guard against null frames

### LINKED ISSUE #5488: [Bug] Bridge malfunction at a designated level
### Game type

Custom level

### Game

TR3

### TRX version

3

### Operating system

Windows 10

### GPU

NVIDIA GeForce GTX Ti 1660

There's a glitch and the game crashes unexpectedly in the bridge area. I've attached the video, the game file, and the game save file.

https://github.com/user-attachments/assets/61ad3fc8-7728-4c96-bc51-824de03d63b9

### Attachments

[Site.zip](https://github.com/user-attachments/files/27618164/Site.zip)

### Before submitting

lahm86:
Resolves #5488.

#### Checklist

#### Description

This avoids trying to read frame bounds in collision tests for objects that contain none.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -25,6 +25,7 @@
 - fixed water height checks when `Fix wall geometry` is not enabled (regression from 1.6)
 - fixed level music tracks continuing to play when returning to the main menu if the option not to play the title music is enabled (#5470, regression from TR1X 4.13 / TR2X 1.3)
 - fixed being unable to smash items that are placed inside geometry but have bounds that extend into regular space (#5483, regression from 1.3)
+- fixed a potential crash when colliding with items that have no animation data (#5488)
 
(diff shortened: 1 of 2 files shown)


rr- (approved):

aredfan (approved):
LGTM.

I'm not seeing the animating 6 object in-game, from the custom level.  Is this something we should worry about?

lahm86:
> LGTM.
> 
> I'm not seeing the animating 6 object in-game, from the custom level. Is this something we should worry about?

Yeah, that's to be expected. As it has no animation data, in OG it would fall forward to the next object in the list and use its frames (kind of similar to the Barkhang prayer wheels before we injected the fixes). OG handling is dangerous as there are no guarantees there is more data to follow; so the TRX draw routine already guards against it, and doesn't draw anything.

PR #5515, opened 2026-05-16
rooms: refactor trigger handling

lahm86:
#### Checklist

#### Description

This splits out the main command switch statement for trigger testing into separate module handlers. I'm not entirely sold on the idea tbh, but it's been a `TODO` for a while so we can decide if we like it or not.

There are no logical changes to the handling other than the following:
- triggering fish/piranhas is now handled in the shoal module; we allow normal trigger handling to set the timer then reset it in the module itself, similar to other objects that repurpose the timer trigger. Deactivating still requires a public function for the sake of kill-all-active triggers.
- added a small helper when triggering items to check for one shot status


rr- (approved):
I'm loving this

aredfan (approved):

PR #5549, opened 2026-05-25
objects/pickup: fix spark-based pickup aid positioning

rr-:
#### Checklist

#### Description

This PR fixes pickup aids being visible through pushblocks or via overlapping rooms.

image

We need to test occlusion in comprehensive scenarios. 
I suspect this fix will cause the pickup aids to no longer work through fences and transparent windows :expressionless:

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,5 +1,10 @@
 ## [Unreleased](https://github.com/LostArtefacts/TRX/compare/trx-1.7...develop) - ××××-××-××
 
+**TR3**:
+- fixed pickup aids appearing from hidden overlapping rooms
+
+
+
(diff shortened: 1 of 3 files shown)


aredfan:
> I suspect this fix will cause the pickup aids to no longer work through fences and transparent windows 😑

I can confirm this is indeed the case. One place to check out this scenario is in Lara's Home, `/tp 41 5 33` and turn right to face the key inside the aquarium.

- [x] I did very light testing and noticed the first large medi pack in Jungle doesn't emit pickup sparks.  `/tp i15`

rr-:
Thanks @aredfan I have changed the fix, and it seems to satisfy all scenarios so far.

aredfan (approved):
@rr- I can confirm the new fix is working great, thank you. 👍

lahm86 (approved):

PR #5573, opened 2026-05-29
objects/cobra: move radius setup to Lua

lahm86:
#### Checklist

#### Description

This replaces the hard-coded level sequence check for Cobra radius setup with a Lua approach.

Test:
- [x] Small Cobras in Nevada should behave normally (single added Lua script)
- [x] All other Cobras should behave normally (no Lua here, they will use the default setup)

### DIFF
--- a/src/trx/game/objects/creatures/cobra.c
+++ b/src/trx/game/objects/creatures/cobra.c
@@ -9,31 +9,31 @@
 static BITE m_CobraBite = { .pos = { 0, 0, 0 }, .mesh_num = 13 };
 
 typedef enum {
-    COBRA_STATE_WAKING_UP = 0,
-    COBRA_STATE_ALERT = 1,
-    COBRA_STATE_BITE = 2,
-    COBRA_STATE_SLEEP = 3,
-    COBRA_STATE_DEATH = 4,
-} M_COBRA_STATE;
+    M_STATE_WAKING_UP,
+    M_STATE_ALERT,
+    M_STATE_BITE,
+    M_STATE_SLEEP,
+    M_STATE_DEATH,
+} M_STATE;
 
 typedef enum {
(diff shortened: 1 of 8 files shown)


rr- (commented):
WDYT of exposing the radii as direct properties, and filling out `radius_setup` in `M_Initialise` from that over `has_reduced_radius`?
I'm fine with the current approach too, just thought of making things more elastic.

Thanks for the property naming cleanup.

lahm86:
> WDYT of exposing the radii as direct properties, and filling out `radius_setup` in `M_Initialise` from that over `has_reduced_radius`? I'm fine with the current approach too, just thought of making things more elastic.
> 
> Thanks for the property naming cleanup.

This is much better imo - thanks ☺️

src/trx/game/objects/creatures/cobra.c
```diff
@@ -1,10 +1,15 @@
 #include <trx/core/utils.h>
 #include <trx/game/creature.h>
-#include <trx/game/game_flow.h>
 #include <trx/game/lara.h>
 #include <trx/game/objects.h>
 #include <trx/game/spawn.h>
-#include <trx/version.h>
+
+// clang-format off
+#define M_DEFAULT_ALERT   (3 / 2)
```
rr-: Won't this evaluate to `1` instead of `1.5`?
I know the objects properties documentation generator evaluates it to 1.5 but because it runs in Python, it does automatic type promotion; I don't think it'll work in C the way we'd want it to and will actually result in a mismatch.

lahm86: Fixed, thanks. Got rid of the fractions in Lua as well for consistency.

aredfan (approved):

rr- (approved):

PR #5576, opened 2026-05-30
objects/pickup: add fast pickup animation

### LINKED ISSUE #1365: Feature request: Add option for faster pickup animation from TR4+
Lara's TR4+ pickup animation is quicker and more efficient than the TR1-3 animation, so an option to use it instead (while keeping her TR1-3 ability to pick up multiple items at once) would be a great QoL feature imo.

lahm86:
Resolves #1365.

#### Checklist

#### Description

This allows Lara to pick items up more quickly, similar to TR4+. The flare pickup is still as per OG TR2/3, which I think is also the case in TR4/5.

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/274


src/trx/game/objects/general/pickup.c
```diff
@@ -464,7 +482,14 @@ static void M_DoAboveWater(const int16_t item_num, ITEM *const lara_item)
...
             Lara_AlignPosition(item, &m_PickupPosition);
-            Lara_AnimateUntil(lara_item, LS(LS_PICKUP));
+            const LARA_TRX_STATE goal_state =
+                g_Config.gameplay.enable_fast_pickups && !is_ducked
+                ? LS_FAST_PICKUP
+                : LS_PICKUP;
+            lara_item->goal_anim_state = LS(goal_state);
+            do {
+                Lara_Animate(lara_item);
+            } while (lara_item->current_anim_state != LS(LS_PICKUP));
```
rr-, line 492: Shouldn't this use `goal_state`?

rr- (approved):

src/trx/game/objects/general/pickup.c
lahm86: This is a slight shortcut to avoid having to replicate `LS_FAST_PICKUP` in other places where `LS_PICKUP` is checked/used (like gun control, flare control, col/state routines etc) - so the fast anim uses the same state as the slow, but we need to have two different goals in the stand idle animations to pick the correct one. `Lara_AnimateUntil` expects goal and target anim state to be the same, so this is just a variant to allow a different goal to the actual target.

aredfan:
> The flare pickup is still as per OG TR2/3, which I think is also the case in TR4/5.

I can confirm this is indeed the case. 👍

aredfan (approved):

PR #5578, opened 2026-05-31
objects/orca: add collidable Dolphin alias

lahm86:
#### Checklist

#### Description

This adds `O_DOLPHIN` as an effective alias for `O_ORCA` but it has collision while underwater to match Sleeping with the Fishes. This is similar to the injection replacements we do in TR2GM for e.g. spider => wolf. This in turn allows us to eliminate `GF_BadIsMod`.

Existing savegames should continue to work.

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/275


src/trx/game/objects/creatures/orca.c
```diff
@@ -185,4 +175,25 @@ static void M_Setup(OBJECT *const obj)
...
+}
+
+static void M_SetupDolphin(OBJECT *const obj)
+{
+    if (!obj->loaded) {
+        return;
+    }
+
+    M_SetupCommon(obj);
+    obj->collision_func = M_Collision;
```
rr-: `M_CollisionDolphin` imo

rr- (approved):

src/trx/game/objects/creatures/orca.c
lahm86: Thanks, fixed.

aredfan:
It might be worth removing these lines in games/tr3-la/strings.json5 as well as other languages if present.

https://github.com/LostArtefacts/TRX/blob/1eb7581b5d18e4f89c6a2bf4f60cea22687f2bda/data/trx/ship/games/tr3-la/strings.json5#L4-L7

lahm86:
> It might be worth removing these lines in games/tr3-la/strings.json5 as well as other languages if present.
> 
> https://github.com/LostArtefacts/TRX/blob/1eb7581b5d18e4f89c6a2bf4f60cea22687f2bda/data/trx/ship/games/tr3-la/strings.json5#L4-L7

Good catch, although removing it would mean `/tp dolphin` and suchlike would fail. I've updated it though to include the translation for all languages in the LA strings for that level only; this is similar to the GM setup.

aredfan (approved):

PR #5582, opened 2026-05-31
objects/shoal: improve usability and fix textures

### LINKED ISSUE #4358: Feature Request: Fish emitters in TR1 & 2
Would it be difficult to implement fish emitters in TR1 and TR2, respectively? Not piranhas, just fish wandering around for aesthetic purposes. TR2 underwater custom levels would heavily benefit from this! It could use sprite sequences like the fire and other emitters do.

lahm86:
Resolves #4358.

#### Checklist

#### Description

This improves usability of shoals in custom levels and fixes the piranha texture bleeding in [aredfan's findings](https://github.com/LostArtefacts/TRX/discussions/2582#discussioncomment-17104138). TR3's dedicated sprite slots are put to use to separate away from sparks, given that these objects are independent of that. By doing so, it lifts the limitation of only supporting two texture variants for tropical fish in any level. Piranhas remain limited to two though given that they interchange per fish rather than between shoals.

Below is the corrected UV mapping for reference, so each fish has its own space and the triangles do not bleed into each other. Note that the first tropical fish ended up squished in OG because of the mapping, so this is now resolved.

<details>

Screenshot 2026-05-30 222243
</details>

I've made a TR1 test level available with 6 tropical fish variants (plus piranhas).

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/276


rr- (approved):

aredfan (approved):
Thank you for fixing the texture issues. 👍

PR #5588, opened 2026-06-01
level/gameplay_objects: remove hard-coded AI bit handling

lahm86:
#### Checklist

#### Description

This removes the hard-coded AI Patrol 1 handling in levels 14 and 15 by injecting additional patrol items, such that enemies placed in the same position will continue to pick up the AI bits and then use the other item as their pathing target.

Other levels, like Aldwych room 37, have setups as below. The enemy is assigned `AI_PATROL_1` from their tile, that item gets killed then they begin patrolling between the other `AI_PATROL_1` and `AI_PATROL_2` items.

`AI_PATROL_1` ----------- `AI_PATROL_1` + `enemy` ----------- `AI_PATROL_2`

HSC and Area 51 use the following setup, so the item was used to assign but also retained. I'm inclined to think this was because the levels are quite close to the original item count limit so they couldn't afford to use similar setups to Aldwych.

`AI_PATROL_1` + `enemy` ----------- `AI_PATROL_2`

I confirmed there are no instances of two or more of the same AI objects on the same tile in any of the OG levels.

A good way to test this is in HSC, go into room 49 and then `/trigger 17`. The guard should patrol as normal.

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/278


rr- (approved):
I don't understand anything, but I'm trusting no regressions were introduced 😅

docs/CHANGELOG.md
```diff
@@ -30,6 +30,7 @@
 - removed the hard-coded strobe light alarm behaviour in Area 51 and moved it to Lua instead
 - removed hard-coded level sequence specific sound effects for animated spikes and moved to animation commands instead
 - removed the hard-coded constraint of only being able to use animated spikes in levels 5 and 7
+- removed the hard-coded `O_AI_PATROL_1` behaviour in High Security Compound and Area 15; refer to migration docs
```
aredfan: `Area 51` 👽

lahm86: Fixed, thanks. I wonder what's in Area 15 👀

aredfan (approved):

PR #5596, opened 2026-06-02
objects/vehicles: move music tracks to Lua

### LINKED ISSUE #4355: Add an option to turn off the hardcoded Jetski music when Lara gets on the vehicle.
The skidoo music plays by default and I don't think you can turn it off currently.

rr-:
#### Checklist

#### Description

This PR makes Snowmobile, mine cart, and RIB music tracks configurable through Lua bringing their setup in line with the existing Quad Bike pattern.

**NOTE:** the properties for the new vehicles provide default values compatible with the OG. This is _different_ from the Quad Bike which got moved entirely to Lua and by default doesn't play anything. Unlike other vehicles, Quad Bike only plays some music in Ganges; on the other hand, the rest of the vehicles only appear in their single respective levels. LMK if we want to change anything here.

**NOTE 2:** I was thinking of introducing slots in the catalog `O_QUAD_BIKE_THEME_1-4` etc., and forgoing custom properties altogether. The problem with this is that we lose ability to change the music tracks from level to level and thus I decided against it.

#### Testing

- [x] Mount a Quad Bike
- [x] Mount a Snowmobile
- [x] Mount a mine cart
- [x] Mount a RIB

Expected outcome: no regressions; TR3 vehicles play more than 1 random tracks

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -4,10 +4,10 @@
 - added `O_DOLPHIN`, which behaves in the same way as `O_ORCA` but has collision underwater
 - added a new console command - `/burn` - to toggle whether or not Lara is on fire
 - added poison Lua property to `O_DART` and `O_DISC`, so that they can poison Lara just like `O_POISON_DART`
-- removed hard-coded fish and piranha setup and moved to Lua instead
-- removed the limit of at most 8 fish/piranha shoals per level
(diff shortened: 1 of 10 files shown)


lahm86 (approved):
Thanks. I am happy re both the notes above, I don't think anything needs to change there.

aredfan (approved):

PR #5598, opened 2026-06-02
objects/lift: add support to move vehicles

### LINKED ISSUE #4353: Make Jetski move up and down on the elevator when Lara isn't riding it.
Currently you can ride up and down in the elevator on the jetski. Similar to the new pushblock mechanic it would be useful to be able to have the jetski move with the lift for potential puzzles and to allow Lara to trigger the elevator with a switch still.

lahm86:
Resolves #4353.

#### Checklist

#### Description

This allows lifts to notify items within or on top that the floor height has changed when they move. Individual items can then react as necessary. This currently applies to skidoos and quad bikes only when Lara is not mounted (mounted vehicles already respond to the height change).

I don't think anything else can technically end up inside a lift - enemies would never go there for example; however, the setup should in theory make it straight-forward to add other items if ever needed.

TR2 test level available.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -4,6 +4,7 @@
 - added `O_DOLPHIN`, which behaves in the same way as `O_ORCA` but has collision underwater
 - added a new console command - `/burn` - to toggle whether or not Lara is on fire
 - added poison Lua property to `O_DART` and `O_DISC`, so that they can poison Lara just like `O_POISON_DART`
+- added the ability for skidoos and quad bikes that are inside or on top of lifts to move when the lift itself moves (#4353)
 - changed hard-coded music tracks from Snowmobiles, mine carts, and RIBs to be configurable via Lua
 - changed hard-coded fish and piranha setup to be configurable via Lua
 - changed hard-coded small Cobra radius setup to be configurable via Lua
--- a/src/trx/game/interpolation.c
+++ b/src/trx/game/interpolation.c
@@ -339,7 +339,8 @@ static void M_InterpolateItems(const double ratio)
     for (int32_t i = 0; i < Item_GetTotalCount(); i++) {
         ITEM *const item = Item_Get(i);
         if (((item->flags & IF_KILLED) || item->status == IS_INACTIVE)
-            && i != lara_vehicle_num) {
+            && i != lara_vehicle_num
+            && XYZ_32_AreEquivalent(item->pos, item->interp.prev.pos)) {
             M_CommitItem(item);
         } else {
             M_InterpolateItem(item, ratio);
--- a/src/trx/game/objects/general/lift.c
+++ b/src/trx/game/objects/general/lift.c
@@ -8,27 +8,29 @@
 #include <trx/game/objects.h>
 #include <trx/game/objects/traps/movable_block.h>
 
-#define LIFT_WAIT_TIME (3 * LOGIC_FPS) // = 90
-#define LIFT_SHIFT 16
-#define LIFT_HEIGHT (STEP_L * 5) // = 1280
-#define LIFT_TRAVEL_DIST (STEP_L * 22)
-#define M_LIFT_NUM_FLOOR_SECTORS 4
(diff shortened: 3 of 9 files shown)


rr- (commented):
I would like to try and avoid moving Lara temporarily to the spot where we want to test for the item, as this sounds like a real design smell worth of getting right now that we expand lift's responsibility.

The directions I'd consider:

1. Add a more robust query context to room-height APIs.  
   `Room_GetHeightEx(…, const ROOM_QUERY *query)` and `floor_height_func(…, const ROOM_QUERY *query)`. The query would carry things like `subject_item`, `subject_pos`, and `ignore_item_num`. Then the lift can base its special-case logic on `query->subject_pos` or `query->subject_item` instead of `Lara_GetItem()`. This feels like the cleanest long-term fix, and the API shape is already following this approach with the presence of `ignore_item_num`.
2. Smaller version of the same idea: add an "occupant/probe" argument only for walkable callbacks.  
   Something like `floor_height_func(item, sample_pos, height, probe_pos)`. The important distinction is between "the point being sampled" and "the actor whose collision space we're resolving". The lift is one of those objects where that distinction matters, and right now Lara is being used as an implicit global for `probe_pos`.
3. Have the lift compute the movement it just performed, package that into the event payload, and let listeners react from that payload, instead of forcing listeners to call `Room_GetHeight()` again.
   For example, send `{ old_y, new_y, delta_y, new_floor_y, new_ceiling_y }` in `event_func(…, data)`. Detached vehicles could then just apply the lift delta or clamp to the supplied floor. While much less invasive and it removes the Lara hack, it's not as elastic, as it is tightly tied to the Y axis and lift internals.

LMK WDYT :)

lahm86:
Thanks. I've gone with a point 3 approach for the time being, but I definitely agree on a bigger `Room_GetHeight` refactor so it's more aware of the context. I think that might also resolve camera issues near lifts (OG). I'm happy to look into this refactor as its own change - will log it 😌 

So for now, we just pass the Y delta but also track which items have been processed otherwise when the lift goes through portals, items could be targeted twice. The other option is just looping over all level items rather than looking through nearby rooms; perhaps the vector is the lesser of two evils there?

rr-:
> Thanks. I've gone with a point 3 approach for the time being, but I definitely agree on a bigger `Room_GetHeight` refactor so it's more aware of the context. I think that might also resolve camera issues near lifts (OG). I'm happy to look into this refactor as its own change - will log it 😌
> 
> So for now, we just pass the Y delta but also track which items have been processed otherwise when the lift goes through portals, items could be targeted twice. The other option is just looping over all level items rather than looking through nearby rooms; perhaps the vector is the lesser of two evils there?

I think we should loop through all items – reallocating a vector on every frame is much costlier. The O(n²) coming from iterating through all items' all sector walkables in practice should be almost null due to majority of items not reacting to events.
Alternatively, allocate a small fixed-size array in the lift's `M_PRIV`, which gives us memory management for free and avoid reallocs.

Let's name `loopend` `loop_end` too for consistency with existing places.

lahm86:
> I think we should loop through all items – reallocating a vector on every frame is much costlier. The O(n²) coming from iterating through all items' all sector walkables in practice should be almost null due to majority of items not reacting to events. Alternatively, allocate a small fixed-size array in the lift's `M_PRIV`, which gives us memory management for free and avoid reallocs.
> 
> Let's name `loopend` `loop_end` too for consistency with existing places.

Thanks again, updated. Just went with an all item loop, so that eliminates the label too.

rr- (approved):

aredfan:
- [x] It's possible for the vehicle to bounce when the lift is moving down. The test level will need to be added to the tr2 base game when playing the recording.

> [record.txt](https://github.com/user-attachments/files/28522781/record.txt)

lahm86:
Thanks, @aredfan - should be sorted out. I also noticed driving back out after the lift going up could end up in the skidoo going through the floor, or the quad hitting an invisible wall. That should be fixed too - portal shenanigans.

aredfan (approved):

PR #5601, opened 2026-06-03
Expose creature damage values in Lua

rr-:
#### Checklist

#### Description

This moves all creatures melee and hitscanner attacks to object properties, enabling changing their values in Lua.

Most shared projectiles like knives, harpoons, or Atlantean bombs remain hard‑coded, as the focus of this PR is placed solely on the creatures module.

One limitation still remains: enemy‑vs‑enemy damage values are heavily reduced and use inconsistent multipliers (10%, 5%, or other) across OG actors that are meant to fight with non-Lara enemies. In this PR, these reduced values continue to be stored in a non-universal way, as separate properties. This means that converting an unsupported enemy into an ally will still cause it to deal the full, overpowered damage intended for Lara instead of the scaled values for enemy combat. Addressing that issue is outside the scope of this PR.

#### Testing

I asked an LLM to look for regressions and it looks good. I suggest doing a random testing of 10-20 different creatures if they continue to behave the same way.


lahm86 (approved):
Brilliant 😃

aredfan:
- [x] TR2: I'm getting a crash at the end of the demos in the PR build.

> 1. launch tr2
> 2. /demo

rr-:
Thanks – fixed, it was a problem with worker 3 loading too eagerly after small refactors. I checked other objects for the same bug, and it was the only one.

aredfan (approved):

PR #5616, opened 2026-06-04
objects/movable_block: add continuous animation option

### LINKED ISSUE #1354: Feature request: Add option for continuous pushing/pulling animations from TR4+
Starting in TR4, Lara is able to push/pull objects continuously (i.e. without having to stop and start again after each tile), which is a great QoL change that would help make moving blocks less tedious in TR1.

lahm86:
Resolves #1354.

#### Checklist

#### Description

This adds continuous animation to pushblocks, so Lara can push/pull as long as action is held and the geometry around her allows for it, similar to TR4+. The config option allows switching in-game, so Lara's regular animations will kick in when it's disabled and her TR4 ones when enabled. Making the original TR1-3 animations seamlessly join together I don't think is feasible.

One other caveat is fly mode is unavailable while Lara is in any of the new animations. This is because the blocks track her movement rather than animating by themselves, so it would result in abandoned pushblocks and potentially softlocks. The fly cheat continues to work once she's finished, and/or if the option is disabled.

Test:
- [x] Have the option disabled and begin pushing/pulling a block. Switch the option on and keep holding action until the end of the original animation - Lara should stop normally then transition again to grab. Subsequent push/pulls should be continuous.
- [x] Have the option enabled and begin pushing/pulling a block. Switch the option off and keep holding action - Lara should not continue moving the block. Subsequent push/pulls should be OG.
- [x] Check behaviour is correct after loading a save made during a push/pull
- [x] Have the option disabled, begin pushing a block then enter the fly cheat - the block should continue to animate itself
- [x] Have the option enabled, begin pushing and try to enter the fly cheat - it should fail
- [x] anything else you can think of!

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/282


rr- (approved):
Heh @ how this works

aredfan (approved):

Issue #1596, opened 2024-05-06
OG bug: Lara retains her underwater hue when re-entering a boat


aredfan:
The first time Lara climbs into a boat from the water, her underwater hue disappear as expected.

It's when re-entering a boat that the underwater hue is retained.  Lara stays like this until she drives the boat into a different room. 

https://github.com/LostArtefacts/TRX/assets/94776985/34a7e4e8-d40e-4d71-8070-ca7178d4f4c9

https://github.com/LostArtefacts/TRX/assets/94776985/017f4fb4-3cef-43b7-a6b2-824c5d911552

PR #5630, opened 2026-06-06
lara/control: add support for damaging dry rooms

lahm86:
#### Checklist

#### Description

This fixes Lara's exposure meter not resetting when going from a damaging room to a normal room - not exactly an issue on develop in OG as the Lua scripts set the damaging flag on all rooms, and the previous control logic did the filtering on water status. I've switched it around so the OG scripts mark underwater rooms only as damaging, and this means we can now offer exposure damage when Lara is above water in custom levels.

I've prepared the change TE side to write these rooms flags, so builders will be able to use the checkboxes in the editor. I'll push that PR if the tests here look good. I've made a test level available from that build for TR1 with various cold and damaging room combos.

### DIFF
--- a/data/trx/ship/games/tr3/scripts/antarc.lua
+++ b/data/trx/ship/games/tr3/scripts/antarc.lua
@@ -1,7 +1,7 @@
 trx.events.after_level_state(function()
   for i = 1, #trx.rooms do
     local room = trx.rooms[i]
-    room.damaging = true
+    room.damaging = room.underwater
     room.cold = true
   end
 end)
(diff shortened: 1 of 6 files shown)


rr- (approved):
Would be great to control the damage value per tick too :D

aredfan:
- [x] If Lara climbs from cold water into the RIB, the cold bar meter will keep draining. It does stop if Lara moves the boat a small distance.

lahm86:
> * [ ]  If Lara climbs from cold water into the RIB, the cold bar meter will keep draining. It does stop if Lara moves the boat a small distance.

Good catch. Turns out this was the same as #1596, we've just not noticed it because of responsive mesh tinting. If you disable that you'll see Lara retains the underwater hue. Fixed, and I've added a changelog entry for it 😌

lahm86:
> Would be great to control the damage value per tick too :D

This would be great - I've added it to the tracker for now ☺️

aredfan (approved):

PR #5636, opened 2026-06-07
Fix Sleeping with the Fishes Diver disappearing too early

rr-:
#### Checklist

#### Description

This PR:

- introduces `O_ANIMATING_EXT`, that is a trimmed-down version of the `O_PROPELLER_2` object from TR3 which is often abused to do stateful animating objects
- enhances the animation logic in `O_ANIMATING_EXT` to handle objects passing through portals
- merges `O_DISPOSABLE_ANIMATING` with `O_ANIMATING_EXT`, introducing `kill_on_trigger` property to achieve feature parity
- replaces the Diver (item 38 and 39) from Sleeping with the Fishes to use the new `O_ANIMATING_EXT_1`
- replaces the beam of light (item 58) from Tinnos to use the new `O_ANIMATING_EXT_1`, and adding a small lua script to set the `kill_on_trigger` property

Related: https://github.com/LostArtefacts/TRXInjectionTool/pull/284

#### Testing

- [x] Save in Tinnos on develop; load the save on this PR
  Expected outcome: the game does not crash and the beam of light is visible
- [x] Save in Sleeping with the Fishes on develop; load the save on this PR
  Expected outcome: the game does not crash and the diver is visible
- [x] Trigger the Diver from the Sleeping with the Fishes
  Expected outcome: he passes through the portal and does not disappear before touching the water surface; the changes persist through saves
- [x] Trigger the beam of light in Tinnos
  Expected outcome: the beam disappears; the changes persist through saves


lahm86 (commented):
Looks good. The only issue is the diver reverts to the bad animation in Sleeping with the Fishes once it has finished - probably best imo to revert #5622 and use the same Lua approach as Tinnos to properly get rid of it ☺️

rr-:
Thanks.
The boolean property is used to control whether to kill the item after it gets triggered, I'd have to introduce a separate property. LMK if this is desirable.
The problem with my PR was is that it routed `O_PROPELLER_2` to `O_ANIMATING_EXT_1` before `luds_diver_animation.bin` patched the faulty `O_PROPELLER_2`'s animation – reordering fixes the problem.

lahm86:
> Thanks. The boolean property is used to control whether to kill the item after it gets triggered, I'd have to introduce a separate property. LMK if this is desirable. The problem with my PR was is that it routed `O_PROPELLER_2` to `O_ANIMATING_EXT_1` before `luds_diver_animation.bin` patched the faulty `O_PROPELLER_2`'s animation – reordering fixes the problem.

Ah of course, my bad. This looks good to me as-is, thanks 😌

lahm86 (approved):

aredfan (approved):

PR #5640, opened 2026-06-08
objects/vehicles: add static collision

### LINKED ISSUE #4285: Make jetski collide with static objects that have collision.
Currently Lara can drive through static objects on the jetski even if they have collision. I think the same can be said about the jetski enemy. Can it be changed so both Lara and the enemy will collide with static objects like they do when hitting a wall or a door. Would massively help with areas that have fences or other clutter objects scattered around that the player would collide with.

lahm86:
Resolves #4285.

#### Checklist

#### Description

This adds properties to the skidoo and quad bike to allow them to collide with static meshes. It's hard collision, so similar to hitting a wall - as such if Lara is going fast enough, she will take damage. Let me know if we think this would be better as a config option; I couldn't think of many affected places in OG (the posts near the hut in Tibet was all that came to mind) so I went with object properties instead.

This doesn't extend to the skidoo driver as it would interfere with pathfinding. Custom levels can block off specific sectors where enemies shouldn't travel anyway. If Lara is riding the black skidoo, collision will apply.

Test level available. The following commands can be run to turn the collision on.

`/lua trx.objects.skidoo_fast.properties.test_static_collision = true`
`/lua trx.objects.quad_bike.properties.test_static_collision = true`

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -15,6 +15,7 @@
 - added animated interaction support to keyholes and puzzle slots (#4471)
 - added an option to control snap interactions for targets such as pickups and switches (Gameplay → Controls → Snap interactions)
 - changed vehicles to allow defining in Lua whether or not they can activate heavy triggers; refer to migration notes
+- change the Skidoo and Quad Bike to allow defining in Lua whether or not they should collide with static meshes (#4285)
 - changed Assault Course Lua record access from `trx.assault_stats` to `trx.assault.stats`
 - changed hard-coded music tracks from Snowmobiles, mine carts, and RIBs to be configurable via Lua
 - changed hard-coded fish and piranha setup to be configurable via Lua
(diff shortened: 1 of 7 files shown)


lahm86:
^ Forgot to update `OBJECTS.md`

aredfan (approved):

src/trx/game/objects/vehicles/common.c
```diff
@@ -228,3 +228,19 @@ void Vehicle_HandleEvent(
...
+
+bool Vehicle_CollideStaticObjects(
+    const ITEM *const vehicle_item, const XYZ_32 pos, const int32_t radius)
+{
+    OBJECT_PROPERTY_VALUE value = {};
+    if (!ObjectProperty_GetItemValue(
+            vehicle_item, "test_static_collision", &value)
+        || !value.as_bool) {
+        return false;
+    }
```
rr-: About that, can we make sure that hot-path getters to properties get cached in `M_Initialise`? This will get every single frame and I'd rather not do that, since the lookup cost of each property is O(n) and isn't very friendly to cpu caching. It's fine now, with low amount of properties, but I think it's best to establish this pattern early.

lahm86: Thanks, updated.

rr- (approved):

PR #5659, opened 2026-06-12
controls: update default keys with quick save and load

Checkm8Croft:
#### Checklist

#### Description

Add function for having NUM_5 for Quick Save and NUM_6 for Quick Load as default keyboard keybinds

Devs please confirm me that these could be good keybinds or should be assigned somthing else


lahm86:
Thanks. I'm not sure if the num pad is a good idea as I think it's quite common for people to use the arrows there for movement. But equally, I'm not sure what would be good alternatives.

This would also need a changelog entry as it's a player-facing change.

Kappa971:
I think generally in older PC games that allowed quick saving, the default keys were F5 for quick saving and F6 or F9 for quick loading.

rr-:
The hotkeys stay purposefully unmapped, as there's a lack of a clear idea on what'd be the best candidates, and numpad is not suitable for the reasons mentioned.

Checkm8Croft:
Fine, my base idea was: having a default hotkey for Quick Save/Load. So no Keypad. Could we use Functions keys? F8-F9 for e.g.

Kappa971:
> The hotkeys stay purposefully unmapped, as there's a lack of a clear idea on what'd be the best candidates, and numpad is not suitable for the reasons mentioned.

>Fine, my base idea was: having a default hotkey for Quick Save/Load. So no Keypad. Could we use Functions keys? F8-F9 for e.g.

Hello? https://github.com/LostArtefacts/TRX/pull/5659#issuecomment-4696703297

rr-:
F8 and F9 are already mapped for other things.

Kappa971:
> F8 and F9 are already mapped for other things.

I looked at some older PC games that support quick saving and these are the keys:
| Game | QuickSave | QuickLoad
--------|-------------|------------|
|Splinter Cell 1-4 	|F5 |F8
|Max Payne 1-2		|F5 |F9
|Unreal 2		|F5 |F8
|Doom 3			|F5 |F9
|Quake			|F6 |F9
|Half Life		|F6 |F7
|Tomb Raider 6		|F5 |F9

rr-:
Thanks. This highlights my point. Technically tr 1-3 f5 and f6 is already considered a quick saving compared to passport, and also it's prevalently f5/f9 so a mix. I don't think there's anything to change here.

PR #5665, opened 2026-06-14
inject/floor_data: re-assign cold and damaging flags

### LINKED ISSUE #5664: [Bug] Lara drowning in The Lost Valley in T-Rex room when high up
### Game type

Original game

### Game

TR1

### TRX version

1.8

### Operating system

Windows 11

### GPU

NVIDIA GeForce GTX 1650 SUPER

Lara starts to lose oxygen and her air bar starts to deplete once she is high up enough in the T-Rex area in TR1 level The Lost Valley.

1. Load attached save
2. Jump into the air from the location you start in upon loading the save.
or proceed towards the secret area above where you acquire one of the cogs.
3. Observe Laras air bar depleting.

Laras air bar should not be depleting because she is not underwater.

### Attachments

[TR1 lost valley drowning in air bug.zip](https://github.com/user-attachments/files/28919511/TR1.lost.valley.drowning.in.air.bug.zip)

### Before submitting

lahm86:
Resolves #5664.

#### Checklist

#### Description

This re-assigns cold and damaging injected flags to unused values; the previous values are referenced in several original levels. Bits 1 and 2 are unused throughout the classic TR series.

The problem was a result of the skybox injections respecting existing room flags, even the ones that were meaningless in the original engines. We don't look at cold/damaging during normal level reading, so hence these flags were slipping in via injections.

~~If we'd prefer to reset the flags data-side, let me know. There are however a lot of rooms in TR1-3 (and 4/5) that have them set, so I'm not sure if it's a good approach.~~

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/290


aredfan (approved):

rr- (approved):
What a mess…

rr-:
TBH I'd rather repackage the flags so that they're always predictable when they come from the injections, rather than rely on the mess that is the OG data.

lahm86:
> TBH I'd rather repackage the flags so that they're always predictable when they come from the injections, rather than rely on the mess that is the OG data.

Makes sense, I can sort that out. Do you think we should go back to `0x100/0x200`, or stick with the lower bits?

rr-:
Either is fine so long as it's under our control :)

lahm86:
Stuck with the low bits, at least we're in a nice order there now :)

Injections updated where we set any room flag, so any that are unrecognised are now stripped. @aredfan - would you mind sanity checking these cases?

- [x] TR1 skyboxes and windy rooms (if one works, they all should)
- [x] TR2 adding water to the plane area in Opera House and the unreachable shark in Wreck
- [x] TR2 missing skybox in Kingdom room 63
- [x] TR3 Thames Wharf cutscene - we removed wind from the bell tower to avoid rain showing there

rr- (approved):

aredfan (approved):

PR #5671, opened 2026-06-14
inject: add texture applicability tests

### LINKED ISSUE #5669: [Bug] Bear textures
### Game type

Original game

### Game

TR1

### TRX version

1.8

### Operating system

Windows 11

### GPU

NVIDIA GeForce GTX 1650

Bear has bat's textures on its limbs.
1. Launch TRX.
2. Start New Game
3. Reach bear pit.

Image

Bear has its standart textures.

### Attachments

Image

[TRX.log](https://github.com/user-attachments/files/28928926/TRX.log)

### Before submitting

lahm86:
Resolves #5669.

#### Checklist

#### Description

This improves applicability tests for texture page overwrite injections by allowing for particular tex infos from the injection source levels to be compared with the runtime levels. Unfortunately this means if there is a mismatch, no texture fixes will be applied (so including wall faces and suchlike). I think we have to guard it this way to avoid this problem continuing to appear.

Unfortunately this impacts a lot of injections, but for testing it should only need a verification that fixes are applied when they should e.g. double check a single face in a level. I'll share a test level that should fail.

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/292


rr- (approved):

lahm86:
^ Restored explicit enum IDs per https://github.com/LostArtefacts/TRXInjectionTool/pull/292#pullrequestreview-4493377083

aredfan (approved):

PR #5678, opened 2026-06-15
objects/generic_trap: add generic trap control

lahm86:
#### Checklist

#### Description

This adds a template for generic trap slots in custom levels, allowing the fundamental properties to be set in Lua and for making no assumptions about animations other than having an on/off state. Test level available with 6 trap types, with various different property setups you can see in the Lua script there.

LMK what we think of this approach. I was able to implement three TR4 traps in the test level, more or less like for like, so I think it will be useful to builders. The only issue I'm having is re `OBJECTS.md` population, as these slots aren't used in OG; I compared with the default traps and each one has something specific enough to keep it separate I felt. Maybe we could have the tooling extend to look for objects like this and have them in their own section?


rr- (approved):
I'll think of something for the docs. Please create a ticket and assign it to me:)

lahm86:
> I'll think of something for the docs. Please create a ticket and assign it to me:)

Thanks!

aredfan (approved):

PR #5684, opened 2026-06-17
lara/flare: set arm before control

### LINKED ISSUE #5683: [Bug] Crash / corrupted quicksave in High Security Compound
### Game type

Original game

### Game

TR3

### TRX version

1.6

### Operating system

Windows 10

### GPU

_No response_

I quicksaved in high security compound and the program silently crashed to the desktop. Reloading that specific save also crashes, so it looks corrupted or malformed. 

I went back to an earlier quicksave and progressed without further issue. I'm still on 1.6 because I'm playing through the game and don't want to risk progression problems.

Not sure if it's relevant, but I removed these lines from **gameflow.json5** to keep my items, before entering the level the first time.

```
{"type": "remove_ammo"},
{"type": "remove_medipacks"},
{"type": "remove_flares"},
```

I'm using TR3 version from GOG, save file attached.

Not to crash.

### Attachments

[save_tr3_q01.zip](https://github.com/user-attachments/files/29039275/save_tr3_q01.zip)

### Before submitting

lahm86:
Resolves #5683.

#### Checklist

#### Description

This ensures that Lara's arm animation has been synced with its frame number before performing effect control, therefore avoiding potential crashes in `Lara_GetJointAbsPosition`. This is a one-frame bug and I'm pretty sure it's OG, it's just that there, it would have processed garbage frame data rather than exiting.

I added arm anim debug info while investigating this and have left it in place. It's not the most readable as the numbers are not relative to any object, given that they can refer to any of the gun or flare animations. It was useful for testing though, and you can see the exact frame below where if you save and load this crash could happen on develop (it's when Lara's arm returns back up to the hold position after e.g. coming to a stop). LMK if we'd prefer to drop the overlay info for now.

<details>

image
</details>


aredfan (approved):

rr- (approved):
Lgtm.

PR #5707, opened 2026-06-25
objects/lift: add functionality to kill Lara

lahm86:
#### Checklist

#### Description

This adds functionality to lifts to kill Lara if they descend directly on top of her. It essentially resolves a softlock in this situation where she would become clamped. This wouldn't impact OG levels as there is space below each lift instance there. Test level available, which includes one facing each direction and a few different gaps below the lift where Lara can stand.

If immunity is on, Lara will instead pop inside the lift. LMK if we want to make the behaviour optional, similar to killer pushblocks.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -14,6 +14,7 @@
 - fixed missing SFX when Lara transitions from hanging to crouching (#5070)
 - fixed Lara's shadow appearing at the top of a pushblock after having pulled it until she returns to a standstill (#1576)
 - fixed Lara being killed by pushblocks if she pulls one onto a trapdoor that is set to be triggered by that pushblock
+- fixed Lara becoming clamped and softlocked if a lift descends on top of her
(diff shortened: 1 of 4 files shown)


rr- (approved):
I don't think we need to make it optional, as this situation was never possible in the OGs.

lahm86:
Sorry to force push, I noticed an issue if she jumps upwards into one that's just above her - she would get pushed into the floor.

aredfan:
Issues:

- [ ] If Lara is in the lift when it's ascending, she gets killed.
- [ ] OG bug: if Lara picks up a flare while the lift is ascending, she'll move through the floor. In the case of the PR, she'll also be killed.

lahm86:
Thank you, @aredfan - should be fixed now. I also noticed a similar issue if she performs a neutral twist, so I've fixed that too. I've kept those as separate commits and re-ordered them for correctness.

aredfan (approved):

PR #5714, opened 2026-06-27
ui/dialogs/new_game: fix new game plus unlock

lahm86:
#### Checklist

#### Description

This fixes the game mode options not being visible after the player has beaten the game.

I've tweaked the UI wording to avoid confusion with not being able to switch it off after beating the game. WDYT of having this as a three-way option instead - off, on, "after completion"? That way players could switch it off again without manually editing the config file.


aredfan (approved):
> WDYT of having this as a three-way option instead - off, on, "after completion"? That way players could switch it off again without manually editing the config file.

This sounds like a good idea to me.

rr- (commented):
I like the proposed three state approach but we'd need a migration path with a comment about which version to remove it in.

lahm86:
Thank you both. Added a second commit to convert to the three-way option.

aredfan (approved):

rr- (approved):

PR #5735, opened 2026-07-04
tools/installer: use safer directory deletion

lahm86:
#### Checklist

#### Description

This updates the installer to catch any errors that occur when attempting to delete directories that aren't required. It will attempt to clear read-only flags on files that would otherwise throw when trying to delete; for other cases e.g. files being open in other programs, it will just skip over them and a final warning message will be logged.

It also amends the download progress value label to remain at a constant width by always showing one decimal place.

I tested this as follows.
- install TR2GM
- navigate into the TR2GM folder and change one of the files to be read-only e.g. a loading picture
- launch the installer again
- deselect TR2GM
- select TR3LA
- the TR3LA install should complete, and the TR2GM folder should be cleared out


rr- (approved):
Thank you 😌

aredfan (approved):

PR #5741, opened 2026-07-06
camera: add flyby support

lahm86:
#### Checklist

#### Description

This adds support for flyby cameras. I have held off adding injection support until we're happy with this - it'll be easy to slot in later.

I'm not sure how I feel about it being a separate phase - it feels logical as Lara usually loses control, but the track path sequences are making me doubt this decision.

There are a few niggles left (that I'm aware of):
- [x] Lara is visible for a frame if a level starts on a flyby
- [x] Pressing look on track path sequences e.g. Coastal room 23, freezes the camera
- [x] Pressing escape will jump to the next camera sequence - this is yet to be made into a config option

Follow-up TODOs (will transfer to tracker after tests)
- In Alexandria, when the camera goes over the fence, the train room is clipped away (no portal here to be fair)
- Add injection support


rr- (commented):
It feels really weird to see this as a phase _and_ `flyby_mode.c`. I think `flyby_mode.c` could stay but orchestrating it should go into `Game_Control`.

lahm86:
> It feels really weird to see this as a phase _and_ `flyby_mode.c`. I think `flyby_mode.c` could stay but orchestrating it should go into `Game_Control`.

Agreed, it's a bit much. I'll tweak things tomorrow 👍

lahm86:
Adjusted the approach to drop the phase and instead just treat it as another camera mode. Resolved the snags above, although I think we should defer data fixes and injection support to a separate PR.

aredfan:
Testing in progress:

- [x] If the look input is held down during the coastal/library "follow Lara" camera, the screen flickers.
- [ ] There is a black screen moment in Temple of Horus that doesn't occur in OG.  This issue can be transferred alongside the other data issues if needs be.

> [record.txt](https://github.com/user-attachments/files/29741554/record.txt)

rr- (approved):

lahm86:
Thanks, @aredfan. The flicker should be resolved.
I've also moved the input reset slightly later which means we can now pause and/or use photo mode during the sequences.

The issue in Horus - yes, let's track that alongside Alexandria as it looks to be a similar cause.

aredfan (approved):
Amazing work, thank you. 🎉 

The Coastal Ruins flyby is so iconic.

PR #5743, opened 2026-07-07
tr4/lara: support poison vision

rr-:
#### Checklist

#### Description

Ports TR4's poison screen wobble. The mechanics are different and are hard to test without actual entity to deal the damage; for now I've added a `/poison` command. This covers only visual changes.

#### Testing

- [x] Run `/poison -t 4096` – screen starts wobbling as poison ramps past 256, HP drains, bar turns poison-colored; `/heal` eases it back out.
- [x] `/poison 1000` (no `-t`) – instant onset that slowly decays (trap-dart behavior)
- [x] Inventory ring and lasersight stay unwarped while poisoned (hard to test as we don't have an inventory ring 😅)
- [x] TR1–3 sanity: `/poison 256` drains HP as before, no screen warp
- [x] TR1-3 existing saves with poison timer


lahm86 (approved):

aredfan:
- [x] There is a vibration affecting the camera. In the recording below, Lara and the environment is shown to be shaking during the poison vision, which doesn't occur in OG.

> [record.txt](https://github.com/user-attachments/files/29739057/record.txt)

rr-:
Fixed and added the ability to turn this visual off, as it's known to cause motion sickness among some players.

aredfan (approved):
> Fixed and added the ability to turn this visual off, as it's known to cause motion sickness among some players.

Great idea. 😌

PR #5747, opened 2026-07-07
Declare Tr4 pickup object mappings

rr-:
#### Checklist

#### Description

This is yet another heavy-hitter – we declare some basic TR4 pickup mappings.

Because there are so many of them (each key has 12 variants, each puzzle item has 12 variants, each puzzle item can have up to two combining item, and each option I have mentioned comes in two flavors – pickup and option), I have decided to aggressively clean things up. This PR introduces `pickups.def` with an X-Macro and replaces all generic places that hardcoded lists of pickups to refer to that.

@lahm86 LMK if you like this approach.

#### Testing

- [x] Inventory ring counts / quantity display
- [x] "Use" button visibility (keys vs quest/pickup items)
- [x] Using keys/puzzles/leadbar on receptacles (incl. Midas, gong, detonator)
- [x] Scion use from inventory ("no" sound); all Scion types
- [x] Flare item pickup + throw (specialized setup intact)
- [x] Puzzle hole → done swap, savegame with used holes
- [x] Save/load round-trip: inventory counts, equipped gun, holster/back meshes
- [x] Legacy save compatibility (tr1–3 release saves)
- [x] Gun quick-draw keys
- [x] Gun cheat
- [x] `/give keys`
- [x] `/give guns`
- [x] `/give moreguns`
- [x] `/give all`
- [x] Gameflow secret rewards / item modifiers (add/remove items, NG+)
- [x] Carrier drops (enemies dropping pickups)
- [x] Sprite-based pickups (TR1/TR2 levels; flare item as sprite)
- [x] Localized object names still applied (strings.json5 overrides)
- [x] Carryover between levels
- [x] Taking items away in levels like Offshore Rig

More?!


rr-:
Oh an re: option and item pickup split. This stays even though TR4 does not need it, in case we'll want to backport things to earlier games, or grant builders the ability to use simplified meshes on the ground like TR3 does.

lahm86:
> LMK if you like this approach.

I do - it's a great way to tackle it 😌

lahm86 (commented):
Looks good. One internal issue is if you run `tools/update_object_docs`, we lose the properties for pickup items. It adds a single table in the main MD file titled "item", and in the TE XML files we lose the IDs.

rr-:
Fixed :relieved:

aredfan:
- [x] There's 2 small differences for the naming which will require TR4 specific strings:

> - `Grenade Gun` instead of `Grenade Launcher`
> - `Flares` instead of `Flare`

rr-:
Updated names.

aredfan (approved):

lahm86 (approved):

PR #5749, opened 2026-07-07
flyby_mode: allow immediate cancellation

lahm86:
#### Checklist

#### Description

This allows cancelling flybys using `/tp` and the fly cheat. If the sequence has nodes that activate heavy triggers, the requests will be ignored until those nodes have been passed, avoiding potential softlocks. I've left the escape key to jump to the next node in the sequence as part of "normal" config/gameplay.

If `/tp` fails e.g. passing an invalid room, the sequence will not be cancelled.

### DIFF
--- a/src/trx/game/camera/flyby_mode.c
+++ b/src/trx/game/camera/flyby_mode.c
@@ -28,7 +28,6 @@ static struct {
         bool spline_to_game;
         bool test_triggers;
         bool pending_trigger_check;
-        bool skip_requested;
         bool look_requested;
         bool look_cancelled;
     } flags;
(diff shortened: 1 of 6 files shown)


aredfan (approved):

rr- (commented):
TBH I thought the escape key would cancel it immediately rather than advance to the next spline position. Maybe I'm missing something.

And it'd be nice to get the fly cheat to cancel, too 😌

lahm86:
Updated, so now requests to cancel are processed immediately. This will cycle through all remaining nodes in the sequence and activate any heavy triggers along the way.

aredfan (approved):

aredfan:
- [x] I noticed an issue rather late, the inventory cannot be opened during the "follow Lara" camera in the Library.  The inventory can be opened in OG in this case.

lahm86:
> * [ ]  I noticed an issue rather late, the inventory cannot be opened during the "follow Lara" camera in the Library.  The inventory can be opened in OG in this case.

Good catch, should be fixed.

aredfan (approved):

rr- (approved):
Thank you, it feels great now!

PR #5753, opened 2026-07-07
tr4/data: define pedestal pickups

lahm86:
#### Checklist

#### Description

This marks up in Lua the pickups in each level that are on plinths/pedestals. The scripts were auto-generated from OG level OCBs.

The additional code in `pickup.c` I'm not sure is necessary. It's supposed to make Lara align closer to the pedestal when they're narrower (like Catacombs) but the -200 seems to always leave her some distance away from them. It'd be good to see what it's like in OG if possible. I was comparing with `i3` in Tomb of Seth and `i92` in Catacombs.


rr- (approved):
LGTM, I'd only expose the enum rather than using `1`.
Just as a note, `src/trx/game/level/finalize/gameplay_objects.c` does set `ocb` property as a raw integer.

lahm86:
> LGTM, I'd only expose the enum rather than using `1`. Just as a note, `src/trx/game/level/finalize/gameplay_objects.c` does set `ocb` property as a raw integer.

Thanks. I was thinking it'd be nice to move away from OCBs and have the properties fully declared and named 😌 It's good that it's there already though for quick access.
The enum is currently private in `pickup.c` - I'll make it public and auto-add it somewhere in the default Lua modules.

lahm86:
Done 😌

rr- (approved):

aredfan:
> The additional code in `pickup.c` I'm not sure is necessary. It's supposed to make Lara align closer to the pedestal when they're narrower (like Catacombs) but the -200 seems to always leave her some distance away from them. It'd be good to see what it's like in OG if possible. I was comparing with `i3` in Tomb of Seth and `i92` in Catacombs.

I've taken a look in OG, I can't see any difference in Lara's position when interacting with the pedestal in Catacombs.  I think we're good.

aredfan (approved):

lahm86:
Thank you for checking, good to know 😌

PR #5759, opened 2026-07-08
lara: add climbing pole

rr-:
#### Checklist

#### Description

Brings climbing pole mechanic over to TR4.


docs/trx/OBJECTS.md
```diff
@@ -2848,6 +2768,86 @@ This page lists documented moveable object properties.
 </tbody>
 </table>
 
+#### O_ANIMATING_7
```
lahm86: Is the ordering change here a side effect of something? Not a big deal.

lahm86 (approved):

docs/trx/OBJECTS.md
rr-: I ran `tools/update_object_docs`. I'll investigate why the order has shifted

aredfan:
- [x] If Lara climbs on the pole by pressing action, then let go. She won't be able to reclimb the pole. I think the bug occurs because there's a subtle shift in Lara's position. In OG, the shift occurs at the end of Lara's "let go" animation whereas in TRX, the shift occurs just as Lara is starting to let go of the pole.


> <details><summary>OG video</summary>
> 
> https://github.com/user-attachments/assets/e548fe16-7d1c-4eb8-beca-eb0daf621d42
> 
> </details>
> 
> [record.txt](https://github.com/user-attachments/files/29802333/record.txt)

docs/trx/OBJECTS.md
rr-: Addressed in https://github.com/LostArtefacts/TRX/pull/5760 – IDK how the ordering on develop could have slipped by the tool, anyway it's now sorted alphabetically rather than by whatever was found first :relieved:

lahm86: Thanks!

rr-:
Thanks.

`LA_POLE_GRAB_TO_STAND` would enter as `LS_STOP` at frame 0, then generic pole `Object_Collision()` would push Lara immediately: x=71168 -> 71289. After changing the TR4 Lara animation injection so anim `342` stays in `PoleIdle` until it finishes, the same replay shows `new_state=113` (`LS_POLE_IDLE`) throughout the release animation, no generic pole collision, and Lara stays at `x=71168` until the final transition to `LS_STOP` which makes the OG.

@lahm86 I'd appreciate a confirmation if this fix is in line with what the OG data has :relieved:

aredfan:
The shift issue is fixed. 👍 

Lara still doesn't reclimb the pole when repeating the above scenario, the action and draw guns inputs are being prevented.

I think it's one of those "busy hands" manifestations we've encountered in the past. A similar bug does occur in OG when Lara backflips and rolls mid-air off the pole, but it's a temporary bug and goes away when she's idle.

rr-:
Added a fix for that too, but this will need testing with walk to pickups / receptacles / etc (with saving and reloading while Lara's transitioning).

lahm86:
> Thanks.
> 
> `LA_POLE_GRAB_TO_STAND` would enter as `LS_STOP` at frame 0, then generic pole `Object_Collision()` would push Lara immediately: x=71168 -> 71289. After changing the TR4 Lara animation injection so anim `342` stays in `PoleIdle` until it finishes, the same replay shows `new_state=113` (`LS_POLE_IDLE`) throughout the release animation, no generic pole collision, and Lara stays at `x=71168` until the final transition to `LS_STOP` which makes the OG.
> 
> @lahm86 I'd appreciate a confirmation if this fix is in line with what the OG data has 😌

The OG anim 342 has state 19, which is `LS_PULL_UP`. This was an empty function IIRC but we added a fix for the vaulting issues in 1-3. Changing the state ID makes sense IMO in this case.

I wonder if we should add a hands free command to the animation as well? That may be a safer fix.

rr-:
Yeah let's do that, that makes a whole lot more sense. Can I ask for a direct push to both branches? :relieved:

rr-:
Edit: pushed to both branches – should be good now.

aredfan (approved):
Climbing poles are working great, thank you. 😌

lahm86 (approved):

PR #5761, opened 2026-07-08
lara: add rope mechanic

rr-:
#### Checklist

#### Description

Ports TR4's swinging rope mechanic.

Rendering uses camera-facing PolyFX ribbon along interpolated node positions (render-only interpolation; physics untouched).
Ropes draw in a global pass (`Rope_DrawAll`, the `DrawRopeList` analog) after room drawing, so segments hanging through non-drawn rooms can't vanish.

Object & catalogs:
- Map TR4 slot 146 → `O_ROPE` (object catalog tool, objects.def, names.def, strings)
- `FIREROPE` (147) stays unmapped – it's a torch-puzzle object, blocked on the torch system.

Physics (`game/rope.c`):
- Bit-exact port of OG's fixed-point rope solver: 24-node rope pool (max 5), pendulum constraint for Lara's weight, per-frame gravity/damping, constant-length mesh rebuild. Includes an atan wrapper reproducing OG's argument pre-shifting for large inputs. I felt ambiguous about including it globally, so left only there.

Lara (`lara/rope.c`, `lara/state/rope.c`, `lara/col/rope.c`):
- Grab from reach/up-jump
- idle/climb/slide/turn
- sprint-pumped swinging with the OG frame-drive logic
- swing-phase-dependent jump-off, fall-off
New `lara->rope` state block, aligned to the rope each frame from `Lara_Animate` like OG.

Savegame `lara->rope.*` persisted; missing keys in old saves default safely.

Fixes surfaced along the way:
TR4 sprite textures were decoded wrong engine-wide: TRX used the TR1–3 rect convention (offset/width/height), but TR4 stores the authoritative rect in the corner fields. The TR4 loader now normalizes corner rects into the canonical fields at load, which should fix all TR4 sprites, not just the rope.

`M_SPRITE_IDX` is hardcoded for now to 16 outside of `SPARK_SPRITE_TYPE` which is intentional, until we get shared sparks.


aredfan:
Issues (testing in progress):

- [x] The red pyramid item representing ropes continue to be drawn.
- [x] While Lara is holding a rope, save and reload, and she gets lowered from where the rope originates.

> [record.txt](https://github.com/user-attachments/files/29812465/record.txt)

rr-:
>  The red pyramid item representing ropes continue to be drawn.

Hmm, I can't reproduce this, but I do `/tp rope` and `/trigger rope` directly, does this matter?

aredfan:
> Hmm, I can't reproduce this, but I do `/tp rope` and `/trigger rope` directly, does this matter?

This issue occurs in both scenarios, whether using commands or crossing triggers. Added a screenshot below in case it helps.

<details><summary>Screenshot, r110</summary>

20260708_191943_Angkor_Wat

</details>

lahm86 (commented):
Looking great!
Got the null mesh showing too, but this gets rid of it.

```diff
diff --git a/src/trx/game/objects/general/rope.c b/src/trx/game/objects/general/rope.c
index e7db5ff9f..76bfcbfa6 100644
--- a/src/trx/game/objects/general/rope.c
+++ b/src/trx/game/objects/general/rope.c
@@ -94,6 +94,7 @@ static void M_Setup(OBJECT *const obj)
     obj->initialise_func = M_Initialise;
     obj->control_func = M_Control;
     obj->collision_func = M_Collision;
+    obj->draw_func = nullptr;
     obj->save_flags = true;
 }
```

rr-:
Both issues should be fixed – thanks. The `draw_func = nullptr` was embarrassingly present on my local machine but not pushed 😅

aredfan (approved):

lahm86 (approved):

PR #5770, opened 2026-07-09
tr4/creature: added scorpion

Checkm8Croft:
#### Checklist

#### Description

This PR adds the scorpion enemy of TR4. This code works both for the small one and the big one

#### Things to fix

- Small Scorpion HP that is too high
- Death Animation of both Scorpions that doesn't start and they don't disappear once killed
- The Scorpion AI that once the first time that bites Lara, keep biting instead of following her


rr-:
Thank you for the PR. We’re not accepting TR4 contributions at the moment, as we want to handle that development ourselves and at our own pace. I’ll close this, but thanks for taking the time to work on it.

PR #5777, opened 2026-07-09
data: add binoculars to TR1-3

lahm86:
#### Checklist

#### Description

This adds binoculars to TR1-3. The feathering gets lost in the injection - I'll try to think of a solution.

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/310


lahm86:
Fixed the feathering, so this should be good for testing.

rr- (approved):
Thanks  :) WDYT of adding it in code and only making it available via a setting that defaults to false? I'm not sure if that's a good approach.

aredfan:
Issue:

- [x] The all items cheat also provides binoculars in TR1-3.

rr-:
While at it can we check if there is a regression with the OG all-weapons cheat (the one that requires jumping back'n'forth)? Thanks

lahm86:
I've updated the approach so we use the `give_item` game-flow sequence like elsewhere, which means we can control there the fact it's not given to Lara in TR4 until after Cambodia, plus it means builders can add/remove it as they please.

I'm a bit stuck on how to hide it in the inventory though if the config option is off. @rr- can you think of a neat way to do this?  The last commit is WIP (and bad).

rr-:
> I've updated the approach so we use the `give_item` game-flow sequence like elsewhere, which means we can control there the fact it's not given to Lara in TR4 until after Cambodia, plus it means builders can add/remove it as they please.
> 
> I'm a bit stuck on how to hide it in the inventory though if the config option is off. @rr- can you think of a neat way to do this? The last commit is WIP (and bad).

I'll have a think about it but likely will be able to chime in only tomorrow.

lahm86:
Thanks. Fixed a little bug in the game-flow setup in the meantime (using `/play` on higher level sequences would miss the first level's inclusion of the binos).

aredfan:
- [x] If the `Crawling` option is disabled, it affects the zoom in/out ability.

rr-:
> > I'm a bit stuck on how to hide it in the inventory though if the config option is off. @rr- can you think of a neat way to do this? The last commit is WIP (and bad).
> 
> I'll have a think about it but likely will be able to chime in only tomorrow.

Done and pushed. :)

aredfan (approved):

lahm86:
> > > I'm a bit stuck on how to hide it in the inventory though if the config option is off. @rr- can you think of a neat way to do this? The last commit is WIP (and bad).
> > 
> > 
> > I'll have a think about it but likely will be able to chime in only tomorrow.
> 
> Done and pushed. :)

Brilliant, thank you 😌

PR #5785, opened 2026-07-11
Add setting for intro FMV timing

### LINKED ISSUE #5658: [Feature] Play opening FMV upon New Game, not when switching

It's only relevant to play the intro FMV if actually starting a new game rather than every time you open TRX or switch to a different game. It would just make the whole thing flow better. Even the remasters knew to make this logical change. 

It would erase tedium and and applies common sense. 

### Extra context

_No response_

### Before submitting

rr-:
#### Checklist

#### Description

Adds `gameplay.intro_fmv_mode` (New Game / Launch) so players can choose between the new default (opening story FMV plays on New Game) and the legacy behavior (plays on every launch/switch). Marks the relevant FMVs with an "intro" flag in each game's gameflow.json5 and gates playback in both the level sequence and the frontend boot sequence accordingly.

Fixes TRX434 / resolves #5658


aredfan:
Issues (testing in progress):

- [x] The game crashes when selecting the PDA
- [x] The game crashes when issuing `set intro launch` then restart the game

rr-:
Fixed the first issue, but I cannot reproduce the second issue.

aredfan:
> Fixed the first issue, but I cannot reproduce the second issue.

The crash occurs if the FPS counter is enabled.

A couple more issues:

- [x] If `Into FMV timing` is set to launch, and the console is opened during the intro fmv, the game crashes.
- [x] If `Into FMV timing` is set to launch, the intro fmv is shown first on start up instead of last after the legal screen and core/eidos fmvs.

rr-:
Thank you – that was it :) fixed both problems.

lahm86 (approved):

aredfan (approved):

PR #5788, opened 2026-07-11
camera/photo-mode: fix frame-step during flybys

rr-:
#### Checklist

#### Description

Advancing a frame while a flyby camera is active did nothing visible: `Camera_PhotoMode_Pause()`/`Resume()` always rebased `g_Camera` onto a stale entry-time snapshot around the simulated step, discarding the flyby's freshly computed position. Exiting photo mode mid-flyby had the same issue via `Camera_PhotoMode_Exit()`.

Both now leave the live camera alone while a flyby is running.

Resolves TRX594.

### DIFF
--- a/src/trx/game/camera/photo_mode.c
+++ b/src/trx/game/camera/photo_mode.c
@@ -35,6 +35,32 @@ static struct {
 } m_PreviousState;
 static BOUNDS_32 m_WorldBounds = {};
 
+static void M_SyncCameraOrientationFromView(void)
+{
+    int16_t angles[2];
+    Math_GetVectorAngles(
+        g_Camera.target.x - g_Camera.pos.x, g_Camera.target.y - g_Camera.pos.y,
+        g_Camera.target.z - g_Camera.pos.z, angles);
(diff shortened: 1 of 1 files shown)


lahm86:
Thanks, looks great. I just noticed one thing:
- enter camera mode during a flyby, then hit roll input twice to rotate 180deg
- advance a frame or two
- press photo mode camera left or camera right inputs - the camera moves in the opposite direction

rr-:
Thanks – good catch. Fixed.

lahm86 (approved):

aredfan:
- [x] Entering and exiting photo mode during a flyby will affect where the camera's end position is.  In the case of Coastal Ruins, the camera ends up in front of Lara instead of behind her. [record.txt](https://github.com/user-attachments/files/29930392/record.txt)

rr-:
This was a bit tricky but should be fixed too.

aredfan:
> This was a bit tricky but should be fixed too.

Your fix did the trick, thank you. 😌

aredfan (approved):

PR #5797, opened 2026-07-12
rope: fix brightness and weave direction

rr-:
#### Checklist

#### Description

The rope drew at twice OG's brightness and with its sprite mirrored.

OG's DrawRope() paints the rope quads a flat 0xFF7F7F7F, which D3D applies as a MODULATE factor over the sprite, i.e. half brightness. The TR4 shader path instead reads an unlit vertex color as a prelit value on the "128 = neutral" scale and doubles it, so the stored 0x7F came out at full bright and the rope stopped sitting in the room's gloom. Halving it to 0x40 survives that doubling and lands back on OG's factor: 0x40/255 * 255/128 = 0.5.

The ribbon's perpendicular used cross(dir, to_cam), which falls on the opposite side of the rope from OG's screen-space perpendicular (-dy, dx), taken with Y pointing down. That mirrored the sprite's U axis and left the rope's weave leaning the wrong way.

Resolves TRX614.

### DIFF
--- a/src/trx/game/rope.c
+++ b/src/trx/game/rope.c
@@ -19,13 +19,17 @@
 #define M_PENDULUM_GRAVITY 0x60000
 #define M_NODE_GRAVITY 0x30000
 #define M_LARA_HAND_OFFSET (-112)
-#define M_SPRITE_IDX 16
 #define M_HALF_WIDTH 24.0f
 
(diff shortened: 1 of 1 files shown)


rr-:
We'll need to test this in a couple of different rooms to make sure it's lit properly :)

aredfan (approved):

lahm86 (approved):

PR #5808, opened 2026-07-12
Rewrite LUA items/rooms public API

rr-:
#### Checklist

#### Description

Replaces the hand-written getter/setter pairs for items and rooms with `FIELD_DESC` reflection introduced in #5805, and mints a public surface in items.lua / rooms.lua. Exposure becomes opt-in: a C member is reachable from a script only if a declaration names it.

The same registry now also drives the docs:
`TRX --dump-lua-api` → `docs/trx/lua/api.json` → the generated reference pages.
It only runs for modules that use the new lua `api` surface, so we can migrate the rest one-by-one.

Enums work the same way now. `ENUM_MAP` is the single source of truth for constant names and values. `trxc.enum` reads from that, while the Lua side just adds descriptions. Every constant must exist and be documented – missing or extra entries throw errors.
As a result, `trx.items.Status` and `trx.rooms.FlipStatus` are finally defined in the reference with their real values.

**Breaking changes:**
- `trx.items.fn` and `trx.rooms.fn` are gone. Use `trx.items.get` and `trx.rooms.get` instead.
- `item.idx` / `room.idx` are removed. Handles are now opaque; use `room.num` instead.
- If you use a stale handle, it throws instead of quietly hitting a recycled slot.
- **I may have missed something that needs capturing in `MIGRATING.md`**, so I'd appreciate a side-by-side comparison of the docs.


aredfan (approved):

docs/trx/MIGRATING.md
```diff
@@ -30,6 +30,38 @@ order: 3
    assets file from https://lostartefacts.dev/pub/tr3-assets.zip, or use the
    shipped `sparks_gfx.bin` injection.
 
+5. **Update scripts that use item handles**
```
lahm86: This'll be a new 1.9 to 1.10 section.

lahm86 (commented):
Looks good. Compared the docs on the items and rooms and looks like everything is captured 👍

docs/trx/MIGRATING.md
rr-: Facepalm. Yes.

rr-: Fixed!

lahm86 (approved):

PR #5838, opened 2026-07-14
camera: fix binoculars with fixed cameras

lahm86:
#### Checklist

#### Description

This allows Lara to break out of fixed cameras or flybys (where she is controllable) when binoculars have been requested.

Resolves TRX651.


rr- (approved):
This doesn't introduce a "jailbreak" out of uninterruptible flybys right?

lahm86:
> This doesn't introduce a "jailbreak" out of uninterruptible flybys right?

I think only if we were to introduce a means to activate binos without going into the inventory, but even then, if she can't be controlled during a flyby the inputs are cleared so preventing hotkeys and whatnot. If Lara can't be controlled, she won't be able to get into the inventory to select them (indeed hitting inventory cancels the flyby anyway).

aredfan (approved):

PR #4671, opened 2026-01-18
camera/los_camera: detect manual camera changes

lahm86:
Resolves #4670.

#### Checklist

- [x] I have read the [coding conventions](https://github.com/LostArtefacts/TRX/blob/master/CONTRIBUTING.md#coding-conventions)
- [x] I have added a changelog entry about what my pull request accomplishes, or it is an internal change
- [x] I have added a readme entry about my new feature or OG bug fix, or it is a different change

#### Description

This ensures the TR3 camera is aware of when the manual camera has been changed, and Lara is idle.

PR #5840, opened 2026-07-14
camera: add TR4 camera mode

lahm86:
#### Checklist

#### Description

This adds an option for TR4's camera, which builds on TR3's by performing checks on whether or not the target angle and elevation have changed since the last frame. This means actions like picking up items or corner shimmying, where Lara's control sets those values, now have the camera properly respond. TR3 is left as-is rather than introducing a separate config fix option for this. It leaves that mode feeling OG to TR3.

TR4 also uses slightly different bounce control, which is gated to the new mode.

Using the manual camera in TR3 mode when Lara is idle should still work (ref #4671).

I also noticed potentially an OG bug with camera bounce in 1-3, where the target Y is altered with both x and y values from the shake. Altering this does change behaviour considerably so I wouldn't want to change it, but it's interesting that it was fixed in TR4.

Resolves TRX580.


lahm86:
Added a changelog to indicate fixing the elevation snapping.

aredfan (approved):

rr- (approved):

PR #5233, opened 2026-03-26
Orca

rr-:
#### Checklist

- [x] I have read the [coding conventions](https://github.com/LostArtefacts/TRX/blob/develop/docs/CONTRIBUTING.md#coding-conventions)
- [x] I have added a changelog entry about what my pull request accomplishes, or it is an internal change
- [x] I have added a readme entry about my new feature or OG bug fix, or it is a different change

#### Description

This adds Orcas to TR3.

I'll need help with the flyer box zones, because the Orcas in this implementation consistently swim near ground and appear very jittery.

PR #5842, opened 2026-07-14
lara/climb: reset interpolation after corner interaction

lahm86:
#### Checklist

#### Description

This fixes Lara's interpolation after climbing around corners. Previously the `prev` values were retained and so even though committing, they would get overwritten when drawing. Maybe we should expose `Interpolation_RememberItem` for re-remembering items in cases like this?

Resolves TRX660.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -2,6 +2,7 @@
 - changed reflections UV mapping to be more correct
 - fixed TR1 and TR2 skyboxes being 2× too bright (regression from 1.9)
 - fixed Lara being unable to use binoculars when fixed cameras or track path flyby sequences are active (regression from 1.9)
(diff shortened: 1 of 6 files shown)


rr- (approved):
I'm okay with that but suggest to go through all `interp` uses across the codebase and route that through `Interpolation_` too.

aredfan (approved):

lahm86:
That's in place now. @aredfan, would you mind sanity checking these other cases?

- [x] HSC starting animation (on the final frame Lara's position is reset so she's in front of the bed instead of interpolating from level start)
- [x] Orcas when they do their 180° flip (ref https://github.com/LostArtefacts/TRX/pull/5233#issuecomment-4140918900)

rr- (approved):

aredfan (approved):
> That's in place now. @aredfan, would you mind sanity checking these other cases?

No problem, both scenarios look good. 😌

PR #5863, opened 2026-07-17
Port TR4 water droplets

rr-:
#### Checklist

#### Description

This PR ports the TR4 water droplets (TRX461):

- Wet Lara – a new FX module gives each of Lara's meshes a wetness value, pinned while the mesh is underwater and shed as falling droplet streaks once she is out. Wetness carries across level transitions and saves like the OG, and a fresh playthrough starts dry.
- The small splash of water pixels when a spent shell lands on a water surface, next to the existing ripple.
- Enabled by default in TR3 and TR4, off in TR1/2, with a toggle under Graphic Options → Visuals → Water droplets.
- The pour-into-scales trigger is left for the TR4 objects work, as the scales object does not exist yet.

One general rendering fix turned up:

- TR4 draws sparks without the sprite flag as short line streaks trailing behind their velocity, where TRX drew flat quads for all of them. The quad path turned out to be the correct TR3 behaviour, so it stays for TR1–3; TR4 now uses streaks, which also covers blood, flare sparks (TODO, but it's pretty :D) and ricochets (TODO). I didn't add tickets for that and would appreciate a cross-check.

#### Testing

- [x] Climbed out of water and watched the drips thin out as Lara dries (the OG never fully dries her quickly – decay only happens through shed drips)
- [x] Fired shells onto water in TR4 for the splash and streak look
- [x] Save/load and level transition carry wetness; old saves load with Lara dry

LMK if we also want to have a way to dry Lara on level start via LUA as part of this PR – I think it'd be super useful for builders to reset her state across zones.


lahm86 (approved):
Being able to dry Lara through Lua sounds good to me 👍

aredfan:
- [x] TR3: the small splash when the shells hit the water is miniscule, I think the spark scaling cap that we removed in [5613](https://github.com/LostArtefacts/TRX/pull/5613) may be in play. [record.txt](https://github.com/user-attachments/files/30123444/record.txt)
- [x] It looks like the water droplets can mess with the TR1/TR2 demos, if enabled. I think it affects the way the enemies are reacting to Lara when she's wet.

rr-:
Thank you – both issues should be addressed.

aredfan (approved):

PR #5865, opened 2026-07-18
inject: add property injection support

lahm86:
#### Checklist

#### Description

This adds the ability to inject properties for objects and items. Test level available generated from TE.

I initially had an odd bug which turned out to be from freeing the property names from the injection - it was in turn leaving dangling pointers in the main property module. There, we now just reference the definitions' names rather than using the ones passed to it when setting values, otherwise ownership becomes difficult to track. The alternative would be duping and then explicitly freeing, WDYT?

Resolves TRX693.


rr- (approved):
This LGTM, and I can try to simplify the switches when consolidating COT and OPT and FT 😌 thank you!

lahm86:
Forgot to push the injection tool update:
https://github.com/LostArtefacts/TRXInjectionTool/pull/316

Essentially that's what TE will be doing, it keeps the tool up-to-date if we ever need to use it.

aredfan (approved):

lahm86:
Added changelog entry for this.

PR #5866, opened 2026-07-18
flyby: fix level initialisation and mic pos

lahm86:
#### Checklist

#### Description

This fixes initialisation of flyby cameras that are included in embedded injections. The number of cameras will not be known until after the embedded injection is initially read, hence a separate call is required for TR1-3 to initialise the gamebuf before the injection is processed.

Test level from TE available.

Resolves TRX695.


aredfan (approved):

lahm86:
I've added the fix for the microphone not updating as discussed on Discord. I think in TR4 the mic was always at the camera pos, so all we needed to do was ensure the position gets updated each frame. If the player chooses mic-at-Lara it will naturally not follow the sequences.

aredfan (approved):

rr- (approved):

PR #5875, opened 2026-07-20
lua/events: let a script claim a flip effect

### LINKED ISSUE #4108: [LUA Feature] Allow floordata triggers to trigger Lua events
Currently, Lua works on firing global events (level load, saving the game etc).

I would like an option which upon calling a floordata triggers, it fires event function from the script (like playing music, triggering door, setting Lara's HP etc)

May require changes to Tomb Editor UI and level file format.

rr-:
#### Checklist

#### Description

`trx.events.on_flip_effect(num, callback)` claims a flip effect number: a level-placed flipeffect trigger with that number dispatches to the script, with the trigger's timer field, under the same gating the stock effect would get. The stock effect for a claimed number does not run. Unclaimed numbers are untouched.

This gives level builders a way to run script code from ordinary triggers placed in any editor, with no editor or file format changes.

A claim follows the lifetime of the script's listeners rather than the listener itself: detaching the handler leaves the number claimed until the level ends.

Resolves #4108 / TRX412.


lahm86 (approved):

aredfan (approved):

SaracenOne:
I'm glad this is getting some attention in the interests of a desire to more or less fully replace the advanced trigger systems from earlier earlier Tomb Raider script extenders level builders used such as those found in TRNG and TREP, but I think the API could benefit from a few key additions or tweaks.

Given that this covers heavy triggers too, it might be a good idea of the API also tracks the specific item which executed the flipeffect too (maybe even the specific sector, but that might not be necessary if implicitly deriving the sector from the item is possible), which if I understand correctly, shouldn't be too much of an issue given that the surrounding function already provides the executing item as a pointer. I feel like this would improve the general robustness of the system.

It might also be worth naming the API with a bit more specificity too given that flipeffects can also be executed via animation commands, since this API only covers execution from floor data. Maybe something like 'on_flip_effect_trigger'?

rr-:
Thanks, both good calls.

- The handler now receives the item that ran the effect (`item_num`), so a heavy trigger can tell which object hit the pad. I left the sector out, since it's derivable from the item.
- On the naming: rather than narrow it to `on_flip_effect_trigger`, I closed the gap you pointed at: the hook now also fires for flipeffects run from animation commands, not just floor data. A claimed number dispatches to the script whichever way the engine runs it, so the name stays accurate.

Pushed as an update to the same commit. @LostArtefacts/core please take another look 🩷

aredfan (approved):

lahm86 (approved):

PR #5881, opened 2026-07-21
ui: link breeze option to droplets and weather

lahm86:
#### Checklist

#### Description

This extends the breeze option to be available if either the braid, weather or water droplets are enabled.

Resolves TRX713.


rr- (approved):

aredfan:
Thank you. 👍 

There is just one issue, the breeze has no effect on the weather/water droplets in TR1/TR2. Perhaps we forgot to backport the TR3 behaviour to these 2 games.  This can be tested in Lost Valley and Tibetan Foothills.

lahm86:
Thanks. I've added a change so that wind will apply to weather/droplets, but only in the Z-direction, so it's faithful to OG TR2 wind.
I'm wondering though if we'd be better to change the breeze option to a mode - off/TR2/TR3+? The TR3 effect is nice with the application in both axial directions IMO. LMK what you think.

rr-:
Yes, good idea – let's do that.

aredfan:
> Thanks. I've added a change so that wind will apply to weather/droplets, but only in the Z-direction, so it's faithful to OG TR2 wind.

Thank you, the breeze is now affecting the weather in TR1/TR2.

The water droplets remains unaffected by the breeze in TR1/TR2.  It's worth checking out Coastal Village if you'd like a TR3 level to compare with.

> I'm wondering though if we'd be better to change the breeze option to a mode - off/TR2/TR3+? The TR3 effect is nice with the application in both axial directions IMO. LMK what you think.

Sounds like a nice idea. It's always good to have flexibility.

lahm86:
TR1/2 droplets should be behaving now. The drops were looking at the wrong room flag (`outside` rather than `wind` - it's the OG `NOT_INSIDE` bad naming that struck us again). Confirmed the flag in tomb4 is `0x20`, so this matches TRX's `wind` named flag.

I'll sort out the new config option tomorrow.

lahm86:
The config option is now a choice between off, TR2 and TR3 ☺️

rr- (approved):

aredfan (approved):

PR #5905, opened 2026-07-23
objects/lift: test ceiling death and climbing states

### LINKED ISSUE #5899: [Bug] Lara can phase into elevators when hanging from a ledge
### Game type

Custom level

### Game

TR2

### TRX version

1.9.2

### Operating system

Windows 10

### GPU

_No response_

This happened while playing the custom level Trouble in Paradise by BradTRe. If Lara is hanging from a ledge/ladder and an elevator comes to occupy the same space as her, she will phase into the elevator. Recently TRX made it so Lara can get squashed when the elevator's floor comes into contact with her - this leads to a funny situation where as soon as the floor comes touches her as she's hanging she gets squashed in midair (see image 1).
1. Rename the save file from .txt to .dat
2. Try to climb the ladder
3. Notice that Lara will phase through the elevator ceiling
4. As she is phasing through the floor, she will get squashed and die

Notice also that Lara's vaulting animation does not take into account that the elevator is moving (see image 2), but this is another bug and I will create another report for this.

If an elevator comes to occupy the same space as Lara, she should let go of the ledge/ladder.

### Attachments

Image 1 (Lara squished by the elevator in midair):
Image
image 2 (Lara is in the vaulting animation to grab onto the ladder):
Image
Save:
[save_tr2_09.txt](https://github.com/user-attachments/files/30290872/save_tr2_09.txt)

### Before submitting

lahm86:
#### Checklist

#### Description

This restores original behavior when Lara is on a ladder and a lift passes through her. This could be useful for OG exploits; making her fall or clip into the lift I feel is too much of a deviation. The squashing behavior is retained only when she's directly under it to avoid softlocks.

Resolves #5899 / TRX732.


rr- (commented):
Hmm if the lift pushes her while she's under/above it I think she should let go. Moving upwards, she would stand on top of the lift then get squished by ceiling, moving downwards she'd start falling (possibly accelerated by the lift) then get squished.

lahm86:
I personally feel the ability to "glitch" when on a ladder should be retained. Do we need a config option?
The ceiling behavior would be an addition too, the squishing only deals with the case when she's below.

rr-:
Yeah, IMO let's gate the lift "sane behavior" additions behind a new option to retain OG lift glitches.

lahm86:
Cool, I'll work on this some more shortly.

lahm86:
Updated now with the following:
- if Lara is on top of a lift that's moving up and the ceiling space becomes too low, she will be squished
- if Lara is in a climbing, hanging or shimmying state and a lift touches her, she will be knocked off the wall
- the entire lift collision function is gated with a config option, so the OG behaviour can be selected (including the clamping softlocks) if people desire
- invulnerability only guards against the above/below death scenarios, where she will be pushed inside the lift

There is one other OG issue (that I'm aware of) remaining. If Lara stands against a wall and holds forward and action as the lift moves up, if vaulting becomes possible, she will do so and push through the lift. On develop this would kill her, now it just gives a bad camera view. I'd like to tackle this separately.

lahm86:
Forgot to add, the test level from the previous killer lift PR has been updated.

aredfan (approved):

rr- (approved):
Thank you 🩷 Does this deserve an e2e test to automate regression checks?

lahm86:
> Thank you 🩷 Does this deserve an e2e test to automate regression checks?

Done. I can look at extending this in the next PR to sort out the remaining vaulting bug.

image

matheusvb3:
I just wanted to say that I decided to experiment with this because I was curious and this also fixes another bug that I was going to report: if you jump and touch the elevator's floor from below as it's moving, Lara would teleport down to the floor under her and die and if she had been squished. =D Thank you for all the loving work!

PR #5910, opened 2026-07-23
objects/lift: add internal collision

lahm86:
#### Checklist

#### Description

This adds collision to the lift walls when Lara is inside and it is in motion. It prevents Lara from being able to hit the geometric walls in the shaft, therefore preventing vaulting. Additionally, it prevents Lara from exiting lifts too early, which can cause camera issues.

I've left the debug drawing that I was using in place (`set debug-bound 1`). LMK if we want to retain that or not.

Resolves TRX742.


aredfan (approved):
> I've left the debug drawing that I was using in place (set debug-bound 1). LMK if we want to retain that or not.

Let's keep it. 👍

rr- (approved):

PR #5921, opened 2026-07-25
lara/state/climb: force corner ladder camera update

lahm86:
#### Checklist

#### Description

This fixes the TR3/TR4 cameras getting stuck when Lara traverses around an outer ladder corner. Loosening the LOS checks in the camera module results in it being able to go OOB in too many normal gameplay scenarios, so retaining this logic in Lara's control is more preferable as it's the driver behind the "unusual" transition.

Resolves TRX593.


aredfan (approved):

rr- (approved):
Thanks. Let's make a TODO to improve this, possibly log in the tracker – the snap is something it'd be great to eliminate, however the camera softlock gone is much better than no fix at all.

lahm86:
TODO added. I'll add it to the tracker as well 👍

PR #5927, opened 2026-07-25
objects/bacon_lara: move anchor room setup to Lua

rr-:
#### Checklist

#### Description

The anchor room moves from the `setup_bacon_lara` sequence event to an optional `anchor_room` object property, which a level editor or a level script can set. Left at its default, it falls back on the room Bacon Lara is placed in.

Resolves TRX753.

#### Testing

I'd appreciate help from Lahm with testing this against a test level which does _not_ define the anchor room.
Otherwise, we should just test for regressions in TR1 Bacon Lara.


lahm86 (approved):
LGTM. Added a link in the tracker to a test level where she starts in the intended anchor room.

aredfan (approved):

PR #5949, opened 2026-07-26
rules: move the exposure values to Lua

rr-:
#### Checklist

#### Description

Introduces trx.rules, a home for gameplay values that belong to no single object and so have no place among an object's properties. Each rule is one row in rules.def; the storage, the defaults, the /rule console command and the Lua bridge are generated from it.

The cold exposure drain rates, damage and max values are the first to be moved.

Resolves TRX416.


lahm86 (approved):

aredfan (approved):

rr-:
Added persisting rules to the savegames, and removed resetting on level load (moved to on playthrough restart).

PR #5953, opened 2026-07-26
objects/flare_item: make TR4 flares faithful

rr-:
#### Checklist

#### Description

TR4 was falling through to the TR3 flare. Its own burns green across every stage, jitters its light below itself rather than above, and gives off nothing else at all: no sparks from the tip in hand or on the ground, and no bubbles underwater.

Lara also refuses to light one while on all fours, and says so.

The lighting is not 100% faithful, but I suspect this might be caused by the not-yet updated flare mesh in Lara's hand.

Resolves TRX474.


lahm86:
Added the TR4 mesh, and linked up the SFX when igniting a flare.

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/321

lahm86 (approved):

aredfan:
Testing (in progress):

- [x] Lara doesn't say "No" when prompted to take out a flare while on all fours.

rr-:
Fixed :relieved:

aredfan (approved):

PR #5961, opened 2026-07-27
objects/floor_switch: add TR4 floor switches

lahm86:
#### Checklist

#### Description

This adds the TR4 floor switch and crowbar switch types. I've made a semi-common way to look-up crowbar receptacles from the inventory, but I'm not a huge fan of it. There will be one more type to come later - Lara prying things off walls with the crowbar. Perhaps a registration approach would be an idea.

Further animations were found to be missing the hands-free command, so those are all updated now (even for switches and such yet to be implemented; I've hopefully I've caught them all).

Resolves TRX793 / TRX794.

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/322


rr- (approved):

lahm86:
^ Rebased for conflicts

aredfan:
Testing (in progress):

- [x] May be a bug. When Lara is at a crowbar switch, press action near it so the inventory opens, then rather than use the crowbar, press escape to close the inventory. Lara uses the crowbar regardless.
- [x] Lara doesn't say "No" when attempting to use a crowbar on the wrong side of the crowbar switch.  Crowbar switches allows Lara to use the crowbar from either side, but it alternatives i.e. she'll have to use one side first before interacting from the other side.

lahm86:
> Lara doesn't say "No" when attempting to use a crowbar on the wrong side of the crowbar switch. Crowbar switches allows Lara to use the crowbar from either side, but it alternatives i.e. she'll have to use one side first before interacting from the other side.

Thanks. And just to confirm, she shouldn't say no if trying to use a regular switch from the wrong side?

lahm86:
Both problems should be fixed 😌 I didn't make her say "no" on the wrong side of normal floor switches.

aredfan (approved):
> Thanks. And just to confirm, she shouldn't say no if trying to use a regular switch from the wrong side?

Lara doesn't say "no" in this scenario, although I've only checked Angkor Wat.

PR #5966, opened 2026-07-27
Port /tp to LUA

rr-:
#### Checklist

#### Description

This ports the /tp command to LUA, giving it `/tp pickup`, `/tp enemy`, `/tp loyal`, `/tp creature` and autocompletion for free.

The object families the console matches names against were made consistent across `/tp`, `/spawn` and `/kill`.

- `/tp` moves to Lua, needing `trx.lara.teleport()`, `trx.rooms.floor_height()` and `query:where()`; `/tp enemy` is a group name like `/tp pickup` rather than a case of its own, and picks the nearest hostile then the next.
- Every family: `pickup`, `creature`, `enemy`, `loyal`, `switch`, `receptacle`, `door`, and the rest, answers to its own name, so `/tp receptacle` and `/spawn door` work. Each command offers only the families its own query keeps, so `/kill` still has no `pickup`.
- Autocompletion offers family names ahead of object names, which used to lose score ties: `/spawn cre` completed to `Secret 1`.
- An angle field wraps when written, so adding a half turn to `item.rot` turns the item round instead of raising.

#### Implementation notes

Internally, `kill.lua` builds `targetable()` as `(q:creature() | q:loyal()):loaded()`, and `:names()` on that is the whole of the suggestion source (`kill.lua:160`). Inside `names()`, a family name is added by testing membership, not by being named:

```lua
for _, group in ipairs(searchable) do
  for _, pair in ipairs(kept) do
    if group.pred(pair[1], pair[2]) then
      add(group.key)
      break
    end
  end
end
```

`kept` is what survived the query – creatures and allies. Every family's predicate is run against them: `creature` and `enemy` hit the wolf, `loyal` hits Winston, and `pickup`'s predicate hits nothing, because no pickup got past `q:creature() | q:loyal()` in the first place. A name is offered when the family has a member present, so `pickup` is absent from `/kill` for the same reason `wolf` is absent from a level with no wolves.
`by_name` works the same way (`query.lua:150`), which is why `/kill pickup` typed by hand also finds nothing – the family is weighed only over the kept set.


src/trx/game/lua/capi/rooms.c
```diff
@@ -176,8 +176,32 @@ static int M_L_RoomsFindValidPos(lua_State *const L)
...
+
+    // Room_GetSector walks from the room it is given, and Room_Get gives back
+    // nullptr for a room outside the level.
+    const lua_Integer room_arg = luaL_checkinteger(L, 2);
+    luaL_argcheck(
+        L, room_arg >= 0 && room_arg <= Room_GetCount() - 1, 2, "unknown room");
+
+    int16_t room_num = (int16_t)room_arg;
+    const SECTOR *const sector = Room_GetSector(pos, &room_num);
+    const int32_t height = Room_GetHeight(sector, pos);
```
lahm86: WDYT of adding an optional arg for fixing tilts in walls (i.e. `Room_GetHeightEx`), maybe defaulted to `true`? It wouldn't ever be useful for modern-compiled levels, but for the vanilla levels it might have some benefit for glitch-hunters. Just a thought, not essential for now IMO.

lahm86 (approved):
Looks good 👍

src/trx/game/lua/capi/rooms.c
rr-: Good idea – added. Also, added a convenience accessor directly on the `Room`: `trx.lara.item.room:floor_height(trx.lara.item.pos)`

aredfan (approved):

lahm86 (approved):

PR #5972, opened 2026-07-27
lint: keep .def tables aligned and sorted

rr-:
#### Checklist

#### Description

The config maps were sorted by hand in e6d182da4, with nothing to keep them that way. The game string alignment check becomes a shared `def_entries` check over every `.def` table, each file carrying a rule for how its columns work.

- Only the config maps are sorted. Input roles and setting tabs appear in the order they are listed, so their order is left alone.
- Alignment now covers `names.def`, `controller.def` and the setting tabs as well, which realigns them.
- `preset_keys` was reading only `map.def` through `map_tr3.def`, and now reads every config map.


lahm86 (approved):
Thanks

aredfan (approved):

PR #5977, opened 2026-07-28
Move Lua game strings into the scripts that show them

rr-:
#### Checklist

#### Description

The text a Lua console command shows lived in `game_strings/entries.def`, a file away from the command itself. `trx.locale.declare()` supersedes `GS_DEFINE()` for anything written in Lua, and every console string a script shows moves into the file that shows it.

- A new function `trx.locale.declare()` takes a table of keys and the text behind them. It runs when the scripts load, before the strings files are read, so a declaration is the fallback and a translated key still reads translated.
- `tools/lint/gen/game_strings` reads the declarations the way it reads `GS_DEFINE()`, so a declared key still lands in `base_strings.json5` and reaches the translators. A key declared twice, or one `entries.def` already has, is reported with its file and line.
- 186 keys move out of `entries.def` into the 42 files that show them, covering every Lua command plus `trx.argparse`, `trx.config` and `trx.console`. The tool reports no changes over the strings files, so every English string and every translation is unchanged.

The first commit regenerates the object catalog docs, which had gone stale on develop.

This change is internal and does not affect players.


lahm86:
> The first commit regenerates the object catalog docs, which had gone stale on develop.

My bad, I'll aim to remember to do this for new objects.

lahm86 (approved):

rr-:
NW. I'll make this a CI/CD check, so we can detect it more reliably :) (I think I already said that but forgot about that)
EDIT: done in https://github.com/LostArtefacts/TRX/pull/5978

aredfan (approved):

PR #5980, opened 2026-07-28
objects/pulley_switch: add

lahm86:
#### Checklist

#### Description

This adds TR4's pulley switch, and converts its OCB handling into explicit properties. The vanilla levels use OCB 1 throughout, so no Lua was needed to be added. I've made a test level available to test the properties. The absolute value of the OCB indicated how many pulls were needed to activate the trigger; if it was negative it meant the pulley was single-use.

The original setup handled the item's control from within Lara's `LS_PULLEY` state routine. This meant that because it behaves like a switch, its control routine could not be relied on for synchronizing animations. It additionally exposed workings to Lara, so I wanted to avoid that altogether.

The other issue was that it's not really a switch as its animation state never changes, so there is no on/off flag as such. That was inferred in the original `trigger_flags`, so that's again more explicit and readable now.

The pulley in Hypostyle Hall is blocked by default until the Obelisk items are aligned properly. That can be tackled later.

Resolves TRX797.


src/trx/game/objects/general/pulley_switch.c
```diff
@@ -0,0 +1,206 @@
...
+{
+    ITEM *const lara_item = Lara_GetItem();
+    M_PRIV *const p = item->priv;
+    if (g_Input.action && p->remaining_pulls != 0) {
+        lara_item->goal_anim_state = LS(LS_PULLEY);
+    } else {
+        lara_item->goal_anim_state = LS(LS_STOP);
+    }
+
+    const bool pulling = Item_TestAnimEqual(lara_item, LA(LA_PULLEY_PULL));
```
rr-: Nit: `is_pulling`

rr- (approved):

src/trx/game/objects/general/pulley_switch.c
lahm86: Thanks, fixed.

aredfan (approved):

PR #5984, opened 2026-07-28
objects/switch: add underwater ceiling switch

lahm86:
#### Checklist

#### Description

I've noticed the camera is not great when using one of these in Race for the Iris (`i58`) - LMK if it's similar in OG, if we know.

Resolves TRX809.


rr- (approved):

aredfan:
I think we might be missing base strings for the underwater ceiling switch.

lahm86:
> I think we might be missing base strings for the underwater ceiling switch.

Fixed, thanks. Oops 🤦‍♂️

lahm86:
Removed the forced fixed camera as while its position was being set in OG, the flag to use it was not actually enabled.

aredfan (approved):

PR #5985, opened 2026-07-28
Bind object properties to the members that keep them

rr-:
#### Checklist

#### Description

Properties like `damage` and `collidable` lived in a table, looked up by name. Objects used them in two different ways: some looked the value up where they needed it, others copied it into their own struct when the item was created. The second way set the rule for everyone, because it meant a property changed later was quietly ignored. Nothing said so anywhere.

A property now points at the field that holds it:

```c
OBJECT_PROPERTY(M_PRIV, damage, M_DAMAGE, "Damage dealt by the ape bite.")
```

The engine fills that field in when the item is created, and again whenever the property changes. The object just reads `p->damage`.

- Changing a property mid-level now works. Before, it did nothing.
- The lookup goes away. Every creature carried the same eight-line `M_GetDamage()`, and initialisers had blocks that copied properties into place - 316 declarations across 126 files.
- Clamps and conversions moved into the property's setter, so they still apply when a value changes later: the lift's speed, the bowl's pour time, the pulley's pull count, the flip slots, the trap masks.
- Values keep the units they are written in. The cobra's `alert_radius` stays in sectors and is squared where it is compared, instead of being stored already converted.
- A few properties have no field to point at and stay in the table: hit points, which the item manager reads for every object; the vehicle settings whose names are built at runtime; and object-level values with no item behind them.
- New lint check: an object that binds a property but never sets `priv_size` has nowhere to put it, and the field reads as zero all level. It caught one, in `bird.c`.

Players see no difference. `docs/trx/OBJECTS.md` and the Tomb Editor catalogs come out of the generator unchanged, so no default and no type moved.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -67,6 +67,12 @@ repos:
     language: python
     files: \.c$
 
+  - id: lint-object-property-binding
+    name: Bound object properties have somewhere to land
+    entry: tools/lint/checks/object_property_binding
+    language: python
+    files: \.c$
+
   - id: lint-gameflow-schema
     name: Gameflow files match the schema
     entry: tools/lint/checks/gameflow_schema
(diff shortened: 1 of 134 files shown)


rr-:
Reopened from the fork as #5986.

PR #5989, opened 2026-07-29
Refactor lifecycle events

rr-:
#### Checklist

#### Description

This PR collapses the five events a level script could listen for as its level loaded:

- `before_level_file`
- `after_level_file`
- `before_item_setup`
- `after_item_setup`
- `after_level_state`

into `on_game_start`, plus a new `on_title_start` for the title screen.

They existed because the engine read what a script set at different points of the load: properties before the items were initialised, allies before anything targeted, room state after the savegame. Property writes now land whenever they are made, so one moment covers all of it.

- 56 shipped scripts move across with only the name changed. `after_item_setup` needed none – it never fired during play.
- Cutscene and demo levels now fire `on_game_start` too.
- The title level's script used to stop working after the first game, because `GF_RunTitle` never ran it. It now runs from the level load itself.
- The old names are removed rather than aliased. A handler runs later than it used to, and breaking loudly beats doing nothing quietly. Migration notes cover the move.
- `max_hit_points` was the last property that had to be set before its item existed. Raising it now carries the creature's current hit points, so it works on a living enemy.

#### Testing

- [x] Start TR3 Antarctica fresh
  The RX worker fights for Lara; underwater rooms damage and every room is cold.
- [x] Start TR4 Karnak and pick up the four plinth items
  They still use the low-plinth animation.
- [x] Start TR1 Atlantis and let Bacon Lara mirror Lara
  She pivots about room 10, as before.
- [x] Save mid-level in each, reload
  Same behaviour after the reload.
- [x] Play a level, quit to the title, start another
  The title screen still loads and its menu works.
- [x] Watch the TR2 cutscene before the Wreck
  Lara changes into the diving suit at the right frame.
- [ ] `/set` an enemy's `max_hit_points` after wounding it
  Its current hit points rise by the same amount, not to full.
- [x] Operate a TR4 pulley switch, saving and reloading mid-sequence
  The pulls already done still count.
- [x] Start a level with NG+ enabled
  Enemies still have double hit points.
- [x] Start Floating Islands
  3D secret meshes match 2D sprites
- [x] Start Floating Islands, save, reload
  3D secret meshes match 2D sprites

### DIFF
--- a/src/trx/game/objects/general/pulley_switch.c
+++ b/src/trx/game/objects/general/pulley_switch.c
@@ -18,7 +18,7 @@ typedef struct {
     bool is_locked;
     bool is_single_use;
     int32_t required_pulls;
-    int32_t remaining_pulls;
+    int32_t pulls_done;
 } M_PRIV;
 
 static const OBJECT_BOUNDS m_Bounds = {
@@ -34,12 +34,19 @@ static const OBJECT_BOUNDS m_Bounds = {
 
 static const XYZ_32 m_Position = { .x = 0, .y = 0, .z = -148 };
 
+// What is left to pull. A requirement raised mid-pull asks for the difference;
+// one lowered past what has already been done leaves nothing to do.
+static int32_t M_RemainingPulls(const M_PRIV *const p)
+{
+    return MAX(0, p->required_pulls - p->pulls_done);
+}
+
 static void M_LoadPriv(ITEM *const item, JSON_READ_IO *const io)
 {
     M_PRIV *const p = item->priv;
     JSON_SHOULD(JSON_READ(io, "is_on", &p->is_on));
     JSON_SHOULD(JSON_READ(io, "is_locked", &p->is_locked));
-    JSON_SHOULD(JSON_READ(io, "remaining_pulls", &p->remaining_pulls));
+    JSON_SHOULD(JSON_READ(io, "pulls_done", &p->pulls_done));
 }
 
 static void M_SavePriv(const ITEM *const item, JSON_WRITE_IO *const io)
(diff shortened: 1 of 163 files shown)


src/trx/game/objects/general/pulley_switch.c
```diff
@@ -34,20 +34,30 @@ static const OBJECT_BOUNDS m_Bounds = {
...
+    return MAX(0, p->required_pulls - p->pulls_done);
+}
+
 static void M_LoadPriv(ITEM *const item, JSON_READ_IO *const io)
 {
     M_PRIV *const p = item->priv;
     JSON_SHOULD(JSON_READ(io, "is_on", &p->is_on));
     JSON_SHOULD(JSON_READ(io, "is_locked", &p->is_locked));
-    JSON_SHOULD(JSON_READ(io, "remaining_pulls", &p->remaining_pulls));
+    // Saved as what is left, which is what earlier saves carry.
```
lahm86: Is this only for the benefit of current save files? The object is unreleased, so would it be easier to read/write `pulls_done`?

src/trx/game/objects/general/pulley_switch.c
```diff
@@ -98,22 +107,23 @@ static void M_ControlWithLara(ITEM *const item)
...
-    if (p->remaining_pulls == 0) {
+    p->pulls_done++;
+    if (M_RemainingPulls(p) == 0) {
         if (p->is_single_use) {
             item->trigger.spent = true;
         } else {
             // Potentially multiple pulls to activate, but one to deactivate.
             p->is_on = !p->is_on;
-            p->remaining_pulls = p->is_on ? M_MIN_PULLS : p->required_pulls;
+            // Turning it back off takes a single pull.
```
lahm86, line 121: Maybe we can collapse the two comments together.

src/trx/game/objects/general/pulley_switch.c
rr-: Yeah, I can simplify that.

aredfan (approved):

src/trx/game/objects/general/pulley_switch.c
rr-: Collapsed :+1:

src/trx/game/objects/general/pulley_switch.c
rr-: Simplified.

lahm86 (approved):

PR #5990, opened 2026-07-29
Fix the Floating Islands secrets, and add trx.objects.swap_sprite

rr-:
#### Checklist

#### Description

The Floating Islands script now swaps the sprites of its second and third secret, through a new `trx.objects.swap_sprite()`.

The level hands the two secrets each other's sprite. The fix we shipped swapped the inventory models to match, which left those wrong instead. Swapping the sprites was not something a script could ask for: an object drawn from a sprite keeps it where a modeled one keeps its meshes, so `swap_mesh` moves the wrong thing when handed one.

Players see the Silver and Jade Dragons the right way round.


lahm86 (approved):

aredfan:
> Players see the Silver and Jade Dragons the right way round.

Whilst the fix makes sense for the dragon's ordering in the stats, I think some players will be more familiar with seeing the jade dragon on the roof of the floating house near the level's start.

Building on top of this PR, perhaps we could do a data fix and swap the jade/silver dragon positions?

rr-:
> > Players see the Silver and Jade Dragons the right way round.
> 
> Whilst the fix makes sense for the dragon's ordering in the stats, I think some players will be more familiar with seeing the jade dragon on the roof of the floating house near the level's start.
> 
> Building on top of this PR, perhaps we could do a data fix and swap the jade/silver dragon positions?

IMO it's an arguably an OG bug because Silver Dragon should be "easier" to get, and it was intentional.

@lahm86  WDYT?

lahm86:
Hooked in an injection to swap the items' positions.
Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/324

rr-:
Testing – secrets should appear just like on develop, with order fixed in the stats (warrants a better changelog entry):

- [x] New play
- [ ] Resume a save made on develop
- [x] Resume a save made on this branch

lahm86:
> Resume a save made on develop

I think this will retain the items' positions from the saves. If we want that to work, I think we'll need to either swap positions in Lua (but then we'd have to detect if it's been picked up or not), or use both the sprite and the model swap that was previously in place?

rr-:
Hmm I think it's OK to let it be if it's too much work.

aredfan (approved):

rr-:
I merged both commits into one and refined the changelog entry to be more truthful what's done. I'd appreciate one final look before shipping :)

aredfan (approved):

PR #5996, opened 2026-07-29
camera: clear the hand-back flag on flyby reset

rr-:
#### Checklist

#### Description

A track path sequence leaves it standing, and the next update would then put the camera on Lara although no sequence is running.

### DIFF
--- a/src/trx/game/camera/flyby_mode.c
+++ b/src/trx/game/camera/flyby_mode.c
@@ -320,6 +320,9 @@ bool Camera_Flybymode_Cancel(void)
 void Camera_FlybyMode_Reset(void)
 {
     m_CurrentSequence = M_NO_SEQUENCE;
+    // Left standing, this would send the next update down the hand-back path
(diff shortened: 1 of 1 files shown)


aredfan (approved):

lahm86 (approved):
Maybe we should just reset `m_State` entirely. I think the reset only runs when a level unloads so we don't really want anything stale to be leftover in theory.

rr-:
Not as written; `Camera_FlybyMode_Reset` is also reached from `Camera_FlybyMode_Deactivate`, which then reads the state it would have wiped:

```c
Camera_FlybyMode_Reset();
...
Viewport_AlterFOV(m_State.initial.fov, m_State.initial.fov_mode);
if (!m_State.flags.spline_to_game) {
    g_Camera.pos = m_State.initial.camera_pos;
    g_Camera.target = m_State.initial.camera_target;
```

lahm86:
Ah of course, thanks 👍

lahm86 (approved):

PR #5999, opened 2026-07-29
objects/generic_switch: add

lahm86:
#### Checklist

#### Description

This adds TR4's six generic switch types, with a mode property to dictate how Lara should interact with them. I've converted OCBs into the Lua scripts for each level applicable levels.

- Negative => hidden pickups
- 0 => regular wall levers
- 3 => single-use shove button
- Anything else => reach-in-to-grab

Note that for ones with pickups, these do nothing yet. I'll follow that up separately: the control is in the pickup module and the mode on the switch means that Lara does not interact with it (otherwise it'd interfere with the pickup logic). I was toying with naming this mode `DECORATIVE` rather than `HIDDEN_PICKUP` as the switch itself knows nothing about the item, but felt it's maybe more familiar for builders. LMK what you think.

The shove button in `joby1a.tr4` should now use the correct animation.

Resolves TRX796.
Resolves TRX818.
Resolves TRX823.


rr- (commented):
How do we feel about doing the pickups entirely in Lua, and collapsing both in-reach branches?

Alternatively, wire the hidden pickup mode to do give her pickups, that would come from declared properties, instead. But do that in generic switch object rather than pickup.

lahm86:
> How do we feel about doing the pickups entirely in Lua, and collapsing both in-reach branches?
> 
> Alternatively, wire the hidden pickup mode to do give her pickups, that would come from declared properties, instead. But do that in generic switch object rather than pickup.

I'd be inclined to keep it separate from switches, as the two have no bearing IMO. I think it was just used for convenience - in theory you could have her reach in for a hidden pickup without the switch object at all, or even place an animating or static in its place.

It'll be a a lightweight change to the pickup module, effectively just another mode alongside the plinth types. Maybe we can see how it looks once I've got that PR sorted out.

rr-:
Ohh got it so there is a separate physical pickup item inside this. Sounds good.

rr- (approved):

lahm86:
Changelog entry added re the shove button camera fix.

aredfan (approved):

PR #6006, opened 2026-07-30
Finish support for the TR4 title sequence

rr-:
#### Checklist

#### Description

The TR4 main menu now plays the title level behind it as the original game does: a flyby sequence sweeps the level, a trigger along its path starts a cutscene, and the cutscene hands over to the next sequence when it ends, cycling for as long as the menu is up. The engine reads the scenes out of `cutseq.pak`, and the title's own Lua script decides which cutscene each flyby leads to, so the order lives in data rather than in engine code.

- A gameflow that names no `main_menu_picture` shows its title level behind the menu instead. This takes over from the title's `flyby_sequence`, which decided both that the level shows and what plays there; the script decides the second now.
- `trx.cutscenes` plays a cutscene, reads and rewrites which ones have run, and sets the field of view and the depth of the cinematic bars a scene is framed with. Three events surround a scene - the trigger, the start and the end - and a handler may answer a trigger with something of its own, which is how the title threads its cycle together.
- The title level is loaded on its own rather than through the level sequencer, so it never dressed Lara or armed her for the scenes she acts in. It does both now, and its gameflow entry may name an outfit like any other level's. A level that holds her but none of the outfit swap objects gets a line in the log rather than a failed assertion.
- The cutscene triggers in the TR4 levels themselves work too, in part: the cameras and the actors run, and a save remembers which scenes have already played, while the touches individual scenes add - hiding an actor part-way through, swapping a mesh - are still to come.

Players see the TR4 main menu the way the original had it. The rest is groundwork for the in-game cutscenes.


aredfan:
Fantastic job, thank you for working on this.

Testing (in progress):

- [x] Cinematic bars doesn't appear in the title screen in OG, It's only used for in-game cutscenes and certain flybys.
- [x] i27 flame emitter is antitriggered and doesn't get triggered again. OG bug, I'm guessing it's a data issue.
- [ ] Lara disappears at the edge of the screen after pushing the mummy onto the spikes. I'd speculate this is an OG bug as the game was built for 4:3 screens.
- [x] Launch tr4, `/mod tr1` and the game crashes.
- [x] The other mods title backdrop is black screen instead of a background image.
- [ ] When the camera goes through the window near the beginning, one of the static objects disappear at the top-right corner.  Guessing it might be a data issue.

> <details><summary>Before</summary>
> 
> Screenshot 1
> 
> </details>
> 
> <details><summary>After</summary>
> 
> Screenshot 2
> 
> </details>

lahm86 (approved):
Very nice ☺️

rr-:
For Lara, I am almost positive the cutscene data cuts short exactly due to the reasons stated. For the static, still investigating.

rr-:
I can confirm the arch static disappearing is our fault, but I'd like to handle this in a separate PR.

aredfan (approved):

PR #6009, opened 2026-07-30
Fix settings and preset dialog text running off screen

### LINKED ISSUE #6000: [Bug] Gameplay settings text is partially hidden in 4:3
### Game type

Original game

### Game

TR3

### TRX version

TRX 1.9.3-469-ga5dd81c

### Operating system

Windows 11

### GPU

NVIDIA GeForce RTX 4060 Laptop 

- Launch TR3 in 4:3 (change visual settings to 4:3 if you're not in 4:3 already) , and go to the gameplay settings: text is partially hidden on both sides of the screen.
- Also very visible when trying to apply a preset.

Settings text should be properly within the game window in 4:3.

### Attachments

Image

Image

[TRX.log](https://github.com/user-attachments/files/30519694/TRX.log)

### Before submitting

rr-:
#### Checklist

#### Description

Settings text no longer runs off both sides of the screen in the longer languages, and the list of settings a preset would change now wraps inside its dialog instead of spilling past it. A dialog too wide for the screen is drawn smaller until it fits, down to a floor of 0.85, and dialogs keep five units clear of the edges.

- 4:3 is where this showed. The canvas is 640 units wide there against 853 at 16:9, and the German, French, Polish and Russian gameplay tabs need more than 640. Nothing that ships today reaches the floor, so no wording had to change; the floor is there to catch a future string that goes further.
- The preset confirmation was a separate fault. Its list was drawn in a box of a fixed 72 units, and sizing a box does not constrain what is drawn inside it, so the frame appeared that narrow while the setting names ran hundreds of units past it on both sides.
- New tests lay both dialogs out at 640x480 for every game and every shipped language, and check that nothing reaches past the screen or past the box it was given. Glyph metrics come from each game's font.bin, so there is no generated table to keep in step, and the suite needs no game data beyond what this repository versions.

Players see the settings screens fit the window at every aspect ratio and in every language.

Before:
20260730_115945_Level_0

After:
20260730_115851_Level_0

Resolves #6000 / TRX843

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -98,6 +98,13 @@ repos:
     additional_dependencies:
       - pyjson5
 
+  - id: lint-test-deps
+    name: Unit tests reach no engine dependency
+    entry: tools/lint/checks/test_deps
+    language: python
+    pass_filenames: false
+    files: \.[ch]$
+
   - id: lint-lua-module-order
     name: Lua module order numbers are unique
     entry: tools/lint/checks/lua_module_order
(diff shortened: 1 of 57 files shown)


aredfan (approved):

rr-:
(Fixed unit tests breaking on missing SDL2 in CI/CD)

lahm86 (approved):

PR #6016, opened 2026-07-30
objects/pickup: add crowbar prying support

lahm86:
#### Checklist

#### Description

This adds support for pickups that need to be pried out of a wall using the crowbar. It also fixes Lara being able to collect hidden receptacle pickups while crouched (unreleased bug).

I've updated Lara's idle animation to add a state change for this pickup type, as OG hardcoded the switch to it. This keeps it in line with all other pickup types in TRX. OG also didn't save the pickup's animation mid-pry, but I'm not sure if it was possible to save during that animation or not - if it was, I'll add a bug fix changelog entry.

If Lara enters the fly cheat mid-pry, the item will return to its original state, otherwise it continues on its own and can't be interacted with.

This involved a minor cleanup in the pickup module, so I'd appreciate a sanity check of the other pickup types as well.
- [x] regular on the floor
- [x] plinths
- [x] hidden behind a receptacle
- [x] flares

Resolves TRX586.

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/326


rr- (approved):

aredfan:
- [x] I found a couple of positions where Lara may step eagerly away from the wall pickup. [record.txt](https://github.com/user-attachments/files/30561512/record.txt)

lahm86:
Should be fixed, thanks.

aredfan:
- [x] The 1st of Lara's positions is resolved, the 2nd remains an issue. In hindsight, it wasn't a good idea to combine 2 bugs in one recording. [record.txt](https://github.com/user-attachments/files/30576678/record.txt)

lahm86:
Hopefully that one's sorted too now.

aredfan (approved):

PR #6020, opened 2026-07-30
Port the /give command to Lua

rr-:
#### Checklist

#### Description

This PR ports the `/give` command to Lua, concluding the process of porting all console commands.

A couple of additional things:

- A few query families were introduced: `gun`, `ammo`, `supply`, `tool`, `key`, `puzzle`, `quest`, `examine`, `collectible` (`O_PICKUP_ITEM_1-4`) and `secret`. They do not cover variants between them: a second state of something Lara already carries, such as a part-full waterskin or an ammo variant, is in none of them.
- A large medipack answers to "big medipack" as well, so `/give big medi` finds it.
- A family answers to its own name, so `/give key` hands over every key item rather than fuzzy-matching the one object that happens to share the word.
- A pickup with several states shares one backpack entry, so `/give scion` is one scion rather than one per state.
- A group that finds nothing to hand over says so instead of playing the cheat's sound over an unchanged backpack, and a count below one is refused in the command's own words.

The scripting side gained what the command needed:

- `trx.lara.inventory`
- `trx.lara.weapons`
- `trx.stats`
- `trx.game.is_ngplus`
- pickup families on `trx.objects.query`

With `/give` gone from C, nothing declares a console command there any more, so the constructor macro that registered one goes, and with it the removal-by-proc that told a C command from a Lua one.

#### Testing

- [x] `/give uzi`
- [x] `/give 5 flare`
- [x] `/give big medi`
- [x] One per family:
    - [x] `/give gun`
    - [x] `/give ammo`
    - [x] `/give supply`
    - [x] `/give tool`
    - [x] `/give key`
    - [x] `/give puzzle`
    - [x] `/give quest`
    - [x] `/give examine`
    - [x] `/give collectible` (for example the Talion in the Ice Palace)
    - [x] `/give secret`
- [x] `/give scion` is one scion, not one per state
- [x] `/give crystal` hands one over; `/give all` and `/give keys` do not
- [x] `/give keys` leaves the crowbar, lasersight, binoculars, waterskins and leadbar behind; `/give all` brings them
- [x] `/give 0 vase` and `/give kettle` are refused, each saying why
- [ ] `/give all` on a level carrying no inventory models reports that it gave nothing (I don't think this is testable in any OG level)
- [x] `/keys`, `/guns`, `/moreguns`, and the same three through `/give`
- [x] `/guns` respects what the level allows where `/moreguns` does not
- [x] A NG+ run tops every weapon's ammunition up
- [x] The item cheat key (`I` by default) still fills the backpack – it now just evaluates `give all`
- [x] Tab on an empty `/give ` offers the keywords and the names together, and on `/give big med` replaces the whole tail
- [x] `/tp pickup` never teleports to a Savegame Crystal


lahm86 (approved):

aredfan:
Issues (testing in progress):

- [x] `/keys` and `/give keys` doesn't provide Lara with a Lead Bar in Palace Midas.

rr-:
Fixed :)

aredfan (approved):

rr-:
I patched `/give secret` to report "Unknown item: secret" in TR2 instead of reporting nothing and doing nothing (secrets are special). `/tp secret` should still work.

aredfan (approved):

PR #6022, opened 2026-07-30
Fix teleporting to receptacle switches

rr-:
#### Checklist

#### Description

A switch set into a wall has its item inside the solid block behind it, with only the model reaching through into the room, so sending Lara to the item itself left her in the wall with her back to the switch. She now goes to a spot off the near face of the model, taken from its bounding box, looking at it.

Resolves TRX844.

#### Testing

Just spot-test teleporting to the said receptacles and a few others, if she's facing the objects as expected.
One win from this is that she will be placed right in front of the levers and switches, ready to interact.
LMK if we would like to improve any other common objects, happy to enhance this PR.


lahm86 (approved):
Thanks ☺️ 
Can't think of any others atm that are as awkward.

rr-:
I also added a special case for push blocks, so that she teleports in front of them instead of on top. This also makes it now possible to teleport near the push block in Vilcabamba, which the command would refuse before.

aredfan:
- [x] I'm getting `bad argument` logs when issuing some teleport commands.  E.g. `/tp receptable` in Karnak, `/tp push` in Natla's Mines.

<details><summary>Screenshots</summary>

20260731_112121_Temple_of_Karnak

20260731_112145_Natla&#39;s_Mines

</details>

rr-:
Thanks – fixed :relieved:

aredfan (approved):

PR #6024, opened 2026-07-31
option/passport: reset game mode on new game

### LINKED ISSUE #960: New game plus doesn't activate if enable_game_modes is false

rr-:
#### Checklist

#### Description

The game mode is global state that outlives a return to the title screen, and starting a new game only cleared it when the game modes dialog was set to always show, so the next new game ran in the mode the previous one was started in. The same held for Play Any Level when no modes were on offer.

The reset was dropped in 2023 to resolve #960, back when beating the last level set the New Game+ flag itself. Completing the game now sets `profile.new_game_plus_unlock`, which the on-completion policy reads, and nothing else raises the flag.

Never therefore means every new game is a regular one, which the option's description now says rather than promising only that the dialog stays hidden.
It's a bit controversial, because it's no longer possible to reach NG+ under this policy. At the same time it's more honest in terms of the UI. I'm inclined to move forward with this, and would be more inclined to remove the Japanese modes which I have never seen anyone play.

Resolves TRX868

#### Testing

Policy: always

- [x] The mode selection is always reflected correctly
    - [x] When the last started game was NG
    - [x] When the last started game was NG+
- [x] Play any level behaves as expected

Policy: on completion

- [x] The mode selection is always reflected correctly
    - [x] When the last started game was NG
    - [x] When the last started game was NG+
- [x] Not shown until the player has beaten the game
- [x] Play any level behaves as expected

Policy: never

- [x] Always starts normal NG, never NG+
- [x] Play any level behaves as expected


lahm86:
> I'm inclined to move forward with this, and would be more inclined to remove the Japanese modes which I have never seen anyone play.

Agreed.

lahm86 (approved):

aredfan (approved):

PR #6027, opened 2026-07-31
lara/col/land: ignore step glitch fix in QWOP state

lahm86:
#### Checklist

#### Description

This avoids the fix for the step glitch interfering with QWOP. It's impossible to step glitch from a QWOP anyway as hitting the wall during that move would reset the QWOP state. Recording available in the ticket.

Resolves TRX870.


rr- (approved):

aredfan:
Unsure if the fix works because it breaks the recording. Approving anyway in case it's working fine.

aredfan (approved):

lahm86:
Interesting, the recording is OK my end before and after. Does it work if you remove the turbo speed (line 4)?

aredfan:
Removing turbo speed from the recording has no effect, at least for the PR build. The recording works as expected in develop and Lara enters the QWOP state.

lahm86:
Recording sorted out on Discord.
Force push above for the changelog conflict.

Issue #5929, opened 2026-07-25
[Bug] Willard appears in a cutscene in Antarctica


WalterBianco:
### Game type

Original game

### Game

TR3

### TRX version

TRX-1.9.3-339-g15cf55f

### Operating system

Windows 10

### GPU

GTX 1050ti

### Describe the bug in plain terms, and write the shortest steps that reproduce it.

In the Antarctica level cutscene, at a certain point Willard also appears at the back window, with the model used for him running away.

### What did you expect instead?

The model used for the later part of the Antarctica cutscene doesn't appear at all in the background until the moment of him running away is shown

### Attachments

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/57744913-6dfb-4f4d-b4d1-2e82948db6e7" />

### Before submitting

- [x] I included enough detail to reproduce this.
- [x] I attached `TRX.log` and any saves, media, or custom level files needed to reproduce this.
- [x] This issue covers one bug.

Issue #3668, opened 2025-07-31
OG bug: disappearing geometry in Atlantean Stronghold


aredfan:
1. launch trub
2. `/play 3`
3. `/tp 16`

<details><summary>Screenshot</summary>

![Image](https://github.com/user-attachments/assets/aaf42384-da8e-465c-8803-beccd783b19c)

</details>

aredfan:
I think the underlying issue here is Atlantean Stronghold has untextured parts on the other side of the alpha textures. This would explain the disappearing geometry.

<details><summary>Image</summary>

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/e6d6567b-8cc4-4519-ae1a-10506405f992" />

</details>

The second issue in Atlantean Stronghold is the cavern hole itself is too big, which differs from The Hive.

<details><summary>Atlantean Stronghold images</summary>

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/16de22f5-8803-4916-b579-3b678bdacebf" />
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/689ac299-ca5a-48ba-af96-d0deb2b47974" />

</details>

<details><summary>The Hive images</summary>

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/b1a2b241-ab9e-49b8-9030-d73b6ebab412" />
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/bdcd8d6c-db61-4b86-ab83-e896f93eeee3" />

</details>

lahm86:
The geometry here is really bizarre. It seems the ceiling is lowered to floor level in the first row of sectors in room 18, which is why there are no textures above.

The visibility portal is also too low, which causes the clipping. Another fine mess 😄 But fixable.

<details>

<img width="800" height="432" alt="Image" src="https://github.com/user-attachments/assets/8be5b519-4e73-4df5-bb81-2819bc21a455" />
</details>

Issue #2452, opened 2025-02-06
OG bug: disappearing polygon faces


aredfan:
In City of Khamoon, these trapdoors disappear and partially reappear depending on the camera angle. I feel #2005 is related but I'm not entirely sure. 

![Image](https://github.com/user-attachments/assets/ec183915-b953-4e3b-a563-ced9a19c7a6a)

Regarding the cat statue, a few polygon faces disappear and reappear, again depending on the camera angle.

![Image](https://github.com/user-attachments/assets/d2135f92-a5a3-4a79-899d-bf22ebd856be)

rr-:
Hi, are we really sure it's not a TR1X error?

aredfan:
> Hi, are we really sure it's not a TR1X error?

Hey, it's OG for sure, I attach the TombATI saves for the locations shown above: [saveati.zip](https://github.com/user-attachments/files/18688970/saveati.zip)

lahm86:
The trapdoor issue is similar to #2438. The items are basically in the wrong room and so the clipping chops away the vertices when they're beyond the room bounds in the room below. The cat is an odd one though as it's in the correct room as far as I can tell.

rr-:
@lahm86 if you choose to work on this, let's hold off until we land the GPU-based projection matrices (to be worked on after the 4.8 release). It should naturally resolve this and related issues with portals.

rr-:
Related: #2438

Issue #2438, opened 2025-02-02
Cowboy sometimes loses his head


rr-:
https://github.com/user-attachments/assets/80d90591-7e8c-40f0-8283-fdeb31fffac3

[save_tr1_00.dat.zip](https://github.com/user-attachments/files/18634503/save_tr1_00.dat.zip)

aredfan:
I can confirm this issue is present in OG with the attached save:

[saveati.zip](https://github.com/user-attachments/files/18642849/saveati.zip)

PR #6029, opened 2026-07-31
items: queue items into the rooms they reach

rr-:
#### Checklist

#### Description

An item is now offered to the neighbouring rooms it reaches into as soon as the level's rooms are in place, and the question is decided from the widest box any of the object's frames reaches rather than from the box of the frame currently being drawn.

Before this, a neighbour was only offered an item when that item's room was recomputed. That never happens for one that stands still, so such an item stayed queued in its own room alone for the whole level and vanished whenever that room dropped out of the draw list, even while the room it was leaning into was still on screen. The box that settled the question came from the frame on screen too, so the answer went stale the moment the item animated. Doing this at `Item_Initialise` is not an option, as it runs before the portals have their bounds.

The Loch Ness monster in Highland Fling shows both halves: it stands in one room and rears up through the ceiling into the one above, and it flickered in and out as the camera turned.

For players, enemies and objects that lean out of the room they belong to no longer come and go with the camera angle.

Resolves TRX874. Ref. f6bd91f785 and d2d20fd49c

#### Testing

- [x] Loch Ness Monster from the recording is now visible at all times

Regressions:

- [x] #2005 Vilcabamba entrance gates; Catacombs room 30, walk Lara forward and watch the trapdoor
- [x] #2438 Cowboy sometimes loses his head (TR1, OG bug)
- [x] #2452 City of Khamoon trapdoors and the cat statue's faces;
- [x] #3668 Atlantean Stronghold: `/play 3`, `/tp 16`
- [x] #1870 TR2 Lara's Home
- [x] The helicopter in Highland Fling crossing portals
- [x] Fish crossing room boundaries
- [x] Bats leaving their spawn room
- [x] The animating diver in Sleeping with the Fishes
- [x] #5929 Willard in the Antarctica cutscene

The changes here make things visible in _more_ rooms, so the risk runs the other way too.
The sharpest test is #2393, which has exact steps: `/play 10`, drop through the opening, kill the barracudas with a conventional weapon, `/tp 51 4 48`, then move the camera. Only the "visible" half is in scope — targeting doesn't go through this code.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -93,6 +93,7 @@
 **TR3**
 - fixed Willard being visible outside the hut at the beginning of the cutscene following Antarctica (resolves #5929)
 - fixed scenery that leans out of its room, such as the streetlight glow in It's a Madhouse!, flickering as the camera turns (OG bug)
+- fixed enemies and objects that reach out of their room, such as the Loch Ness monster in Highland Fling, flickering as the camera turns (OG bug)
 - fixed Lara, when on fire, not extinguishing at the right water depth compared with OG (regression from 1.1)
 - fixed the waterfall and drowning mist bunching up in one spot instead of spreading out along the water (regression from 1.2)
 - fixed a crash when a level holds fewer raptors than its raptor emitters can send out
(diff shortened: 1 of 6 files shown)


lahm86 (approved):

aredfan:
Testing (in progress):

- [x] Vilcabamba entrance gates flickers in and out. [record.txt](https://github.com/user-attachments/files/30589836/record.txt)

rr-:
Fixed

aredfan (approved):

Issue #5975, opened 2026-07-28
[Bug] (TR1) Animated interactions on: can't pick up the hidden large medipack in Sanctuary of the Scion


drydenj:
### Game type

Original game

### Game

TR1

### TRX version

1.9.3-412-g23e04e7

### Operating system

Arch

### Describe the bug in plain terms, and write the shortest steps that reproduce it.

I've attached save files standing at the items. There's also two other pickup locations I've found to have minor issues with Animated interactions on, as below.

With Animated interactions off, all items can be picked up without problem.

With Animated interactions on:

- save_tr1_00: You cannot pick up the hidden large medipack inside Set's ear in Santuary of the Scion. I've tried from multiple angles to no avail.
- save_tr1_01 (tomb of tihocan): At this angle, you cannot pick up the small medipack at your feet. Lara won't move at all if you hold down action. Note also that Lara's is unable to move again for a couple of seconds after trying to pick the item up. Try holding look + left/right down and you'll observe her face/body snap to the side instantly after she unfreezes.
- save_tr1_02 (obelisk of khamoon): Only with your back to the decoration object's mesh, you cannot pickup the items. Lara will continuously sidestep one way and then the other way if you hold down action.

Not sure if I should separately report the other two, particularly the tomb of ticohan instance due to lara becoming frozen momentarily. It's late. I can test stable + older versions from feature introduction (1.0) later.

### What did you expect instead?

Lara should be able to pick up the items.

### Attachments

[save_tr1_.zip](https://github.com/user-attachments/files/30439393/save_tr1_.zip)

### Before submitting

- [x] I included enough detail to reproduce this.
- [x] I attached `TRX.log` and any saves, media, or custom level files needed to reproduce this.
- [x] This issue covers one bug.

PR #6053, opened 2026-08-02
lara: fix movement during animated interactions

lahm86:
#### Checklist

#### Description

This fixes two issues with animated interactions. Firstly, when in the controlled walk states, during long interactions she would stop after one animation cycle. In fixing that, it became evident that Lara was moving too quickly towards her targets, which was a result of her position being set during interaction control and then again in `LaraAnimate`. TR4 prevents animation position shifts while Lara is being controlled.

~~This also fixes the Tihocan case in #5975; the other two problems there are caused by static collision, which is a separate beast.~~

I added interaction debug to the UI while investigating this, visible when `debug-pos` is enabled. It shows `target_item_num / is_moving / move_count`. LMK if we'd prefer not to retain that.

Resolves TRX891.

Tests:
- [x] Switches, all above-water types
- [x] Pickups, all modes
- [x] Flares
- [x] Puzzle slots
- [x] Keyholes
- [x] TR4 interactive doors
- [x] Climbing pole


lahm86:
^ Tweaked changelog (`Tomb1Main` rather than `TR1X`)

lahm86:
Ignore the Tihocan comment, it doesn't fix it after all (I had walk-to off). Will look into that separately, it's likely because it's so close to the edge.

aredfan (approved):

rr- (approved):

PR #6068, opened 2026-08-04
Add script-defined trigger zones

rr-:
#### Checklist

#### Description

This PR introduces `trx.zones` Lua module, which lets a script mark out a piece of the level and hear about what walks into it. A zone is a box or a sphere in world space, or a single sector the way a floor trigger covers one, and it reports what enters it, what stays, what leaves, and what a flyby camera passes through.

```lua
local arena = trx.zones.box(min, max, { watch = "items", name = "arena" })  -- a box in world space
local bell = trx.zones.sphere(centre, 2048)                                 -- a sphere around a point
local plate = trx.zones.tile(pos)                                           -- one sector, in its own room

arena:on_enter(function(item) end)         -- something stepped in
arena:on_tick(function(item) end)          -- and again every frame it stays
arena:on_exit(function(item) end)          -- it left, or was destroyed inside
arena:on_flyby_enter(function() end)       -- a flyby camera crossed in
arena:on_flyby_exit(function() end)

arena:contains_item(trx.lara.item)         -- the test the zone makes each frame
arena:contains_point(pos)                  -- and the same test for a place
arena:occupants()                          -- who was inside as of the last frame
arena:disable()                            -- stop testing it, without forgetting who is inside
arena:clear_occupants()                    -- forget, so anyone still inside enters again
arena:remove()                             -- and take the hooks attached to it

trx.events.on_zone_enter(function(zone, item) end)   -- every zone at once, and siblings for the rest
trx.zones["arena"], trx.zones[1], #trx.zones         -- by name, by place, how many
```

A zone watches Lara unless it is made with `watch = "items"`, and then it answers for anything the level holds. Each frame settles who is inside every zone before it reports anything, and the exits go out before the enters, so a script following Lara from one zone into the next hears her leave the first before she enters the second. Something destroyed while it is inside counts as leaving, reported while its handle still resolves. Zones belong to the level that made them and are not written to savegames.

Two smaller pieces come with it:

- `trx.math.round_to_sector()` snaps a position or a single coordinate to the corner of the sector it stands in. It rounds west and north on both sides of the origin, which is easy to get wrong when it is written as a modulo.
- `trx.events.on_level_unload()` runs as the engine releases a level, while its world is still available. It gives scripts a chance to clean up what they set up before the next level takes over. Previously, level scripts weren't dropped until after the old level was gone and the new one had loaded, leaving modules with no opportunity to release level-specific state. With `Level_Unload`, cleanup happens while the ending level is still current.

Nothing changes for players on its own. For mod authors, the zones are a new module, and its transitions travel through the engine's event registry, so a zone hook carries an `events.Listener` and detaches through `trx.events` like any other.


src/lua/api/math.lua
```diff
@@ -102,6 +102,51 @@ local angle = trx.math.atan(lara.pos.z - pos.z, lara.pos.x - pos.x)]],
...
 
+api.define("math.round_to_sector", {
+  description = [[
+Snaps a position back to the corner of the sector it stands in, the way the
+level's own geometry is laid out. A whole position keeps its height: a sector
+is a column, and rounding it is about the ground plan rather than how far up
+the position sits. A single coordinate rounds on its own, which is what an axis
+at a time needs.
+
+The corner is always the one to the west and the north, on both sides of the
```
lahm86: Nit: I think it's South rather than North i.e. -Z

aredfan (approved):

src/lua/api/math.lua
```diff
@@ -102,6 +102,51 @@ local angle = trx.math.atan(lara.pos.z - pos.z, lara.pos.x - pos.x)]],
...
+  },
+  returns = {
+    type = { "math.Vec3", "math.Distance" },
+    description = "The corner of the sector, in whichever of the two came in.",
+  },
+  examples = {
+    [[local corner = trx.math.round_to_sector(trx.lara.item.pos)
+trx.zones.box(corner, {
+  x = corner.x + trx.math.WALL_L,
+  y = corner.y + trx.math.WALL_L,
```
lahm86: I wonder if a comment somewhere about -Y being up and +Y being down would be helpful for the box setup example. Maybe not needed, but it might not be obvious to everyone.

src/lua/api/zones.lua
```diff
@@ -0,0 +1,985 @@
...
+A zone watches Lara and nobody else, unless it is made with `watch = "items"`,
+and then it watches everything the level holds - enemies, pickups, anything
+with a position. Its hooks hand over whatever set it off. To listen in one
+place rather than zone by zone, `trx.events.on_zone_enter` and its siblings
+hear about all of them.
+
+A flyby camera passing through a zone is its own pair of hooks,
+`trx.zones.Zone:on_flyby_enter` and `trx.zones.Zone:on_flyby_exit`, since a
+camera is not an item and there is nothing to hand
+a handler but the zone. Every zone answers for a flyby, a tile by its room, and
```
lahm86: This last sentence is a bit confusing imo. Not sure what "a tile by its room" means on its own, and using "looked at" for the camera might be a bit ambiguous (maybe just "checked").

lahm86 (commented):
Very cool, just a few nitpicks.

src/lua/api/math.lua
rr-: Fixed; also fixed math.Box declaration.

src/lua/api/math.lua
rr-: Fixed

src/lua/api/zones.lua
rr-: Reworded

lahm86 (approved):
Great stuff 👏

PR #6077, opened 2026-08-04
objects/pickup: expose effect properties 

lahm86:
#### Checklist

#### Description

This exposes properties on pickup items to configure whether or not they glow and rotate like TR3 artefacts, and whether or not pickup aids are shown. The ability to set glow colour in Lua using hex strings is now in place as a result. I've added injection support as well, but will raise a TE injector PR after our next stable, hence these colour properties remain hidden from the XML files for the time being.

The glow in TR3's artefacts matches the OG curve exactly when using the hex value in the scripts provided here. Using `0xFF` gives a slightly different curve (not noticeable to my eye, only in the logs 😄) so I just opted to keep it at `0xF8` for the benefit of the doubt. LMK what you think.

We'll just need to confirm the artefacts behave correctly in each TR3 level they appear. And of course, TR4's quest items should now behave normally.

Resolves TRX914.

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/329

### DIFF
--- a/data/te/tr1/Properties/default.xml
+++ b/data/te/tr1/Properties/default.xml
@@ -158,6 +158,8 @@
     </moveable>
     <moveable id="84-94,110-113,126,129-132,141,142,144,187,241-246,248,249,251-256">
         <property category="TRX" defaultValue="Normal" description="Pickup animation mode." displayName="Pickup Mode" entries="Normal, Low Pedestal, High Pedestal, Hidden Reach-in, Crowbar" internalName="pickup_mode" type="Enum" />
+        <property category="TRX" defaultValue="true" description="Show a twinkle effect above the item." displayName="Show Pickup Aid" internalName="show_pickup_aid" type="Bool" />
+        <property category="TRX" defaultValue="0" description="How much to rotate the item by each frame while it's active, in engine angle units." displayName="Rotation" internalName="rotation" maxValue="16384" minValue="-16384" type="Int" />
     </moveable>
     <moveable id="145">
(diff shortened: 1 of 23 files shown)


rr-:
Two nits:

1. I think the scripts may be spinning the artefacts at twice the OG speed. `DEG_90` is `0x4000`, so `trx.math.DEG_90 // 8` is `2048`, but the code this replaces used `item->rot.y += 1024`. It should be `DEG_90 // 16` (or `DEG_360 // 64`).
2. The documented range on rotation is not enforced. `pickup.c:947` declares "minimum -16384; maximum 16384" with a plain `OBJECT_PROPERTY`, so Lua and injection accept anything an `int16_t` holds while the generated TE property caps at ±16384. In other places, the range sentence appears it comes with a check (`lift.c` speed, `big_bowl.c` `flip_slot`). The solution is to either allow `int16_t` unchecked (and lose the TE catalog bindings), or enforce it like so:
    ```c
    static const char *M_CheckRotation(const TRX_VALUE *const in)
    {
        return ABS(in->as_int) > DEG_90 ? "rotation is beyond a quarter turn"
            : nullptr;
    }

    OBJECT_PROPERTY_CHECKED(
        M_PRIV, rotation, 0, M_CheckRotation,
        "How much to rotate the item by each frame while it's active, in "
        "engine angle units. Value range: "
        "minimum -16384; maximum 16384."),
    ```

lahm86:
Thanks, @rr-. Fixed 😌

rr- (approved):

aredfan (approved):

PR #6081, opened 2026-08-04
Add a key to name each level

rr-:
#### Checklist

#### Description

Every level now carries a key: the name of the file it loads, without the directory and without the extension, in lower case. `data/WALL.TR2` reads back as `wall`, and a level that loads no file of its own reads `nil`.

```lua
local level = trx.game.current_level
level.key -- "wall"
```

Before this, a script keying its own data by level had `trx.game.Level.num`, a position that moves as soon as a game flow gains a level, or `trx.game.Level.path`, which is where the file sits on this install.

The second commit shares the path split the key needed: `File_GetBaseName` in core/filesystem, which the game string language scan, the overlay image scan and the level cache now call.

### DIFF
--- a/src/trx/core/filesystem.c
+++ b/src/trx/core/filesystem.c
@@ -127,14 +127,23 @@ char *File_GetParentDirectory(const char *path)
     return nullptr;
 }
 
-char *File_GetStem(const char *const path)
+const char *File_GetBaseName(const char *const path)
 {
     if (path == nullptr) {
         return nullptr;
     }
 
     const char *const last_delim = MAX(strrchr(path, '/'), strrchr(path, '\\'));
-    const char *const name = last_delim != nullptr ? last_delim + 1 : path;
+    return last_delim != nullptr ? last_delim + 1 : path;
+}
+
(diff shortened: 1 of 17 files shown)


aredfan (approved):

src/trx/core/filesystem.c
```diff
@@ -119,14 +119,36 @@ char *File_GetParentDirectory(const char *path)
         return nullptr;
     }
 
-    char *last_delim = MAX(strrchr(path, '/'), strrchr(path, '\\'));
+    const char *const last_delim = MAX(strrchr(path, '/'), strrchr(path, '\\'));
     if (last_delim != nullptr) {
-        return String_Format("%.*s", last_delim - path, path);
+        return String_Format("%.*s", (int)(last_delim - path), path);
```
lahm86: `int32_t`

lahm86 (approved):

src/trx/core/filesystem.c
rr-: Done

PR #6092, opened 2026-08-06
Add a Lua random module

rr-:
#### Checklist

#### Description

This PR adds `trx.random`, drawing from the sequence the simulation itself runs on, which the savegame already carries, so what a script draws comes back the same after a reload.

```lua
trx.random.random()                  -- a fraction of one
trx.random.randint(1, 6)             -- a whole number, both bounds included
trx.random.choice(seq)               -- one item out of a list
trx.random.choices(seq, weights, k)  -- several of them, with shares
trx.random.angle()                   -- a direction, anywhere around the turn
trx.random.chance(0.25)              -- whether something that likely happens
```


rr-:
I deliberately did not expose `Random_GetDraw` to avoid confusion, since nothing in Lua is exposed to drawing phase yet, and would only render the API confusing.

lahm86 (approved):

aredfan (approved):

PR #6094, opened 2026-08-06
music: default non-ambient floor data music as one shot

### LINKED ISSUE #5312: [Bug] Music repeats
### Game type

Original game

### Game

TR3

### TRX version

1.5

### Operating system

Windows 10

### GPU

GTX 1050ti

When passing through areas already passed, the trigger music gets triggered to play again. Happened on multiple occasions when playing through Area 51 in different rooms. Nowhere else did this happen when replaying other levels.

The specific trigger music plays only one time and doesn't when Lara goes over the same location again.

### Attachments

https://github.com/user-attachments/assets/8c32ccfd-adde-4ba4-bca6-cb01dbf783d0

### Before submitting

lahm86:
#### Checklist

#### Description

This defaults all non-ambient floor data triggered music tracks as being one shot in TR3 and TR4, aside from in TR3's gym level. If the config option to fix one-shot triggers is switched off, then these tracks can be replayed.

Marking FD triggers in the original levels as one-shot isn't feasible, as ones with additional commands would be impacted, e.g. making a door one shot, or timed switches usable only once.

I'd like to streamline the `Music_Trigger` function as a separate task - if we like the idea of rules, I can look at extending that to cover all the engine version quirks.

Resolves #5312 / TRX161.

### DIFF
--- a/data/trx/ship/games/tr3-la/scripts/_game.lua
+++ b/data/trx/ship/games/tr3-la/scripts/_game.lua
@@ -1,0 +1,5 @@
+trx.events.on_game_start(function()
+  local level = trx.game.current_level
+  local use_one_shot = level ~= nil and level.type ~= trx.game.LevelType.GYM
+  trx.rules.set("music.is_one_shot_default", use_one_shot)
+end)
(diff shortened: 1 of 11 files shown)


rr- (approved):
Yeah it'd be ideal to make the C side engine agnostic.

aredfan (approved):

PR #6097, opened 2026-08-06
Rewrite the config layer as a runtime option registry

rr-:
### Checklist

### Description

This PR rewrites the config module. Every setting is now an option registered at runtime that owns its own value, and `g_Config` becomes a const projection offering convenience **read-only** access.

**The registry**

An option was a row in a static table whose `target` pointed into `g_Config`, so `g_Config` was where a setting lived and anything could write one behind the config layer's back. The reads are untouched; the writes that went around the layer are compile errors until they go through the option. An config option write says which option has changed and the change event carries them, so `Shell_HandleConfigChange` can now just ask whether an option it monitors is in the payload, rather than diffing the whole struct against `g_SavedConfig`.

**Holds**

A mod, a level or the engine can now hold an option at a value, and a held option cannot be edited in the menus. That one state replaces enforced, overridden and disabled, which were three names for it. A write made while a hold is in force lands on the hold rather than on the option, so the player's own choice survives underneath, comes back when the hold is released, and is what the file saves either way.

**Bounds**

An option states its range in `map*.def`, and `Config_Sanitize` is a loop over the registry holding every option to it. Before this, a range was written wherever it was needed: `visuals.fov` had a `CLAMP` in `Config_Sanitize` and a min/max in its settings tab. The tabs drop their min/max now. Two cases a range cannot state stay behind in `M_SanitizeSpecial`: `aspect_mode`'s set and fps being 30 or 60.

**UI**

Each tab `.def` becomes just a list of keys: which option it shows. The option knows the rest – its name, what it accepts, the values an enum row offers. What is left over is behaviour, and it lives in `UI_SETTING_HANDLER`, registered from the file that owns it:

- `format_value`
- `can_change_value`
- `request_change_value`
- `is_available`
- `is_visible`
- `is_enum_value_available`
- `delta_slow`, `delta_fast`
- `enum_order`

What goes with that:

- `UI_SETTINGS_OPTION`
- `X_UI_CFG_ENUM`, `X_UI_CFG_DYN_ENUM`, `X_UI_CFG_RGB888`
- the tabs' row arrays, and the 24 enum value tables that listed exactly the enum's own values
- `setting_helpers/enums.c`, `setting_helpers/enums.h`, `setting_helpers/handlers.h`

**File sections**

A module can own a section of the settings file: it names the key and says how to read and write it. Input layouts and the gym's track records use this, so `config/priv.c` no longer knows how either is spelled or reaches into `game/input.h` and `game/gym.h`.

**Duplicate declarations**

Seven settings were declared in `map.def` and again in a game's own `map*.def`, meaning to override. That didn't work as expected. Each game now declares its own, and a setting a game declares twice is an error where it used to warn and keep the first. TR3 now starts with save crystals healing and TR4 with target lock off, which is what both were already written to do.

### Testing

- [x] Fresh launch with no TR*X.json5: the file is written, holds the defaults, and reads back unchanged on the next launch
- [x] A pre-rewrite TR*X.json5 is read with no value lost, and rewritten without dropping keys this build has no option for
- [x] A hand-edited value outside an option's bounds is brought to the nearest one it accepts, and the file is rewritten with it
- [x] A hand-edited value of the wrong type falls back to the default
- [x] Custom input layouts survive a restart (the input module's own section)
- [x] Gym track records survive a restart (the gym's own section)

**Handling each type**

- [x] String (the language setting)
- [x] On/off switch
- [x] Color
- [x] Enum (texture filter, ally hostility policy etc.)
- [x] Dynamic enum (outfits, bar looks)
- [x] Percentage
- [x] Integer

**Every ui tab**

- [x] Change one setting, leave the dialog, restart, and confirm it stuck
- [x] The rows are the same as pre-rewrite
- [x] Hidden options are hidden
- [x] Greying out options works
- [x] Nothing overflows the dialog at 4:3, in every shipped language
- [x] Options defaults is restored to the expected values
- [x] The presets tab applies a preset and the affected rows follow it

**Taking effect without a restart**

- [x] Master, sound, music, ambient, cutscene and FMV volumes
- [x] Fullscreen, window size, vsync, renderer and texture filtering
- [x] FOV, draw distance, fog
- [x] UI scale, bar scale, bar look, bar colors, bar positions
- [x] Language: titles, values and descriptions all re-read
- [x] Control layout changes, keyboard and controller
- [x] FPS counter, texture filter, and other F-keys - Flipping each such setting should survive restarts

**Holds**

- [x] A level with enforced_config: the row is marked, cannot be changed, and the value never reaches the settings file
- [ ] Leaving that level puts the player's own value back
- [x] A level with hidden_config: the row is absent from the dialog entirely
- [x] trx.config.override from a script greys the row; releasing it restores what the player chose, not the default
- [x] A demo runs with its own settings and gives them back when it ends
- [x] A script cannot hold a setting the game flow enforces

**`/mod` and game change**

- [x] /mod to another mod and back: the config is re-read, and no value from the previous game is left behind
- [x] A setting one game has and the other does not disappears from the dialog and stays in the file
~~- [ ] Options a script declared go when the scripts do, and their rows with them~~ not this branch yet

**Dynamic enums**

- [x] visuals.lara_outfit: "Default" (the null value) is offered and sticks
- [x] An outfit named in the settings file that the current mod does not supply. Confirm what the dialog shows and what the file holds afterwards
- [x] ui.bar_look and the bar colour rows: the values a mod supplies appear, and the order the options are registered in against the order the values are seeded does not matter

**Console**

- [x] `/set` by full path and by the last segment alone
- [x] Tab completion over option names
- [x] A value out of range is refused with a message rather than clamped silently
- [x] `/set` on an enforced option is refused


lahm86:
- [x] _Leaving that level puts the player's own value back_
  I don't think we support `enforced_config` per level in the gameflow, but indeed leaving the game and removing the enforcement leaves the player's original choice untouched. LMK if I've picked this up wrongly.
- [x] _A value out of range is refused with a message rather than clamped silently_
  Tried with `/set visuals.fov 1000`, `/set visuals.fog-end -256` and both get clamped (consistent with develop). Setting the wrong type does indeed give a bad invocation, like `/set visuals.fov foo`. Just wanted to clarify if we were expecting this type of clamping to fail.

rr-:
> * _A value out of range is refused with a message rather than clamped silently_
>   Tried with `/set visuals.fov 1000`, `/set visuals.fog-end -256` and both get clamped (consistent with develop). Setting the wrong type does indeed give a bad invocation, like `/set visuals.fov foo`. Just wanted to clarify if we were expecting this type of clamping to fail.

IMO yes, I think it's more ergonomic to have it clamp.

rr-:
> I don't think we support enforced_config per level in the gameflow, but indeed leaving the game and removing the enforcement leaves the player's original choice untouched. LMK if I've picked this up wrongly.

We only do that with `trx.config.override` LUA API – sorry for the confusion. But actually, this will not be reset on level load, which I think is fair, as it should be script-controlled after all.

A real bug is that I don't think this emits a change event – overriding water color does not update it immediately. I'll look into that.

rr-:
Fixed. As an effect, the following should become immediately observable rather than requiring a level restart:
```
/lua trx.config.override("visuals.water_color", "ff0000")
/lua trx.config.restore("visuals.water_color")
```

lahm86 (approved):

aredfan (approved):

PR #6100, opened 2026-08-06
Let a game declare settings of its own, and add water color presets

### LINKED ISSUE #1619: Offer to choose between PC or PS1 water colours
Info from Arsunt on this:

> Just extracted water colors from TR2 PS1 levels. I'll add them as individual customizable parameters for each PC level.
> For Lara's Home (for example) the color on PS1 is the same as PC
> But for Venice levels, yes. It is green rather than cyan
> My script is parsing whole levels data right now. I'll provide exact colors per each level file soon
> 
> https://www.htmlcsscolor.com/hex/CCFF80
> - Venice
> - Bartoli's Hideout
> - Opera House
> ----------
> https://www.htmlcsscolor.com/hex/CCFF98
> - Temple of Xian
> ----------
> https://www.htmlcsscolor.com/hex/CCFFCC
> - Floating Islands
> - Dragon's Lair
> ---------- 
> https://www.htmlcsscolor.com/hex/B2E6E6
> - The Great Wall
> - Tibetan Foothills
> ---------- 
> https://www.htmlcsscolor.com/hex/80FFFF
> - All other PS1 levels
> ---------- 
> https://www.htmlcsscolor.com/hex/80C0FF
> - PC hardware renderer uses this color for all levels
> ---------- 
> https://www.htmlcsscolor.com/hex/AAAAFF
> - PC software renderer mode uses this color for all levels 
> ---
> Colors ordered by tone: from green to blue, and then to lavender

rr-:
#### Checklist

#### Description

A game's own scripts can now add settings with `trx.config.declare()`. A declared setting is saved and loaded with the player's own, takes its text from the game strings, and appears in the settings menu on the tab the declaration names. Before this, a game could read and write the settings the engine defines but could not add one, so anything a mod wanted the player to set had to ride on an existing setting.

```lua
-- add a setting of the game's own, and the row that shows it
trx.config.declare(spec)

-- called with the value now, and whenever the setting moves
local watcher = trx.config.on_change(key, fn)
-- drop a watcher before the script that made it is done
watcher:detach()
```

- A declaration carries no text: the engine derives `settings/<key>/title` and its neighbours from the key and looks them up in the game strings.
- A row names the setting it sits before or after rather than a position, so it stays put when the tab is reordered.
- `format_value`, `is_available`, `is_visible`, `can_change_value` and `request_change_value` are callbacks a row can carry, for what data cannot say.
- A game's declarations go when the game does.

The water color preset is the first feature to use this, and it demonstrates why it belongs in a script: every game has its own tints, and the feature is too small to justify building into the engine for every custom level to carry. To support this, the game ships with a `common.water_color` script beside the executable. It declares the setting and stores the color while a preset is selected, and each game simply calls it with its own colors.

Using that script, players get a water color preset under Graphic Options → Visuals, offering the underwater tints from TR1, TR2, and TR3, shipped per level where the PlayStation releases varied them. Picking one holds the water color below it; Custom gives the player's own back.

Resolves #1619

#### Testing

**TR1**

- [x] The preset changes the water color from the settings menu
- [x] The console rejects a preset the game does not offer (`/set visuals.water-color-mode ps1`)
- [x] "Custom" gives the player's own color back, and the color row is greyed while a preset is picked
- [x] Try to apply TR2/TR3 PS1/PC presets – shouldn't offer anything regarding water settings

**TR2 / TR3**

- [x] The PS1 preset follows the level
- [x] A preset survives a relaunch
- [x] A replay that uses a PS1 preset is loaded correctly, when main game is set to use PC
- [x] Try to apply TR2/TR3 PS1/PC presets – should work within the same game engine

### DIFF
--- a/data/trx/ship/cfg/base_strings.json5
+++ b/data/trx/ship/cfg/base_strings.json5
@@ -1821,6 +1821,19 @@
         "visuals.water_color": {
             "title": "Water color",
             "description": "Color of the water.",
+        },
+        "visuals.water_color_mode": {
+            "description": "Which release's underwater tint to use. While a preset is picked it holds the water color below; Custom leaves that color to you.",
+            "title": "Water color preset",
+            "values": {
+                "custom": "Custom",
+                "dos": "DOS",
+                "pc": "PC",
+                "pc_hardware": "PC (hardware)",
+                "pc_software": "PC (software)",
+                "ps1": "PS1",
+                "tombati": "TombATI",
+            }
         }
     },
     "objects": {
--- a/data/trx/ship/cfg/base_strings-de.json5
+++ b/data/trx/ship/cfg/base_strings-de.json5
@@ -1821,6 +1821,19 @@
         "visuals.water_color": {
             "title": "Wasser-Farbe",
             "description": "Farbe des Wassers.",
+        },
+        "visuals.water_color_mode": {
+            "description": "\\{review}Welcher Unterwasser-Farbton verwendet wird. Solange eine Vorgabe gewählt ist, hält sie die Wasserfarbe unten fest; Benutzerdefiniert überlässt diese Farbe Ihnen.",
+            "title": "\\{review}Wasserfarbvorgabe",
+            "values": {
+                "custom": "\\{review}Benutzerdefiniert",
+                "dos": "\\{review}DOS",
+                "pc": "\\{review}PC",
+                "pc_hardware": "\\{review}PC (Hardware)",
+                "pc_software": "\\{review}PC (Software)",
+                "ps1": "\\{review}PS1",
+                "tombati": "\\{review}TombATI",
+            }
         }
     },
     "objects": {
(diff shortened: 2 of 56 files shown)


aredfan (approved):

data/trx/ship/cfg/base_strings.json5
```diff
@@ -1821,6 +1821,19 @@
         "visuals.water_color": {
             "title": "Water color",
             "description": "Color of the water.",
+        },
+        "visuals.water_color_mode": {
+            "description": "Which release's underwater tint to use. While a preset is picked it holds the water color below; Custom leaves that color to you.",
```
lahm86, line 1826: Will need an `en-gb` entry for `colour`

lahm86 (approved):

data/trx/ship/cfg/base_strings.json5
rr-: Thanks – good catch. Updated :)

aredfan (approved):

PR #6119, opened 2026-08-08
Fix Lara voiding when a door shuts as she climbs up

### LINKED ISSUE #6047: [Bug] Lara can teleport if climbing onto a ledge that becomes blocked by a (timed) door
### Game type

Custom level

### Game

TR3

### TRX version

1.6

### Operating system

Windows 10

### GPU

_No response_

This happened while playing the level [Underground Mystery](https://trcustoms.org/levels/3885) by Mahetus, but I believe this also happens in the originals (my memory is a little fuzzy). If Lara climbs onto a ledge and it gets blocked mid-animation by a timed door, she will teleport to the ceiling.
1. Download the custom level from the link
2. Rename the attached save file from .txt to .dat and place it in the saves folder
3. After loading the save, immediately jump forward and grab onto the ledge, then get up
4. As Lara is getting up, the door will close, causing her to get teleported to the ceiling (and die afterwards)

Instead of being teleported into the ceiling, she should maybe get ejected from the ledge.

### Attachments

https://github.com/user-attachments/assets/991238e6-c03c-454c-86fb-e7d35e6e17e4

[save_tr3_18.txt](https://github.com/user-attachments/files/30623539/save_tr3_18.txt)

### Before submitting

rr-:
#### Checklist

#### Description

A door that shut while Lara was pulling up onto its tile teleported her into the ceiling. Her position is only re-based at the end of the pull-up animation, so by then the doorway was blocked and she landed inside the wall, where normal wall glitch mechanics took over.

The invisible block is now lifted while she is climbing onto the tile as well as while she stands on it, and both sides of the doorway open together so that she crosses it in one piece.

https://github.com/user-attachments/assets/78de0f84-d733-4f39-be4a-8cfdab4e8d9c

Resolves #6047

### DIFF
--- a/src/trx/game/objects/general/door.c
+++ b/src/trx/game/objects/general/door.c
@@ -1,5 +1,6 @@
 #include <trx/game/objects/general/door.h>
 
+#include <trx/config.h>
 #include <trx/game/const.h>
 #include <trx/game/game_flow/common.h>
 #include <trx/game/game_flow/sequencer.h>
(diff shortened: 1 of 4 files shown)


src/trx/game/objects/general/door.c
```diff
@@ -77,17 +77,30 @@ static SECTOR *M_GetRoomRelSector(
...
 
-    int16_t room_num = lara->room_num;
-    const SECTOR *const lara_sector = Room_GetSector(lara->pos, &room_num);
-    return lara_sector == sector;
+    if (M_TestBlockedSector(p, lara->pos, lara->room_num)) {
+        return true;
+    }
+
+    XYZ_32 pending_pos;
+    return Item_GetPendingOrigin(lara, &pending_pos)
```
lahm86, line 107: I wonder if we should guard it with `g_Config.gameplay.wall_glitch_mode == WALL_GLITCH_FIXED`, just in case builders or others want to retain the "glitchy" behaviour.

lahm86 (approved):
Thanks, nice solution 👍

src/trx/game/objects/general/door.c
rr-: Sounds good, I'll do that shortly.

rr-: Done :)

aredfan (approved):

PR #6124, opened 2026-08-08
Remember where the inventory ring was left

### LINKED ISSUE #3273: Remember last-used item in inventory ring
When reopening the inventory ring, allow the option for the previous item showing in the centre, instead of the compass (TR default).

Intended to help users quickly access frequently needed items, as has come up with TRX' optionality features being migrated in-game.

rr-:
#### Checklist

#### Description

A ring now opens on the entry it was left on, under a new option that ships off (Gameplay → General → Remember inventory position). It takes three values:

- Off: every opening starts at the first entry, as in the original game.
- Main ring: only the item ring is remembered.
- All rings: the item and keys rings are remembered.

The first entry comes back where Lara no longer carries what was remembered, and each level starts there.

Before this, every opening began at the first entry, so reaching a frequently used item meant the same scrolling each time.

What is kept is the object a ring was left on rather than its position, because a ring is rebuilt as Lara's belongings change and a position would come back pointing at something else.

All rings and the item duplication fix ask opposite things of the keys ring, one to open where it was left and the other to send it back to its first entry, so the memory wins and the fix greys out while it is chosen.

Resolves #3273

#### Testing

- [x] The new modes work as expected
- [x] Starting a new level forgets the preselected items
- [x] Reloading a save forgets the preselected items
- [x] Item auto selection is behaving as expected
    - [x] On: using a key/puzzle item preselects the item Lara needs regardless of the memory mode
    - [x] Off: using a key/puzzle item preselects according to the memory mode
- [x] Item duplication glitch is behaving as expected
    - [x] Inventory ring mode = off: available and works
    - [x] Inventory ring mode = main ring: available and works
    - [x] Inventory ring mode = all rings: ~~unavailable, ignored~~ available and works


lahm86 (commented):
`Inventory ring mode = all rings: unavailable, ignored`
Interestingly, the duplication glitch still works in this scenario.

rr-:
> `Inventory ring mode = all rings: unavailable, ignored` Interestingly, the duplication glitch still works in this scenario.

Do you think we should just unblock it in the UI then?

lahm86:
Sounds good 👍

rr-:
Changed the UI to no longer grey-out the item dupe glitch option.

lahm86 (approved):
Thanks

data/trx/ship/cfg/base_strings.json5
```diff
@@ -1365,6 +1370,10 @@
             "title": "Restore PS1 enemies",
             "description": "Adds the mummy that appears in the PlayStation version of City of Khamoon, room 25.\nChanging this option will require restarting the game.",
         },
+        "gameplay.ring_memory_mode": {
+            "title": "Remember inventory position",
+            "description": "Opens a ring on the entry it was left on, rather than on its first one.\n\n- Off: every opening starts at the first entry, as in the original game.\n- Main ring: only the item ring is remembered.\n- All rings: the item and keys rings are remembered, which overrides the item duplication fix for the keys ring.\n\nThe first entry comes back if Lara no longer carries what was remembered, and each level starts there.",
+        },
```
aredfan, line 1376: The base strings will need adjusting re the duplication part.

rr-: Thanks – good catch. Fixed that as well.

aredfan (approved):

PR #6125, opened 2026-08-08
Restore the backpack cutscene in Angkor Wat

rr-:
#### Checklist

#### Description

Angkor Wat's backpack cutscene now plays as it does in the original game. The backpack and the skelly's hand beside it appear, young Lara wears no backpack until she picks one up part-way through, and the scene leaves her on the floor beyond the crawlspace rather than standing inside it.

To achieve this:

- The level loader now keeps object slots the catalog cannot name, because a cutscene reaches its actors by slot number.
- The level script tells the engine where to place Lara after the crawlspace scene.
- Young Lara uses the level’s torso until frame 1350, when she gets the backpack. The skin system handles the swap correctly across outfit changes without affecting custom outfits with a new mesh-swap API.
- Two small Lua additions support that behavior and are useful on their own:
    ```lua
    trx.cutscenes.frame_num                    -- frame on screen, nil if none is playing
    trx.lara.set_mesh(mesh, object, mesh_num)  -- put another object's mesh on one of Lara's
    trx.lara.clear_mesh(mesh)                  -- give her own back
    ```

Everything specific to the scene lives in `angkor1.lua`. The engine knows nothing about backpacks :)

### DIFF
--- a/data/trx/ship/games/tr4/scripts/angkor1.lua
+++ b/data/trx/ship/games/tr4/scripts/angkor1.lua
@@ -5,3 +5,14 @@ trx.events.on_game_start(function(is_save)
   trx.objects.animating_16.properties.collidable = false
   trx.lara.holsters_visible = trx.lara.has_pistol_weapon
 end)
+
+local BACKPACK_CUTSCENE = 5
+
+trx.events.on_cutscene_start(function(cutscene_num)
+  if cutscene_num == BACKPACK_CUTSCENE then
+    -- The scene starts while Lara is crawling, and the crawlspace she triggers
+    -- it from is no place to stand up in. The original engine carries her out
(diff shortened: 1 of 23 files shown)


lahm86:
- [x] Launch into level 1 with outfit set to default, then hit `T` - classic TR4 Lara gains young Lara's torso
- [x] `Walk` + `T` from there, young Lara regains her backpack

rr-:
Fixed the above issues + the lighting issue with skelly's arm.

lahm86 (approved):

rr-:
Also fixed Lara's arm getting bad frames, by using angular interpolation with quaternions instead of linear.

aredfan (approved):

PR #6133, opened 2026-08-08
objects/pickup: handle O_SCION_ITEM_1 in pickup module

lahm86:
#### Checklist

#### Description

This moves the handling for `O_SCION_ITEM_1` fully to the pickup module, and it now becomes a mode which is the same as `PICKUP_MODE_PLINTH_HIGH`, except it invokes Lara's special animation.

It also means they now show pickup aids (and can use the other pickup properties). LMK if we'd prefer to switch that off in the two OG levels.

Resolves TRX976.


rr- (commented):
Shouldn't we change `O_SCION_ITEM_1` to be declared as `X_PICKUP_MISC`? And we could piggyback fixing `give.lua` to treat Scion 1 like Lead Bar.

lahm86:
> Shouldn't we change `O_SCION_ITEM_1` to be declared as `X_PICKUP_MISC`? And we could piggyback fixing `give.lua` to treat Scion 1 like Lead Bar.

Yeah could do, I just kept is as a special so it gets added to `m_QuestObjects` - so you'd prefer to drop `X_PICKUP_SPECIAL` and just move scion to `X_PICKUP_MISC`, so it becomes part of `m_ToolObjects`?

rr-:
No, I missed that :sweat: revealing a gap in the family naming – X_PICKUP_MISC could become X_PICKUP_TOOL and MISC could be removed altogether.

Ideally I'd like to collapse all scion pickups to quest objects themselves, and trigger the special plint animation via lua. If we don't fancy doing that here, let's leave it as X_PICKUP_SPECIAL, or just hardcode it next to the sentinel in m_QuestObjects family.

lahm86:
Gotcha. I've added the explicit `X_PICKUP_SPECIAL` define in `pickup.c` to remind us it's in use. Agreed it'd be great to get the anim triggering moved to Lua, I'll log that in the tracker 👍

rr- (approved):

aredfan (approved):

PR #6156, opened 2026-08-09
Keep the music intact when the audio output is changed

### LINKED ISSUE #2489: Music corrupts if Win11 audio output is changed during playback
https://github.com/user-attachments/assets/2623da89-a145-4b22-bc53-50c29b15915f

On Win11 the shortcut for audio output device "on the fly" changing is <kbd>CTRL</kbd>+<kbd>WIN</kbd>+<kbd>V</kbd>. It's an overlay that's sadly not captured in the video but here's a screenshot from reddit:

Image

The issue is also present in TR1X.

rr-:
#### Checklist

#### Description

The mixer now grows its buffer to match SDL’s requested length, preventing a heap overflow and corrupted audio when switching to a device with a longer WASAPI period. FMVs were unaffected because they use their own audio device and buffer.

Resolves #2489

#### Testing

Before merging, we should ideally reproduce the original issue on develop, and confirm the fix.
Switching to a device whose period differs from the one the game started on is what matters; a Bluetooth or virtual output is likelier to differ than another built-in one.

##### Audio output switching (Windows)
- [x] Music keeps playing cleanly when the output device is changed mid-track with <kbd>Ctrl</kbd>+<kbd>Win</kbd>+<kbd>V</kbd>
- [x] Switching back to the original device leaves the music intact
- [x] Switching several times in a row leaves the music intact
- [x] Sound effects still play correctly after a switch
- [x] An FMV still plays with sound after a switch

##### No regression elsewhere
- [x] Music, sound effects and FMVs play normally with no device switching at all

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -109,6 +109,7 @@
 - Added an option to keep the music playing when Lara dies (Sound Options → Misc → Play music after death) (#4221)
 - Fixed the sound stuttering when a piece of music starts playing (#3094)
 - Fixed the game pausing the first time a sound effect is played, which was most noticeable with longer ones (#2286)
+- Fixed the music turning to noise when the system audio output is changed while the game is running (#2489)
 - Fixed music briefly restarting from its beginning when loading a save (#1265)
 
 **Options and menus**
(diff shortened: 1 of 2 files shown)


lahm86 (approved):
I am only able to test by switching headsets at runtime, and this is behaving well.

aredfan (approved):

PR #6166, opened 2026-08-10
lara/common: improve pickup embed fix

lahm86:
#### Checklist

#### Description

This improves the position alignment test for Lara to check the overall destination space rather than comparing with her current Y position, as otherwise sloped floors are not taken into account.

`/tp i97` in the [mentioned level](https://trcustoms.org/levels/3912).

Resolves TRX1013.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Added the ability for Lara to shimmy more quickly (Gameplay → Controls → Fast shimmying) (#5638)
 - Added the ability for Lara to pull up from ledges more quickly (Gameplay → Controls → Fast pull up) (#4857)
 - Improved state change handling when shimmying is requested while in the slow swing-in state on thin ledges (#5161)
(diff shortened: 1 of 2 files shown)


rr- (approved):

aredfan:
- [ ] I think the issue is still persisting. In the alcove, have Lara facing outwards, then collect the pickup.

lahm86:
Hmm, it seems OK on my end - do you have the fix option toggled off?

lahm86:
Or if you have crawling enabled, that will bypass it too (as she can recover from the clamp)

aredfan:
> Hmm, it seems OK on my end - do you have the fix option toggled off?

I do have the `Fix pickup embed glitch` enabled, but the outcome remains the same.

> Or if you have crawling enabled, that will bypass it too (as she can recover from the clamp)

Yes the crawling does free Lara when she is embedded on the spot.

lahm86:
Thanks. I've changed it now to look at whether or not Lara is crouched, rather than just whether or not crawling is enabled. Better to not clamp at all rather than relying on the recovery.

aredfan (approved):

PR #6189, opened 2026-08-12
objects/smashable: hold Lara at unbroken walls

rr-:
#### Checklist

#### Description

A breakable wall now stops Lara where it stands, rather than letting her vault, crawl or roll through it (Gameplay → Fixes → Fix breakable wall clipping). Before this she could pass through one and arrive on the far side with the wall still intact.

The floor data speaks of whole sectors, so an item standing between two of them leaves the ground it covers reading as open, and the tests that ask what lies ahead answer as if it were not there. An object can now say that it bars the way, through a block_func beside the height functions it already has, and `Room_IsPathBlocked` puts the question to the items nearby. The collision info, the two probes the movement states use, the vault's sampler and the pull-up's test of the ledge above all ask it, each answering as it would for a wall.

Resolves TRX1024


aredfan:
- [x] When the fix is enabled, if Lara runs towards a window, she'll stop a small distance from the window and slide instead when closing the gap. [record.txt](https://github.com/user-attachments/files/30976036/record.txt)

rr-:
Should be fixed, along with Lahm's finding about the inability to smash Bartoli windows through sprinting.

aredfan (approved):

lahm86 (approved):

PR #6194, opened 2026-08-12
Move flip rooms a group at a time

### LINKED ISSUE #5336: [Feature] Multiple flipmaps per level

It would be nice to be not restricted to 1 flipmap per level for the classic games (TR123) and allow up to 15 (What I assume the maximum is based on TombEditor)

This would allow for much more flexibility and creativity for custom level builders.

### Extra context

Image

rr-:
#### Checklist

#### Description

TR4 flip maps now support independent groups. Each group can be toggled and saved separately, while levels that don't specify groups behave exactly as before. This fixes a couple of Angkor Wat triggers.

Flip groups are also accessible through Lua and the console:

```lua
trx.rooms.flip(3)           -- move that group alone
trx.rooms.flip()            -- move every group
trx.rooms.is_flipped(3)     -- whether that group is showing its pairs
trx.rooms.is_flipped()      -- the same for the group that moved last
trx.rooms.flip_group_count  -- how many groups a level can hold
```

`/flip 3` moves that group, `/flip` on its own moves them all, and each says which it changed rather than reporting a silent success.

To eventually extend support to TR 1-3, injection version 10 adds a byte to the room-properties edit for setting a room's flip group. Since that edit writes all room properties at once, patches not changing the group use 255 to preserve the level's original value. Ref https://github.com/LostArtefacts/TRXInjectionTool/pull/341

Resolves #5336 / TRX173


lahm86:
Looks good 👍 I think there might be a room mesh update issue though:
- `/play seth`
- `/tp empty`, `/keys`
- Use the slot (performs flip map 2)
- Head through, skip the flyby and go down to the sphinx
- The geometry is unflipped but the mesh is flipped. This area is flip group 1

rr-:
Should be fixed :relieved:

lahm86 (approved):

aredfan (approved):

PR #6197, opened 2026-08-12
gun/rifle: add an M16 hip fire option

### LINKED ISSUE #3861: Enhancement Request: Improvements to M16
As it stands, the M16 in TR2 is one of the less reliable weapons due to its quirk with animations: Lara tends to switch to her shoulder animation variant for the M16 very often and during combat, making its reliable DPS extremely low. To remedy this issue for the HK Gun in FLEP, there is an option present called "StillHipsFire".

The option in FLEP is essentially about making it so Lara does not change the HK to her shoulder when she stops moving while shooting with the M16. In essence, it only allows the shoulder animations to play if Lara is stationary, in which case starting to move would switch it to the hip animations, but not vice versa [from hip to shoulder as is the current behaviour].

The option could be named something like "M16 aiming variants" and have an option for "original" and "enhanced", with the enhanced version being the one that retains the hip animations when moving. It would overall make the M16 much more viable in combat due to TR2's abundance of gun enemies. [Yes, one could argue the rough aiming conditions are a drawback in favour of the high damage value, but I'd argue it makes the M16 hard to use in combat that TR2 offers, which favours constant movement with gun enemies in most slots.]

rr-:
#### Checklist

#### Description

Lara can now keep firing the M16 and MP5 from her hip while the action key is held, and shoulders the gun only when she opens fire standing still. Before this, stopping mid-burst always switched her to the shoulder animation, and the transition cost her the shots it took.

Hip fire keeps its own ballistics, so the option trades damage per shot for uninterrupted fire.

Resolves #3861


rr-:
(This patch is consistent with FLEP, including lower accuracy when shooting from the hip.)

aredfan (approved):

lahm86 (approved):

PR #5850, opened 2026-07-15
lara: define TR4 extra meshes

lahm86:
#### Checklist

- [x] I have read the [coding conventions](https://github.com/LostArtefacts/TRX/blob/develop/docs/CONTRIBUTING.md#coding-conventions)
- [x] I have added a changelog entry about what my pull request accomplishes, or it is an internal change
- [x] I have added a readme entry about my new feature or OG bug fix, or it is a different change

#### Description

This adds the extra meshes required for TR4, such as the torch and waterskin, into the outfit system. It also adds Lara's speech heads there, later required for cutscenes. That API is pretty much just a placeholder for now, but it removes the data roadblock in the meantime.

Below is a script for testing that can be pasted into e.g. `karnak1.lua` - Lara will cycle through her new kit every couple of seconds.

<details>

```lua
local swap_tests = {
  [trx.lara.Mesh.HAND_L] = {
    trx.lara.ExtraMesh.WOODEN_TORCH,
    trx.lara.ExtraMesh.SHOVEL,
    trx.lara.ExtraMesh.JERRYCAN,
    trx.lara.ExtraMesh.SANDBAG,
    trx.lara.ExtraMesh.WATERSKIN,
  },
  [trx.lara.Mesh.HAND_R] = {
    trx.lara.ExtraMesh.CROWBAR,
    trx.lara.ExtraMesh.BINOCULARS,
    trx.lara.ExtraMesh.HOOK_AND_POLE,
    trx.lara.ExtraMesh.DETONATOR,
  }
}

local hands = {
  trx.lara.Mesh.HAND_L,
  trx.lara.Mesh.HAND_R,
}

local outer = 1
local inner = 1
local timer = -1

trx.events.before_control(function()
  timer = timer + 1
  if timer % 60 ~= 0 then
    return
  end

  trx.lara.clear_equipment(hands[1])
  trx.lara.clear_equipment(hands[2])

  local hand = hands[outer]
  local mesh = swap_tests[hand][inner]
  trx.lara.set_extra_equipment(hand, mesh)

  inner = inner + 1
  if inner > #swap_tests[hand] then
    inner = 1
    outer = outer + 1

    if outer > #hands then
      outer = 1
    end
  end
end)
```
</details>

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/315

Resolves TRX598.

PR #6205, opened 2026-08-13
cutseq: expose a scene's actors to scripts

rr-:
#### Checklist

#### Description

A script can now hide a scene's actors and swap the mesh a node draws, which is how TR4 gives a speaking actor the head that says the line. The slots those heads live in are named in the catalog, so a script picks one by name rather than by number.


lahm86 (approved):
We added Lara's speech heads in #5850 to the outfits injection, with the view to support young/classic swaps in the cutscenes (and indeed to ignore in TR1-3 outfits). `Lara_Skin_SetSpeechFace` - although I think the implementation will need tweaking, maybe it should just return the mesh index. Understand if integration with that is to come later.

aredfan (approved):

PR #6208, opened 2026-08-13
pathing: cross TR4's jump and monkey overlaps

rr-:
#### Checklist

#### Description

TR4 numbers a box in eleven bits and spends the rest of the word on what it takes to cross the overlap: a jump, or a swing along the monkey bars. The box search reads those, and a creature that can do neither is not offered them, so its LOT setup says which it can.

A creature crossing one is off the boxes until it lands, so the box it is over stops holding it back and its height comes from the room. Monkey swinging hangs it off the ceiling the same way. Its own control asks whether the way ahead is bars, which nothing answered before.

### DIFF
--- a/src/trx/game/pathing/types.h
+++ b/src/trx/game/pathing/types.h
@@ -24,6 +24,7 @@ typedef enum {
     LOT_SETUP_QUADRUPED,
     LOT_SETUP_JUMPER,
     LOT_SETUP_CLIMBER,
+    LOT_SETUP_ACROBAT,
     LOT_SETUP_FLYER,
 } LOT_SETUP_TYPE;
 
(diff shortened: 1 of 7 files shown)


src/trx/game/pathing/types.h
```diff
@@ -24,6 +24,7 @@ typedef enum {
     LOT_SETUP_QUADRUPED,
     LOT_SETUP_JUMPER,
     LOT_SETUP_CLIMBER,
+    LOT_SETUP_GUIDE,
```
lahm86: I think other enemies have these abilities - skeletons can leap gaps and the red ninjas with swords can monkeyswing. Maybe `LOT_SETUP_ACROBAT` would be an idea for a name rather than coupling it with guides? Just a name though, so not a big issue.

lahm86 (approved):

aredfan (approved):

src/trx/game/pathing/types.h
rr-: Loving it :D

rr-: Done :+1:

PR #6216, opened 2026-08-13
Add TR4 waterfalls

rr-:
#### Checklist

#### Description

This ports TR4's waterfalls: the three animated waterfall meshes in slots 423 to 425, and the mist emitter in slot 385, which had no name in TRX until now.

- The waterfall meshes scroll their texture at the rate the OG uses, so the water runs rather than standing still.
- OCB controls the looping sound (sand or water), and can also make a waterfall switchable so it disappears when its trigger is inactive.
- The mist emitter spawns mist and plays its loop while Lara is nearby. Before this it ran the TR1 path and spawned splash effects instead.

Scrolling reuses the UV rotate system instead of duplicating it. Each scrolling texture stores its own speed and wrap period at load time, so waterfalls can scroll independently of the level's UV rotate speed.

Resolves TRX525. LMK if we'd rather move the OCB to a dedicated property now, rather than later.


lahm86 (approved):
Dedicated property sounds good, happy for it to happen later (guess we should log it)

rr-:
> Dedicated property sounds good, happy for it to happen later (guess we should log it)

Done on this branch; I'd appreciate confirming if I got all the instances right :)

aredfan (approved):

PR #6219, opened 2026-08-13
Generate Lara's golden form instead of shipping models for it

rr-:
#### Checklist

#### Description

This PR adds a golden form to every existing outfit, making it no longer necessary to inject a duplicate model for each outfit separately.

Previously, gold required six duplicate outfits, each with its own Lara model, gun map, braid meshes and modes. They occupied six of the thirty-two outfit slots, and outfits without a golden copy, such as Natla, could not be gilded. Instead, outfits can now just set `gold_color`, which defaults to the flat gold tile used by the shipped golden models in all four games.

It also adds a new option to turn Lara to gold next to the outfit option, and convenience console commands:

```
/golden          toggle gold
/golden on
/outfit          show current outfit
/outfit natla    change outfit
/outfit -        restore the level's outfit
```

For mod authors, the golden outfits and their supporting gun map, braid mode, braid offset and extra meshes are removed. `extra_outfits` no longer accepts `LS_EXTRA_MIDAS_KILL`, and the gold braid meshes leave `trx.lara.ExtraMesh`. Outfits can now specify the objects supplying their extra, gun and leg meshes, allowing gilded outfits to use their minted copies. See the migration guide for details.

#### Implementation details

Underneath, golden meshes are now minted from the current outfit's vertices and faces on first use and kept until the level unloads. They live in their own object, so other actors sharing Lara's meshes, such as Bacon Lara, remain unchanged. These objects are only used for mesh swaps, and occupy no slots; to make this possible, a new API was created `Object_Mint()` which hands the caller a runtime-only slot. The slot created this way must not be applied to an `ITEM`, and will not survive savegames; its sole purpose is to carry out mesh swaps. In the future, it might be expanded to let Lua declare its own AI objects (!).

Regenerating `lara_outfits.bin` without the old golden outfits also frees their mesh objects for custom outfits.

Resolves TRX1070

#### Testing

- [x] Midas hand death, gilding her mesh by mesh, with and without a braid
- [x] Golden Lara setting on and off mid-level, and across a level change or a savegame
- [x] Bacon Lara stays unglided while Lara wearing a Bacon Lara outfit is gold
- [x] `/golden`, `/golden on`, `/golden off`
- [x] `/outfit`, `/outfit {name}`, `/outfit -`


aredfan:
- [x] It's possible for Lara's back weapon to not turn to gold:

> 1. launch tr1
> 2. `/play 1`, `guns`
> 3. either have Lara draw her shotgun, or do nothing as this step can be skipped
> 4. `/golden`, and observe the shotgun hasn't turned to gold

- [x] It's possible for Lara's back weapon to retain the gold when the command is reversed:

> 1. launch tr1
> 2. `/play 1`
> 3. `/golden`
> 4. `/guns`
> 5. either have Lara draw her shotgun, or do nothing as this step can be skipped
> 6. `/golden` and observe the shotgun has not been reverted

- [x] The game crashes with the following steps:

> 1. launch tr1
> 2. `/play 0`

- [x] Angkor Wat: Young Lara's pre-backpack torso doesn't turn to gold.

docs/CHANGELOG.md
```diff
@@ -187,6 +194,8 @@ Showcase: https://youtu.be/DKpqz_Yum6o
 - Fixed a new game, or a level started from Play Any Level, running in the game mode the previous game was started in
 
 **Developer console**
+- Added the `/outfit` console command, which shows or changes what Lara is wearing
+- Added the `/golden` console command, which casts Lara in gold
```
aredfan: Couple of duplicate entries added under 1.10.

aredfan (approved):

src/trx/game/lara/skin/types.h
```diff
@@ -46,10 +45,16 @@ typedef struct {
     bool is_defined;
     OBJECT_ID mesh_obj_id;
     OBJECT_ID joints_obj_id;
+    OBJECT_ID extra_obj_id;
+    OBJECT_ID guns_obj_id;
+    OBJECT_ID legs_obj_id;
     LARA_SKIN_GUN_MAP *gun_map;
     LARA_SKIN_BRAID braid;
     bool is_selectable;
     bool is_reflective;
```
lahm86: I think we can get rid of `is_reflective` now 😌

lahm86 (commented):
Thanks 👍 
I'll follow this up with removing all the redundant meshes from the injections.

src/trx/game/lara/skin/types.h
rr-: Updated :)

lahm86 (approved):

PR #6245, opened 2026-08-15
data: give crystal on game start

lahm86:
#### Checklist

#### Description

This gives Lara a save crystal when starting Jungle if that mode is enabled. Do we want to extend it to the other games? I feel like it's not needed there as it's in the same boat as her starting shotgun, flares etc, so is very game-specific.

Resolves TRX1101.

### DIFF
--- a/data/trx/ship/games/tr1-ub/scripts/_game.lua
+++ b/data/trx/ship/games/tr1-ub/scripts/_game.lua
@@ -5,3 +5,5 @@ require("common.water_color").declare({
     dos = "99B2FF",
   },
 })
+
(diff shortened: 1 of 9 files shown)


rr- (commented):
I think this should apply to all games as it's more of a property of the pickups SG mode rather than TR3.

To avoid duping the same logic between games, I'd ship this in trx/modules, and require it from _game.lua in each game.

lahm86:
Now in place for all games 👍

rr- (approved):
Dope

aredfan (approved):

PR #6267, opened 2026-08-15
items/anim: avoid land/water SFX in swamp rooms

### LINKED ISSUE #6248: [Bug] Walking backwards in mud/quicksand causes the wet footsteps sound to play
### Game type

Original game

### Game

TR3

### TRX version

1.10

### Operating system

Windows 10

### GPU

_No response_

This does not happen in OG.
1. `play jungle`
2. `tp 12 27 73`
3. Walk backwards through the mud

The footstep sound shouldn't play.

### Attachments

Unfortunately I can't record something with audio, but if you follow the steps you should be able to hear it.

### Before submitting

lahm86:
#### Checklist

#### Description

This avoids land and water specific SFX playing in swamp rooms.

I'd appreciate a sanity check in OG TR3 that the only sound she makes is from wading forward (that SFX has mode set to `ACE_ALL`).

Resolves #6248 / TRX1115.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,7 +1,8 @@
 ## [Unreleased](https://github.com/LostArtefacts/TRX/compare/trx-1.10...develop) - ××××-××-××
 
 **Lara's movement**
-- Fixed Lara entering the step down animation when walking backwards in a swamp room (OG bug) (#6248)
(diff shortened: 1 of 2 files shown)


rr- (approved):

aredfan (approved):
> I'd appreciate a sanity check in OG TR3 that the only sound she makes is from wading forward (that SFX has mode set to `ACE_ALL`).

Can confirm this is indeed the case in OG TR3. 👍

PR #6277, opened 2026-08-16
output: keep effect sizes off the raster grid

rr-:
#### Checklist

#### Description

Water particles, lightning bolts and the sparks that hold a fixed size took their size from the viewport the scene is rasterized on, which the supersampling factor multiplies. A billboard displacement is a world space offset, so the specks and the bolts grew with the factor while the sparks shrank. They now convert from the screen pixels of the original's 640-wide, 80 degree reference viewport, which also stops them from following the resolution and the upscaling factor. `fx/fire.c` already kept that reference privately and now shares it.

This means that the lightnings will become thinner, and water particles will become larger on bigger desktop resolutions. Please LMK if this is acceptable - if not, we'll have to tweak by feel I'm afraid.

The wireframe and the debug portal lines are drawn in the pixels the driver rasterizes with, and thinned rather than grew, so they take the factor instead.

Resolves TRX1106.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -40,6 +40,11 @@
 - Changed the save crystal behavior to give Lara a crystal when starting a game, if the mode is set to Saving (pickups), in line with the TR3 PS1 version (Gameplay → General → Crystal mode)
 - Fixed the PS1 water color preset showing as unnamed text in The Lost Artefact and Golden Mask (#6265)
 
+**Rendering**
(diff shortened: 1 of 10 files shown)


lahm86 (approved):
LGTM

aredfan:
The wireframe, debug, and water particles all appear as expected. 👍 

> This means that the lightnings will become thinner, and water particles will become larger on bigger desktop resolutions. Please LMK if this is acceptable - if not, we'll have to tweak by feel I'm afraid.

The lightning is indeed quite thin. If there's a way to preserve the lightning's faithfulness then I'm all for it.

aredfan (approved):

PR #6283, opened 2026-08-16
Add XYZ_32 helpers

rr-:
#### Checklist

#### Description

Resolves TRX1113 / TRX204 / TRX1142.

Adds shared math functions to replace roughly 200 places where the game was manually combining `Math_Sin` and `Math_Cos`, then updates the game to use those new functions.

- `XYZ_32_RotateYaw` / `XYZ_32_UnrotateYaw`:
  Turns an offset into the world's axes, and back along an item's.
- `XYZ_32_Rotate` / `XYZ_32_Unrotate`:
  The same by a whole orientation, for an item that pitches or rolls.
- `XYZ_32_OffsetLocalYaw` / `XYZ_32_OffsetLocal`:
  The point the offset reaches, for callers that have the origin.

Repeated/stacked positioning now keeps its rounding exactly the same. A position calculated only once might still end up off by 1 unit because of rounding.

Additionally:

- The free camera in photo mode no longer slowly moves when you tilt it up or down.
- Some old rotation math (`TRIGMULT2` and `TRIGMULT3`) was removed. Camera/view rotation is now handled by `M_RotYXZ`, with its matrix transposed for the view calculation.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Changed Lara's outfit setting to offer only the outfits the current level can dress her in
 - Fixed the icons beside the volume settings, which now show a note for the music and a speaker for the sound effects
 - Fixed the game flashing over the black bars beside the picture when a frame is advanced in photo mode (regression from 1.10)
+- Fixed the photo mode camera drifting upwards and overshooting when it is moved while pitched up or down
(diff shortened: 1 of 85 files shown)


lahm86 (approved):
Brilliant 😌

aredfan (approved):

PR #6290, opened 2026-08-17
Report startup failures rather than ending the run

rr-:
#### Checklist

#### Description

Startup failures now return a `RESULT` to the shell instead of exiting at the point of failure. This covers:

- SDL,
- window and graphics setup,
- shaders,
- catalogs,
- arguments,
- mods,
- subsystem loading,
- path resolution,
- settings.

The shell reports these errors in one place while preserving the startup stage and stack.

- Settings parse errors now include the file, line, and column, with game flow and player settings errors handled consistently.
- Recoverable failures should no longer bring the game down.
- Malformed settings fall back to defaults.
- Save compression errors affect only the failed save.
- Replays for games that are not installed are refused.
- Shiva and the monkey are skipped when their replacement meshes are missing.

Additionally, subsystem modules now have names, allowing the shell to identify which module failed during startup. By default, the name comes from the module's source file, but registration can override it. `GamePath_Resolve` also reports unresolved paths instead of terminating the game, while callers where a missing path is expected use `GamePath_PeekResolve`.

```c
RESULT Subsystem_LoadAll(void);
RESULT GamePath_Resolve(
    GAME_DYNAMIC_PATH path, const char *rel, const char **out_path);
RESULT Config_Read(const char *default_path, const char *enforced_path);
RESULT ConfigFile_Read(const char *default_path, const char *enforced_path);
RESULT TRX_GL_Context_Attach(void *window_handle);
RESULT TRX_GL_Program_Init(TRX_GL_PROGRAM *program);
RESULT TRX_GL_Program_AttachShader(
    TRX_GL_PROGRAM *program, GLenum type, const char *path);
RESULT TRX_GL_Program_Link(TRX_GL_PROGRAM *program);
```

Players who hit a broken driver, shader, mod or settings file are told what is wrong and where, instead of the game closing on them. Mod authors get the same for a file their mod ships that the game cannot read.

The missing `--mod` case still exits early, since no reporting session exists that soon.

### DIFF
--- a/src/tests/fakes/settings.c
+++ b/src/tests/fakes/settings.c
@@ -135,10 +135,10 @@ void FS_CloseDirectory(FS_DIR *const dir)
     closedir((DIR *)dir);
 }
 
-bool ConfigFile_Read(
+RESULT ConfigFile_Read(
     const char *const default_path, const char *const enforced_path)
 {
-    return false;
+    return OK;
 }
 
 bool ConfigFile_WasFound(void)
--- a/src/tests/fakes/settings.c
+++ b/src/tests/fakes/settings.c
@@ -64,15 +64,17 @@ char *GamePath_ExpandVars(const char *const path)
 
 // Cached and valid for the run, as the engine's is: a caller may hold what it
 // was given.
-const char *GamePath_Resolve(
-    const GAME_DYNAMIC_PATH path, const char *const rel)
+RESULT GamePath_Resolve(
+    const GAME_DYNAMIC_PATH path, const char *const rel,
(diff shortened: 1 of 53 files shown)


rr-:
Sanity checks to perform:
- [x] Remove an essential file and observe what the game does
- [x] Remove the config file, should load defaults
- [x] Damage the config file, should exit with an error
- [x] Check restoring a setting to its default (`/set xxx -`) works

lahm86 (approved):

aredfan (approved):

PR #6291, opened 2026-08-17
Fix the photo mode camera near vertical

rr-:
#### Checklist

#### Description

This PR makes the photo mode camera rotate smoothly through vertical and continue into an upside-down view, while also making roll frame-rate independent like pitch and yaw. Previously, the camera would flip back and forth at vertical, and rotation near vertical became coarse because the view reconstructed its orientation from the camera target rather than photo mode’s own angles.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -8,19 +8,20 @@
 - Fixed Lara's rope-grab reach being shorter from certain directions (OG bug)
 
 **UI**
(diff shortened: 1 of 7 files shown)


lahm86 (approved):

aredfan:
Issue:

- [x] When photo mode is open and without moving the camera, press the look button and notice the screen flicker for a frame or two.

aredfan (approved):

PR #6317, opened 2026-08-18
Add an instant save/load screen option

rr-:
#### Checklist

#### Description

Adds an option to open the save and load screens instantly (Graphic Options → UI → Instant save/load screen, off by default). When enabled, pressing the save or load key shows the save slots on the first frame, and leaving the screen returns to the game just as quickly.

This skips the usual transition effects:

- the inventory ring spinning in and out,
- the passport opening and turning to the appropriate page,
- the fade behind the ring,
- the menu sounds associated with these animations.

Because the screens now appear immediately, menu input is ignored until all keys are released. This prevents keys held during gameplay from immediately moving through the save slots. Buffering clears the input history as the ring closes, which could otherwise cause a held key to register again. Inventory buffering continues to work as expected with the normal inventory ring.

Also fixes a dialog appearing too low for a single frame when opened with a screen heading. The heading space was applied one frame late, which the inventory ring animation previously hid.

Resolves TRX1170


lahm86 (approved):

aredfan:
LGTM. 👍 

If feasible, wdyt of having the inventory ring objects not drawn, with the UI appearing by itself, overlaying a dark, monochrome, or carpet screen. It'll probably also hide the last frame of the inventory ring closing when loading a save.

aredfan (approved):

rr-:
I'll demo this and ask for a re-review soon, as a separate commit, to see how we feel about it.

rr-:
Done. I think it indeed is better, but please LMK :) I kept the original dialog offset and didn't exactly center it on screen.

aredfan (approved):
> Done. I think it indeed is better, but please LMK :) I kept the original dialog offset and didn't exactly center it on screen.

I think the new instant save/load screen looks pretty good, especially with the carpet. The dialog offset is adequate imo.

Many thanks for working on this. 👍

lahm86 (approved):

PR #6318, opened 2026-08-19
Add the PlayStation fog and lighting models

### LINKED ISSUE #5348: [Feature] Set PSX lighting as high contrast on Tomb Raider III

When selecting "high contrast" mode in Tomb Raider I and II, the lighting becomes the same as in the PS1 version. I suggest the same should be done with Tomb Raider III to maintain a consistent standard. I also suggest adding the PS1 version's lighting formula for Lara, because in brightly lit environments, Lara appears shine too brightly; this options could be optionals, adding an option "Lara Lighting: responsive or PS1.

Add an extra option allowing to set Lara lighting mode PS1 and Responsive.

_No response_

### Extra context

_No response_

rr-:
#### Checklist

#### Description

Adds two rendering options that more closely match the original PlayStation hardware, and gives each preset its own dither mode.

PlayStation fog blends toward the far color along 1/z and is calculated once per vertex, while PC fog ramps with distance per fragment. The PlayStation `saturate` lighting model clips each color channel independently, allowing brightly lit surfaces to deepen in color instead of washing out toward white.

For players:

- Graphic Options → Rendering → PlayStation fog, enabled by the PlayStation presets.
- Graphic Options → Rendering → Lighting model, with Flat, Overbright, and Saturate modes. The TR3 PlayStation preset uses Saturate, while the TR3 PC preset uses Flat.
- Lighting contrast is now available only in TR1 and TR2, since TR3 and TR4 scale dynamic light sources independently.
- F9 cycles the lighting model in games that support it, and lighting contrast otherwise. Settings files using the binding's old name remain compatible.
- Each preset now selects its dither mode: PS1 for PlayStation presets and off for PC presets.
- TR3 now defaults to the new curve and has a one-off migration that discards brightness and gamma settings.
- TR4 colors are now more saturated in very bright areas, but mostly remain unchanged.

Screenshot comparison:

### The base workaround people have been using
20260819_111329_Nevada_Desert

### TR3 PC implementation
20260819_111221_Nevada_Desert

### TR4 PC implementation
20260819_111211_Nevada_Desert

### PS1 faithful implementation, now used by default in TR3 and TR4
20260819_111216_Nevada_Desert

Resolves #5348 / TRX182


lahm86 (approved):

aredfan:
Proposal for `Lighting model` description cleanup, which breaks the description down to bullet points, and clarifies which game version that each setting originates from.  Please correct me if I got it wrong.

> "Chooses how far vertex lighting scales the texture.\n\n- Overbright is what the PC renderers did: a fully lit vertex takes the texel to twice its brightness, with the excess added on afterwards (TR4 PC).\n- Saturate is the PlayStation's, doubling the texel as well but letting each color clip where it lands, which keeps bright surfaces in their own color instead of washing them out (TR3/4 PS1).\n- Flat stops at the texture's own brightness, which is dimmer (TR3 PC)."

**Edit:** Sorry, forgot to mention that the above option will require a en-gb translation as well for the two `colour` uses.

rr-:
> Proposal for `Lighting model` description cleanup, which breaks the description down to bullet points, and clarifies which game version that each setting originates from. Please correct me if I got it wrong.
> 
> > "Chooses how far vertex lighting scales the texture.\n\n- Overbright is what the PC renderers did: a fully lit vertex takes the texel to twice its brightness, with the excess added on afterwards (TR4 PC).\n- Saturate is the PlayStation's, doubling the texel as well but letting each color clip where it lands, which keeps bright surfaces in their own color instead of washing them out (TR3/4 PS1).\n- Flat stops at the texture's own brightness, which is dimmer (TR3 PC)."
> 
> **Edit:** Sorry, forgot to mention that the above option will require a en-gb translation as well for the two `colour` uses.

Done :)

aredfan (approved):
> Done :)

Thank you. 😌

PR #6336, opened 2026-08-20
Stop the moving spikes at a pushed block

### LINKED ISSUE #6308: [Feature] Make pushable blocks stop moving spikes as soon as Lara begins moving them

This can happen in Temple of Puna, but I'm not sure if there's any other place in the classic games where this can manifest. If Lara uses a pushable block to stop a moving spike trap, they will completely ignore the block while Lara is pushing it, only stopping once she has finished the pushing animation. It would make more sense if the moving spikes trap took into account whether the block pushing animation has already begun and stop moving. Conversely, it would make sense if Lara was unable to push the block once the spikes have gone down far enough - right now she is able to push it _through_ the spikes even though they've begin hurting her, which will result in her death as she is in the pushing animation.

1. `play puna`
2. `tp 45 -17 50`
3. Slide down the slope
4. Wait for the spikes to get closer to Lara
5. Begin pushing the block
6. Notice the spikes will stop further down than they should
7. Notice also that you can begin pushing the block after the spikes have gone down far enough to begin hurting Lara

This is something that also happens in OG (vanilla behavior).

I'm not sure if this is feasible, but it would be more logical if the spikes considered whether Lara is already pushing the block.

### Extra context

https://github.com/user-attachments/assets/7a5eae0f-6ec7-447f-8772-cf02ac6da960

### Before submitting

rr-:
#### Checklist

#### Description

Fixes moving spikes ignoring a pushable block while Lara is pushing it. The spikes now stop while the block occupies the sector ahead and resume if the block moves elsewhere.

Resolves #6308 / TRX1169

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -78,6 +78,7 @@
 - Fixed Lara being able to turn too quickly when wading (regression from 1.0) (TRX1129)
 - Fixed Lara's sliding SFX continuing after she leaves a slope (#6294 / TRX1155, regression from 1.0)
 - Fixed underwater blood clouds appearing in the air when the security lasers hurt Lara at the water surface (OG bug) (#6323 / TRX1180)
+- Fixed the ceiling spikes not stopping on a pushable block until Lara finishes pushing it (OG bug) (#6308 / TRX1169)
 
 **TR4**
 - Added the ability to skip in-game cutscenes (TRX1051)
--- a/src/trx/game/objects/traps/movable_block.c
+++ b/src/trx/game/objects/traps/movable_block.c
@@ -1109,6 +1109,49 @@ static void M_Setup(OBJECT *const obj)
     obj->save_position = true;
 }
 
+static bool M_IsSameSquare(const XYZ_32 pos_1, const XYZ_32 pos_2)
+{
+    return (pos_1.x >> WALL_SHIFT) == (pos_2.x >> WALL_SHIFT)
+        && (pos_1.z >> WALL_SHIFT) == (pos_2.z >> WALL_SHIFT);
+}
+
+static bool M_HasStartedMoving(const ITEM *const item)
+{
+    const ANIM *const anim = Item_GetAnim(item);
+    const ANIM_FRAME *const current = Item_GetBestFrame(item);
+    if (anim == nullptr || anim->frame_ptr == nullptr || current == nullptr) {
+        return false;
+    }
+
+    return current->offset.x != anim->frame_ptr->offset.x
+        || current->offset.z != anim->frame_ptr->offset.z;
+}
+
+bool MovableBlock_TestSquareClaimed(const XYZ_32 pos)
+{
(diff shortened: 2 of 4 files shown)


lahm86:
I don't think this is working as expected as `Room_GetWorldSector` won't be able to work with the sectors that the pushblock alters in Puna until it's at that height, because of the way the rooms are tiered. But if we were to use the sector obtained from `Room_GetSector` then it would mean the spikes would stop much too far up, and the trap would lose its timed aspect.

<details>

image
</details>

I'm not sure of an alternative though, other than object-object collision checks.

rr-:
Thanks – redone, should be much better now.

lahm86:
Thanks. The spikes stop as soon as Lara starts to pull the block, so even if it's still near the top of the room, then it resumes when she stops pulling it.

WDYT of this patch (can probably be tidied)? It means it will continue as normal until the bottom of the spikes (appear to) touch the block.

```diff
diff --git a/src/trx/game/objects/traps/movable_block.c b/src/trx/game/objects/traps/movable_block.c
index 2aa570fae..c15ffe47d 100644
--- a/src/trx/game/objects/traps/movable_block.c
+++ b/src/trx/game/objects/traps/movable_block.c
@@ -1137,10 +1137,14 @@ bool MovableBlock_TestSquareClaimed(const XYZ_32 pos)
             continue;
         }
 
+        if (M_IsSameSquare(item->pos, pos) && item->pos.y - WALL_L <= pos.y) {
+            return true;
+        }
+
         const M_PRIV *const p = item->priv;
         const XYZ_32 target =
             XYZ_32_OffsetYaw(item->pos, p->interaction_rot, WALL_L);
-        if (M_IsSameSquare(item->pos, pos) || M_IsSameSquare(target, pos)) {
+        if (M_IsSameSquare(target, pos) && target.y - WALL_L <= pos.y) {
             return true;
         }
     }
diff --git a/src/trx/game/objects/traps/spike_ceiling.c b/src/trx/game/objects/traps/spike_ceiling.c
index 24c410ac9..d8e903cff 100644
--- a/src/trx/game/objects/traps/spike_ceiling.c
+++ b/src/trx/game/objects/traps/spike_ceiling.c
@@ -39,7 +39,8 @@ static void M_Move(const int16_t item_num)
     const XYZ_32 pos = { item->pos.x, item->pos.y + p->speed, item->pos.z };
     const SECTOR *const sector = Room_GetSector(pos, &room_num);
 
-    if (MovableBlock_TestSquareClaimed(item->pos)) {
+    if (MovableBlock_TestSquareClaimed(
+            (XYZ_32) { item->pos.x, item->pos.y + WALL_L, item->pos.z })) {
         Sound_StopEffect(SFX_SPIKE_WALL);
         return;
     }
```

rr-:
Thanks – folded that in. :)

lahm86 (approved):

aredfan (approved):

PR #6346, opened 2026-08-20
level/format_tr4: add statistics probe

lahm86:
#### Checklist

#### Description

This adds probing to the level 4 format reader to allow stats to be calculated.

Unsupported items are assigned as `O_CAMERA_TARGET` to allow the stats scanner to run - do we want to re-introduce `O_DUMMY` for this instead?

Resolves TRX1201.

### DIFF
--- a/src/trx/game/level/format/format_tr4.c
+++ b/src/trx/game/level/format/format_tr4.c
@@ -1,6 +1,7 @@
 #include <trx/core/benchmark.h>
 #include <trx/core/file.h>
 #include <trx/core/memory.h>
+#include <trx/core/strings.h>
 #include <trx/core/utils.h>
 #include <trx/debug.h>
 #include <trx/game/const.h>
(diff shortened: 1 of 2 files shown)


src/trx/game/level/format/format_tr4.c
```diff
@@ -264,6 +255,7 @@ static RESULT M_ReadItemsTR4(LEVEL_CONTEXT *const ctx, TRX_FILE *const file)
         item->init_flags = tr4_item->flags;
         if (item->object_id == NO_OBJECT) {
             LOG_WARNING("Unsupported TR4 item object %d on item %d", obj_id, i);
+            item->object_id = O_CAMERA_TARGET;
```
rr-, line 258: We can use the minted object IDs for this, maybe?

rr- (approved):

src/trx/game/level/format/format_tr4.c
lahm86: That seems to cause issues with reloading saves unfortunately. Maybe camera target is OK to continue with, although it might also cause save issues once we implement the missing objects (but tbh those saves would be redundant at that point anyway).

rr-: Right. I'd like to be able to mint save-surviving object IDs at some point, but now's not the time for that :) Let's keep camera target.

aredfan (approved):

PR #6348, opened 2026-08-20
Skip TR4 cinematics with look

rr-:
#### Checklist

#### Description

The look input skips cutscenes and cutscene levels, as the original TR4 does. With cinematic skips enabled, it also ends a flyby sequence the level marks unbreakable, which only the inventory input did before.

The setting description changed with it, and its translations are machine-made and marked for review.

Resolves TRX1093


lahm86:
I don't think it's working for flybys currently. I think we should perhaps drop the `Camera_FlybyMode_RequestLook` approach and just combine the escape and look options to trigger `FlybyMode_Cancel`. That's the safer way to stop them as it will ensure heavy triggers still get activated. It will also prevent cancelling track cameras which can result in the flickering we saw in #5741. These are more like locked cameras IMO similar to Angkor room 58.

Maybe we could also extend this to Von Croy's tutorials. E.g. the input is accepted in the first cutscene in level 1, but is not in his tutorial a few steps forward.

rr-:
Thanks, applied, though I'd appreciate an ingame test with the "Skip cinematics" option using escape and look inputs separately, if it behaves as expected for cutscenes and for flybys, as it's a bit confusing.

lahm86 (approved):
LGTM ☺️ Thank you

aredfan (approved):

Issue #5351, opened 2026-04-07
[Feature] Allow for customisation of HUD through LUA


Gizzy-91:
### What would you like to see?

A large one so please bare with me!

I was thinking something like a draw routine for individual existing elements such as ammo, pickup sprite/model, health/air/sprint/poison/exposure meters, extending to custom sprites/models that can be referenced from a WAD, and images from the ../images folder. Adding the ability to turn off individual components would be nice too 👍 

### What problem would this solve?

This effectively allows for total customisation control over the HUD for builders & players

### Extra context

My LUA is rotten but was thinking some pseudo-code like this?
For images:
`local image = "../images/coolPic.webp"`
`trx.ui.hud.draw(image, xPos, yPos, rotAngle, scale)`

Models/sprites can follow a similar pattern, getting the item slot using trx.items[n] or similar alias. Checking the existing docs, trx.items.Item doesn't have a mesh property exposed, not sure how sprites would be referenced either!

PR #6355, opened 2026-08-21
Add experimental Lua UI and persistent storage layer

rr-:
#### Checklist

#### Description

Adds persistent Lua state and scriptable HUD UI, then uses both to restore the Race for the Iris timer in TR4.

`trx.store` exposes two persistent tables:

```lua
trx.store.level.race_frames = 0      -- reset when a level loads
trx.store.game.visited_karnak = true -- preserved for the playthrough
```

Store values may be numbers, strings, booleans, or tables containing supported values. Table identity is preserved across saves, including shared and cyclic references. Unsupported values are omitted during serialization, with their keys reported in the log.

`trx.ui` allows level scripts to contribute widgets to the game's existing UI regions through a new `trx.events.on_ui_draw` event, fired once per screen region on each rendered frame:

```lua
trx.events.on_ui_draw(function(location)
  if location == trx.ui.Location.TOP_CENTER then
    trx.ui.stack({ spacing = { v = 2 } }, function()
      trx.ui.label("00:12")
      trx.ui.bar({
        type = trx.ui.BarType.PROGRESS,
        value = 40,
        max_value = 100,
      })
    end)
  end
end)
```

Container widgets take a callback for their contents rather than requiring scripts to manually open and close UI scopes. This keeps the scene balanced even when a callback raises an error. Drawing bindings are only valid during `on_ui_draw`; measurement APIs remain available elsewhere.

UI dimensions use canvas units exposed by `trx.ui.canvas` and `trx.ui.safe_area`. Script widgets are inserted into the region currently being built by the game, allowing them to participate in the same layout as native elements such as health bars and item names rather than being drawn as a separate overlay.

`trx.game.LOGIC_FPS` is also exposed so frame-based script timers can convert elapsed frames to seconds.

As part of the API cleanup, `trx.game.trx_version` has been renamed to `trx.game.TRX_VERSION`. This is a breaking change and is documented in the migration guide.

Race for the Iris now implements its original elapsed-time display entirely in its level script, timing gameplay between the opening and closing cutscenes.

Resolves TRX87
Resolves TRX1109
Part of #5351 / TRX185


src/lua/api/game.lua
```diff
@@ -20,6 +20,24 @@ api.unit("game.Seconds", {
...
 })
 
+api.const("game.TRX_VERSION", {
+  value = raw.TRX_VERSION,
+  -- The real version names this checkout, so the reference prints a sample
+  -- instead.
+  sample = "TRX 1.9.3-42-g0f4c2a1",
+  type = "string",
+  description = "What this build reports as its version: `1.9.3` for a release, and the tag with "
+    .. "the commits since it for a development build.",
```
lahm86: Nit: "since _then_..."

src/lua/api/store.lua
```diff
@@ -0,0 +1,47 @@
...
+
+A store holds numbers, strings, booleans and tables of those. A value of any
+other type, a function among them, is dropped when the game is saved and the
+key it sat at is named in the log. A value is checked as the game is written
+rather than as a script assigns it, so a value a save cannot hold is reported
+later than the line that stored it.
+
+Put one table in two places and you get one table back, not two copies. Change
+it through the first place and the second place shows the change. A table that
+holds itself works too.
```
lahm86: `A table that holds itself works too.` - is it the kind of setup below? Maybe we could say self-referencing or something along those lines, it took me a few reads to understand this.
```
A = {}
A["self"] = A
```

src/lua/api/store.lua
```diff
@@ -0,0 +1,47 @@
...
+
+An item handle, and any other handle the engine owns, is not a value a store
+takes. A script that wants to name an item across a save stores something the
```
lahm86: I'd maybe say `should store` rather than `stores`, as present tense sort of implies there would be automatic conversion.

lahm86 (commented):
Great stuff! Just a few text nitpicks below.

src/lua/api/store.lua
rr-: Correct, that's exactly this setup; reworded a bit.

src/lua/api/store.lua
rr-: Fixed

src/lua/api/game.lua
rr-: Fixed

lahm86 (approved):

aredfan:
The timer in Race for the Iris looks and works great. 👍 

- [ ] OG bug, the timer doesn't stay still, it appears to be readjusting its position when a digit changes.
- [ ] Text scaling, here's a couple of screens comparing the timer's scale in OG and TRX. Personally I think TRX already looks nice, but mentioning just in case.

> <details><summary>OG screenshot</summary>
> 
> OG
> 
> </details>
> 
> <details><summary>TRX screenshot</summary>
> 
> 20260821_142602_Race_for_the_Iris
> 
> </details>

rr-:
Both out of scope for now until we land new TR4 UI IMO.

aredfan (approved):

PR #6375, opened 2026-08-23
objects/falling_block: add TR4 control

lahm86:
#### Checklist

#### Description

This adds TR4's collapsible tiles, which do not have any animation data and instead are wiggled by the engine and shattered into pieces when they fall. It also fixes a bug whereby Lara could slide on a walkable that's sitting directly on a steep slope e.g. Karnak i7.

Some require heavy triggering, and that's available in a property.


lahm86:
The falling blocks no longer explode, but I believe they are supposed to shatter into pieces when they hit the ground. This is potentially best as a follow-up as it touches on static mesh shattering too. LMK what you think.

rr-:
Thanks 🩷 I'd try to land it here, unless it's too annoying?

rr- (approved):
LGTM it'd be nice to clean up the commits :)

lahm86:
Thanks, commits are squashed/re-ordered and branch rebased. There is a rule in place now for the shatter rotation as discussed, so in-game can be toggled with `/rule fx.rotate_debris`; on by default.

rr-:
Lovely – thank you!

aredfan (approved):

PR #6396, opened 2026-08-25
objects/traps: make underwater propeller untriggerable

Checkm8ra1n:
#### Checklist

#### Description

This PR adds the possibility to untrigger underwarter propellers in TR2, since before wasn't possible before with 
turning off a switch or using antitriggers. 
This is also a feature requested from a TRX custom level builder


lahm86:
The propeller in Offshore Rig can be switched off, so I think if there are issues elsewhere, it will be because of missing state changes in the animation data. The engine should not need to hard-code specific behaviour for this, the propeller is either on or off and state control is offloaded to the animation data.

rr-:
Closing for the reasons we talked about on Discord.

PR #6403, opened 2026-08-26
lara/state/crouch: reset torso angle when crouched

### LINKED ISSUE #6402: [Bug] Lara's arms can become strange when looking around while crouched
### Game type

Original game

### Game

TR3

### TRX version

1.10.2

### Operating system

Windows 10

### GPU

_No response_

This is an OG bug that also happens in 4, 5 and the remasters. While Lara is crouched, her torso rotates with her head, causing some problems: her right arm is supposed to be resting on her leg (see image below), but moving around can cause it to clip through her leg, and her left arm is supposed to be touching the ground, but it can end up looking like it's resting in the air, which is is strange.

Lara's arms shouldn't move when she is looking around while crouched to avoid the problems described above.

### Attachments

This is what her arms should look like while in a crouch stance:
Image

### Before submitting

matheusvb3:
#### Checklist

#### Description

Prevents Lara's torso from rotating with the view when looking around while crouched. Previously this caused her arms to leave their intended positions while in the crouch animation.

Fixes #6402

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -60,6 +60,7 @@
 - Changed a missing or unknown `lara_outfit` in a level to fall back to the default outfit, rather than stopping the game from starting (TRX1087)
 - Removed the golden outfits, which the engine now produces from any outfit, freeing their model slots for outfits of your own (TRX1070)
 - Fixed a crash when switching mods (TRX1239)
+- Fixed Lara's arms leaving their intended positions when looking around in the crouched stance (OG bug) (#6402)
 
 **Saves and settings**
 - Added smoke, sparks, mist and bubbles to saves (Gameplay → General → Save effects)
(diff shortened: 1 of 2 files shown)


rr- (approved):

docs/CHANGELOG.md
```diff
@@ -60,6 +60,7 @@
 - Changed a missing or unknown `lara_outfit` in a level to fall back to the default outfit, rather than stopping the game from starting (TRX1087)
 - Removed the golden outfits, which the engine now produces from any outfit, freeing their model slots for outfits of your own (TRX1070)
 - Fixed a crash when switching mods (TRX1239)
+- Fixed Lara's arms leaving their inteded positions when looking around in the crouched stance (#6402)
```
lahm86: Typo - `intended`

lahm86 (approved):

docs/CHANGELOG.md
aredfan: We'll also need `(OG bug)` at the end of the changelog entry, so it'd appear like this:

> Fixed Lara's arms leaving their `intended` positions when looking around in the crouched stance `(OG bug)` (#6402)

aredfan (approved):

docs/CHANGELOG.md
matheusvb3: Thanks!

PR #6429, opened 2026-08-26
Reflect catalog identities by author names

rr-:
#### Checklist

#### Description

Built-in catalog names now match `names.def`, so `O_SHOTGUN_ITEM` becomes `shotgun`, making `trx.catalog.objects.shotgun` consistent with `trx.objects.shotgun`.
When `names.def` has no entry, the lower-case C name without its prefix is used.

- Built-in keys are now derived directly from `names.def` and context prefixes.
- Values recognize every name they were defined under.
- Bulk enums include constants missing from their explicit lists.
- `trxc.enum.get(type, name)` matches names exactly as written.

For mod authors, catalog and object names now agree, while old C-style names remain aliases. Existing scripts and CSV files keep working.


docs/trx/MIGRATING.md
```diff
@@ -90,6 +90,18 @@ order: 3
      `trx.catalog.objects.EXPLOSIVE_OPTION` no longer exist
    - the slots are free for an object of your own to take
 
+9. **The catalogs report the name an author writes**
```
lahm86, line 93: This is maybe too much in terms of implementation details (builders don't need to know about names.def or C, I feel). I think this entry could be simplified to be a bit more understandable.

docs/trx/MIGRATING.md
```diff
@@ -90,6 +90,18 @@ order: 3
...
+   A catalog constant is the name `objects/names.def` gives a thing rather
+   than its C spelling, so `trx.catalog.objects.SHOTGUN_ITEM` is now
+   `trx.catalog.objects.SHOTGUN`, and it is the word `trx.objects.shotgun`
+   already answered to. The old spelling still resolves, so no script has to
+   be rewritten:
+   - `trx.catalog.objects.SHOTGUN_ITEM` and `trx.catalog.objects.SHOTGUN` are
+     the same constant, in any case
+   - a catalog value written back as text now reads `shotgun` where it read
```
lahm86, line 101: This in particular took me a few reads - is it saying this won't work with older versions? Maybe we don't need that, it's agiven with any change imo.

rr-: It normalizes the keys to drop the `_ITEM` as per `names.def`, and that's why it normalizes `SHOTGUN_ITEM` into `shotgun`.

I actually would like some discussion around it, as I am a bit on the fence whether it's the right call to make.
On one hand we want i18n strings we provide in `names.def`, and `strings.json5` to be keyed with "nice" keys. The second argument to `X_OBJ_NAME_DEFINE` exists solely to make the strings "nicer" for builders to type out, and it already conflicts with the way we define _catalog_ entries. In its present form, the PR makes it possible for catalog to refer _both_ by its `O_…` canonical name _and_ `names.def` name, lessening this inconsistency; and serializing the enum back to a string, results in the `names.def` choice being picked back.
On another hand, this change rolls out the `ENUM_MAP` based reflection of _all_ catalog identities, and thanks to this we already know `SFX_LARA_NO` is `lara_no`.

I think we should actually make this change _more_ aggressive and just drop the key from `names.def`, instead using the reflected enum type as everywhere else going forward, meaning the migration notice will be much more impactful as it'll touch all `strings` files, especially those around `KEY_ITEM_1`. But at least it'll be self-consistent across all catalogs and we will go from three systems competing for namespacing (`O_REVOLVER_AMMO_ITEM` C naming, `revolver_ammo_item` derived C-enum naming, and `revolver_ammo` manual `names.def` naming) to two.

WDYT?

lahm86: Thanks. Yeah, that makes sense and I think it's definitely a good change to remove confusion at this level between the different types.

I was more curious about the end of this section - `...so a file a build of this version writes is not one an older build reads`. Just a wording issue imo, I feel this is difficult to understand and if it just means you can't run scripts that have been designed for the new exe on older exes, that's true of any change so perhaps doesn't need to be listed. But LMK if I'm not understanding that properly.

rr-: Yeah that's a very roundabout way to say the same.
I'll send the `names.def` refactor, then rewrite this PR, along with fixing the shitty prose :)

PR #6439, opened 2026-08-27
Name a catalog identity by its key

rr-:
#### Checklist

#### Description

The shotgun had two names: the catalog called it `O_SHOTGUN_ITEM`, while objects called it `shotgun_item`, with its name at `objects/shotgun_item/name`. `Catalog_GetKey` now returns `shotgun_item`, so the same identity has the same spelling everywhere it is read or written.

- This works for every catalog: object, sample, music track, Lara state, Lara animation, and item action IDs all have prefixes that can be removed without collisions.
- Prefix stripping now lives in `catalog/keys.c`. `objects/names.c` already needed the same logic, so it now shares the implementation instead of keeping its own copy.
- The C spelling remains an alias, so `games/tr1/catalog_objects.csv` and files using names such as `O_WOLF` continue to bind as before.
- `ENUM_MAP` is deliberately untouched. Catalog identities are run-time data and the enum map reflects what the executable was built with, so the two are kept apart.

Nothing reads these keys yet, so this does not change anything on disk. It lands now because savegames and injection files are about to start storing them; after that, changing the spelling would become a file-format migration instead of an internal cleanup.

The Lua API is unchanged. Catalog enums already declare `strip = "O_"`, so Lua was already exposing the prefixless spelling. `just lua-api-dump` produces byte-identical output.

Supersedes #6429

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -320,6 +320,7 @@ common_sources = [
   'trx/game/camera/photo_mode.c',
   'trx/game/camera/spline.c',
   'trx/game/camera/vars.c',
+  'trx/game/catalog/keys.c',
   'trx/game/catalog/manager.c',
   'trx/game/catalog/table.c',
   'trx/game/clock/common.c',
--- a/src/tests/meson.build
+++ b/src/tests/meson.build
@@ -1259,6 +1259,7 @@ unit_tests += {
     'game/enum.c',
     'game/game_strings/entries.c',
     'game/gun/registry.c',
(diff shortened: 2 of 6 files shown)


lahm86 (approved):

rr-:
Reverted the enum map touch-up, as it is not going to be needed in the following PRs, after all – enums continue to only state the C names. I have yet to confirm if they need to be declared at all.

aredfan (approved):

PR #6441, opened 2026-08-28
Fix overbright shadows ringing

rr-:
#### Checklist

#### Description

Fixes a bug where sprite shadow and other unlit effects would have bad ringing when the lighting model was set to overbright.

Resolves TRX1296

### DIFF
--- a/data/trx/ship/cfg/shaders/meshes.glsl
+++ b/data/trx/ship/cfg/shaders/meshes.glsl
@@ -226,9 +226,10 @@ void main(void) {
     vec3 L = lightIn * (overbright || saturate ? 255.0 / 128.0 : 1.0) + lr.add;
     gOut.add = overbright ? max(L - vec3(1.0), vec3(0.0)) * (64.0 / 255.0)
                       : vec3(0.0);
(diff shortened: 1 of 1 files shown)


docs/CHANGELOG.md
```diff
@@ -105,6 +105,7 @@
 - Fixed the inventory background showing black, and objects around it going missing, in a headless run
 - Fixed the underwater view wobbling less at higher supersampling values (TRX1207)
 - Fixed sprite shadows breaking apart on uneven ground (Graphic Options → Visuals → Shadows shape) (#6388 / TRX1244)
+- Fixed ringing around sprite shadows when the lighting model is set to overbright (Graphic Options → Rendering → Lighting model) (TRX1296)
```
aredfan: I think we can omit a changelog entry because sprite shadows are unreleased in tr4, and the overbright lighting model is unreleased in tr3.

aredfan (approved):

docs/CHANGELOG.md
rr-: Thanks – good point; dropped.

lahm86 (approved):

PR #6452, opened 2026-08-29
Address saved identities by key

rr-:
#### Checklist

#### Description

Savegames can now name catalog identities by key, beside the slot number they already write. This lets moved mod content keep resolving to the same object or music track instead of following an old slot number to the wrong place.

```c
bool SaveGame_WriteIdentity(
    JSON_WRITE_IO *io, const char *slot_field, const char *key_field,
    CATALOG_CONTEXT context, CATALOG_ID id);
RESULT SaveGame_ReadIdentity(
    JSON_READ_IO *io, const char *slot_field, const char *key_field,
    CATALOG_CONTEXT context, CATALOG_ID *out_id);
```

Savegame reads take the key when one is present and fall back to the number, so saves remain compatible in both directions. A record naming something the game no longer has is skipped and counted instead of blocking the whole save. Music track flags also gain identity records, because their old positional list can shift when a game adds tracks.

#### Testing

- [x] Save and reload in each game.
  Expected outcome: items, music and the rest resume as before.
- [x] Load a save written by an earlier build.
  Expected outcome: it loads, reading the numbers it carries.
- [x] Load a save written here on an earlier build.
  Expected outcome: it loads, ignoring the keys it does not know.

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -877,6 +877,7 @@ common_sources = [
   'trx/game/savegame/file.c',
   'trx/game/savegame/file_read.c',
   'trx/game/savegame/file_write.c',
+  'trx/game/savegame/identity.c',
   'trx/game/savegame/manager.c',
   'trx/game/savegame/resume.c',
   'trx/game/savegame/vars.c',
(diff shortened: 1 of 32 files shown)


aredfan:
Issue:

- [x] The PR build doesn't start up. No pop-up message shows, no log or dmp is created.

lahm86 (approved):

aredfan (approved):

PR #6454, opened 2026-08-29
Let scripts control fog and fog bulbs

rr-:
#### Checklist

#### Description

Adds `trx.fx.fog_bulbs` for reading a level's fog bulbs, and `trx.fx.fog_color` for holding distance fog to a color. Each bulb exposes its position and radius, and lets scripts set its color and density.

```lua
trx.fx.fog_color = trx.math.color(245, 200, 60)
trx.fx.fog_color = nil                             -- back to the level color

for _, bulb in pairs(trx.fx.fog_bulbs) do
  bulb.color = trx.math.color(0, 64, 192)
  bulb.density = 128
end
```

TR4 now uses flip effect 28 to control fog colors, with the timer selecting a color from `tr4.fog`. Timer 100 disables fog bulbs. Mods can customize these colors or reuse the effect in their own scripts. Angkor Wat uses the effect at startup and at the waterfall.

Fog bulbs now initialize with the level fog color. Their level data does not contain color information: density is stored in the red channel, while the remaining channels are unused.

Levels with a custom fog color no longer make fog opaque when transparent fog is enabled. Game flows containing an unknown `death_tile` value now fail to load instead of leaving the level without death tile behavior.

Resolves TRX658

##### TR4, The Angkor Wat and Race for the Iris

- [x] The fog is sand colored at the start of the level, and changes at the waterfall
- [x] Fog bulbs are drawn in the fog color, and disappear where the original turns them off
- [x] Saving inside a changed fog color and loading the save keeps that color
- [x] Leaving the level and coming back puts the level fog color back (hubs)

##### Settings

- [x] Turning fog bulbs off in the settings removes them, in every game
- [x] Fog color, fog distance and water color in the settings still take effect
- [x] A level that names its own fog color keeps that color, unless transparent fog is turned on

### DIFF
--- a/data/trx/ship/cfg/shaders/common.glsl
+++ b/data/trx/ship/cfg/shaders/common.glsl
@@ -23,6 +23,7 @@
 #define VERT_OVERBRIGHT        0x1000u
 #define VERT_TEX_WRAP          0x2000u
 #define VERT_ADDITIVE          0x4000u
+#define VERT_NO_FOG            0x8000u
 
 #define LIGHTING_CURVE_FLAT       0
 #define LIGHTING_CURVE_OVERBRIGHT 1
--- a/data/trx/ship/cfg/shaders/meshes.glsl
+++ b/data/trx/ship/cfg/shaders/meshes.glsl
@@ -606,7 +606,8 @@ void main(void) {
     }
 
     // Fog
-    if ((gIn.flags & VERT_NO_LIGHTING) == 0u && uLightingEnabled != 0) {
+    if ((gIn.flags & (VERT_NO_FOG | VERT_NO_LIGHTING)) == 0u
+        && uLightingEnabled != 0) {
         texColor = applyFog(texColor, length(gIn.eyePos.xyz));
         // The OG skips fog bulbs on additive polys (AddTriClippedSorted
         // excludes drawtypes 2 and 5 from OmniFog).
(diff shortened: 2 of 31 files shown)


aredfan:
Issue:

- [x] In the original game, if fog bulbs are disabled, the distance fog colour reverts to black.  The exception to this behaviour is Valley of the Kings, KV5 and Desert Railroad which keep their assigned fog colour.
- [x] In Angkor Wat, make a save at the start, then go to the waterfall room and make a 2nd save, then reload the first save. The waterfall's blue fog is carried back to the level's start. Reloading the same save twice restores the sand coloured fog. 
[record.txt](https://github.com/user-attachments/files/31609034/record.txt)
- [x] Keep the save from the previous scenario, relaunch the game, when in the title screen, issue `/load 1` and the environment ends up looking darker.

rr-:
>  In the original game, if fog bulbs are disabled, the distance fog colour reverts to black. The exception to this behaviour is Valley of the Kings, KV5 and Desert Railroad which keep their assigned fog colour.

I'm thinking this should be treated as an improvement on TRX part; WDYT?

aredfan:
> I'm thinking this should be treated as an improvement on TRX part; WDYT?

I would argue that since the fog bulbs and distance fog are tied to the same trigger set, it'd make sense to keep them ring-fenced under one option. OG called the option `Volumetic FX`.

My thinking (maybe bias) is keeping the option ring-fenced would allow a return to black distant fog, which better suits dark levels such as Cairo and Giza.

______________

New issue:

- [x] The fog also affects the horizon in levels such as City of the Dead and The Sphinx Complex. Both locations are close to the level's start.

<details><summary>Screenshots</summary>

20260830_074702_City_of_the_Dead
20260830_074722_The_Sphinx_Complex

</details>

rr-:
I think I got it fixed

lahm86 (approved):

aredfan (approved):

PR #6466, opened 2026-08-30
Add a /copy console command

### LINKED ISSUE #6372: [Feature] Make `/version` copy to clipboard automatically

For added ease of bug reporting, this would be handy to use with the GH template.

Some of the build strings are quite long, especially with dev builds, and having it all in the clipboard directly would facilitate things.

WDYT?

### Extra context

_No response_

### Before submitting

rr-:
#### Checklist

#### Description

`/copy` runs another console command and copies its output to the clipboard. It also shows the output in the console, so players can put `/copy pos`, `/copy version`, or any other command result straight into a bug report without typing it out by hand.

For mod authors, the console module now exposes the pieces needed to build wrapper commands:

```lua
local text = trx.console.eval("pos", { capture = true })
trx.console.copy(text)
local names, start_at, end_at = trx.console.complete("po", 2)
trx.console.register({ name = "wrap", complete = trx.console.complete, run = run })
```

`trx.console.eval` can return command output, `trx.console.copy` writes text to the clipboard, and `trx.console.complete` gives Lua commands the same completions as the console prompt. Lua commands can also provide their own completion function instead of using parser-based completion.

Resolves #6372 / TRX1226

##### The command

- [x] `/copy version` copies the version the game reports
- [x] `/copy pos` copies the level and Lara's position while a level is played
- [x] `/copy` on its own says what it expected
- [x] `/copy pos` outside a level reports that the command did not run, and the clipboard keeps what it held
- [x] Ctrl+V in the console pastes back what `/copy` put there

##### Completion

- [x] Tab after `/copy ` offers the command names
- [x] Tab in `/copy giv` completes to `give`
- [x] Tab in `/copy give ` offers what `/give` accepts


lahm86 (approved):

aredfan (approved):
Backtombs 😓

PR #6500, opened 2026-09-04
tr4/data: add pickup aid

lahm86:
#### Checklist

#### Description

This adds the pickup aid to TR4, using the same sparks based approach as TR3. Pickup aids are also now disabled on hidden reach-in, sarcophagus and crowbar-pry modes as they don't make sense on those; the latter would show the pickup aid on the floor mid-tile, and while we could calculate the absolute position, it would need extra logic to work out which way is up, ensuring the sparks aren't too close to the wall, etc. LMK if we want to re-address that; I feel they're more of a puzzle than a regular pickup so think it's fine to omit.

Resolves TRX1355.

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/359


rr- (approved):
Agree on all accounts

aredfan (approved):

PR #6538, opened 2026-09-07
game/phase/phase_pause: resume audio after fadeout

### LINKED ISSUE #6536: [Bug] Audio resumes before the fadeout effect has finished when unpausing the game
### Game type

Original game

### Game

TR3

### TRX version

1.10.2

### Operating system

Windows 10

### GPU

_No response_

When unpausing the game with the fade effect enabled, the audio will resume before the fadeout has ended. This can cause problems with sounds that loop, such as the MP5 shooting sound, making them momentarily louder (possibly because the reverb effect isn't being applied). In in-game cutscenes, this can cause the audio to momentarily desync from the video, causing a small stutter - after the fadeout has ended, the audio corrects itself (this does not happen with FMVs because both the audio and the video resume at the same time).

I am unsure if this affects the other games, but it does affect TR3.

The audio should resume after the fadeout has ended.

### Attachments

I can't upload a video showing this, but it's very noticeable when using the "Story so far" feature to rewatch cutscenes.

### Before submitting

matheusvb3:
#### Checklist

#### Description

This fixes the audio resuming before the fadeout has ended if the fade effect is enabled for the pause screen.

Fixes https://github.com/LostArtefacts/TRX/issues/6536

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -41,6 +41,7 @@
 - Changed vertex snapping to offer Disabled, 320x240, and Upscale Res modes (#6278 / TRX1137)
 - Fixed a setting description showing a question mark in place of a key that is bound to a combination, such as Alt+Enter (TRX1136)
 - Fixed ability to open the inventory ring while a flyby sequence has Lara's control (TRX1057)
+- Fixed the game audio resuming before the fadeout effect has ended when unpausing with this option enabled (#6536)
 - Fixed several dialogs running past the screen edges or overlapping headings at large text sizes (#6293 / TRX1154)
 - Fixed the save and load dialogs covering the inventory headings at large text sizes
 - Fixed the game flashing over the black bars beside the picture when a frame is advanced in photo mode (regression from 1.10) (TRX1068)
--- a/src/trx/game/phase/phase_pause.c
+++ b/src/trx/game/phase/phase_pause.c
@@ -32,6 +32,7 @@ typedef struct {
     } ui;
     GF_ACTION action;
     FADER fader;
+    bool resume_pending;
 } M_PRIV;
 
 static void M_RemoveText(M_PRIV *const p)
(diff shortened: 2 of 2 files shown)


rr-:
This should become two PR's not a single one as it's two logical changes.

Please split this into two.

Also, a regression: The exit-to-title prompt no longer appears during normal gameplay, only during cutscenes.

matheusvb3:
Thank you for informing me of the regression, I did catch it and put the `!` there to negate the condition, in fact the build I have here has the correct behavior (no prompt in FMVs, shows prompt in Lara's Home and Jungle), but I forgot to amend the commit...

The PR has been split into two 👍

rr- (approved):

lahm86 (approved):

docs/CHANGELOG.md
```diff
@@ -41,9 +39,9 @@
 - Changed the F9 key to cycle the lighting model in TR3 and TR4, and the lighting contrast in TR1 and TR2
 - Changed the save, load, level select and mod dialogs to show scroll arrows in every game, as TR1 does (TRX547)
 - Changed vertex snapping to offer Disabled, 320x240, and Upscale Res modes (#6278 / TRX1137)
-- Changed screenshots taken inside photo mode to not include debug overlay elements
 - Fixed a setting description showing a question mark in place of a key that is bound to a combination, such as Alt+Enter (TRX1136)
 - Fixed ability to open the inventory ring while a flyby sequence has Lara's control (TRX1057)
+- Fixed the game audio resuming before the fadeout effect has ended when unpausing with this option enabled (#6536)
```
aredfan, line 44: I think it goes without saying that fades need enabling. Let's simplify the changelog entry like this.

> Fixed the game audio resuming before the fadeout effect has ended when unpausing ~~with this option enabled~~ (#6536)

aredfan (approved):

docs/CHANGELOG.md
rr-: Fixed directly in develop

PR #6541, opened 2026-09-07
Map rendered colors to the level palette

rr-:
#### Checklist

#### Description

Adds an optional Palette quantization setting to the rendering tab. When it is enabled, rendered colors snap to the nearest color in the level palette. With dithering enabled, nearby palette colors blend into smoother shades. The setting applies to the scene and the UI.

#### Implementation details

The renderer builds a lookup table when a level loads and uses it during the composite pass. PlayStation mode still applies its 15-bit color reduction after palette mapping.

##### Palette quantization

- [x] Enable Palette quantization during gameplay.
  Expected outcome: The setting takes effect without a level reload.
- [x] Change levels while Palette quantization is enabled.
  Expected outcome: Colors use the new level's palette.

##### Dithering

- [x] Enable Palette quantization with Software renderer dithering.
  Expected outcome: Colors shade between palette colors instead of banding.
- [x] Enable Palette quantization with PlayStation dithering.
  Expected outcome: The 15-bit color reduction remains visible.
- [x] Disable dithering while Palette quantization is enabled.
  Expected outcome: Colors snap directly to the palette.


lahm86 (commented):
Looks great.
If used in TR4 however, the screen goes black - perhaps because the levels have no palettes?

data/trx/ship/cfg/base_strings.json5
```diff
@@ -1554,6 +1554,10 @@
             "title": "Affine texture mapping",
             "description": "Drops perspective correction from texturing, as the PlayStation had none, so textures warp and swim across large surfaces when the camera moves.",
         },
+        "rendering.enable_palette_quantize": {
+            "title": "Palette quantization",
+            "description": "Maps every rendered color to the nearest color in the level palette. With dithering on, the palette shades between its colors.",
+        },
```
aredfan, line 1560: > Maps every rendered `colour` to the nearest `colour` in the level palette. With dithering on, the palette shades between its `colours`.

We'll need a British translation for the description.

rr-: Thanks – added.

rr-:
> Looks great. If used in TR4 however, the screen goes black - perhaps because the levels have no palettes?

That's right – I've added a quick palette derivation.
Doesn't seem to be too bad on the performance.

20260908_133506_Angkor_Wat

aredfan (approved):

lahm86 (approved):

PR #6550, opened 2026-09-08
Fix FMV and in-game cutscene input discrepancies

### LINKED ISSUE #6537: [Bug] Pause behavior is inconsistent between FMVs and in-game cutscenes
### Game type

Original game

### Game

TR3

### TRX version

1.10.2

### Operating system

Windows 10

### GPU

_No response_

When pausing in-game cutscenes, if you press the <kbd>esc</kbd> button you will get a prompt asking if you wish to exit to the title screen. This prompt does not appear if you are watching an FMV, which creates a strange inconsistency in behavior.

The behavior should be the same for FMVs and in-game cutscenes.

### Attachments

This is very noticeable when using the "Story so far" feature.

### Before submitting

matheusvb3:
#### Checklist

#### Description

This makes the FMV and in-game cutscene pause logic more consistent:

- FMV audio and video only resume after the fadeout has ended, similar to how it is for in-game cutscenes (that is, after 7c0d534bb551c2311c866738cd9bb62595af42c4). Previously they resumed immediately when <kbd>P</kbd> was pressed.
- Pressing <kbd>esc</kbd> will take precedence over pressing pause in FMVs, similar to how it is for in-game cutscenes (this is noticeable when using slowdown)
- Pressing <kbd>esc</kbd> while paused during an in-game cutscene no longer prompts the player if they want to exit to title, similar to how it is for FMVs. This seems more logical to me since cutscenes only play at the very end or the very beginning of levels and can be skipped anyway, please tell me if this isn't a good change.

Fixes #6537

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -42,6 +42,7 @@
 - Changed the save, load, level select and mod dialogs to show scroll arrows in every game, as TR1 does (TRX547)
 - Changed vertex snapping to offer Disabled, 320x240, and Upscale Res modes (#6278 / TRX1137)
 - Changed screenshots taken inside photo mode to not include debug overlay elements
+- Changed the pause behavior to be consistent between FMVs and in-game cutscenes (#6537)
 - Fixed a setting description showing a question mark in place of a key that is bound to a combination, such as Alt+Enter (TRX1136)
 - Fixed ability to open the inventory ring while a flyby sequence has Lara's control (TRX1057)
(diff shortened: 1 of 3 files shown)


rr- (approved):

matheusvb3:
In relation to [this reply](https://github.com/LostArtefacts/TRX/pull/6538#issuecomment-5591483607): I amended to get rid of the negation entirely, that's clearer but I sometimes don't immediately notice I'm using an unnecessary negated conditional, I apologize for the confusion

lahm86 (approved):

aredfan (approved):

PR #6558, opened 2026-09-10
inventory_ring: fix an item at the screen edge

rr-:
#### Checklist

#### Description

Fixes a bug where an item stays at the edge of the screen while the inventory changes from one ring to another, then disappears at once. Before this, the camera pitched a fixed 45 degrees to take the ring out of view, which is not far enough on a viewport that sees more of the scene above and below the center.

The angle now follows the field of view the inventory draws with.

Resolves TRX1404

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -51,6 +51,7 @@
 - Fixed the game flashing over the black bars beside the picture when a frame is advanced in photo mode (regression from 1.10) (TRX1068)
 - Fixed dialogs sitting against a band of screen that no longer matched what the overlay was drawing there, which could leave them overlapping the heading or the item name
 - Fixed the inventory ring's button hints and item count overlapping the line of text the game puts at the foot of the screen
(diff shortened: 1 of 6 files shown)


lahm86 (approved):

aredfan:
The bug does still persist in the PR build. To observe the issue, wait until the back of the compass is facing the screen, then switch rings to the one above.

rr-:
Thanks, changed the approach. Please LMK if it looks OK – worth trying with various items in the ring, as the angle is now derived mathematically from bounding boxes of the rings' content.

aredfan:
> Thanks, changed the approach. Please LMK if it looks OK – worth trying with various items in the ring, as the angle is now derived mathematically from bounding boxes of the rings' content.

The issue is fixed, thank you. I tested with the TR4 compass, flare, and a couple of puzzle items. I also checked the TR3 inventory for good measure.

aredfan (approved):

PR #6561, opened 2026-09-11
Fix Lara's braid after a load

rr-:
#### Checklist

#### Description

This PR fixes three braid issues when loading a level or save:

- The braid tip no longer drops away for a few frames after loading.
- The braid is now visible on the first frame after loading.
- Braid velocity is now stored directly on each segment, fixing stale velocity during setup.

Resolves TRX1304.

#### Testing

- [x] Save and load while wading; the braid stays intact.
- [x] Load a level; the braid is visible on the first frame.
- [x] Switch outfits while wading, including one with two pigtails; the braid is placed correctly.
- [x] Walk, run, jump, and roll; the braid behaves as before.
- [x] Wade, surface swim, and dive; the braid behaves as before.

### DIFF
--- a/src/trx/game/lara/hair.c
+++ b/src/trx/game/lara/hair.c
@@ -7,6 +7,7 @@
 #include <trx/core/utils.h>
 #include <trx/debug.h>
 #include <trx/game/game.h>
+#include <trx/game/interpolation.h>
 #include <trx/game/lara.h>
 #include <trx/game/lara/mesh.h>
 #include <trx/game/lara/pose.h>
@@ -322,6 +323,11 @@ static void M_Control(
 
             Matrix_Pop();
         }
+
+        // The chain has just been built from scratch, so there is no earlier
+        // pose to blend from; without this the first frames drawn after a
+        // level load stretch the braid between the two.
+        Interpolation_RememberBraid();
         return;
     }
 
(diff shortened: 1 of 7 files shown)


aredfan (approved):

src/trx/game/lara/hair.c
```diff
@@ -923,11 +928,12 @@ void Lara_Hair_Draw(void)
...
-        // Until the first control pass has chained the segments off the head,
-        // they still sit at their bone offsets near the world origin - drawn,
-        // the welded ring would stretch from there to the head.
+        // A draw can come before the first control pass, with the segments
+        // still at their bone offsets near the world origin. Chain them off
+        // the head here, so the braid is drawn in place rather than left out
+        // of the frame.
         if (m_IsFirstHair[i]) {
-            continue;
+            M_Control(false, i, braid->setup[i].position, braid->count > 1);
```
lahm86: Just wondering if we could call `Lara_Hair_Control` from somewhere during startup/save loading instead of making changes in the draw routine - or are we not so worried about that, as it's only going to happen on one draw frame?

rr-: I figured the latter but you're right, that way is much cleaner. I'll refactor it today :)

rr-: Thanks – done.

rr-:
@aredfan I have included new fixes for TRX1437 :)

lahm86 (approved):

aredfan:
> @aredfan I have included new fixes for TRX1437 :)

Thank you again, the braid fixes are working great. 👍

aredfan (approved):

PR #6564, opened 2026-09-11
workflow: set macOS 12 as minimum version for the app bundle

Checkm8ra1n:
#### Checklist

#### Description

This PR sets on the macOS Build Workflow, the minimum version for TRX from Sonoma to Big Sur, that is the real minimum version that TRX supports.


rr-:
How does not supporting older versions manifest?

Checkm8ra1n:
When I try to opening it on previous version, from the .app, it appears an error from the system that the minimum version is Sonoma, because the workflow runs on Sonoma. While if I run it from the executable, I saw that the real minimum version is Big Sur

rr-:
Can you please confirm the game runs with this change?

PR builds do not include macOS, so CI cannot check this for us.

What we need is the .app from the built bundle, started on the older macOS, and the game reaching the title screen. I don't think the bare executable is enough: it already started before this change, so it does not tell us whether the bundle does. The libraries in the bundle come from the build runner and carry their own minimum version.

Please say which macOS version you tested on, and attach TRX.log from the run. Thanks

Checkm8ra1n:
I'm currently trying to compiling TRX, from my main OS to try to make an Application as the Workflow does, for then testing it on a VM of macOS Big Sur.

But I have this error of compilation:
```
[2/445] Compiling C object TRX.p/trx_game_lua_capi_struct.c.o
FAILED: [code=1] TRX.p/trx_game_lua_capi_struct.c.o 
clang --target=x86_64-apple-darwin -ITRX.p -I. -I../src -I/opt/local/include -I/opt/local/include/SDL2 -I/usr/local/Cellar/uthash/2.4.0/include -fdiagnostics-color=always -Wall -Winvalid-pch -Wextra -Wpedantic -std=c2x -O3 -Wshadow -Wno-unused -Wno-unused-parameter -Wunused-result -Wunused-value -DMESON_BUILD -DGLEW_NO_GLU -DSDL_MAIN_HANDLED -fms-extensions -fno-omit-frame-pointer -fmacro-prefix-map=../src/trx/= -fdebug-prefix-map=../src/trx/=src/ -DDWST_STATIC -DPCRE2_STATIC -DPCRE2_CODE_UNIT_WIDTH=8 -DDEBUG=0 -Wno-microsoft-anon-tag -Wno-gnu-binary-literal -Wno-gnu-empty-initializer -Wno-gnu-zero-variadic-macro-arguments -mmacosx-version-min=11.0 -D_THREAD_SAFE -MD -MQ TRX.p/trx_game_lua_capi_struct.c.o -MF TRX.p/trx_game_lua_capi_struct.c.o.d -o TRX.p/trx_game_lua_capi_struct.c.o -c ../src/trx/game/lua/capi/struct.c
../src/trx/game/lua/capi/struct.c:475:33: error: use of undeclared identifier 'lua_newuserdatauv'
  475 |     LUA_STRUCT_REF *const ref = lua_newuserdatauv(L, sizeof(LUA_STRUCT_REF), 0);
```

rr-:
@Checkm8ra1n we merged the fixes from yesterday; you'll need Lua 5.4 to get the game to compile natively.

Checkm8ra1n:
Ok thanks for clarifying.
Btw this is how I tested the compatibility with Big Sur:
- I created a macOS Big Sur VM
- I compilled from my host TRX with this commands:
```
meson setup build src \
  --prefix=/tmp/TRX.app \
  --bindir=Contents/MacOS \
  --buildtype release \
  --cross-file tools/shared/mac/x86-64_cross_file.txt \
  -Dc_args="-I/usr/local/Cellar/uthash/2.4.0/include -mmacosx-version-min=11.0" \
  -Dc_link_args="-mmacosx-version-min=11.0"
```
and `ninja -C build install`

- I copied the `TRX.app` generated in `/tmp` in the VM
- When I copied it the game into the VM, macOS did let me opening it; when before the situation was that the application had a 🚫 symbol where if you try to opening it, the system showed an error where you need macOS Sonoma

rr-:
Thanks; my biggest question is, did the game play fine with this fix?

Checkm8ra1n:
I'm sorry but I realized that when I try to run the game (on both my compiled version and the normal one) I have this:
```
Symbol not found: _CVBufferCopyAttachments
  Referenced from: /Users/checkm8/Documents/TRX.app/Contents/MacOS/../Frameworks/libavutil.60.dylib (which was built for Mac OS X 15.0)
```
I asked to a LLM and it said that the problem is that the Symbol _CVBufferCopyAttachments was introduced with macOS Monterey (12.0)
So we need to set macOS 12.0 instead of 11.0
The last time I played TRX on Big Sur was like 1 y ago, so may things are changed.
LMK if is it ok and I rename the PR in case

rr-:
Yes it's OK :) so long as it's verified that it's working as intended

Checkm8ra1n:
Perfect, now I'm creating a Monterey VM, for testing it correctly

Checkm8ra1n:
I updated the workflow, I tested it on my fork and it generates correctly the dmg and TRX runs perfectly with no compatibility errors.
Now code linters doesn't run successfully, may because I used meson with backslashes?

rr-:
Only two tiny problems:

  .github/workflows/job_build_macos.yml:121: trailing whitespace
  .github/workflows/job_build_macos.yml:137: trailing whitespace

rr- (approved):

rr-:
Thanks for looking into it :+1:

PR #6626, opened 2026-09-17
music+data: fix TR3 gym music

lahm86:
#### Checklist

#### Description

This fixes one-shot music in TR3's gym repeating. OG hard-coded track 2 to always play, so the handling for this has been moved to Lua. It will also now play as an overlay track so that Lara doesn't interrupt it while crossing other triggers in that level - LMK if we like that or not.

Resolves TRX1503.

Ref: https://github.com/LostArtefacts/TRXInjectionTool/pull/380

Tests:
- [x] Use the button in room 79, wait for the music to finish then use the button again - it should restart
- [x] None of the other music triggers should restart


rr- (approved):

rr-:
Note: potentially affects TR2 and TR1 gym, too.

lahm86:
> Note: potentially affects TR2 and TR1 gym, too.

TR1+TR2 exit early from this function, so the change only touches on TR3.

aredfan (approved):
> LMK if we like that or not.

I think it's a great change, thank you. 👍

PR #6654, opened 2026-09-20
Let packages move the whole TRX installation with `TRX_DIR`

### LINKED ISSUE #6651: 1.11 crashes on FreeBSD
### Game type

Original game

### Game

All games

### TRX version

1.11

### Operating system

FreeBSD 16 amd64

### GPU

_No response_

Data was collected from TRX windows installer from GOG tr1, tr2 and tr3

After loading tr1 or tr2 menu, games crashes.

work fine as in 1.10.2 version

### Attachments

TRX output and lldb bt:

[trx-bt.txt](https://github.com/user-attachments/files/32432725/trx-bt.txt)
[trx.out.txt](https://github.com/user-attachments/files/32432724/trx.out.txt)

### Before submitting

rr-:
#### Checklist

#### Description

This PR lets packages keep the executable and installed files in separate locations with `TRX_DIR`. For example:

```sh
export TRX_DIR="/usr/local/share/TRX"
exec /usr/local/bin/TRX.bin "$@"
```

This fixes installs where the game could not find shipped files such as `modules/` and `scripts/`.
The PR also makes missing files and script errors stop startup with an error instead of being ignored.

Specifically, it now:

- Reports errors from startup scripts and `_game.lua` instead of continuing.
- Reports the paths searched when `scripts/` is missing.
- Stops when shipped weapon or inventory-ring modules cannot find their JSON5 files.

Normal installs are unchanged. Mod startup scripts that raise now report the error and stop the run.

Resolves #6651

##### Split install

- [ ] Point `TRX_DIR` at an installation tree and run the executable from elsewhere. The game starts.
- [ ] Set `TRX_DIR` without setting `TRX_GAMES_DIR`. The game loads games from `$TRX_DIR/games`.
- [ ] Set both variables. The game uses `TRX_GAMES_DIR`.

##### Broken installs

- [x] Remove `scripts/`. The game reports the searched paths and exits.
- [x] Remove a game's `inv_ring.json5`. The game reports the file and exits.
- [x] Add a syntax error to a startup script. The game reports the error and exits.


lahm86 (approved):

nunotexbsd:
Thanks for this changes that affects packagers.

One case I'm not sure about: `games/<id>/` mixes shipped files (inv_ring.json5, gameflow.json5, catalog_*.csv, injections/) with user-supplied ones (levels/, music/, fmv/). With TRX_DIR pointing at the package tree and TRX_GAMES_DIR at the user's, does the engine merge both, or does TRX_GAMES_DIR replace the shipped set entirely?

At this point, trx freebsd port only install TRX binary and wrapper (that points its variables to where data and games files are). Maybe after this pull is merged, port will have to install data files from its distfile into local/share/TRX.

Well, let's this be merged and I'll see what we can do at packager side.

Thanks

rr-:
Yes -- Each game can define its own data files which reside in `games/`, and a coule of top-level directories such as `shaders`, `scripts` and `modules` are meant to be game-agnostic and a part of the engine data itself.

aredfan (approved):

PR #6658, opened 2026-09-20
Show TR4 loading screens

rr-:
#### Checklist

#### Description

This PR adds scripted viewpoint previews shown before levels and when opening the main menu that serve as loading screens in TR4.

- Adds the missing TR4 title loading screen definition.
- Prevents duplicate title previews and restores the camera state afterward.
- Adds an original-style duration bar showing the preview’s remaining time.
- Fixes a crash caused by Lara’s meshes surviving level unload.
- Keeps the title logo hidden until the ring enters view, then fades it in.

Regular level previews last two seconds; the title menu preview lasts one second.
Resolves TRX493


aredfan:
Issues (testing in progress):

- [x] The `Loading screen` option will need to be unhidden for TR4.
- [x] I'd like to propose adjusting the TR4's Loading screen default configuration to `Always`, we can see how it goes.
- [x] If Lara is underwater when reloading a save, or ending a level, the loading screen gets tinted. The latter issue can be tested in Sacred Lake, room 87. **Edit:** It can also occur when exiting to title.

docs/gameflow.schema.json
```diff
@@ -628,6 +631,60 @@
       ],
       "additionalProperties": false
     },
+    "seq_loading_camera": {
+      "type": "object",
+      "description": "Shows a fixed view of the level before play starts instead of a loading picture. The loading screen option enables or disables this view. This sequence is equivalent to the TR4 LoadCamera command. TR4 only.",
```
lahm86: Seems to work well in other games as-is, so maybe we can drop "TR4 only" at the end.

aredfan:
New issues (testing in progress):
LMK if any of these issues are out of scope so I can raise a separate issue for it.

- [ ] Inside Menkaure's Pyramid, the puzzle pickup embedded to the wall appear to be lower than its expected position.

<details><summary>TRX screenshot</summary>

20260921_074300_Inside_Menkaure&#39;s_Pyramid

</details>

<details><summary>OG screenshot</summary>

OG1

</details>

- [ ] In OG, loading screens show the level as it is. For example, pickups are either present or collected, flip groups being untriggered or triggered, puzzle receptacles being empty or full. A good example of this is in Temple of Poseidon.

<details><summary>OG screenshots, before and after</summary>

OG2
OG3

</details>

rr-:
Fixed the flip maps and the wording.

The Menkaure pickup height is a separate issue, and I think it affects every TR4 pickup. OG lifts a pickup by its own height when the level starts. `InitialisePickUp` (`TOMB4/game/init.cpp:489`):

```c
ocb = item->trigger_flags & 0x3F;
if (ocb == 0 || ocb == 3 || ocb == 4)
    item->pos.y_pos -= bounds[3];
```

The lift applies to surface pickups: normal, plinth high, and plinth low. Wall-mounted pickups stay where the level places them. TRX already has the collision logic in `pickup.c`, including plinth bounds and `M_FindPlinthBounds`. `PICKUP_MODE` also covers all OCB modes. However, `M_Initialise` does not apply the lift, leaving pickups one mesh height too low. IDK if this is intentional due to custom levels?

The same routine also handles two unrelated behaviours:

- OCB 128 makes the pickup rotate.
- OCB 256 clears `mesh_bits`.

The lift is a bit magic and I don't like it; I'd probably opt-in for lifting the pickup items in `_game.lua` on any TR4 level start rather than hardcode it into C. Either way, it warrants a separate ticket.

CC @lahm86

lahm86:
I think the `item->pos.y_pos -= bounds[3];` lifting we already have covered in `Object_DrawPickupItem` (so fixing ones sitting below the baseline or floating above).

The crowbar pickups rely on frame offsets for positioning them correctly so use a different draw routine - is it possible that the pickup modes that are set in Lua aren't getting applied in time before the screen is shown?

https://github.com/LostArtefacts/TRX/blob/0e16477d8e8e5cf19dc96034bd9ac9e6736efcf0/src/trx/game/objects/general/pickup.c#L975

docs/trx/game_flow/SEQUENCES.md
```diff
@@ -130,6 +130,52 @@ default game flow for examples.
...
     <td>Double</td>
   </tr>
+  <tr valign="top">
+    <td rowspan="10"><code>loading_camera</code></td>
+    <td><code>source_x</code></td>
+    <td>Integer</td>
+    <td rowspan="10">
+      Shows a fixed view of the level before play starts instead of a loading
+      picture. The loading screen option enables or disables this view. This
+      sequence matches the TR4 <code>LoadCamera</code> command. TR4 only.
```
lahm86: Sorry, another "TR4 only" mentioned here.

rr-: Fixed :)

rr-:
> I think the `item->pos.y_pos -= bounds[3];` lifting we already have covered in `Object_DrawPickupItem` (so fixing ones sitting below the baseline or floating above).
> 
> The crowbar pickups rely on frame offsets for positioning them correctly so use a different draw routine - is it possible that the pickup modes that are set in Lua aren't getting applied in time before the screen is shown?
> 
> https://github.com/LostArtefacts/TRX/blob/0e16477d8e8e5cf19dc96034bd9ac9e6736efcf0/src/trx/game/objects/general/pickup.c#L975

Sorry, that was exactly right. Refactored.

Also, in the process, I discovered a curiosity about the OGs – the loading screen seems to show the last level's scene, rather than the scene from the level about to be loaded. This makes little sense IMO for savegames. TRX shows the level about to be played.

lahm86 (approved):
Looks great. Agreed that it makes much more sense to show the screen for the level that's starting and not the previous.

aredfan:
I agree with the change for loading screens to show the next scene.
The prior issues are fixed, thank you. 👍 

New issues:
- [ ] The loading screen is a black screen when starting Angkor Wat and Karnak.
- [ ] Lara and her shadow can appear in loading screens which isn't the case in OG. It'll be worth hiding Lara's footprints too if feasible.

If the latter issue is unfixable then we can consider it an enhancement instead.

aredfan:
Looks great, thank you.
Just one more issue remaining.

- [ ] It looks like a previous bug has resurfaced. The `Inside Menkaure's Pyramid` puzzle item is lower than its expected position.

aredfan:
The prior issues are resolved, thank you again. I found a new issue while retesting.

- [ ] The `Cambodia, 1984` text is briefly visible after the loading screen ends, and before the intro cutscene starts.

aredfan (approved):

PR #6659, opened 2026-09-20
option/combine: support item combining

rr-:
#### Checklist

#### Description

Inventory items can now be combined when Lara is carrying their matching partner. The partner-selection ring opens over the original item, and combining replaces both pieces with the resulting object—for example, the two Eye of Horus pieces become the complete Eye.

Keyholes no longer accept the individual halves, since they only work when combined.

Recipes are defined in `object_combos.json5`, so mods can add their own combinations. Laser sights are still unsupported because they attach to weapons instead of creating a separate object.

Resolves TRX482


lahm86 (approved):

aredfan:
**Issues:**

- [ ] Many inventory objects such as the passport and pistols now show a `Use` prompt. Is this intended?
- [ ] When 2 items are combined, the ring can temporarily show a gap until the ring is spun again. This can be tested in Tomb of Semerkhet. Use the all items cheat, then combine the 2 puzzle pieces.

aredfan (approved):

